Issue #185, opened 2012-06-12
multi_json gem version requirement


flah00:
Someone more familiar with the multi_json code base should update the gemspec, to include a minimum version for multi_json. It appears that MultiJson#dump is relatively new (4/12) and the lack of a version causes issues for people with older versions. I know 1.3.6 has #dump, but I'm uncertain what the minimum version is... I could submit a PR for version 1.3.6, but that seems inaccurate and therefore unwanted.

PR #250, opened 2012-09-26
refs: #185 gemspec requires multi_json >= 1.3.2

gruis:
MultiJson.dump was introduced in version 1.3.2.

An alternative approach would be to use MultiJson.encode if
MultiJson.dump is not available.


dblock:
Note https://github.com/ruby-grape/grape/pull/1623

Issue #396, opened 2013-04-29
can not stub/mock helper methods


unorthodoxgeek:
because of the odd and extremely complex way helpers are defined in grape, there is no way I can find to mock/stub helper methods. This is a very serious flaw in this framework IMHO, as testing it proves very difficult.
for instance, I have a helper method named current_user, it returns the obvious value.

I want to test an api request, once when logged in, once when logged out. There's absolutely no way I could find which allows me to do something.stub(:current_user).and_return(user)

This makes it VERY difficult to test grape, on the verge of actually making it impossible.

If the helper was more sensibly made, as a simple module applied on the request class, there would be an easy way to stub it thus test everything which is needed for the development using this library.

PR #397, opened 2013-04-29
Adds Grape::Endpoint.setup for pre-run endpoint config.

mbleigh:
This pull request addresses #396 by providing a `Grape::Endpoint.setup` class method which can be used to manipulate an Endpoint just before it begins executing. The use of this in a test suite would go something like this:

Let's say I have a `current_user` helper that I want to stub. Let's say I don't even have it implemented yet, but I don't want it to error out while I'm building my library. Simple!

``` ruby
describe MyAPI do
  let(:app){ MyAPI }

  context 'with a logged in user' do
    before do
      Grape::Endpoint.setup do |endpoint|
        endpoint.stub(:current_user).and_return mock(:User, name: "Bob Example")
      end
    end

    after{ Grape::Endpoint.setup nil }
  end
end
```

I'm entering this as a pull request for a few reasons:
1. It adds code that will be conditionally executed every time any endpoint is called. I don't take that super lightly.
2. Right now there is only the ability to do a single `setup` step which precludes nesting or multiple layers of setup. I'm not sure if this is the right way to go, but I wanted to start with a straightforward implementation.

Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Next Release
 #### Features
 
 * [#617](https://github.com/intridea/grape/pull/617): Running tests on Ruby 2.1.1, Rubinius 2.1 and 2.2, Ruby and JRuby HEAD - [@dblock](https://github.com/dblock).
+* [#397](https://github.com/intridea/grape/pull/397): Adds `Grape::Endpoint.before_each` to allow easy helper stubbing - [@mbleigh](https://github.com/mbleigh).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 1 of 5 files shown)


fbandrey:
For latest versions of RSpec this is works:
```
...
  Grape::Endpoint.before_each do |endpoint|
    allow(endpoint).to receive(:current_user).and_return double(:user, id: 123, ...)
  end
...
```

thisismydesign:
https://stackoverflow.com/a/71306761/2771889

My preferred way is a more RSpec-like stubbing:

```rb
# helper
module AuthHelper
  def current_user
    # ...
  end
end

# api
module API
  module V1
    class Auth < Grape::API
      helpers AuthHelper
    end
  end
end

# spec
before do
  allow_any_instance_of(AuthHelper).to receive(:current_user).and_return(user)
end
```

dblock:
@thisismydesign Is this something useful for the docs? Care to contribute?

Issue #559, opened 2014-01-27
params 'missing'  due to rack bug


myitcv:
[This commit in `rack`](https://github.com/rack/rack/commit/b0593078ce792a380779528a6a135c066aa03515) fixes an issue whereby all requests which use a temp file (all requests larger 128 kb?) fail to be properly parsed.

The issue manifests itself in `grape` in the following way:

```ruby
params do
  optional :picture, type: String
end
patch '/' do
  picture_from_params = params[:picture] # nil - but should not be
  picture_from_env = env['rack.request.form_hash']['picture'] # not nil - as expected

  env["rack.request.form_input"].eql? env["rack.input"] # false - THIS IS THE BUG fixed by the commit in rack
  env["rack.request.form_input"].equal? env["rack.input"] # true
end
```

Indeed this could have been the root cause of a few issues raised in the Google Group. 

Not so much an issue from `grape`'s perspective apart from it being the place where this error manifests itself. Definitely something to be aware of when  running `Ruby 1.9+` (commit comment suggests this doesn't affect `Ruby 1.8.x`)

So more of an FYI in case anyone else gets bitten by this. 

The issue is further complicated by the fact that running `rack/test` doesn't cross this codepath. Presumably because temp files are not used in tests??

This issue is present in `rack == 1.5.2`, fixed in [`master`](https://github.com/rack/rack/tree/master) as of the time of this post, but not yet released (expected in `1.5.4` I believe)

dblock:
Grape doesn't lock to a particular Rack version, so now that Rack 1.6.0 has shipped, this works. I committed the test with https://github.com/intridea/grape/commit/c72181dcb7d32e37dfd7a26084aeb6738cfa58a3.

dblock:
Note that https://github.com/rack/rack/pull/756 back-ported the fix into Rack 1.5.3.

TangMonk:
same problem

PR #563, opened 2014-01-29
Rack expectations test

myitcv:
Per the description #559, this test demonstrates a bug in `rack 1.5.2` whereby `params` is not correctly populated when `rack.input` is a `Tempfile`

`rack.input` is a `Tempfile` in various circumstances. In our case it was with a [`puma`](https://github.com/puma/puma) webserver frontend serving [requests with a large body](https://github.com/puma/puma/blob/master/lib/puma/server.rb#L667-L669)

This test is designed to fail against `rack 1.5.2` - but works against `rack#master` ([`8dc31fec37` at the time of writing](https://github.com/rack/rack/commit/8dc31fec37f6f53a5f0f8e24274da9035abd92b2))

### DIFF
--- a/spec/grape/rack_expectations_spec.rb
+++ b/spec/grape/rack_expectations_spec.rb
@@ -1,0 +1,27 @@
+require 'spec_helper'
+
+describe 'Rack expectations' do
+
+  it 'demonstrates the Tempfile bug' do
+    input = Tempfile.new 'rubbish'
+    begin
+      app = Class.new(Grape::API) do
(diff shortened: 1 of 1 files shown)


dblock:
Grape doesn't lock to a particular Rack version, so now that Rack 1.6.0 has shipped, this works. I committed the test with some minor changes with https://github.com/intridea/grape/commit/c72181dcb7d32e37dfd7a26084aeb6738cfa58a3.

PR #614, opened 2014-04-03
regex validator refuses params with nil value

### LINKED ISSUE #608: Regexp Validation
Hi

I am write an API using Grape. 
Writing my rspec tests, I noticed a problem with a regexp validation of a param.
My Api is :

``` ruby
class Methods < Grape::API
  helpers do
    def results
      {
        method:  route.route_method,
        id:      params[:id]
      }
    end
  end

  params do
    requires :id, regexp: /^\d{5}$/
  end

  namespace :tests do
    [:get, :post, :put, :delete].each do |method|
      desc "Renvoie le parametre Id via #{method}"
        route(method, method) do
          results
        end
    end
  end
end
```

The Rspec file :+1: 

``` ruby
describe "Test" do

  params = [[nil, 'aucun param', :missing],
            [{ :id => nil }, 'id is nul', :invalid],
            [{ :id => '' }, 'id is vide', :invalid],
            [{ :id => 'A' }, 'id is "A"', :invalid],
            [{ :id => '54' }, 'id is "54"', :invalid],
            [{ :id => 78 }, 'id is 78', :invalid],
            [{ :id => 1234 }, 'id is 1234', :invalid],
            [{ :id => 567890 }, 'id is 567890', :invalid],
            [{ :id => "0000" }, 'id is "0000"', :invalid],
            [{ :id => 0000 }, 'id is 000', :invalid]
  ]

  params.each do |pp|
    it "si #{pp[1]}" do
      get "/tests/get", pp[0]
      expect(response.status).to eql(403)
      expect(response.body).not_to be_empty
      jRes = JSON.parse(response.body, symbolize_names: true)
      expect(jRes).to have(1).items
      expect(jRes).to have_key(:error)
      expect(jRes[:error]).to be_eql("id is #{pp[2]}")
    end
  end
```

The problem is with the test : :id => null

The API returns a 200 and a result : 

``` javascript
{
  "method" : "GET",
  "id" : null
}
```

If I want to test the id is null from a browser, I enter as an URL :+1: 

``` javascript
http://xxxxxx/test/get?id
```

I have a the same result, so the require and the regexp are not done...

dm1try:
Fixes #608

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Next Release
 
 * Your contribution here.
 
+#### Fixes
+
+* [#614](https://github.com/intridea/grape/pull/614): Params with `nil` value are now refused by `RegexpValidator` - [@dm1try](https://github.com/dm1try).
+
 0.7.0 (4/2/2013)
 =================
 
--- a/lib/grape/validations/regexp.rb
+++ b/lib/grape/validations/regexp.rb
@@ -2,7 +2,8 @@ module Grape
   module Validations
     class RegexpValidator < SingleOptionValidator
       def validate_param!(attr_name, params)
-        if params[attr_name] && !(params[attr_name].to_s =~ @option)
+        if params.has_key?(attr_name) &&
+          (params[attr_name].nil? || !(params[attr_name].to_s =~ @option))
           raise Grape::Exceptions::Validation, param: @scope.full_name(attr_name), message_key: :regexp
         end
       end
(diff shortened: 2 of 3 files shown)


calfzhou:
why refuse `nil` or blank value params? Why not leverage `allow_blank` to refuse `nil` if you don't like it. If blank value is allowed, the blank value should not go through `regexp` validation. Otherwise, user need modify the regular expression to some complex form to support nil value.

dblock:
@calfzhou Yes yes, this should support `allow_blank` like everything else. Would appreciate an issue/test/PR.

PR #630, opened 2014-04-16
Add Hack on Nitrous Button

arunthampi:
This will enable developers to work on Grape in the cloud on Nitrous.IO


dblock:
The links merged here broke, so I removed it. If you want to re-add it I probably want to see it on the new http://www.ruby-grape.org instead, which is https://github.com/ruby-grape/ruby-grape.github.io.

PR #642, opened 2014-05-13
Added support for coercion to Set or Set[Other]

jordansexton:
I like Virtus's support for coercion to Sets as seen here:
https://github.com/solnic/virtus#collection-member-coercions

They can be treated the same as Arrays in the case of `nil`, and changing line 42 from `@option[0]` to `@option.first` behaves the same for either.


dblock:
Closed via https://github.com/intridea/grape/pull/995.

PR #730, opened 2014-08-22
Issues/677 require regexp validator bug

zdennis:
For more information (see the actual Github [issue](https://github.com/intridea/grape/issues/677)) that this describes.

Based on the failing specs in this PR I think the way validation errors are handled may be able to be simplified which solves the original issue I had (see [this comment](https://github.com/intridea/grape/issues/677#issuecomment-53008917)). If that's the case then this code here can be even further simplified (and likely most of it deleted).

If that's not the case then I will continue exploring the validation chain logic and update the PR to intelligently group validation errors that must run together and allow others to be short-circuited.

_Note: I do not expect this to be accepted as is, using it currently to get feedback on what the final outcome should be_.

### DIFF
--- a/lib/grape/dsl/validations.rb
+++ b/lib/grape/dsl/validations.rb
@@ -8,7 +8,7 @@ module Grape
       module ClassMethods
         def reset_validations!
           settings.peek[:declared_params] = []
-          settings.peek[:validations] = []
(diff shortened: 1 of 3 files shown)


dm1try:
Apparently the original issue already is already fixed in master. The provided tests are passed.

PR #735, opened 2014-08-25
Proposal for shared params

dspaeth-faber:
Like discussed in #716 I would suggest this DSL for shared params instead of using Helpers.

### DIFF
--- a/spec/grape/api/helpers_spec.rb
+++ b/spec/grape/api/helpers_spec.rb
@@ -1,22 +1,35 @@
 require 'spec_helper'
 
 describe Grape::API::Helpers do
-  module SharedParams
-    extend Grape::API::Helpers
+  module Grape
+    class API
+      module HelpersSpec
+        module SharedParams
+          extend Grape::SharedParams
 
-    params :pagination do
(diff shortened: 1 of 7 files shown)


spec/grape/api/helpers_spec.rb
```diff
@@ -1,22 +1,35 @@
...
-      helpers SharedParams
       format :json
 
+      shared_params 'a name' do
+        params :pagination_more do
+          optional :per_page, type: Integer
+        end
+      end
+
+      include_params Grape::API::HelpersSpec::SharedParams, 'a name'
```
dmitry: Why is there are `include_params` required? Why not simply:

```ruby
        shared_params :pagination_more do
          optional :per_page, type: Integer
        end
```

dm1try:
So @dblock there are no other opinions :)
I can push this forward and finish the PR if needed.

spec/grape/api/helpers_spec.rb
dm1try: @dmitry good point :+1:

dblock:
Thanks @dm1try

PR #784, opened 2014-10-20
Fix bug that would overwrite the content of the response body when using multiple presenters

mfunaro:
When using multiple presenters that make use of Grape Entity, if the user does not specify the root when calling the present method, the contents added to the body by the first presenter would be overwritten by the second presenter. 

This issue becomes apparent when the user does not pass a root to the present method but instead opts to use the 'root' inside the Grape Entity being provided to present method.


dblock:
I ran into https://github.com/intridea/grape/issues/1023 which was caused by this change. I am going to change `respond_to?('merge')` to a `kind_of?(Hash)`, which might break something (but doesn't break specs, so I am going to go ahead). Please comment there if you think that's wrong.

PR #809, opened 2014-11-11
Don't use `.format` suffix in paths if you only have one format

ajvondrak:
For example, say you have the following API:

``` ruby
require 'grape'

class API < Grape::API
  format :json

  get :hello do
    { hello: 'world' }
  end
end

run API
```

If you hit `/hello`, you'd get back the JSON `{"hello":"world"}`. This makes sense: you specifically say that your format is JSON.

However, before this commit, you could also hit `/hello.xml`, which would confusingly return the _JSON_ `{"hello":"world"}`. Why bother exposing this path as a valid API endpoint when we're just going to ignore the format anyway?

In this commit, `/hello.xml` (or `/hello.json` or `/hello.foobar`) will all be 404s. We don't even entertain the notion of an optional format path parameter.

Old behavior is still preserved if you add different content types after your `format :xxx` declaration. So, if you have the following API:

``` ruby
require 'grape'

class API < Grape::API
  format :json
  content_type :xml, 'application/xml'

  get :hello do
    { hello: 'world' }
  end
end

run API
```

then `/hello`, `/hello.json`, `/hello.xml`, and `/hello.foobar` all return the JSON `{"hello":"world"}`. I don't know why you'd want this behavior either, but it was being enforced by the specs, so I preserved it.

What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#779](https://github.com/intridea/grape/pull/779): Fixed using `values` with a `default` proc - [@ShPakvel](https://github.com/ShPakvel).
 * [#799](https://github.com/intridea/grape/pull/799): Fixed custom validators with required `Hash`, `Array` types - [@bwalex](https://github.com/bwalex).
 * [#784](https://github.com/intridea/grape/pull/784): Fixed `present` to not overwrite the previously added contents of the response body whebn called more than once - [@mfunaro](https://github.com/mfunaro).
+* [#809](https://github.com/intridea/grape/pull/809): Removed automatic `(.:format)` suffix on paths if you're using only one format (e.g., with `format :json`, `/path` will respond with JSON but `/path.xml` will be a 404) - [@ajvondrak](https://github.com/ajvondrak).
 * Your contribution here.
 
 0.9.0 (8/27/2014)
--- a/README.md
+++ b/README.md
@@ -196,12 +196,12 @@ run Twitter::API
 
 And would respond to the following routes:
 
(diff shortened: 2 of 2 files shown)


ur5us:
I just tried to upgrade one of our apps from Grape 0.9.0 to ~> 0.10.0. It took me hours to debug the issue as to why `.json` in the URL wouldn't work, I simply missed how important this change was. Also, this change breaks Swagger UI as described in https://github.com/tim-vandecasteele/grape-swagger/issues/187. I'm not sure whether it's just our installation of Swagger UI but pointing the [Swagger UI demo](http://petstore.swagger.wordnik.com/) at our own API didn't work either (probably that's outdated, too?). Moreover, using `grape-swagger-rails` is not really an option as it'd downgrade the `grape-swagger` gem.

Changing `format :json` to 

```
content_type :json, 'application/json'
default_format :json
```

actually fixes the problem. Now, with that change the following URLs will work:

```
/users.json
/users
```

and both will return the same. To be honest, that's what I'd expect when using `format :json`. It should actually respond to both URLs, with and without the suffix. Moreover, for any other format I'd actually expect a `406 Not acceptable` header instead of an `JSON` response. So in my opinion this change goes too far for the whole suffix discussion and not far enough when it comes to asking for a response format which is actually not supported.

dblock:
I had a similar sentiment a few times. I would be open to undoing this change carefully as you describe, if you want to try a PR.

ur5us:
Alright, though I'll probably aim for several PRs as a lot of the things I've described are actually different/non-related things.

PR #817, opened 2014-11-17
Inspect the options hash on a route object

penguinpowernz:
Don't want to parse the output from `to_s`.


penguinpowernz:
No time sorry.

PR #875, opened 2015-01-03
Version specs

rodzyn:
Hi, I've started to implement routing specs and have some questions.

Routing module is using "nest" method from API class, I think this coupling between is not quite necessary. I get rid of this class, the only thing what it is doing is calling provided block (name_inheritable methods) then making instance_eval of passed block as parameter - so I did the same. 
However, side effect of "nest" method is reset_validations! method (from Grape::DSL::Validations module) and I'm not if I should mimic it's behaviour too. 

Second thing, method "versions" suggest that we could expect probably an array of versions, but it keeps just latest defined version. Maybe we should change it in some way?

### DIFF
--- a/lib/grape/dsl/routing.rb
+++ b/lib/grape/dsl/routing.rb
@@ -35,10 +35,13 @@ module Grape
             fail Grape::Exceptions::MissingVendorOption.new if options[:using] == :header && !options.key?(:vendor)
 
             @versions = versions | args
-            nest(block) do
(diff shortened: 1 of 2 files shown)


dblock:
This is good, and the change with `nest` makes sense. Can you squash your commits please, I'll merge.

rodzyn:
Ok, thanks. Squashed.

dblock:
Tanks for contributing pending specs, looking forward to more!

PR #876, opened 2015-01-04
Close #797

### LINKED ISSUE #797: declared(params) as Hashie::Mash
It would make the code a little prettier if it was:

`params.key`

instead of:

`params[:key]`

rodzyn:
Hashie::Mash for declared params

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Next Release
 
 * [#871](https://github.com/intridea/grape/pull/871): Fixed Grape::Middleware::Base#response - [@galathius](https://github.com/galathius).
 * [#559](https://github.com/intridea/grape/issues/559): Support Rack 1.6.0 to parse requests larger than 128KB - [@myitcv](https://github.com/myitcv).
+* [#876](https://github.com/intridea/grape/pull/876): declared(params) now returning Hashie::Mash - [@rodzyn](https://github.com/rodzyn).
 * Your contribution here.
 
 0.10.1 (12/28/2014)
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Build Status](http://img.shields.io/travis/intridea/grape.svg)](https://travis-ci.org/intridea/grape)
 [![Dependency Status](https://gemnasium.com/intridea/grape.svg)](https://gemnasium.com/intridea/grape)
 [![Code Climate](https://codeclimate.com/github/intridea/grape.svg)](https://codeclimate.com/github/intridea/grape)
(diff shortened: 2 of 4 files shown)


dblock:
This needs a CHANGELOG entry please, and possible documentation in README.

rodzyn:
Added.

dblock:
Merged, thanks.

jwkoelewijn:
Mash does not seem to play well with Rails::StrongParameters, because of the reimplementation of #respond_to?:
  - Mash#respond_to? will always return true for methods ending in '!' or '?'
  - Mash#identifier? will return true or false, based on the identifier being present in the Mash
  - Mash#identifier! will make sure the identifier key is present in the Mash

Rails::StrongParameters utilizes the #sanitize_for_mass_assignment method, which raises an
 ActiveModel::ForbiddenAttributesError whenever the provided attributes respond to :permitted?
 (which it does for all Mash'es, because it ends with a '?') AND when !attributes.permitted?
 (which always returns true when the 'permitted' key is not present in the Mash):

```ruby
module ActiveModel
  class ForbiddenAttributesError < StandardError
  end

  module ForbiddenAttributesProtection
    protected
    def sanitize_for_mass_assignment(attributes)
      if attributes.respond_to?(:permitted?) && !attributes.permitted?
        raise ActiveModel::ForbiddenAttributesError
      else
        attributes
      end
    end
  end
end
```
 
This results in always raising a ActiveModel::ForbiddenAttributeError

I do not see how to write a simple test-case for this and for now I reimplemented the #declared(params) helper in my own helpers utilizing a regular Hash.

kuraga:
@jwkoelewijn @dblock Are `Hashie::*` useful for us, at all?

jwkoelewijn:
I would propose to use the standard Hash and add a line or two to the readme to show how one could overwrite the #declared(params) method, for instance by wrapping it in a Hashie::Mash.

I would argue that the need for a Mash is not standard, rather an exception.

Apart from that, I never felt the need to access parameters as methods....

dblock:
Please see https://github.com/Maxim-Filimonov/hashie-forbidden_attributes, what am I missing?

jwkoelewijn:
Ok, so there is a gem to fix a bug introduced by a change to introduce mashes. What's still not entirely clear though is the added value of using this Mash thing in the first place...

dblock:
To be fair Mash was used all over the place _before_ Rails4 strong attributes :) But I agree, Mash is "everything to everyone", and possibly there's a better fix that provides Mash semantics by mixing parts of Hashie (indifferent access? method access?) without needing another gem, I would welcome a PR like that.

PR #877, opened 2015-01-04
fix #726 - add failing spec

### LINKED ISSUE #726: JSON endpoints accept 'text/plain' in JSON format
According to the Grape documentation, this endpoint should only accept JSON requests, and all other formats should result in a `406` response code.

> The following API will only respond to the JSON content-type and will not parse any other input than application/json, application/x-www-form-urlencoded, multipart/form-data, multipart/related and multipart/mixed. All other requests will fail with an HTTP 406 error code.

```
  format :json
  default_format :json
end
```

I am finding that posting `text/plain` in JSON format does not result in a 406. Debugging within the endpoint shows that `env['CONTENT_TYPE']` is indeed `text/plain`.

Morred:
#726


dnesteryuk:
@Morred thank you for this PR! https://github.com/ruby-grape/grape/pull/1589 contains a similar test and an actual fix for the issue :blush:

PR #878, opened 2015-01-05
Check supported format even when there is no Content-Type given

rodzyn:
I found this "thing" investigating #726 .

I'm not sure why there is a check with

``
!request.form_data? || !request.media_type
``
But removing it does not break test suite (one change in the very bottom of this commit) and gives us desired feature I guess

### DIFF
--- a/lib/grape/middleware/formatter.rb
+++ b/lib/grape/middleware/formatter.rb
@@ -48,7 +48,6 @@ module Grape
       # store read input in env['api.request.input']
       def read_body_input
         if (request.post? || request.put? || request.patch? || request.delete?) &&
-           (!request.form_data? || !request.media_type) &&
            (!request.parseable_data?) &&
            (request.content_length.to_i > 0 || request.env['HTTP_TRANSFER_ENCODING'] == 'chunked')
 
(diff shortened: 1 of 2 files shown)


dblock:
The reason why that code is there is because there's parsing going on outside of Grape in Rack (https://github.com/rack/rack/blob/master/lib/rack/request.rb#L201) or Rails, however this may be outdated. I think putting this back makes the params parsed twice, but you're right we don't have any specs.

Can you dig a bit in the codebases of Rails/Rack just to understand what's going on there? I am not saying this change isn't right, but it's quite sensitive in terms of where it is :)

loren:
I just stumbled across this as well via https://groups.google.com/forum/?fromgroups#!searchin/ruby-grape/content-type/ruby-grape/7p5JhyCwejw/ubOCdSz4Hm0J when trying to set a default request content type.

dblock:
Anyone wants to add something here? What should we do with this?

rodzyn:
Since https://github.com/ruby-grape/grape/pull/1589/files has addressed the issue I'm closing this one. It's been here way too long 🤣

PR #880, opened 2015-01-05
Optional default lambdas with params (Close #785)

### LINKED ISSUE #785: Default param values from endpoint variables
You can access things like `cookies` and `request` from inside of an endpoint definition, but for my particular application, I need to give the requester the ability to override some params that should otherwise default to Rack environment values.

Example:

``` ruby
# config.ru
require 'grape'

class API < Grape::API
  format :json
  rescue_from :all

  params do
    optional :id, type: Integer, default: -> { cookies[:id] }
    optional :ip, default: -> { request.ip }
  end

  get do
    {
      params_id: params[:id],
      endpoint_id: cookies[:id],
      params_ip: params[:ip],
      endpoint_ip: request.ip,
    }
  end
end

run API
```

```
$ rackup
$ curl 'localhost:9292' && echo
{"error":"undefined local variable or method `cookies' for #<Grape::Validations::ParamsScope:0x007faec2846fe8>"}
$ curl 'localhost:9292?id=1' && echo
{"error":"undefined local variable or method `request' for #<Grape::Validations::ParamsScope:0x007faec2846fe8>"}
$ curl 'localhost:9292?id=1&ip=whatever' && echo
{"params_id":1,"endpoint_id":null,"params_ip":"whatever","endpoint_ip":"127.0.0.1"}
```

 I could accomplish this with one-off lines like `params[:id] ||= cookies[:id]` in the endpoint, but I have to do this for a bunch of different endpoints, so I'd like to just be able to reuse the params (via `use :such_and_such`).

How hard would it be to allow defaults access to these `Grape::Endpoint` instance variables? It looks like you'd have access to `cookies`, `headers`, `env`, `request`, etc. [at the point where you run the validators](https://github.com/intridea/grape/blob/master/lib/grape/endpoint.rb#L229), and default values are populated [via validation](https://github.com/intridea/grape/blob/master/lib/grape/validations/validators/default.rb#L10). But I don't want to just kludge a patch in there myself if there's a more elegant way of doing it than passing every-variable-under-the-sun to the `validate!` method.

Let me know what you think!

rodzyn:


README.md
```diff
@@ -704,6 +704,15 @@ params do
   optional :non_random_number, type: Integer, default:  Random.rand(1..100)
 end
 ```
+Default value provided by lambda can take zero, one or two arguments
+
+```ruby
+params do
+  optional :random_number, type: Integer, default: -> () { Random.rand(1..100) }
+  optional :random_number, type: Integer, default: -> (request) { Random.rand(1..100) }
+  optional :random_number, type: Integer, default: -> (request, cookies) { Random.rand(1..100) }
```
dblock: I think the proc should only take `request`, from which you should be able to get to cookies and what not. Request is basically everything, and if it's not we should be passing in context or a hash of options.

lib/grape/endpoint.rb
```diff
@@ -232,6 +232,7 @@ def run(env)
 
       route_setting(:saved_validations).each do |validator|
         begin
+          validator.context = Hashie::Mash.new(request: request, cookies: cookies)
```
dblock: Assigning anything on a validator inside a request seems bad. It now holds onto the request, and the validator is attached to the endpoint. So when is request going to be GCed?

README.md
rodzyn: Yeah, good point

lib/grape/endpoint.rb
rodzyn: Hmmm... The request will be GCed with validator object I think. We can make context nil after validation but it's a bit over engineering the thing.
Do you see here any use case/example that could lead to memory leak/stack overflow?

rodzyn:
@dblock What do you think about validate! method taking request object instead of params hash as an argument?

lib/grape/endpoint.rb
dblock: I think the validator object belongs to the endpoint, which is never GCed, an endpoint is a class thing. So yes, you need to find a way to pass this along, maybe via ENV?

rodzyn: I just pushed some new ideas below

dblock:
I think it's a lot better, but I am worried about the backward incompatible change. I am also not in love with the fact that we check whether the proc takes an argument or not and we have to check for it.

What context to validators execute in? Maybe we can expose `request` explicitly in the context of the evaluation, so basically add a `request` method to `Validator::Base`?

rodzyn:
I see this change more like refactoring. Also, what I understand from the code and documentation, we are rather exposing validate_param! method for any custom validators, that's why I didn't care about backward compatibility.
To be sure for the future commits - what's the policy about backward compatibility? I assumed that before 1.0 release it's not the problem to change some methods.

Everything regarding creation of the validators are done before the request (during rackup). Then, created validators are executed from the endpoint where I can pass request object instead of just params. So the env is in the same place where request, outside the validator object. I'm not sure if I got the point but you can paste some even pseudo-code example if it's not clear.

dblock:
I'll try to be clearer. We *can* break backwards compatibility but we would rather not when possible. That's more of a general statement. I think the problem is that validators take `params` and we want to change that. The question is what do we want to change that into, right?

I think we want to change it into something where someone writing a validator can just call `request` anywhere inside the validator. So this should become a method on `Validations::Base`. I am happy to take a stab at it, but it might take a bit of time :)

dblock:
Bump @rodzyn, what should we do about this PR?

rodzyn:
@dblock as I remember it's breaking backward compatibility (passing request instead of params) and it was quite tough to do it other way around so probably can close it for now

dblock:
OK, I'll close this.

Issue #723, opened 2014-08-18
'optional' Array with nested 'requires' params 


ipkes:
Grape v 0.8.0 <br/>
<b>Headers:</b>
Content-Type: application/json 
Accept: \*/\*
Accept-Encoding: gzip,deflate,sdch
Accept-Language: en-US,en;q=0.8,ru;q=0.6 <br/>
<b>Validation</b>:
<pre>
params do
  requires :docs do
    requires :type, type: String
    optional :files do 
      requires :href, type: String
    end
  end
end
</pre>
<b>Package json(with xml same problem)</b>:
<pre>
{
  "docs": [
       {"type": "aaa"}, 
       {"type": "bbb", "files": [{"href": "some_href"}] }
  ]
}
</pre>
<b>Result</b>:
<pre>
&lt;error&gt;
  &lt;message&gt;docs[files][href] is missing&lt;/message&gt;
&lt;/error&gt;
</pre>
This
<pre>
{
  "docs": [
      {"type": "aaa"}
  ]
}
</pre>
and this:
<pre>
{
  "docs": [
     {"type": "bbb", "files": [{"href": "som_href"}] }
  ]
}
</pre>
Works fine!
Help plz! What I'm doing wrong? Sorry if subject duplicate.

justfalter:
I think this needs to be re-opened. Optional Array and Hash groups do not validate properly when the required parameter is missing. See pull request #883 for specs.

dblock:
I think you're right @justfalter, reopening.

NeMO84:
Any update on this?

dblock:
Nobody has submitted a fix, please contribute.

pocman:
Hi,

I have a similar issue : 
```
requires :route, type: Array, allow_blank: false do
    optional :calendar, type: Array, allow_blank: false do
        requires :starts_at, type: Time, allow_blank: false
        requires :ends_at, type: Time, allow_blank: false
    end
end
```
`"route": [{},{}]` is valid.
`"route":[{},{"calendar": [{"starts_at": "2016-05-02T09:00:00+00:00","ends_at":"2016-05-02T12:00:00+00:00"}]}]` returns 
```
{
  "error": "route[0][calendar][0][starts_at] is missing, route[0][calendar][0][ends_at] is missing"
}
```

ipkes:
[...]

PR #883, opened 2015-01-07
Add failing specs that demonstrate issue with optional groups that have required parameters.

justfalter:
I believe this demonstrates issue #723.


dm1try:
Fixed in the master branch. I checked using your failing specs. Thanks!

PR #885, opened 2015-01-08
Allows values to be a kind of the coerced type in range

u2:
My code like this ```requires :month, type: Integer, values: 1..12``, the Range values coercion is workable.Before it workable, now it failed.

https://github.com/u2/grape/commit/0df3bda5d288d03d1647170ab2cbcb645001a324#diff-45b62b4ed0aed7ceff346c347f76570e

After ```!(@values.is_a?(Proc) ? @values.call : @values).include?(params[attr_name])``` chang to ```(param_array - values).empty?```, the Range values coercion raise a error ```no implicit conversion of Range into Array```

### DIFF
--- a/lib/grape/validations/validators/values.rb
+++ b/lib/grape/validations/validators/values.rb
@@ -10,6 +10,7 @@ module Grape
         return unless params[attr_name] || required_for_root_scope?
 
         values = @values.is_a?(Proc) ? @values.call : @values
+        values = values.to_a if values.is_a? Range
         param_array = params[attr_name].nil? ? [nil] : Array.wrap(params[attr_name])
         unless (param_array - values).empty?
           fail Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message_key: :values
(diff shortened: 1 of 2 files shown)


dblock:
I think we should just support `Range` explicitly with code that looks at `values` and says "are you a range, if so do this ...". As you pointed out having a big range would make a side effect of converting it to an array every time, which can't be good ;)

This will need a CHANGELOG entry and a note in the README that `Range` is explicitly supported as a type.

dblock:
This is https://github.com/intridea/grape/issues/881.

u2:
I think it's easy check a value in a Range or not.But it's not easy check the array params in a Range or not, unless converting it to an array every time.

dblock:
Can we make this lazier though? At least half of the use-case is not an Array/Range scenario.

ajvondrak:
BTW, I think this can be closed since https://github.com/intridea/grape/pull/899 was merged.

u2:
Thanks.

dblock:
Thanks, closing.

Issue #722, opened 2014-08-16
Group param with a required param defaults to Array


jdp:
I'm trying to use a required param inside of a group, like this:

``` ruby
params do
  group :a do
    requires :b
  end
end
```

As an aside, my actual use case looks more like:

``` ruby
params do
  group :a do
    requires :b
  end
  group :c do
    requires :d
  end
  exactly_one_of :a, :c
end
```

But if I send a request that should satisfy that first example, like one of these:

``` shell
$ curl "localhost:9292/case" -d "a[b]=c"
$ curl "localhost:9292/case" -d '{"a":{"b":"c"}}' -H "Content-Type: application/json"
```

I get a response saying `a` is invalid:

``` json
{"error":"a is invalid"}
```

I made [an example here](https://gist.github.com/jdp/cb76505f83dadd90609e) that reproduces the bug. I'm using grape version 0.8.0.

jrichter1:
I'm going to give it a shot to make it raise an error if no type is specified. I'll issue a PR when done and we'll see what comes out of it.

dnesteryuk:
The solution was implemented in the attached PR.

PR #886, opened 2015-01-10
Make the type in group of required params required (#722)

jrichter1:
Made it so groups with required params will demand a type to be set to Array or Hash, errors otherwise (made a new exception for that).
Covered with some tests, also had to modify several existing tests to support this.

### DIFF
--- a/lib/grape/dsl/parameters.rb
+++ b/lib/grape/dsl/parameters.rb
@@ -38,6 +38,13 @@ module Grape
         orig_attrs = attrs.clone
 
         opts = attrs.last.is_a?(Hash) ? attrs.pop : {}
+        type = opts[:type]
+
+        # check type for optional parameter group
+        if attrs && block_given?
+          fail Grape::Exceptions::MissingGroupTypeError.new if type.nil?
+          fail Grape::Exceptions::UnsupportedGroupTypeError.new unless [Array, Hash].include?(type)
+        end
 
         validate_attributes(attrs, opts, &block)
 
--- a/lib/grape/locale/en.yml
+++ b/lib/grape/locale/en.yml
@@ -33,4 +33,6 @@ en:
         at_least_one: 'are missing, at least one parameter must be provided'
         exactly_one: 'are missing, exactly one parameter must be provided'
         all_or_none: 'provide all or none of parameters'
+        missing_group_type: 'group type is required'
+        unsupported_group_type: 'group type must be Array or Hash'
 
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -2166,11 +2166,11 @@ describe Grape::API do
     it 'groups nested params and prevents overwriting of params with same name in different groups' do
       subject.desc 'method'
       subject.params do
-        group :group1 do
+        group :group1, type: Array do
           optional :param1, desc: 'group1 param1 desc'
           requires :param2, desc: 'group1 param2 desc'
         end
-        group :group2 do
+        group :group2, type: Array do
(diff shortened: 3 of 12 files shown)


lib/grape/locale/en.yml
```diff
@@ -33,4 +33,5 @@ en:
         at_least_one: 'are missing, at least one parameter must be provided'
         exactly_one: 'are missing, exactly one parameter must be provided'
         all_or_none: 'provide all or none of parameters'
+        invalid_group_type: 'group of required parameters must have a specified type of Array or Hash, type provided: %{group_type}'
```
dblock: I don't think we want to have the type provided in the message, typically error messages should be fixed, and potentially additional data should indicate the source. I would make this just  `group of required parameters must be Array or Hash`.

spec/grape/api_spec.rb
```diff
@@ -2166,11 +2166,11 @@ def static
     it 'groups nested params and prevents overwriting of params with same name in different groups' do
       subject.desc 'method'
       subject.params do
-        group :group1 do
+        group :group1, type: Array do
```
dblock: Why aren't we just defaulting to `Hash`?

dblock:
Can you please elaborate why this change for required and *not* optional params? I don't see the difference. 

This needs CHANGELOG, either way..

spec/grape/api_spec.rb
jrichter1: I was acting your comment on the issue not to change the default, if I got that correctly, that is :)

jrichter1:
There is no specific reasoning behind this, other than what I understood from the issue. It seemed to me that it is specifically targeted at required param groups. 
If you tell me to change it for optional as well, then I shall do so.

dblock:
Lets step back (and I re-read everything, I realize I totally contradicted myself :)): #722 says that we want an error when the type is not provided so that people aren't confused. Try to do that, but I do think it should be for any group, not just a `required` group. I think it should raise a clear `MissingGroupTypeError` (feel free to suggest a different name) in the case the type is missing and it should raise an explicit `UnsupportedGroupTypeError` in the case the type is not `Hash` or `Array`. What do you think?

jrichter1:
Sounds good to me, I'll get to it then.

jrichter1:
So I finally got to take a stab at it, what do you think?

lib/grape/dsl/parameters.rb
```diff
@@ -38,7 +38,15 @@ def optional(*attrs, &block)
 
         validations = {}
         validations.merge!(attrs.pop) if attrs.last.is_a?(Hash)
+        type = validations[:type]
         validations[:type] ||= Array if block_given?
+
+        # check type for optional parameter group
+        if attrs && block_given?
+          fail Grape::Exceptions::MissingGroupTypeError.new if type.nil?
+          fail Grape::Exceptions::UnsupportedGroupTypeError.new if type != Array && type != Hash
```
dblock: Maybe more idiomatic to say `[Hash, Array].include?(type)`, more future-proof.

dblock:
I like it, squash the commits, see my minor comment and lets merge this.

jrichter1:
Done and done. Cheers!

dblock:
I forgot something, this needs a section in UPGRADING, please. Sorry about that.

jrichter1:
Should be updated now

dblock:
Merged.

DavidHooper:
The README.md needs updating to reflect this change.

dblock:
Right, @jrichter1 could you please do that?

PR #894, opened 2015-01-14
Bug/893 fix globals middleware and example logger middleware

### LINKED ISSUE #893: Very difficult to access params for logging
The Globals middleware appears to be intended to alleviate this, but it is broken with no specs.

pboling:
Fixes #893

### DIFF
--- a/lib/grape/middleware/globals.rb
+++ b/lib/grape/middleware/globals.rb
@@ -4,9 +4,9 @@ module Grape
   module Middleware
     class Globals < Base
       def before
-        @env['grape.request'] = Grape::Request.new(@env)
+        request = Grape::Request.new(@env)
         @env['grape.request.headers'] = request.headers
-        @env['grape.request.params'] = request.params if @env['rack.input']
+        @env['grape.request.params'] = request.params
       end
     end
   end
--- a/lib/grape/middleware/globals.rb
+++ b/lib/grape/middleware/globals.rb
@@ -1,8 +1,17 @@
+# This Middleware is not loaded by grape by default.
+# If you intend to use it you must first:
+#   require 'grape/middleware/globals'
+# See the spec for examples.
+#
 require 'grape/middleware/base'
 
 module Grape
   module Middleware
     class Globals < Base
+      # NOTE: If you have Grape mounted as a Rack endpoint in a Rails stack action dispatch may have moved the params
+      # If your params are not showing up,
+      #   then you may need to override this method and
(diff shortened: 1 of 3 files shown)


dblock:
Generally, how to do logging with Grape comes up often, so thanks for looking into this.

Regarding this specific implementation, today the Globals middleware assigns `env['grape.request']`, which can give you `env['grape.request'].params` and the rest, no? So could you have written the logger you suggest here with just that? That's just for my understanding.

More importantly, if that middleware is not included today, I think we should delete it rather than extend it and move the suggested logger into a new gem called grape-logger-middleware or something like that. Would you like to do that?

kuraga:
@dblock +1. But then, should we remove `@env['grape.request.params']`, too? What's the difference between headers and params?

dblock:
If the middleware is not currently used, then it doesn't matter how it changes (especially if we're going to delete it).

pboling:
Yeah, I'd like to do that.  Currently the Globals middleware doesn't work at all, as written it will raise an exception every time because the `request` local is [never set before it is used](https://github.com/intridea/grape/blob/master/lib/grape/middleware/globals.rb#L8)

So, yes, I will make this into a gem, but I would like to not have to set grape.request in the gem, since then any other gem would also be doing the same thing, and we all be up in each other's business.

I propose:

* Globals becomes an **optional** standard (loaded only when requested) in grape.
* Grape gem extensions will load it the middleware, and then utilize the preset 'env['grape.request']`, modifying and preserving it for each middleware in the stack.

Questions:

* Are there any grape gem extensions doing similar things now?
* How would the gem extension both to insert itself into the stack, and make sure it is inserted *after* the Globals middleware?

dblock:
A while ago we got rid of preserving `request` across middleware and others. I think the best course of action is not to re-introduce Globals, but to implement `request` as `Grape::Request.new(env)` whenever you need it. It comes with params and everything else you may need and keeps middleware independent, which is how it's designed for.

I don't know about other grape middleware's out there, but I am sure they exist.

I think the consumer (developer) should insert the middleware manually and it should just be part of instructions.

pboling:
Makes sense.  If I have time I will send a separate PR that removes Globals.

I will create a gem for a logger.  Would you consider a section in the readme for listing grape extensions?

dblock:
We currently have this in the wiki: https://github.com/intridea/grape/wiki, people can add their own.

pboling:
@dblock :+1:

dblock:
If you don't think you'll get to the Globals removal, please open an issue? Appreciate it.

PR #895, opened 2015-01-15
Add two specs for missing arrays and hashes

adamcooper:
A value of nil was chosen in the case of optional hashes with required
keys.  A value of nil should bypass the optional requirement whereas a
default value of hash or array might trigger the nested required flag.

The corresponding issue is here: https://github.com/intridea/grape/issues/892


dblock:
See @dm1try's comment in https://github.com/intridea/grape/issues/892#issuecomment-74437376, is this still legit?

u2:
See #994.It can be closed, I think.

dblock:
Closing, @adamcooper please confirm.

PR #896, opened 2015-01-15
Updating README.md section of 'Stubbing Helpers' to use up to date RSpec...

xmjw:
The Read Me for this was using the old RSpec .stub method, which was a little confusing, I think we should have the latest approach (allow... to... receive) in the documentation.


dblock:
This is good, thanks!

Issue #881, opened 2015-01-06
values validator can no longer accept a range


smathieu:
We're currently migration from grape `0.9.0` to `0.10.1`. This validator that use to work now stopped working:

```ruby
    params do
      requires :score, type: Integer, values: 0..10, desc: "..."
    end
```

relevant stacktrace:

```
     TypeError:
       no implicit conversion of Range into Array

     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/validations/validators/values.rb:14:in `validate_param!'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/validations/validators/base.rb:17:in `block in validate!'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/validations/attributes_iterator.rb:15:in `block (2 levels) in each'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/validations/attributes_iterator.rb:14:in `each'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/validations/attributes_iterator.rb:14:in `block in each'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/validations/attributes_iterator.rb:13:in `each'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/validations/attributes_iterator.rb:13:in `each'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/validations/validators/base.rb:15:in `validate!'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/endpoint.rb:235:in `block in run'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/endpoint.rb:233:in `each'
     # /Users/simon/.rvm/gems/ruby-2.2.0@rainforest/gems/grape-0.10.1/lib/grape/endpoint.rb:233:in `run'
[...]

PR #899, opened 2015-01-16
Fix for #881

ajvondrak:
I added some regression tests and avoid `to_a`-ing the `Range` at all. What do you think?


dblock:
This is good. You have to look at CI failures on those other platforms (we do support them :)) and squash the commits please.

ajvondrak:
Woops, hadn't noticed the build failing. :) Ruby 1.9.3 doesn't have a `Range#size` method, as it turns out.

Anyway, fixed (assuming Travis says so) & squashed.

dblock:
Thanks, merging.

u2:
:+1:

PR #901, opened 2015-01-17
add support for scoped callbcaks within a version block

kushkella:
Please refer https://github.com/intridea/grape/issues/898. I haven't really understood in depth how Grape works internally, but I think this should solve the problem. I would really appreciate if you could give some feedback. 

I haven't documented this PR since I'm not sure if this is the correct fix. After I receive some feedback, I'll document this PR. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Next Release
 * [#876](https://github.com/intridea/grape/pull/876): Call to `declared(params)` now returns a `Hashie::Mash` - [@rodzyn](https://github.com/rodzyn).
 * [#879](https://github.com/intridea/grape/pull/879): The `route_info` value is no longer included in `params` Hash - [@rodzyn](https://github.com/rodzyn).
 * [#881](https://github.com/intridea/grape/issues/881): Fixed `Grape::Validations::ValuesValidator` support for `Range` type - [@ajvondrak](https://github.com/ajvondrak).
+* [#901](https://github.com/intridea/grape/pull/901): Fix: callbacks defined in a version block are only called for the routes defined in that block - [@kushkella](https://github.com/kushkella).
 * Your contribution here.
 
 0.10.1 (12/28/2014)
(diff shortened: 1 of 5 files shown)


dblock:
I think this is the perfect fix. Add CHANGELOG, update README and UPGRADING (explicitly call this a regression and that the behavior is fixed in the next version).

kushkella:
I've documented everything as you mentioned. It would be great if you could merge this and release a new version of this gem if possible. Thanks once again!

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Next Release
 * [#559](https://github.com/intridea/grape/issues/559): Support Rack 1.6.0 to parse requests larger than 128KB - [@myitcv](https://github.com/myitcv).
 * [#876](https://github.com/intridea/grape/pull/876): declared(params) now returning Hashie::Mash - [@rodzyn](https://github.com/rodzyn).
 * [#879](https://github.com/intridea/grape/pull/879): route_info is not included in params Hash anymore -  [@rodzyn](https://github.com/rodzyn).
+* [#901](https://github.com/intridea/grape/pull/901): callbacks defined in a version block are only called for the routes defined in that block -  [@kushkella](https://github.com/kushkella).
```
dblock: Start with `Fix: callbacks ...` and fix the extra space after the -, please.

README.md
```diff
@@ -2143,6 +2143,39 @@ GET /123        # 'Fixnum'
 GET /foo        # 400 error - 'blah is invalid'
 ```
 
+When a callback is defined within a version block, it is only called for the routes defined in that block.
```
dblock: it's

spec/shared/versioning_examples.rb
```diff
@@ -107,6 +107,38 @@
     end
   end
 
+  context 'with before block defined within a version block' do
+    it 'should only call before block that is defined within the version block' do
```
dblock: Change "should ..." to "calls ..."

dblock:
Made minor comments. Can you please update with those and squash the commits. Thank you.

kushkella:
I've made the changes, squashed the commits and rebased on master. Thank you.

dblock:
Merged, thank you.

Issue #904, opened 2015-01-20
Errors caused by invalid body content are not processed by rescue handlers


croeck:
Hi all,

I ran into a more or less severe issue today. I am running grape directly behind rack (**NO rails**) and found an issue that has also been described in issue #389, which however was run behind rails.

Here is the problem:
If I fire a request against my API with the body set to an random non-JSON value, for instance simply `test`, the `MultiJson` parser returns an `MultiJson::ParserError` error with the message `795: unexpected token at 'test'` that occurs in the [Grape::Middleware::Formatter](https://github.com/intridea/grape/blob/567e779e19109de2e5732dd4d12d5713aa6b2edc/lib/grape/middleware/formatter.rb#L85) and is then thrown as `400` response message.

What I would expect is that my `rescue_all` handler gets notified of the error and can handle the situation. Instead, [Grape::Middleware::Error.call](https://github.com/intridea/grape/blob/567e779e19109de2e5732dd4d12d5713aa6b2edc/lib/grape/middleware/error.rb#L26) catches the thrown error, which then gets processed in the [Grape::ErrorFormatter::Base](https://github.com/intridea/grape/blob/567e779e19109de2e5732dd4d12d5713aa6b2edc/lib/grape/error_formatter/base.rb#L39).

In there, two lines (actual comments) perfectly explain the next issue:

     # env['api.endpoint'].route does not work when the error occurs within a middleware
     # the Endpoint does not have a valid env at this moment

Ok, so there the call crashes again and raises a `NoMethodError`. This error then finally gets processed by my rescue_all handler. Nevertheless I receive a response with the status `500` instead of the actual expected `400`.

Even if the exception in the [Grape::ErrorFormatter::Base](https://github.com/intridea/grape/blob/567e779e19109de2e5732dd4d12d5713aa6b2edc/lib/grape/error_formatter/base.rb#L39#L40) would be prevented, the call would not reach the rescue handlers.

In my opinion the rescue handlers should get a chance to handle the error since it is directly related to the request body.

If a have some spare time I'll try to write a spec for the issue in the next couple of days.

croeck:
Ok, I wrote a small spec and found out that the issue is related to the content type definition that is sent by the user. When specifying `application/json`, the request fails. But when no content type is specified, the request is processed and fails (as expected) because of the missing param.

dblock:
It's definitely a problem, would love it to see this fixed.

croeck:
[...]

PR #906, opened 2015-01-21
Invalid body parse error not rescued by handlers

croeck:
Please see issue #904 for a more detailed description. The spec with content_type `application/json` fails and is not processed by `rescue_from :all`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Next Release
 * [#901](https://github.com/intridea/grape/pull/901): Fix: callbacks defined in a version block are only called for the routes defined in that block - [@kushkella](https://github.com/kushkella).
 * [#886](https://github.com/intridea/grape/pull/886): Group of parameters made to require an explicit type of Hash or Array - [@jrichter1](https://github.com/jrichter1).
 * [#912](https://github.com/intridea/grape/pull/912): Extended the `:using` feature for param documentation to `optional` fields - [@croeck](https://github.com/croeck).
+* [#906](https://github.com/intridea/grape/pull/906): Fix: invalid body parse errors are not rescued by handlers - [@croeck](https://github.com/croeck).
 * Your contribution here.
 
 0.10.1 (12/28/2014)
--- a/lib/grape/locale/en.yml
+++ b/lib/grape/locale/en.yml
@@ -35,4 +35,10 @@ en:
         all_or_none: 'provide all or none of parameters'
         missing_group_type: 'group type is required'
         unsupported_group_type: 'group type must be Array or Hash'
+        invalid_message_body:
+          problem: "message body does not match declared format"
+          resolution:
+            "when specifying %{body_format} as content-type, you must pass valid
+            %{body_format} in the request's 'body'
+            "
 
--- a/lib/grape/middleware/formatter.rb
+++ b/lib/grape/middleware/formatter.rb
@@ -81,6 +81,8 @@ module Grape
                 end
                 env['rack.request.form_input'] = env['rack.input']
               end
+            rescue Grape::Exceptions::Base => e
+              raise e
             rescue StandardError => e
               throw :error, status: 400, message: e.message
             end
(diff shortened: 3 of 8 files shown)


lib/grape/middleware/formatter.rb
```diff
@@ -81,6 +81,9 @@ def read_rack_input(body)
                 end
                 env['rack.request.form_input'] = env['rack.input']
               end
+            rescue MultiJson::ParseError
```
dblock: This is specific to the JSON parser, so it should be in it, not here. Then you can raise a `Grape::Exception::InvalidMessageBody` error and re-rescue and just re-raise anything that is `Grape::Exception::Base` here. 

You might want to look at other parsers while you're at it, maybe the XML one has a similar problem and can be fixed in the same way?

lib/grape/locale/en.yml
```diff
@@ -35,4 +35,10 @@ en:
         all_or_none: 'provide all or none of parameters'
         missing_group_type: 'group type is required'
         unsupported_group_type: 'group type must be Array or Hash'
+        invalid_message_body:
+          problem: "message body does not match declared format '%{body_format}'"
```
dblock: I think this one should be without any values, so "message body does not match declared format" without the `body_format` value, then the resolution can contain that. It makes message expectations more consistent. I don't feel strongly about it though, so feel free to leave it too.

spec/grape/exceptions/body_parse_errors_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      subject.params do
+        requires :beer
+      end
+      subject.post '/beer' do
+        'beer received'
+      end
+    }
+
+    # non-JSON body
+    body = 'test'
```
dblock: This declares something in a scope that's not very clear, since who knows what `context` is :) I would just hard-code `test` below in the two places you need it and not worry about it. Same for headers.

Otherwise use `let(:body) { ...}`, which is the way RSpec tells us to declare things like this.

dblock:
This is good work, see my comments.

lib/grape/locale/en.yml
croeck: Makes sense to me :)

croeck:
Found (and fixed) the same issue to appear when using the XML parser.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Next Release
 * [#901](https://github.com/intridea/grape/pull/901): Fix: callbacks defined in a version block are only called for the routes defined in that block - [@kushkella](https://github.com/kushkella).
 * [#886](https://github.com/intridea/grape/pull/886): Group of parameters made to require an explicit type of Hash or Array - [@jrichter1](https://github.com/jrichter1).
 * [#912](https://github.com/intridea/grape/pull/912): Extended the `:using` feature for param documentation to `optional` fields - [@croeck](https://github.com/croeck).
+* [#906](https://github.com/intridea/grape/pull/906): Fix: invalid body parse errors are not rescued by handlers [@croeck](https://github.com/croeck).
```
dblock: Nitpick: this is missing a dash. before your github name.

lib/grape/middleware/formatter.rb
```diff
@@ -82,6 +82,7 @@ def read_rack_input(body)
                 env['rack.request.form_input'] = env['rack.input']
               end
             rescue StandardError => e
+              raise e if e.is_a? Grape::Exceptions::Base
```
dblock: I think the more idiomatic way (and technically faster) would be to:

```ruby
rescue Grape::Exception::Base
    # ignore
rescue StandardArray => e
   ...
```

dblock:
See my minor comments. Will merge after, great work.

lib/grape/middleware/formatter.rb
croeck: Shouldn't it be:

    rescue Grape::Exception::Base => e
        raise e
    rescue StandardArray => e
        ....

If we don't raise the error originating from the parser it can never be rescued and the follwing calls are likely to experience other errors?

dblock: Totally. My bad.

croeck:
Ok, fixed your comments and the build just passed :)

dblock:
Merged, thanks.

PR #907, opened 2015-01-22
Add the ability to default a path version

ajvondrak:
In support of porting my legacy application, I needed a path-versioned API. Basically, I was doing something similar to [this blog post](http://code.dblock.org/2013/07/19/evolving-apis-using-grape-api-versioning.html).

Problem is, the legacy API also responded *without* a version in the path. So, we supported `/v1/foo` and `/v2/foo`, but *also* `/foo`, which behaved just like `/v2/foo`. I needed a way to "default" the version in a URL path, which is kind of tricky because it affects Route definitions. Instead of working around with something kludgey (e.g., looked into mounting the same "core logic" in multiple places, but [that has its own problems](https://github.com/intridea/grape/issues/570)), I poked around and figured out how to implement this feature.

Let me know how it looks, or if I'm barking up the wrong tree with a feature like this.

P.S.: Not sure what CHANGELOG entry to make until a pull request is set up so I can link it from the CHANGELOG. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,30 @@ Using this versioning strategy, clients should pass the desired version in the U
 
     curl -H http://localhost:9292/v1/statuses/public_timeline
 
+You may optionally specify the `:default` option for a path version, which will make the version in the URL optional. For example, the following API will respond to either the URL `/v1/foo` or just `/foo`. In either case, the value of `env['api.version']` will be `'v1'`.
+
+```ruby
+class API < Grape::API
+  version 'v1', using: :path, default: 'v1'
+
+  get :foo do
+    # ...
+  end
+end
+```
+
+This is also useful for "evolving" your APIs (see [this blog post](http://code.dblock.org/2013/07/19/evolving-apis-using-grape-api-versioning.html)). For example, the following API responds to `/v1/foo` or `/v2/foo`, but `/foo` will behave like `/v2/foo`:
+
+```ruby
+class API < Grape::API
+  version 'v1', 'v2', default: 'v2'
+
+  get :foo do
+    # ...
+  end
+end
+```
+
 ### Header
 
 ```ruby
--- a/lib/grape/middleware/versioner/path.rb
+++ b/lib/grape/middleware/versioner/path.rb
@@ -31,13 +31,13 @@ module Grape
             path = Rack::Mount::Utils.normalize_path(path)
           end
 
+          env['api.version'] = default_version if default_version
+
           pieces = path.split('/')
           potential_version = pieces[1]
           if potential_version =~ options[:pattern]
             if unrecognized_version?(potential_version)
-              if default_version
-                env['api.version'] = default_version
-              else
+              unless default_version
                 throw :error, status: 404, message: '404 API Version Not Found'
               end
             else
--- a/lib/grape/middleware/versioner/path.rb
+++ b/lib/grape/middleware/versioner/path.rb
@@ -34,10 +34,15 @@ module Grape
           pieces = path.split('/')
           potential_version = pieces[1]
           if potential_version =~ options[:pattern]
-            if options[:versions] && !options[:versions].find { |v| v.to_s == potential_version }
-              throw :error, status: 404, message: '404 API Version Not Found'
+            if unrecognized_version?(potential_version)
+              if default_version
+                env['api.version'] = default_version
+              else
+                throw :error, status: 404, message: '404 API Version Not Found'
+              end
+            else
+              env['api.version'] = potential_version
             end
-            env['api.version'] = potential_version
           end
         end
 
@@ -45,6 +50,16 @@ module Grape
 
         def prefix
           Rack::Mount::Utils.normalize_path(options[:prefix].to_s) if options[:prefix]
+        end
+
+        def default_version
+          options[:version_options] && options[:version_options][:default]
+        end
+
+        def unrecognized_version?(potential_version)
+          options[:versions] && !options[:versions].find do |v|
+            v.to_s == potential_version
+          end
         end
       end
     end
--- a/lib/grape/path.rb
+++ b/lib/grape/path.rb
@@ -28,6 +28,10 @@ module Grape
       !!(settings[:version] && settings[:version_options][:using] == :path)
     end
 
+    def optional_path_versioning?
+      uses_path_versioning? && !!settings[:version_options][:default]
+    end
+
     def has_namespace?
       namespace && namespace.to_s =~ /^\S/ && namespace != '/'
     end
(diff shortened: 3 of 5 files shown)


dblock:
The code is clean, but I feel like when you default to a version you shouldn't be responding to both paths, but redirecting to the default with a 301 or 302. The whole idea of responding to two different paths with the same data is the beginning of something bad (eg. caching). Did you try that or chose to deliberately not do it? Would you do it for a new API this way?

Now, from experience and what is also a matter of preference, I would not roll path versioning out. You already have `/api` and you're actually in the best place possible IMO. I would make header-based v2 so that you continue responding with v1 on /api and if the right header is specified, serve v2. Path versioning is something I regret for every single API I had written :)

ajvondrak:
Good feedback; I'd like to avoid whatever hacks I can. But I think my API is unfortunately a little "special".

To give you some background, it's a real-time ad auction API. Basically, the endpoints boil down to some variant of `get :gimme_an_ad`, which will run an auction with constantly-changing inventory and present the winner. This means a couple things:

1. We expressly do **not** want to cache. We literally do a `before { header 'Cache-Control', 'private, max-age=0, no-cache' }`. The GET endpoints are side-effecting because they trigger auctions which have different results each time (not really REST, but it's what we have to work with...). So if path versioning "breaks" caching, so much the better in our case!

2. We're very sensitive to latency. A 301 or 302 might be semantically correct, but we'd really rather avoid the extra HTTP round-trip.

You're right, of course. If I were starting over clean-slate, I'd probably version using headers or params, because those are easier to default (i.e., using the cascading technique in your blog post or using a default param value). But yet *another* consideration I have is legacy: we have some customers hitting `/v1/**` URLs, some hitting `/v2/**`, some hitting `/**`. Getting them all to switch URLs turns out to be like pulling teeth (and don't get me started on having them switch their GET params!).

(Speaking of regrets, I regret our specific API even has versions, because v1 & v2 are completely compatible with each other, so it was strictly unnecessary to split them up in the first place. But nothing to do about it now.)

As such, if I have to work with path versioning, I at least wanted to be able to default/cascade them like header or param versions.

If you feel like that's too much rope for users to hang themselves with, one way I thought about working around this was making the path param optional by hand, like

```ruby
class API < Grape::API
  get '/(:path_version/)gimme_an_ad', requirements: { path_version: /v1|v2/ } do
    env['api.version'] = params[:path_version] || 'v2'
    # ...
  end

  get '/(:path_version/)gimme_another_ad', requirements: { path_version: /v1|v2/ } do
    env['api.version'] = params[:path_version] || 'v2'
    # ...
  end
end
```

I'd of course prefer to have a simple `version 'v1', 'v2', default: 'v2'` declaration. Or some other alternative, if you've got one. :)

ajvondrak:
BTW, I was looking through the docs and noticed https://github.com/intridea/grape#http-status-code was out of place - stuck in the middle of the versioning section.

dblock:
I think the way this is implemented here is not generic enough, `default` does something I don't expect in a more "classic" API. Maybe you can refactor the path versioner so that you can derive from it and substitute the API specifics that you're trying to accomplish? Or monkey patch a very small section of it?

Please do PR a doc fix :)

ajvondrak:
> I think the way this is implemented here is not generic enough, `default` does something I don't expect in a more "classic" API.

Could you clarify what you mean by this? Like, `default` should be an option for header/param/accept-header versions, too?

> Maybe you can refactor the path versioner so that you can derive from it and substitute the API specifics that you're trying to accomplish? Or monkey patch a very small section of it?

Well, I already had to modify part of the path versioner (as you can see). I guess it's entirely feasible to make a new versioner middleware and declare the version like `version :v1, :v2, using: :optional_path` or something.

However, modifying/replacing that middleware alone is not enough. The bigger problem is that `version :v1, :v2, using: :path` will cause the endpoints to compile with routes like `'/:version/foo'`. Ultimately, this string gets compiled into a regexp using `Rack::Mount::Strexp` [here](https://github.com/intridea/grape/blob/567e779e19109de2e5732dd4d12d5713aa6b2edc/lib/grape/endpoint.rb#L180), which gets added to the `Rack::Mount::RouteSet`. Without explicitly making the `:version/` optional (by wrapping it in parens), the regexp turns out like `/\A\/(?<version>v1\|v2)\/foo\Z/`, which won't successfully match `/foo` alone.

So AFAICT (and from my own testing), we'll never even *reach* the path-versioner middleware from `GET /foo` if `Grape::Path` doesn't do some work to make the path version optional - Rack rejects it out of the gate. Which is why that change is in there.

dblock:
I mean that I would expect `default: true` to do a redirect, otherwise this is not in the spirit of HTTP + API, I wouldn't expect anyone else wanting the API to respond on both routes just by adding a switch. 

I think you should pursue a refactor that would minimize the hacking you need to do on your side. A lot of this should be extracted in strategy-like behavior in Grape, then you can plug the custom strategy you want.

Makes sense?

ajvondrak:
So if I get you right...you mean maybe something akin to the versioner middlewares, but for `Grape::Path` instead? So we can make "plugins" for the logic that's currently in `Grape::Path`, like:
* prefixing
* mounting under new paths
* namespaces
* format suffixes
* path versions
* **(new)** optional path versions (maybe not in the official code base, I could just add it for my project using the extension mechanism)

(This would probably still entail some tweaks to the versioner middleware for my use case, but I guess as long as it's possible...)

Do I understand you right? Sounds like more work than this initial fix for my problem (which isn't much code). :) But maybe more "correct", I guess?

dblock:
Yes, right on.

ajvondrak:
I'll see what I can do...might take me a little while.

Devil's advocate: at what point do we just dump all that and go the Sinatra route of allowing regexps & custom route matcher classes? I guess the largest difference being that Grape's DSL influences all sorts of things about the route (mounting, prefixing, namespacing, versioning are all exposed via DSL rather than regexp-ing the routes). Just a passing thought.

ajvondrak:
In the meantime, I figured out a quick & dirty way to monkeypatch this for my use case that I thought might be instructive:
```ruby
module OptionalPathVersions
  extend ActiveSupport::Concern

  module ClassMethods
    def version(*versions, default:)
      before_validation do
        env['api.version'] = params.delete(:optional_version) || default
      end
      @versions = versions
    end

    def route(methods, paths = ['/'], route_options = {}, &block)
      paths = Array.wrap(paths).map! do |path|
        path = path.to_s
        if path.empty? || path == '/'
          '/(:optional_version)'
        else
          "/(:optional_version/)#{path.sub(%r{^/+}, '')}"
        end
      end
      route_options[:requirements] ||= {}
      route_options[:requirements][:optional_version] = /#{@versions.join('|')}/
      super(methods, paths, route_options, &block)
    end
  end
end

class API < Grape::API
  include OptionalPathVersions

  version 'v1', 'v2', default: 'v2'

  get :foo do
    "#{version} bar"
  end
end

# GET /foo -> v2 bar
# GET /v1/foo -> v1 bar
# GET /v2/foo -> v2 bar
```

dblock:
I think we always want 1st class DSLs for things, that's the "Grape" way.

I'll close this PR, thanks for taking the time and looking forward to the future progress!

ajvondrak:
> I think we always want 1st class DSLs for things, that's the "Grape" way.

Sounds right to me.

> I'll close this PR, thanks for taking the time and looking forward to the future progress!

Alright. Thanks for all the discussion!

PR #910, opened 2015-01-25
Unify Grape::DSL::Parameters#requires and #optional. Breaking changes maybe

kuraga:


dblock:
I like this. I am worried about this whole `validations[:type] ||= Array if block_given?` business, but lets just trust our tests :)

PR #912, opened 2015-01-28
Added support for `optional` param documentation with `using`

croeck:
Without these changes the `using` does only work for `requires` and not with `optional` parameters. This restriction is currently *not* documented.

The feature / fix is required to provide a more complete DRY approach when using `grape-swagger` and `grape-entity`. With these changes, one can document the possible fields and values, e.g. when using `PATCH` update, without requiring them to be present.


dblock:
Odd that we only did it for `requires`. This looks good, thank you.

swistaczek:
:+1:

PR #913, opened 2015-01-30
Fixed issues related to invalid accept headers

croeck:
Fixed that invalid accept headers are (1) not processed by rescue handlers and (2) result in status 500 (no method) when http_codes are defined. The fix is based on changing the former `throw` to raising a new error, which then can be processed.

**To (1):**  
The actual behavior of cascading the request remains unchanged as the headers are included in the raised error. As long as the error is not rescued it results in the same error messages (404 or 406).
The error message that is generated consists of two parts: The problem `Invalid accept header` and what actually went wrong, eg. `API vendor or version not found`. The cause of the problem was included as the resolution of the problem, as it always describes what the user did wrong:
- Accept header must be set
- Accept header must not contain ranges ("*")
- 406 Not Acceptable <-- not quite perfect, shouldn't it be *No matching media type found* ?
- API vendor or version not found
- *The error message when `Rack::Accept::MediaType.new env['HTTP_ACCEPT']` fails*

**To (2):**  
As commented in the source code, `lib/grape/error_formatter/base.rb` does raise an error if the failure originated from the middleware and not a route. All the `throw` statements that were used before have caused those errors if http_codes were defined in the route description.

Please let me know if there are any issues with these changes.
Cedric

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Next Release
 * [#901](https://github.com/intridea/grape/pull/901): Fix: callbacks defined in a version block are only called for the routes defined in that block - [@kushkella](https://github.com/kushkella).
 * [#886](https://github.com/intridea/grape/pull/886): Group of parameters made to require an explicit type of Hash or Array - [@jrichter1](https://github.com/jrichter1).
 * [#912](https://github.com/intridea/grape/pull/912): Extended the `:using` feature for param documentation to `optional` fields - [@croeck](https://github.com/croeck).
+* [#913](https://github.com/intridea/grape/pull/913): Fix: Invalid accept headers are not processed by rescue handlers - [@croeck](https://github.com/croeck).
+* [#913](https://github.com/intridea/grape/pull/913): Fix: Invalid accept headers cause internal processing errors (500) when http_codes are defined - [@croeck](https://github.com/croeck).
 * Your contribution here.
 
 0.10.1 (12/28/2014)
--- a/lib/grape/error_formatter/base.rb
+++ b/lib/grape/error_formatter/base.rb
@@ -41,7 +41,7 @@ module Grape
           http_codes = env['rack.routing_args'][:route_info].route_http_codes || []
           found_code = http_codes.find do |http_code|
             (http_code[0].to_i == env['api.endpoint'].status) && http_code[2].respond_to?(:represent)
-          end
+          end if env['api.endpoint'].request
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Next Release
 * [#901](https://github.com/intridea/grape/pull/901): Fix: callbacks defined in a version block are only called for the routes defined in that block - [@kushkella](https://github.com/kushkella).
 * [#886](https://github.com/intridea/grape/pull/886): Group of parameters made to require an explicit type of Hash or Array - [@jrichter1](https://github.com/jrichter1).
 * [#912](https://github.com/intridea/grape/pull/912): Extended the `:using` feature for param documentation to `optional` fields - [@croeck](https://github.com/croeck).
+* [#913](https://github.com/intridea/grape/pull/913): Fix: Invalid accept headers are (1) not processed by rescue handlers and (2) result in 500 when http_codes are defines - [@croeck](https://github.com/croeck).
```
dblock: Typo: defineD, right?

You can split this into two separate lines, to avoid the whole (1) (2) thing.

lib/grape/error_formatter/base.rb
```diff
@@ -41,7 +41,7 @@ def present(message, env)
           http_codes = env['rack.routing_args'][:route_info].route_http_codes || []
           found_code = http_codes.find do |http_code|
             (http_code[0].to_i == env['api.endpoint'].status) && http_code[2].respond_to?(:represent)
-          end
+          end unless env['api.endpoint'].request.nil?
```
dblock: `if env['api.endpoint'].request`

lib/grape/middleware/versioner/header.rb
```diff
@@ -86,6 +83,12 @@ def available_media_types
           available_media_types = available_media_types.flatten
         end
 
+        def rack_accept_header
+          Rack::Accept::MediaType.new env['HTTP_ACCEPT']
+          rescue RuntimeError => e
```
dblock: This is mis-aligned I think, the `rescue` is usually at the `def` level to be a function-based rescue.

dblock:
Solid, nice work @croeck. Amend by my nitpicky comments and I'll merge.

croeck:
Changes are now included, thanks for your comments!

dblock:
Merged rebased via https://github.com/intridea/grape/commit/046450e172e8acae0a0982f09aa4765985a0d931. Thank you.

Issue #915, opened 2015-02-06
Allow blank not work when type Datetime


u2:
```
optional :remind_at, type: DateTime, allow_blank: true
```
When the params is `"remind_at"=>""`, grape return a message `to_do[remind_at] is invalid`.

u2:
I am not sure this is justifiable or not.But it's very normal for me to 
```
requires :to_do, type: Hash do
  optional :remind_at, type: DateTime, allow_blank: true
end

to_do.update(params[:to_do])
```
If before to_do have a `remind_at`, then I want to clean it, How I can implement it.

And then what about Integer/Float, and so on?

dblock:
I agree that DateTime, Integer and everything else should support `allow_blank`. Thanks for looking into this.

u2:
Updated.Added `DateTime/Date/Numeric/Boolean` type support `allow_blank`.

u2:
`allow_blank` default value is `true`?I think it should be clear in doc.

dblock:
Yes, default values definitely need to be documented. Please.

dblock:
This feature request was closed by https://github.com/intridea/grape/pull/916. Would welcome any future doc and whatever other updates. Thx.

PR #916, opened 2015-02-06
Fix allow_blank supporting DateTime/Date

u2:
#915

### DIFF
--- a/lib/grape/validations/validators/coerce.rb
+++ b/lib/grape/validations/validators/coerce.rb
@@ -29,9 +29,12 @@ module Grape
         # allow nil, to ignore when a parameter is absent
         return true if val.nil?
         if klass == Virtus::Attribute::Boolean
-          val.is_a?(TrueClass) || val.is_a?(FalseClass)
+          val.is_a?(TrueClass) || val.is_a?(FalseClass) || (val.is_a?(String) && val.empty?)
         elsif klass == Rack::Multipart::UploadedFile
           val.is_a?(Hashie::Mash) && val.key?(:tempfile)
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -7,4 +7,5 @@ group :development, :test do
   gem 'guard'
   gem 'guard-rspec'
   gem 'guard-rubocop'
+  gem 'byebug'
```
kuraga: Remove it, please.

dblock:
Needs CHANGELOG, too, please.

Gemfile
u2: Thank you.I forgot it.

kuraga:
Hm... Is `val == ""` dangerous, isn't?

Can `val` be a non-`String` here?
If can then `"" == val` maybe?
Or `val.empty?` with some additional conditions?

Which values of `val` are possible?

u2:
Before I wrote `val.is_a?(String) && val.empty?`,but I thought it's the same meaning of `val == ""`.
Yes, you are right.This val can be a non-`String`.I will change it asap.

lib/grape/validations/validators/coerce.rb
```diff
@@ -29,9 +29,12 @@ def _valid_single_type?(klass, val)
         # allow nil, to ignore when a parameter is absent
         return true if val.nil?
         if klass == Virtus::Attribute::Boolean
-          val.is_a?(TrueClass) || val.is_a?(FalseClass)
+          val.is_a?(TrueClass) || val.is_a?(FalseClass) || val == ''
```
kuraga: @u2 And here, too?

u2: sorry..

dblock:
Merging, thanks.

PR #918, opened 2015-02-10
Reset endpoint options between calls

yesmeck:
I'm using Grape and Rable with grape-rabl gem, I have following code:
```ruby
class Comments < Grape::API
  resources :comments do
    post do
      comment = Comment.new(params)
      if comment.save
        render rabl: 'comments/show'
      else
        car.errors.to_hash
      end
    end
  end
end
```
When first request save the comment succesfully, whatever the nex request save the comment succesfully or not, the rabl template will alway be rendered. Because  grape-rabl write the template info to the endpoint options (https://github.com/LTe/grape-rabl/blob/master/lib%2Fgrape-rabl%2Frender.rb#L4), so I thinks the endpoint options should be reset between calls.

### DIFF
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -324,5 +324,10 @@ module Grape
     def afters
       namespace_stackable(:afters) || []
     end
+
+    def initialize_copy(other)
+      super
+      self.options = other.options.deep_dup
+    end
   end
 end
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -766,6 +766,19 @@ describe Grape::Endpoint do
     expect(last_response.body).to eq('Hello')
   end
 
+  it 'reset options between calls' do
+    subject.get('/options') do
+      if params[:rabl]
+        env['api.endpoint'].options[:route_options][:rabl] = params[:rabl]
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


dblock:
I am 50/50 on this. I think grape-rabl should be using ENV. Could you open an issue there, please?

Either way this needs a CHANGELOG and the rubocop build to be fixed, please.

yesmeck:
If endpoint isn't supposed to change, I think we can just make the endpoint options read only. 

I have open an issue on grape-rabl.

yesmeck:
It seems grape-rabl is hard to store template info in env directly, because when we write code like `get '/', rabl: 'index'`, the template info is have to be stored in route_options.

dblock:
I think the declarative part should be stored on the endpoint, but the runtime part should not.

yesmeck:
@dblock Yes, I figured it out, I'm composing a new PR to grape-rabl.

yesmeck:
As https://github.com/LTe/grape-rabl/pull/36 merged, should we still freeze the `@options` or make it read only? If the answer is yes, grape will not compatible with old version grape-rabl before 0.3.1.

dblock:
IMO, I think we should keep it as is.

dblock:
Closing this, LMK if you think we should still do something here.

PR #924, opened 2015-02-12
merge options and block

masarakki:
When using [wine_bouncer](https://github.com/antek-drzewiecki/wine_bouncer) gem, 
I have to use `authorizations` options in `desc` method like this:

```ruby
   desc 'protected method with required public scope',
   authorizations: { oauth2: [{ scope: 'public' }] }
   get '/protected' do
      { hello: 'world' }
   end
```

but any options will be ignored if I use `desc` method with block like this:

```ruby
   desc 'protected method with required public scope',
   authorizations: { oauth2: [{ scope: 'public' }] } do
     success Entity::Hello
   end
   get '/protected' do
      { hello: 'world' }
   end
```

So I want to merge options and block.
If you think it is not good, I want some easy way to add DSL method like `authorizations` in `desc` block.
`desc_container` is difficult for hack.

```ruby
   desc 'protected method with required public scope' do
     authorizations oauth2: [{ scope: 'public' }]      # add hash configuration
     oauth2 scope: 'public'                                     # and/or add wrapper like `success`
     success Entity::Hello
   end
   get '/protected' do
      { hello: 'world' }
   end
```

### DIFF
--- a/lib/grape/dsl/configuration.rb
+++ b/lib/grape/dsl/configuration.rb
@@ -28,7 +28,7 @@ module Grape
             end
 
             config_class.configure(&config_block)
-            options = config_class.settings
+            options = options.merge(config_class.settings)
           else
             options = options.merge(description: description)
           end
--- a/spec/grape/dsl/configuration_spec.rb
+++ b/spec/grape/dsl/configuration_spec.rb
@@ -70,6 +70,22 @@ module Grape
           expect(subject.namespace_setting(:description)).to eq(expected_options)
           expect(subject.route_setting(:description)).to eq(expected_options)
         end
+
+        it 'can merge options and block' do
+          expected_options = {
+            description: 'The description',
+            detail: 'more details',
+            params: { first: :param },
+            authorizations: { oauth2: [] }
+          }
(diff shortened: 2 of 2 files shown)


dblock:
I really think we want 1st class DSLs here, that was the goal from refactoring `params`, `desc` and others to take a block. You want to take that out of `desc` though, so something like this:

```ruby
desc 'reticulate splines'
authorizations oauth2: ...
oauth2 scope: :public
```

You can accomplish this with `route_settings` now, there's a decent example in https://github.com/intridea/grape/blob/master/UPGRADING.md:

```ruby
module Grape
  module Extensions
    module SortExtension
      def sort(value)
        route_setting :sort, sort: value
        value
      end
    end

    Grape::API.extend self
  end
end
```

To retrieve this value at runtime from within an API, use `env['api.endpoint'].route_setting(:sort)` and when introspecting a mounted API, use `route.route_settings[:sort]`.

This should let you create a proper grape-wine_bouncer gem that has documented and maintainable integration, what to try that?

dblock:
Closing this without merging, thanks for your suggestion @masarakki.

PR #936, opened 2015-02-22
fix weird behaviour for optional groups with default params

dm1try:
apply defaults for inner params only if parent param is present

fixes #615 (see for more info)

/cc @dblock  take a look

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 0.11.1 (Next)
 =============
 
+#### Fixes
+
+* [#936](https://github.com/intridea/grape/pull/936): Fixed default params processing for optional groups - [@dm1try](https://github.com/dm1try).
+
 * Your contribution here.
 
 0.11.0 (2/23/2015)
--- a/lib/grape/validations/validators/default.rb
+++ b/lib/grape/validations/validators/default.rb
@@ -11,12 +11,12 @@ module Grape
       end
 
       def validate!(params)
+        return unless @scope.should_validate?(params)
+
         attrs = AttributesIterator.new(self, @scope, params)
-        parent_element = @scope.element
         attrs.each do |resource_params, attr_name|
           if resource_params[attr_name].nil?
             validate_param!(attr_name, resource_params)
-            params[parent_element] = resource_params if parent_element && params[parent_element].nil?
           end
         end
       end
--- a/spec/grape/validations/validators/default_spec.rb
+++ b/spec/grape/validations/validators/default_spec.rb
@@ -41,20 +41,6 @@ describe Grape::Validations::DefaultValidator do
         end
 
         params do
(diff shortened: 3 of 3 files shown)


dblock:
Looks good, just needs CHANGELOG.

dm1try:
@dblock up

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 0.11.0 (2/23/2015)
 ==================
 
+* [#936](https://github.com/intridea/grape/pull/936): Fixed default params processing for optional groups - [@dm1try](https://github.com/dm1try).
```
dblock: I released 0.11.0 today, so this needs to bump into Next Release, sorry for the hassle.

u2:
```
      it 'optional array if empty array params' do
        post '/optional_array', { optional_array: [] }.to_json
        expect(last_response.body).to eq({ optional_array: [] }.to_json)
      end
```

```
     Failure/Error: expect(last_response.body).to eq({ optional_array: [] }.to_json)
       
       expected: "{\"optional_array\":[]}"
            got: "{\"optional_array\":null}"
       
       (compared using ==)
```

dm1try:
Yes, it is right but you forgot to provide `application/json` content type in your spec.
You cannot send an empty array using http query key/value pairs this is why your spec fails.

So
```ruby
      it 'optional array if empty array params' do
        header 'Content-Type', 'application/json'
        post '/optional_array', { optional_array: [] }.to_json
        expect(last_response.body).to eq({ optional_array: [] }.to_json)
      end
```
should work

u2:
```
        subject.params do
          optional :optional_array, type: Array do
            optional :foo_in_optional_array, default: 'bar'
          end
        end
        subject.post '/optional_array_no_declared_params' do
          { no_declared_array: params[:no_declared_array] }
        end

      it 'optional array if params not declared' do
        post '/optional_array_no_declared_params', { no_declared_array: [{id: 2}] }.to_json
        expect(last_response.body).to eq({ no_declared_array: [{id: 2}] }.to_json)
      end
```

```
     Failure/Error: expect(last_response.body).to eq({ no_declared_array: [{id: 2}] }.to_json)
       
       expected: "{\"no_declared_array\":[{\"id\":2}]}"
            got: "{\"no_declared_array\":null}"
       
       (compared using ==)
```

dm1try:
> it's weird.

For me it seems normal. You send the json payload with `application/x-www-form-urlencoded` content type, it does not make a sense. You should remove  `.to_json` if you use this content type or provide `application/json` content type as above if you wanna use a json payload.

u2:
Thank you for your explication. :-)

dm1try:
@u2 sure) so #937 can be closed I think.

u2:
Yes.

Issue #929, opened 2015-02-15
optional Array validate error


u2:
when
```
            params do
              optional :photos, type: Array do
                optional :id, type: Integer
                optional :photo_path, type: String
                optional :width, type: Integer
                optional :height, type: Integer
                optional :description, type: String, default: ''
              end

              optional :videos, type: Array do
                optional :id, type: Integer
                optional :video_path, type: String
                optional :width, type: Integer
                optional :height, type: Integer
                optional :description, type: String, default: ''
                optional :persistent_id, type: String
              end
            end

            post '/array' do
            end
```
The params like below:
```
parameters: {"vedios"=>[nil], "photos"=>[nil]}
```

It return the error:
```
NoMethodError (undefined method `[]' for nil:NilClass):
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/validations/validators/default.rb:17:in `block in validate!'
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/validations/attributes_iterator.rb:15:in `block (2 levels) in each'
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/validations/attributes_iterator.rb:14:in `each'
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/validations/attributes_iterator.rb:14:in `block in each'
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/validations/attributes_iterator.rb:13:in `each'
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/validations/attributes_iterator.rb:13:in `each'
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/validations/validators/default.rb:16:in `validate!'
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/endpoint.rb:235:in `block in run'
  /Users/u2/.rvm/gems/ruby-2.1.2/bundler/gems/grape-5e48701fbbdb/lib/grape/endpoint.rb:233:in `each'
[...]

PR #937, opened 2015-02-24
Fix empty array with default value validates

u2:
#929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 0.11.1 (Next)
 =============
 
-* Your contribution here.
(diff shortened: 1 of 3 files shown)


dm1try:
Hi @u2, thanks for looking into this issue.
Those changes do not fix the behavior if optional array is not provided
```ruby
get '/optional_array_with_default_values' do
 { videos: params[:videos] }
end

it 'set nil for an optional array if it is not provided' do
  get('/optional_array_with_default_values')
  expect(last_response).to eq { { videos: nil }.to_json } # fails
end
```

see #936

u2:
@dm1try Yes.

dblock:
Closing in favor of https://github.com/intridea/grape/pull/936, LMK if I am doing something wrong ;)

PR #939, opened 2015-02-25
Pass thru the Error .status value when using rescue_from :all

### LINKED ISSUE #938: rescue_from :all  &  Grape::Exceptions::ValidationErrors
I was writing an integration test of our API and was having trouble understanding why I was getting a 500 response code on validation tests instead of the 400 response I was expecting.  Eventually I tracked it to our use of `rescue_from :all`.  

It would be cool if we could still use `rescue_from :all` but for situations where the error is a validation error the returning response code would be a 400?  

Thanks!!

johnsinco:
This fixes #938 by checking to see if the Error object passed down to error_response has a :status method that returns an Integer.  If so, the value is passed into error_response by handle_error to set the http status code vs the default of 500.   This fixes a situation where a validation error will return as a 500 status due to `rescue_from :all` without an attached block handler.

### DIFF
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -59,7 +59,7 @@ module Grape
       end
 
       def handle_error(e)
-        error_response(message: e.message, backtrace: e.backtrace)
+        error_response(message: e.message, backtrace: e.backtrace, status: e.try(:status))
       end
 
       def error_response(error = {})
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -59,7 +59,8 @@ module Grape
       end
 
       def handle_error(e)
-        error_response(message: e.message, backtrace: e.backtrace, status: e.try(:status))
+        status = (e.respond_to?(:status) && e.status.is_a?(Integer)) ? e.status : nil
+        error_response(message: e.message, backtrace: e.backtrace, status: status)
       end
 
       def error_response(error = {})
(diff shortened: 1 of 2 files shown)


dblock:
Since `status` is not a built-in thing into an exception, I think we want to leave this as is by design. The `rescue_all` block should, IMO behave the same way for everybody, future and past exceptions, and end up with a 500 catch-all error unless you explicitly handle it. 

Another way of looking at it is that an exception shouldn't dictate an HTTP status code, they appear in a different context.

That said, I think you can achieve what you want in a cleaner way today by explicitly handling Grape validation errors, something like this:

```ruby
rescue_from Grape::Exceptions::Validation do |e|
          Rails.logger.warn "#{e.class.name}: #{e.message} (#{Api::Util::BacktraceHelpers.find_app_first(e.backtrace)})"
          rack_response({
            type: 'param_error',
            message: 'Invalid parameters.',
            detail: { e.params.join(', ') => [e.message] }
          }.to_json, 400)
        end
          rack_response({
            type: 'param_error',
            message: 'Invalid parameters.',
            detail: e.errors.inject({}) do |h, (k, v)|
              # JSON does not permit having a key of type Array
              h[k.count == 1 ? k.first : k.join(', ')] = v
              h
            end
          }.to_json, 400)
        end
```

I am going to close this, thank you for suggesting it, appreciate your work. If anyone feels strongly about this otherwise, please add your comments.

PR #942, opened 2015-03-04
Fix: Operate on a cloned Hash when specifying param presence

croeck:
Operate on a cloned Hash when specifying `presence: true` for parameters to prevent issues with re-used entities. Includes spec and the minor fix.

Should be self explaining this time :)


dblock:
Thx

PR #944, opened 2015-03-05
Fix: route_param cannot accept string that contains a period, interprets it as a format extension

sentience:
When a Grape::API is set to support just a single format (e.g. `format :json`), it does not need URL extensions to specify the format of a request. Under such conditions, it should be possible to pass a string value that contains a period (.) to a `route_param`, which would otherwise be recognized as a format extension.

### DIFF
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -147,7 +147,7 @@ module Grape
         options[:path].map do |path|
           prepared_path = prepare_path(path)
           anchor = options[:route_options].fetch(:anchor, true)
-          path = compile_path(prepared_path, anchor && !options[:app], prepare_routes_requirements)
+          compiled_path = compile_path(prepared_path, anchor && !options[:app], prepare_routes_requirements)
           request_method = (method.to_s.upcase unless method == :any)
 
           Route.new(options[:route_options].clone.merge(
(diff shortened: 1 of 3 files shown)


sentience:
The fix I’ve included works by allowing `.` as a character in all `route_param` values when the endpoint does not take format extensions.

Really, I think this character should also be allowed in all `route_param`s that aren’t at the end of the request path. Ideally, the request path handling would be smart enough to allow `.` even in an end-of-path `route_param` while still allowing for a format extension. There’d be some nasty edge cases to handle, though, and from what I can see the current implementation isn’t structured in a way that would make this easy.

lib/grape/endpoint.rb
```diff
@@ -165,10 +165,17 @@ def prepare_routes
...
     def prepare_path(path)
-      path_settings = inheritable_setting.to_hash[:namespace_stackable].merge(inheritable_setting.to_hash[:namespace_inheritable])
       Path.prepare(path, namespace, path_settings)
     end
 
+    def path_settings
+      inheritable_setting.to_hash[:namespace_stackable].merge(inheritable_setting.to_hash[:namespace_inheritable])
+    end
+
+    def uses_specific_format?
```
sentience: This method is identical to `Path#uses_specific_format?`. There’s probably a better way to structure things so that this duplication is not necessary.

lib/grape/endpoint.rb
```diff
@@ -165,10 +165,17 @@ def prepare_routes
...
+      !!(path_settings[:format] && path_settings[:content_types].size == 1)
+    end
```
sentience: This method is identical to `Path#uses_specific_format?`. There’s probably a better way to structure things so that this duplication is not necessary.

dblock:
This is generally resolved by specifying different `requirements` for the route. Is this not possible here? Do we still want the code change.

If this is still needed, for this to be merged:

* Update CHANGELOG
* Squash your commits

sentience:
Aha! Yes, setting `requirements` for the `route_param` indeed overrides the default regex, and therefore makes it possible to accept strings containing `.`. Thanks for the pointer. :)

Issue #874, opened 2015-01-02
spec/grape/dsl/routing_spec.rb needs specs


dblock:
```
Pending:
  Grape::DSL::Parameters#use does some thing
    # Not yet implemented
    # ./spec/grape/dsl/parameters_spec.rb:45
  Grape::DSL::Parameters#use_scope does some thing
    # Not yet implemented
    # ./spec/grape/dsl/parameters_spec.rb:49
  Grape::DSL::Parameters#includes does some thing
    # Not yet implemented
    # ./spec/grape/dsl/parameters_spec.rb:53
  Grape::DSL::Parameters#group does some thing
    # Not yet implemented
    # ./spec/grape/dsl/parameters_spec.rb:107
  Grape::DSL::Parameters#params does some thing
    # Not yet implemented
    # ./spec/grape/dsl/parameters_spec.rb:111
  Grape::DSL::RequestResponse.rescue_from does some thing
    # Not yet implemented
    # ./spec/grape/dsl/request_response_spec.rb:105
  Grape::DSL::Routing.version does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:18
  Grape::DSL::Routing.route does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:63
  Grape::DSL::Routing.namespace does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:116
  Grape::DSL::Routing.group does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:120
  Grape::DSL::Routing.resource does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:124
  Grape::DSL::Routing.resources does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:128
  Grape::DSL::Routing.segment does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:132
  Grape::DSL::Routing.routes does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:136
  Grape::DSL::Routing.route_param does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:140
  Grape::DSL::Routing.versions does some thing
    # Not yet implemented
    # ./spec/grape/dsl/routing_spec.rb:144
``

PR #947, opened 2015-03-06
Add missing parameters specs.

dabrorius:
- Add missing parameters specs
- Replace single line #group method with alias_method call
- Add comment that explains what #use method does

This is related to #874 even though name of the issue is misleading - these are not routing specs.


dblock:
Nice work, thanks.

PR #948, opened 2015-03-09
Update README with new MiniTest base class

brianstorti:
Minitest 5.0 renamed `MiniTest::Unit::TestCase` to `Minitest::Test`.


dblock:
Thanks, merged.

Issue #934, opened 2015-02-19
Idea: Provide constants for HTTP status codes


swistaczek:
Hi,
It would be nice if we could provide some kind of constants including HTTP status codes like Rails does ( http://guides.rubyonrails.org/layouts_and_rendering.html#the-status-option ).

dm1try:
`rack` already has smth similar, see [1](https://github.com/rack/rack/blob/e98a9f7ef0ddd9589145ea953948c73a8ce3caa9/test/spec_utils.rb#L413), [2](https://github.com/rack/rack/blob/aec5471cfbea791b72dd9a352514df5e50f10435/lib/rack/utils.rb#L580)

you can define a helper and use it easily
```ruby
helpers do
  def with_status(code, &block)
    status Rack::Utils.status_code(code)
    yield if block_given?
  end
end

get '/goodies' do
  with_status :payment_required do
    { error: 'give me some money' }
  end
end
```

swistaczek:
@dm1try thanks for information :), what do you think to support this behaviour:
```status :method_not_allowed```

If method status detect that given argument is Symbol then it will try resolve correct HTTP code based on some rails-like symbol to http status code mapping table.

dm1try:
`rails` [uses the same `rack` helper](https://github.com/rails/rails/blob/6d72485b6977804a872dcfa5d7ade2cec74c768e/actionpack/lib/action_dispatch/http/response.rb#L180) that I mentioned above.

If you are talking about "built-in sugar" for `#status` then I don't have any strong opinion because I don't use *named* statuses :)

dblock:
I would be down with enabling this in Grape via `Rack::Utils.status_code(status)`.

dabrorius:
Is somebody already working on this one? I would like to give it a shot.

dblock:
@dabrorius Give it a shot!

whatasunnyday:
@dblock Looks this issue can be closed.

dblock:
Yep.

PR #950, opened 2015-03-09
Allow symbols for status codes

dabrorius:
This is a very simple implementation of feature requested in #934.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 * [#936](https://github.com/intridea/grape/pull/936): Fixed default params processing for optional groups - [@dm1try](https://github.com/dm1try).
 * [#942](https://github.com/intridea/grape/pull/942): Fixed forced presence for optional params when based on a reused entity that was also required in another context - [@croeck](https://github.com/croeck).
+* [#950](https://github.com/intridea/grape/pull/950): Status method can now accept one of Rack::Utils status code symbols (:ok, :found, :bad_request, etc.). - [@dabrorius](https://github.com/dabrorius).
 
 * Your contribution here.
 
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -94,7 +94,13 @@ module Grape
       #
       # @param status [Integer] The HTTP Status Code to return for this request.
       def status(status = nil)
-        if status
+        if status.is_a? Symbol
+          if Rack::Utils::SYMBOL_TO_STATUS_CODE.keys.include?(status)
+            @status = Rack::Utils.status_code(status)
+          else
+            fail ArgumentError, "Status code :#{status} is invalid."
+          end
+        elsif status
           @status = status
         else
           return @status if @status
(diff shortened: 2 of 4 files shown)


dblock:
I like it. Rebase and squash please?

dabrorius:
Done

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 * [#936](https://github.com/intridea/grape/pull/936): Fixed default params processing for optional groups - [@dm1try](https://github.com/dm1try).
 * [#942](https://github.com/intridea/grape/pull/942): Fixed forced presence for optional params when based on a reused entity that was also required in another context - [@croeck](https://github.com/croeck).
+* [#950](https://github.com/intridea/grape/pull/950): Status method can now accept one of Rack::Utils status code symbols (:ok, :found, :bad_request, etc.) - [@dabrorius](https://github.com/dabrorius)
```
dblock: OCD: missing period after the line.

lib/grape/dsl/inside_route.rb
```diff
@@ -94,7 +94,13 @@ def redirect(url, options = {})
       #
       # @param status [Integer] The HTTP Status Code to return for this request.
       def status(status = nil)
-        if status
+        if status.is_a? Symbol
+          if Rack::Utils::SYMBOL_TO_STATUS_CODE.keys.include?(status)
+            @status = Rack::Utils.status_code(status)
+          else
+            fail ArgumentError, "Status code :#{status} is not valid."
```
dblock: You want to specialize this error and make it localizable too, https://github.com/intridea/grape/blob/master/lib/grape/locale/en.yml.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 * [#936](https://github.com/intridea/grape/pull/936): Fixed default params processing for optional groups - [@dm1try](https://github.com/dm1try).
 * [#942](https://github.com/intridea/grape/pull/942): Fixed forced presence for optional params when based on a reused entity that was also required in another context - [@croeck](https://github.com/croeck).
+* [#950](https://github.com/intridea/grape/pull/950): Status method can now accept one of Rack::Utils status code symbols (:ok, :found, :bad_request, etc.). - [@dabrorius](https://github.com/dabrorius)
```
dblock: OCD: period missing after `daborius)`, not inside the phrase, see how the other lines above are :)

lib/grape/dsl/inside_route.rb
dabrorius: This is an error that only developer should see, not the end user. Is there really need to localize it?
I used this as reference https://github.com/intridea/grape/blob/master/lib/grape/dsl/inside_route.rb#L28

dabrorius:
@dblock I changed the error message but didn't localize it yet. As I commented before, I'm not sure if it makes sense to allow user to translate a message that API end-users (as far as I understand) will never see. If it really needs to be localized I can change it but I wanted to make sure you really want that.

dblock:
Fair enough, merging.

swistaczek:
Good one @dabrorius  :+1:

dabrorius:
Thanks :)

PR #952, opened 2015-03-10
Raise error if unknow status code is passed to #status

dabrorius:
- Raise error if given status code is not included Rack::HTTP_STATUS_CODES
- Raise error if given status is not a Fixnum or Symbol

PR #953, opened 2015-03-10
Allow all Fixnum statuses

dabrorius:
This is a fixup for #952
I didn't edit changelog because #952 change description still seems valid.

### DIFF
--- a/spec/grape/dsl/inside_route_spec.rb
+++ b/spec/grape/dsl/inside_route_spec.rb
@@ -118,9 +118,8 @@ module Grape
             .to raise_error(ArgumentError, 'Status code :foo_bar is invalid.')
         end
 
-        it 'raises error if unknow status code is passed' do
-          expect { subject.status 210 }
-            .to raise_error(ArgumentError, 'Status code 210 is invalid.')
+        it 'accepts unknown Fixnum status codes' do
+          expect { subject.status 210 }.to_not raise_error
         end
 
         it 'raises error if status is not a fixnum or symbol' do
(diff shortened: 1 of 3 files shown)


dblock:
@dm1try Double-check this for me please? Merge at will.

dblock:
@dabrorius I think we need an explicit test for being able to set a status to a value not in that Rack list.

spec/grape/dsl/inside_route_spec.rb
```diff
@@ -118,11 +118,6 @@ def initialize
             .to raise_error(ArgumentError, 'Status code :foo_bar is invalid.')
         end
 
-        it 'raises error if unknow status code is passed' do
-          expect { subject.status 210 }
-            .to raise_error(ArgumentError, 'Status code 210 is invalid.')
```
dblock: So this test should be here, but assert the opposite, `.to_not raise_error` ...

dabrorius: Ok, I thought we will just pretend like that feature never existed :)

spec/grape/dsl/inside_route_spec.rb
```diff
@@ -118,9 +118,8 @@ def initialize
...
+        it 'accepts unknown Fixnum status codes' do
+          expect { subject.status 210 }.to_not raise_error
```
dblock: Can I be annoying? I think we also need an integration level test that says an endpoint actually returns this status. Thanks.

dabrorius: Sure, I'll see what I can do.

dabrorius:
@dblock I added few integration specs. Let me know if that's what you had in mind.

dblock:
Good. Will merge on :green_book:

Issue #955, opened 2015-03-12
Feature request: support present_hash


u2:
My scenario is that we want to cache `present` like below:
```
result = cache(key: [entry.id, entry.updated_at], expires_in: 30.minutes) do
   present entry, with: Entities::Entry::Entry
end
result = Oj.load(result)
present result
present entry_user, Entities::Entry::DetailEntryUser
```
The result is only `entry_user`, grape do not support `present` a hash, just support `present key, value`, and I don't want to write something for traversing the result and `present` all the keys again.So what about adding support for `present_hash hash`.
I am sure it's reasonable or not, it's just a suggestion.If it's not reasonable, I will write some extend lib.

dblock:
Before you begin, you should be able to `present` a Hash by extending Hash, no? I would try that first. 

Generally I think it's bad that you have an object half the time and a hash the other time through the same code. 

I would cache the whole thing and would check out [garner](https://github.com/artsy/garner) for your caching problems.

u2:
Yes, it's not good that `present` an hash and then `present` an object.But, for some special scenario, like below
```
          to_dos_group = to_dos.group_by{|t| "#{t.end_at.year}-#{t.end_at.month}" if t.end_at }

          json = Jsonify::Builder.plain do |j|
            j.months(to_dos_group) do |group|
              category = group[0].to_s
              todos = group[1]
              j.year category.split('-')[0]
              j.month category.split('-')[1]
              j.count category.sum(&:left_number)
              j.to_dos(todos) do |todo|
                j.id todo.id
                j.title todo.title
                j.to_do_category_id todo.to_do_category_id
                j.to_do_category_name todo.to_do_category_name
                j.event_identifier todo.event_identifier
                j.description todo.description
                j.remind_at todo.remind_at.utc.iso8601 if todo.remind_at
                j.end_at todo.end_at.utc.iso8601 if todo.end_at
                j.expired todo.expired
              end
            end

            present Oj.load(json).merge(....)
```
We present a hash first, then we could not present a object, just can `merge` a hash.

Yes, garner[https://github.com/artsy/garner] is a good choice for caching the whole thing,but for the most time we want to cache fragment.

dblock:
For the purpose of this proposal, do you agree that `present Hash` should work? I mean we don't want another method called `present_hash`. If you do, see whether it actually doesn't work already (there was some work on HEAD in that direction).

u2:
Yes, I have read the README again.
[...]

PR #956, opened 2015-03-13
Support present hash

u2:
#955

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 0.12.0 (Next)
 =============
 
+#### Features
+
+* [#956](https://github.com/intridea/grape/issues/956): Support `present` with `Grape::Presenters::Presenter`  - [@u2](https://github.com/u2).
+
 #### Fixes
 
 * [#936](https://github.com/intridea/grape/pull/936): Fixed default params processing for optional groups - [@dm1try](https://github.com/dm1try).
--- a/README.md
+++ b/README.md
@@ -1939,6 +1939,31 @@ Grape will automatically detect that there is a `Status::Entity` class and use t
 representative entity. This can still be overridden by using the `:with` option or an explicit
 `represents` call.
 
+You can present `hash` with `Grape::Presenters::Presenter` to keep things consistent.
+
+```ruby
+get '/users' do
+  present { id: 10, name: :dgz }, with: Grape::Presenters::Presenter
+end
+````
+The response will be
+
+```ruby
+{
+  id:   10,
+  name: 'dgz'
+}
+```
+
+It has the same result with
+
+```ruby
+get '/users' do
+  present :id, 10
+  present :name, :dgz
+end
+```
+
 ### Hypermedia and Roar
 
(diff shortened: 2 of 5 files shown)


lib/grape/dsl/inside_route.rb
```diff
@@ -184,29 +184,37 @@ def body(value = nil)
...
+
+          representation = { root => representation } if root
+        end
 
-        representation = { root => representation } if root
         if key
           representation = (@body || {}).merge(key => representation)
         elsif entity_class.present? && representation.respond_to?('merge')
           representation = (@body || {}).merge(representation)
+        elsif hash_present && representation.respond_to?('merge')
```
dblock: Should this be `respond_to?(:merge)`?

dblock:
So, the real question given this code is why isn't `present <hash>` a noop to return the hash? What are we gaining by running it through `present`?

dblock:
Another question is what does `present <hash>, with: Presenter` do and does it make any sense?

u2:
Yes, `present <hash>` just merge `body` and return the `body`, no other thing.

lib/grape/dsl/inside_route.rb
u2: No, removed already.

u2:
Yes, maybe it's better to `present <hash>, with: Presenter`.I will try to rewrite it.

dblock:
I think I am missing something. What is the difference in the result between `present <hash>` and `present <hash>, with: Presenter`?

u2:
For the result, there is no difference.But `present <hash>, with: Presenter` fit the grape style.Just add a `Presenter`, and use `with: Presenter` for represent the object itself.It can present anything, not only `Hash`.

dblock:
Hm, I can see the argument for a noop class like this. Do you think it belongs in `Middleware`? It's not a Middleware.

(and sorry to be difficult :)

u2:
I was hesitant to about that, but I am not sure where I can place `Presenter `.

dblock:
Maybe just `Grape::Presenters::Presenter`?

u2:
Ok, I will change it right now. :)

spec/grape/dsl/inside_route_spec.rb
```diff
@@ -268,6 +268,46 @@ def initialize
...
+            end
+
+            describe 'instance' do
+              before do
+                subject.present hash_mock1, with: Grape::Presenters::Presenter
+                subject.present hash_mock2, with: Grape::Presenters::Presenter
+              end
+
+              it 'presents both dummy presenter' do
+                puts subject.body.inspect
```
dblock: Stray `puts` here. 

Lets put this into a separate _spec file too.

dblock:
I think the code here looks perfect. This needs an entry in README. Maybe just describe something around using this presenter to keep things consistent?

u2:
I have updated README.

dblock:
Merging, thanks for your patience.

swistaczek:
Hey, I still dont get idea behind this change :dancer:

dblock:
It's entirely to make returning data consistent, if you `present` an object you should also be able to `present` anything else. Sometimes it makes sense if your code path figures out what to present, then presents it.

PR #957, opened 2015-03-16
make `regexp` allowing nil value

calfzhou:
for the discussion in #614


dblock:
So this is a potentially breaking change for users, right? If so it needs an example and a paragraph in CHANGELOG and we should bump the second part of the version number.

calfzhou:
@dblock does it look good?

dblock:
Merged via https://github.com/intridea/grape/commit/f3e82d9f30307bd0a6524e4135391bb2fb5c2db4 (some README and CHANGELOG updates).

PR #962, opened 2015-03-20
Make sure to set documentation attribute even for `default: false`

ajvondrak:
I wasn't seeing documentation attributes get propagated for default values of `false` as in
```
param :foo, type: Boolean, default: false
```

Whether or not this still propagates the default value correctly with grape-swagger / swagger-ui is tangential. Just noticed that the guard was being perhaps a bit too liberal with excluding values.

Made a quick fix + added a couple regression tests.

### DIFF
--- a/spec/grape/validations/params_scope_spec.rb
+++ b/spec/grape/validations/params_scope_spec.rb
@@ -9,6 +9,70 @@ describe Grape::Validations::ParamsScope do
     subject
   end
 
+  context 'setting a default' do
+    let(:documentation) { subject.routes.first.route_params }
+
+    context 'when the default value is truthy' do
+      before do
+        subject.params do
+          optional :int, type: Integer, default: 42
+        end
+        subject.get
+      end
+
+      it 'adds documentation about the default value' do
+        expect(documentation).to have_key('int')
+        expect(documentation['int']).to have_key(:default)
+        expect(documentation['int'][:default]).to eq(42)
+      end
+    end
+
+    context 'when the default value is false' do
+      before do
+        subject.params do
+          optional :bool, type: Virtus::Attribute::Boolean, default: false
+        end
+        subject.get
+      end
+
+      it 'adds documentation about the default value' do
+        expect(documentation).to have_key('bool')
+        expect(documentation['bool']).to have_key(:default)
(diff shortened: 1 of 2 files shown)


dm1try:
@ajvondrak looks good, pls fix the RuboCop offense

ajvondrak:
Oh, woops. :) Pushed.

dm1try:
@ajvondrak sorry for nitpicking, could you squash your commits into one?:) Then I'll merge. Thanks!

ajvondrak:
Not a problem. Squashed & pushed.

Also looks like the Travis build still failed, but I can't work out why. There were a bunch of "allowed failures", and then the one core dump in Rubinius (looks like a problem with rake?). Any clues? If I'm lucky, it'll just go away with this next build, haha.

dm1try:
@ajvondrak ok, now it's :green_apple:

spec/grape/validations/params_scope_spec.rb
```diff
@@ -9,6 +9,58 @@ def app
...
+      it 'when the default value is truthy' do
+        subject.params do
+          optional :int, type: Integer, default: 42
+        end
+        subject.get '/' do
+        end
+        expect(documentation).to have_key('int')
+        expect(documentation['int']).to have_key(:default)
+        expect(documentation['int'][:default]).to eq(42)
+      end
```
dm1try: I relook this and `it 'when ...` definition seems unusual for me :)
maybe we can remove 'propagates documentation' context and split 'its' definitions to own contexts
```ruby
context 'when the default value is truthy' do
  before do
    subject.params do
      optional :int, type: Integer, default: 42
    end
    subject.get
  end
  it 'adds documentation to the value' do
   ...
  end
```
What do you think?

ajvondrak: D'oh. Good point. Will invert.

dm1try:
Great! I think this need a CHANGELOG entry also

dm1try:
Closed via fb0eae78ed. Thanks for your patience :)

ajvondrak:
Thanks for adding the changelog. Hadn't gotten back to this over the weekend, would've added it. But hey, not complaining. :)

dm1try:
@ajvondrak sure :)

PR #973, opened 2015-03-31
Less allocated objects on each request

u2:
http://www.sitepoint.com/unraveling-string-key-performance-ruby-2-2/
https://github.com/rack/rack/pull/737

```ruby
module V1
  module App
    class Ping < Grape::API
      format :json
      get '/ping' do
        present :ping, :pong
      end
    end
  end
end
```

```ruby
require 'rails_helper'
require 'memory_profiler'

describe ::API::API do
  describe "GET /v1/api/app/ping.json" do
    it "return 200" do
      report = MemoryProfiler.report do
        get "/v1/api/app/ping.json"
      end

      report.pretty_print
      expect(response.status).to eq(200)
    end
  end
end
```
before:
```grape_upstream/lib x 10369233```

after:
```grape/lib x 10366633```

```(10369233 - 10366633)/10369233.to_f * 100  #=> 0.02507417858196455```% fewer objects per request.

Maybe there are some missed optimizations.I will updated later.

### DIFF
--- a/lib/grape.rb
+++ b/lib/grape.rb
@@ -42,6 +42,10 @@ module Grape
   autoload :Validations,         'grape/validations'
   autoload :Request,             'grape/http/request'
 
+  module Http
+    autoload :Headers,           'grape/http/headers'
+  end
+
   module Exceptions
     autoload :Base,                           'grape/exceptions/base'
     autoload :Validation,                     'grape/exceptions/validation'
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -96,7 +96,7 @@ module Grape
 
     def call(env)
       result = @route_set.call(env)
-      result[1].delete('X-Cascade') unless cascade?
+      result[1].delete(Grape::Http::Headers::X_CASCADE) unless cascade?
       result
     end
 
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: lib/grape.rb [hunk 1]
```diff
@@ -27,6 +27,17 @@
 require 'i18n'
 require 'thread'
 
+CONTENT_TYPE = 'Content-Type'.freeze
```
dblock: I don't like how these are globals. They are all HTTP headers, so create a module, like `Grape::HTTP::Headers` and stick them in there?

u2:
I am not sure the `Grape::HTTP::Headers` suitable or not.

u2:
Why `eql Hash.new` passed, and `eql {}` failed?
I think `Hash.new.eql?({}) #=> true`.

dm1try:
> and eql {} failed?
 because here `{}` was interpreted as a block, use parenthesesю

u2:
Yes, thank you. @dm1try

lib/grape.rb
```diff
@@ -26,6 +26,7 @@
 require 'virtus'
 require 'i18n'
 require 'thread'
+require 'grape/http/headers'
```
dblock: Why is this required?

lib/grape.rb
```diff
@@ -133,6 +134,8 @@ module Presenters
   end
 end
 
+include Grape::Http::Headers
```
dblock: Maybe group with the above `require`, but in theory this shouldn't even be necessary since this class should be found anyway, no?

u2: Not be necessary.I just want to write `HTTP_ACCEPT_VERSION` instead of `Grape::Http::Headers::HTTP_ACCEPT_VERSION`, less words,but maybe it's not a good idea.

u2:
Could you close this PR, I create a another one #977

Issue #889, opened 2015-01-12
Should error! and error_response have the same method signature?


whatasunnyday:
I was trying to use `error!` inside of a `rescue_from` block and got a `NoMethodError`. I later found `error_response` and noticed that you don't get the same level of customization that you do from `error!`.

I'm proposing that `error_response` and `error!` have the same method signature. Their current implementations are as follows:

```ruby
def error_response(error = {})
  status = error[:status] || options[:default_status]
  message = error[:message] || options[:default_message]
  headers = { 'Content-Type' => content_type }
  headers.merge!(error[:headers]) if error[:headers].is_a?(Hash)
  backtrace = error[:backtrace] || []
  rack_response(format_message(message, backtrace), status, headers)
end

def error!(message, status = nil, headers = nil)
  self.status(status || namespace_inheritable(:default_error_status))
  throw :error, message: message, status: self.status, headers: headers
end
```

My use case was to use `error_response` like I'm using `error!` below:

``` ruby
error!({id: 'not_found', message: 'Not Found.'}, 404)
```

Instead, I'm skipping using `error_response` and returning a `Rack::Response` object like this:

```ruby
Rack::Response.new({
  id: 'interal_error',
  message: 'Interal server error. Check logs.' 
}.to_json, 500, { 'Content-type' => 'text/error' }).finish
```

Apologies if I'm misunderstanding the API and there's a simple way to do the same thing.

dblock:
There's a lot of legacy around this, so most of the reasons for it to be this way are historical and not practical. Try making a PR that changes things the way you describe, I think it's definitely a worthy endeavor.

whatasunnyday:
@dblock I just got a chance to work on this and I was a bit confused. One big difference between the two method signatures is that `error_response` accepts a backtrace and `error!` doesn't.

When I run this example, I don't see a backtrace from /error. I do see one for /standard_error. It isn't clear to me if rack or grape is providing the backtrace for /standard_error.

```ruby
module Example
  class API < Grape::API
    format :json

    get :standard_error do
      raise StandardError.new("hello")
    end

    get :error do
      raise Grape::Exceptions::Base
    end
  end
end
```

From what I can tell, the formatter shouldn't be swallowing the backtrace:
https://github.com/intridea/grape/blob/master/lib/grape/error_formatter/json.rb

Is rack providing the backtrace or grape?

Is is it the default behavior for formatters to swallow backtraces? In that case, it makes sense for the backtrace to be dropped as an argument.

[...]

PR #974, opened 2015-04-01
error! in rescue from blocks

whatasunnyday:
I was having a similar problem with `ValidationErrors#to_json` as I was with `ValidationErrors#as_json` (see: #889). If there's preferred fix, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1031,7 +1031,7 @@ You can rescue a `Grape::Exceptions::ValidationErrors` and respond with a custom
 ```ruby
 format :json
 subject.rescue_from Grape::Exceptions::ValidationErrors do |e|
-  rack_response e.to_json, 400
+  error! e, 400
 end
 ```
 
@@ -1442,16 +1442,28 @@ class Twitter::API < Grape::API
 end
 ```
 
-You can rescue all exceptions with a code block. The `error_response` wrapper
+You can rescue all exceptions with a code block. The `error!` wrapper
 automatically sets the default error code and content-type.
 
 ```ruby
 class Twitter::API < Grape::API
   rescue_from :all do |e|
-    error_response({ message: "rescued from #{e.class.name}" })
+    error!("rescued from #{e.class.name}")
   end
 end
 ```
+
+Optionally, you can set the format, status code and headers.
+
+```ruby
+class Twitter::API < Grape::API
+  format :json
+  rescue_from :all do |e|
+    error!({ error: "Server error.", 500, { 'Content-Type' => 'text/error' } })
+  end
+end
+```
+
 
 You can also rescue specific exceptions with a code block and handle the Rack
 response at the lowest level.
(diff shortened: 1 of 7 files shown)


whatasunnyday:
Also, it's possible to remove error_response and sub in error! but it was getting pretty confusing. There seems to be a bit of coupling with the method signature of error_response and the rest of the code. I can send another PR for that or append it here.

README.md
```diff
@@ -1031,7 +1031,7 @@ You can rescue a `Grape::Exceptions::ValidationErrors` and respond with a custom
 ```ruby
 format :json
 subject.rescue_from Grape::Exceptions::ValidationErrors do |e|
-  rack_response e.to_json, 400
+  error!(e, 400)
```
dblock: Generally I would omit the parenthesis here.

lib/grape/exceptions/validation_errors.rb
```diff
@@ -24,7 +24,7 @@ def each
         end
       end
 
-      def as_json
+      def as_json(_ = {})
```
dblock: Make it `_opts = {}`, better not to have anonymous parameters.

UPGRADING.md
```diff
@@ -15,6 +15,45 @@ end
 
 See [#957](https://github.com/intridea/grape/pull/957) for more information.
 
+#### Replace error_response with error! in rescue_from blocks
```
dblock: Just to be clear, this isn't required, right? I mean the old code still works, so maybe you want to call that out?

dblock:
This needs a CHANGELOG entry and please squash your commits.

UPGRADING.md
whatasunnyday: As of now, no. Do you think the correct move moving forward is to remove error_response all together since its now undocumented? If so, I think its worth mentioning that it will be deprecated and adding a TODO to refactor error! to replace error_response.

lib/grape/exceptions/validation_errors.rb
whatasunnyday: Cool, should I do the same with to_json?

UPGRADING.md
dblock: It would be nice to have an actual deprecation warning in there?

lib/grape/exceptions/validation_errors.rb
dblock: Yes. I actually don't think we need that `to_json` method at all, check?

whatasunnyday:
@dblock Thanks for your fast response. I've applied your recommendations. Let me know if you have anymore feedback.

dblock:
I like this, merging.

u2:
> 
0.28.0 (10/12/2014)
> 
New features
> 
New cop ExtraSpacing points out unnecessary spacing in files. (@blainesch)

This feature has been released in `0.28.0`, I think it's a bug of rubocop, anyway I create a commit for updating robocop version.

binary1101:
still doesn't work in 0.14.0
```
  format :json
  helpers do
    def server_error!
      error!({ error: 'Server error.' }, 500, { 'Content-Type' => 'text/error' })
    end
  end

  rescue_from :all do |e|
    server_error!
  end
end
```

dblock:
@binary1101 how does it not work exactly? Open a new issue please

PR #975, opened 2015-04-01
Remove Ruby 1.9.3 from travis.

whatasunnyday:
Support for Ruby 1.9.3 is officially over and security patches will not be backdated. It might be worth removing it from the travis build matrix.
https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/


dblock:
Fair enough.

PR #980, opened 2015-04-07
Add eager loading

u2:
http://blog.plataformatec.com.br/2012/08/eager-loading-for-greater-good/
Eager loading may be greater good.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 #### Features
 
+* [#980](https://github.com/intridea/grape/issues/980): Grape is now eager-loaded - [@u2](https://github.com/u2).
 * [#956](https://github.com/intridea/grape/issues/956): Support `present` with `Grape::Presenters::Presenter`  - [@u2](https://github.com/u2).
 
 #### Fixes
--- a/lib/grape.rb
+++ b/lib/grape.rb
@@ -20,6 +20,7 @@ require 'active_support/ordered_hash'
 require 'active_support/core_ext/object/conversions'
 require 'active_support/core_ext/array/extract_options'
 require 'active_support/core_ext/hash/deep_merge'
+require 'active_support/dependencies/autoload'
 require 'grape/util/content_types'
 require 'multi_json'
 require 'multi_xml'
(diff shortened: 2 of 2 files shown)


dblock:
I definitely think we want to have this in CHANGELOG, please. Would you also care to provide some explanation of why you think this could be better?

u2:
> To be clear, the question you should be asking is: "if I don't eager load this module, when will it be autoloaded"? If the answer is: "during a request", you have to eager load it.

dblock:
I am sold ;)

u2:
Sorry, what's the meaning of `I am sold`? I don't understand some English slang.
For example:
```ruby
config.autoload_paths += %W(#{config.root}/extras)

# root/extras/foo.rb
class Foo
end

# root/app/models/blog.rb
class Blog < ActiveRecord::Base
end
```
Things are ok in development
```ruby
defined?(Blog)
# => nil 
defined?(Foo)
# => nil 

Blog
# => Blog (call 'Blog.connection' to establish a connection) 
Foo
# => Foo 

defined?(Blog)
# => "constant" 
defined?(Foo)
# => "constant"
```
But on the production, the situation is a little different…
```ruby
defined?(Blog)
# => "constant"

defined?(Foo)
# => nil

Blog
# => Blog (call 'Blog.connection' to establish a connection) 
Foo
# => Foo 

defined?(Blog)
# => "constant" 
defined?(Foo)
# => "constant"
```
Even before we try to use Blog for the very first time, the class is already loaded and the constant is known. Why is that so? Because of eager loading.

u2:
:), I will add CHANGELOG later.

dblock:
"I am sold" means "you sold me the reason to do this", which means "I agree with this change because your argument makes sense". So yes, update CHANGELOG and we're good to go.

u2:
updated.

CHANGELOG.md
```diff
@@ -14,7 +15,6 @@
 * [#957](https://github.com/intridea/grape/pull/957): Regexp validator now supports `allow_blank`, `nil` value behavior changed - [@calfzhou](https://giihub.com/calfzhou).
 * [#962](https://github.com/intridea/grape/pull/962): The `default` attribute with `false` value is documented now - [@ajvondrak](https://github.com/ajvondrak).
 * [#974](https://github.com/intridea/grape/pull/974): Add error! to rescue_from blocks - [@whatasunnyday](https://github.com/whatasunnyday).
-* Your contribution here.
```
dblock: Put this back please?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 #### Features
 
+* [#980](https://github.com/intridea/grape/issues/980): Added eager loading - [@u2](https://github.com/u2).
```
dblock: Lets say "Common Grape classes are now eager-loaded."

u2:
ok.

dblock:
Merged.

PR #981, opened 2015-04-09
Replaced map..flatten with flat_map

tancnle:
In these use cases, `flat_map` is more concise and performant


tancnle:
Oops what a rookie mistake! I just realised that `flat_map` is only does 1 level of flattening.

Issue #863, opened 2014-12-22
Build named path using url helper


kinko912000:
This issue was as follows

http://stackoverflow.com/questions/27578341/how-to-build-named-helper-using-grape-on-rails/27585384

reprah:
From the research I've done regarding how named routes and route sets are generated, they're pretty closely involved with Rails::Railtie and I'm unsure how a plain Rack app would be able to have this feature. @dblock did you have any broad ideas for how this can be implemented that I can look into?

dblock:
I think a route should figure out what its helper method is [here](https://github.com/intridea/grape/blob/master/lib/grape/endpoint.rb#L153) and then a `method_missing` implementation would emit a method when something wants `..._path` by looking up a route that matches in `.routes`.

reprah:
Ok, I was approaching this thinking "how can I get a Grape API's routes to end up in the url_helpers module when `mount` is called in the routes file?", since that's what seems to happen for Rails engines and this seems like an analogous situation.

Were you thinking the method_missing method that finds tthe right Grape route should be defined in a separate module that Grape provides (and a user would have to include)? Or something else entirely?

dblock:
Yes to the module, but ultimately Endpoint would extend it.

reprah:
@dblock Hi, I've got a working implementation for path helper methods and would like your feedback about the overall idea before I keep moving forward.

I want to support helper methods for both simple cases (api_v2_statuses_path returns '/api/v2/statuses') and more complicated ones (e.g., users_path(id: 2, format: 'xml') would return '/users/2.xml') that need a programmer's input via arguments. Here's the general flow:

1.) There's a UrlHelpers module. Its method_missing looks up a Grape::Route that responds to the missing method, and sends the method/arguments to that route

2.) When a Grape::Route is initialized, it determines its helper name by looking at route_path and filling in any dynamic segments  with values in @options. For example 'api/:version/tasks' becomes 'api_v1_tasks_path'

3.) Grape::Route then defines that helper method on itself. This method returns the path. The method should take arguments so the programmer can pass in things like an ID to generate the correct path

I thought this was the most straightforward thing to do - the UrlHelpers module just finds the matching route like you described, and Grape::Route does the hard work of figuring out the method name and generating a method that returns the correct path.

dblock:
That sounds good, and that said you cannot have any ambiguity in these, ie. users_path either takes (aka requires) parameters or doesn't (cause it either has a dynamic portion or it doesn't).

[...]

PR #983, opened 2015-04-11
Implement named route helpers

reprah:
Hi,

This is a response to issue #863; it adds path helper functions to Grape projects. I also have a [fork of grape-on-rack](https://github.com/reprah/grape-on-rack/tree/named-route-helpers#get-redirected-grape-route-helpers-demonstration) to demonstrate it (there are new instructions in its README). If there's anything I can provide to make it more obvious how the feature works, or if I need to explain why I did something a certain way, please let me know.

Summary: Grape::Route objects have dynamically named methods that return their paths, and there's a module called NamedRouteMatcher that does nothing but use its `method_missing` to match a method name to those routes and returns a path. The module's included in Grape::Endpoint and can be included by itself into another project.

I tried to provide thorough test coverage, and all new files have been Rubocop'd.

I'd like your thoughts on two things in particular:

- The difference between including the route helpers module at the Grape::Endpoint class level vs. [extending the individual Grape::Endpoint instances](https://github.com/intridea/grape/blob/master/lib/grape/endpoint.rb#L188)
- When it's appropriate to make something a private method in this project

Thanks for any feedback.

### DIFF
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -50,11 +50,15 @@ module Grape
         end
       end
 
-      protected
+      def all_routes
+        subclasses.flat_map(&:prepare_routes)
+      end
 
       def prepare_routes
         endpoints.map(&:routes).flatten
       end
+
+      protected
 
       # Execute first the provided block, then each of the
       # block passed in. Allows for simple 'before' setups
--- a/lib/grape/named_route_matcher.rb
+++ b/lib/grape/named_route_matcher.rb
@@ -1,0 +1,25 @@
+module Grape
+  module NamedRouteMatcher
+    def method_missing(method_id, *arguments)
+      segments = arguments.first || {}
+
+      route = Grape::API.all_routes.detect do |r|
+        route_match?(r, method_id, segments)
+      end
+
+      if route
+        route.send(method_id, *arguments)
+      else
+        super
+      end
+    end
+
+    def route_match?(route, method_name, segments)
+      return false unless route.respond_to?(method_name)
+      fail ArgumentError,
+           'Helper options must be a hash' unless segments.is_a?(Hash)
+      requested_segments = segments.keys.map(&:to_s)
+      route.uses_segments_in_path_helper?(requested_segments)
+    end
+  end
+end
--- a/lib/grape.rb
+++ b/lib/grape.rb
@@ -7,7 +7,9 @@ require 'rack/auth/basic'
 require 'rack/auth/digest/md5'
 require 'hashie'
 require 'set'
+require 'active_support'
 require 'active_support/version'
+require 'active_support/core_ext/class'
 require 'active_support/core_ext/hash/indifferent_access'
 
 if ActiveSupport::VERSION::MAJOR >= 4
@@ -37,6 +39,7 @@ module Grape
     autoload :API
     autoload :Endpoint
 
+    autoload :NamedRouteMatcher
     autoload :Route
     autoload :Namespace
 
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -10,6 +10,7 @@ module Grape
     attr_reader :env, :request, :headers, :params
 
     include Grape::DSL::InsideRoute
+    include Grape::NamedRouteMatcher
 
     class << self
       def before_each(new_setup = false, &block)
--- a/lib/grape/route.rb
+++ b/lib/grape/route.rb
@@ -1,8 +1,114 @@
 module Grape
   # A compiled route for inspection.
   class Route
+    attr_reader :helper_names, :helper_arguments
+
     def initialize(options = {})
       @options = options || {}
+      @helper_names = []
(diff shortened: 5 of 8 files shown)


dblock:
First, this is some excellent work, thank you and very well done. The code is clean, this could be merged as is (I will do a more thorough review, but generally it looks very, very good).

We've discussed continuing a path of modularizing Grape. I wonder whether this could be implemented in a separate gem (eg. `grape-route-helpers`). What do you think?

kuraga:
Excellent! Changelog, readme?

dblock:
+1 @kuraga

lib/grape/api.rb
```diff
@@ -50,12 +50,16 @@ def cascade(value = nil)
...
+        subclasses.flat_map(&:prepare_routes)
+      end
 
       def prepare_routes
         endpoints.map(&:routes).flatten
       end
 
+      protected
+
       # Execute first the provided block, then each of the
```
dm1try: why `.prepare_routes" scope visibility was changed to `public`?

lib/grape/named_route_matcher.rb
```diff
@@ -0,0 +1,25 @@
+module Grape
+  module NamedRouteMatcher
+    def method_missing(method_id, *arguments)
+      segments = arguments.first || {}
+
+      route = Grape::API.all_routes.detect do |r|
```
dm1try: Can be `Grape::API.all_routes` is memoized? or they change in the runtime?

lib/grape/named_route_matcher.rb
```diff
@@ -0,0 +1,25 @@
...
+        route_match?(r, method_id, segments)
```
dm1try: Could this `method_missing` use the "helper name" matching before do its job? It's too greedy for me. Seems you already have this method https://github.com/intridea/grape/pull/983/files#diff-64d352c6d46153e1e4ff964b0ff4ecdeR46 , maybe this matching can be done using this info?

dm1try:
@reprah  good job!

+1  to the implementation in a separate gem

dm1try:
one more question, we have this spec (".prepare_routes should be called once")- https://github.com/intridea/grape/blob/489cdc94a4e36a40f6707d55c565232336aae4fc/spec/grape/dsl/routing_spec.rb#L212
I mean, could be any side effects when you are using `.prepare_routes` directly instead of call `.routes`?

reprah:
@dblock @dm1try Thanks the brief reviews and recognizing the effort I put into this. The way you all respond to pull requests and users in general is a big reason why I want to write things for Grape in the first place (apart from the fact I use it regularly).

I do remember dblock saying before that Grape was trying to become more modular. This can be its own gem - I can only think of two downsides:

1.) From a code standpoint, it would require reopening class definitions. It was also useful to have the entire Grape test suite uncover edge cases in the code I added.

2.) From a community standpoint, you might get duplicate requests from people wanting named route helpers (if they don't realize/don't look for a gem that does this).

I'd be happy with having my own gem and incorporating advice from code reviews here. If I add to Grape's README or wiki and link my gem, maybe that can prevent some of number 2. Thoughts?

dblock:
Thanks for the kind words. Make a new gem, PR a link to the README. We should be treating such gems as *supported extensions*, so this should be one. At some point you should PR some very minimal tests into Grape proper that check that the extensions still work with the last shipped release of the extension.

Anything that cannot be done via a simple `extend Module` should be refactored in Grape to allow that.

reprah:
Is there another gem that's an "extension" to Grape that I can use as a reference? Does the word extension have any special meaning here (i.e. does it mean the gem interacts with the grape gem in a specific way)?

dblock:
@reprah I don't think so, you're pioneering something new!

dblock:
@reprah Did you get a change to move forward with this?

reprah:
@dblock I'm happily working on it! Thanks for checking in. I found some patterns I like in other grape-related gems, and I have a good idea of what kinds of tests would verify my gem still works with the current version of Grape.

dblock:
Bump, someone attempted something similar in https://github.com/intridea/grape/pull/1040, maybe we should converge on an implementation?

reprah:
How could we approach that?

I do like how their implementation allows custom method names.

dblock:
I think whoever releases a gem first wins :) Then we can iterate on it.

reprah:
Ha! Sounds good.

reprah:
I pushed my gem today, repository is here: https://github.com/reprah/grape-route-helpers
I did experience some regressions with testing with 0.12.0, so I think your idea of adding tests to Grape that test that my gem works with it is probably a good idea. I'd only need to verify that some of the attributes on Grape::Route objects have the expected format, I think.

dblock:
Cool. Open a PR with a 1-liner about named routes linking the gem from the README. It's a frequent-enough request that we should make it easily discoverable. 

I'll close this.

dblock:
Also add it to the WIKI.

PR #985, opened 2015-04-14
Demonstrate failing specs due to rack-accept parsing of media-type headers with underscores

jerrod:
The purpose of this pull request is to illuminate the issues with the Header versioning strategy when the vendor header uses an underscore `_`.

I have a pull request to address the parsing issues for rack-accept, however this change has the effect of breaking the recognition of Content-Types when using media_type headers.

Unfortunately I do not see a good way to address both of these issues while still relying on rack-accept for media_type and vendor parsing, so this is meant to start the conversation about how to address these issues.

[see also: https://github.com/mjackson/rack-accept/pull/17 ]

### DIFF
--- a/spec/grape/middleware/versioner/header_spec.rb
+++ b/spec/grape/middleware/versioner/header_spec.rb
@@ -4,276 +4,412 @@ describe Grape::Middleware::Versioner::Header do
   let(:app) { lambda { |env| [200, env, env] } }
   subject { Grape::Middleware::Versioner::Header.new(app, @options || {}) }
 
-  before do
-    @options = {
-      version_options: {
-        using: :header,
-        vendor: 'vendor'
+  context "with a vendor name utilizing an underscore" do 
+   before do
+      @options = {
+        version_options: {
+          using: :header,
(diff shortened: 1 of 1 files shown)


dblock:
It does look though that the problem is rack-accept and https://github.com/mjackson/rack-accept/pull/17 fixes it. I suggest reaching out to the contributors of rack-accept and getting someone to take a look to get that merged? @mjackson really.

jerrod:
I agree - my main concern with just having rack-accept patched and the dependency updated in Grape is that once we started using our own patched rack-accept we started also seeing the need to pass a very specific media type with Content-Type on any request with a body - e.g. passing application/json no longer worked an instead we needed to match the Accept header - as in application/vnd.whole_vendor_name-V1+JSON which seems wrong.

I am unable to determine if this is due to rack-accept or if this is handled in Grape, so any insight would be great if you have any.

dblock:
Run this test you have here with code from https://github.com/mjackson/rack-accept/pull/17, does the spec pass?

Issue #552, opened 2014-01-14
endpoints will be repeated after the reload


tumayun:
env
```
grape (0.6.1)
goliath (1.0.3)
rack (1.5.2)
```
posts.rb
```ruby
class Posts < Grape::API
  version 'v1', :using => :path
  format :json

  params do
    requires :id, type: Integer
  end

  resource 'shopping-carts' do
    get "/users/:id" do
    end

    post "/users" do
    end

    put "/users/:id" do
    end

    delete "/users/:id" do
    end
  end
end
```

```ruby
2.0.0p247 :001 > Posts.endpoints.size
 => 4 
2.0.0p247 :002 > load '/home/tumayun/workspace/shopping-cart-service/app/apis/posts.rb'
 => true 
2.0.0p247 :003 > Posts.endpoints.size
 => 8 
2.0.0p247 :004 > load '/home/tumayun/workspace/shopping-cart-service/app/apis/posts.rb'
 => true 
2.0.0p247 :005 > Posts.endpoints.size
 => 12 
2.0.0p247 :006 > load '/home/tumayun/workspace/shopping-cart-service/app/apis/posts.rb'
 => true 
2.0.0p247 :007 > Posts.endpoints.size
 => 16 
```

dm1try:
Fixed in #988

PR #988, opened 2015-04-16
Fixed duplicating endpoints

u2:
#552

### DIFF
--- a/lib/grape/dsl/routing.rb
+++ b/lib/grape/dsl/routing.rb
@@ -120,7 +120,8 @@ module Grape
             }).deep_merge(route_setting(:description) || {}).deep_merge(route_options || {})
           }
 
-          endpoints << Grape::Endpoint.new(inheritable_setting, endpoint_options, &block)
+          new_endpoint = Grape::Endpoint.new(inheritable_setting, endpoint_options, &block)
+          endpoints << new_endpoint unless endpoints.any?{ |e| e.equals?(new_endpoint) }
 
           route_end
           reset_validations!
--- a/spec/grape/dsl/routing_spec.rb
+++ b/spec/grape/dsl/routing_spec.rb
@@ -87,12 +87,17 @@ module Grape
             .to change{ subject.endpoints.count }.from(0).to(1)
         end
 
+        it 'does not duplicate identical endpoints' do
+          subject.route(:any)
+          expect { subject.route(:any) }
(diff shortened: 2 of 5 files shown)


spec/grape/dsl/routing_spec.rb
```diff
@@ -82,9 +82,17 @@ class Dummy
...
-        it 'defines a new endpoint' do
-          expect { subject.route(:any) }
-            .to change{ subject.endpoints.count }.from(0).to(1)
+        context 'defines new endpoint' do
+          it 'defines a new endpoint' do
+            expect { subject.route(:any) }
+              .to change{ subject.endpoints.count }.from(0).to(1)
+          end
+
+          it 'uniq the duplicating endpoint' do
```
dblock: You don't need to put these two tests together, just add a new test at the same level that says 'does not duplicate identical endpoints'.

dblock:
I understand the case of a reloader, but is this generally correct? What happens when you re-define an endpoint with all the same arguments, you probably would want the second one, not the first one, to be defined or to see a warning/error?

Maybe a better approach could be to freeze endpoints or something like that? Not sure if it can be accomplished.

Definitely needs a CHANGELOG, thanks.

lib/grape/dsl/routing.rb
```diff
@@ -120,7 +120,8 @@ def route(methods, paths = ['/'], route_options = {}, &block)
             }).deep_merge(route_setting(:description) || {}).deep_merge(route_options || {})
           }
 
-          endpoints << Grape::Endpoint.new(inheritable_setting, endpoint_options, &block)
+          new_endpoint = Grape::Endpoint.new(inheritable_setting, endpoint_options, &block)
+          endpoints << new_endpoint unless endpoints.any?{ |o| o.options == new_endpoint.options && o.inheritable_setting.to_hash == new_endpoint.inheritable_setting.to_hash }
```
dblock: I would want a comparison operator on `Grape::Endpoint` so we could maybe write `endpoints.include?(...)` or `endpoints.any? { |e| endpoint.equals?(e) } # or == `.

u2: Yes, The the comparison of `Endpoint` is meaningful.

u2:
Yes.I can't agree more.Before, I did not think so much about that.

dblock:
Merging.

PR #995, opened 2015-04-20
Added support for coercion to Set or Set[Other]

u2:
> I like Virtus's support for coercion to Sets as seen here:
https://github.com/solnic/virtus#collection-member-coercions

> They can be treated the same as Arrays in the case of `nil`, and changing line 41 from `@option[0]` to `@option.first` behaves the same for either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 #### Features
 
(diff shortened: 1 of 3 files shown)


u2:
#642

u2:
I think there are many works for going on `Set`.

dblock:
Perfect, thanks.

PR #996, opened 2015-04-20
Treat optional Set params default to Array

u2:
Is it reasonable for adding `Set` nested params?

### DIFF
--- a/spec/grape/validations/params_scope_spec.rb
+++ b/spec/grape/validations/params_scope_spec.rb
@@ -224,10 +224,15 @@ describe Grape::Validations::ParamsScope do
       get '/optional_type_array'
     end
 
+    before do
+      class Tweet < Set
+      end
+    end
+
     it 'errors with an unsupported type' do
       expect do
         subject.params do
-          group :a, type: Set do
+          group :a, type: Tweet do
             requires :b
           end
         end
@@ -235,7 +240,7 @@ describe Grape::Validations::ParamsScope do
 
       expect do
         subject.params do
-          optional :a, type: Set do
+          optional :a, type: Tweet do
             requires :b
           end
         end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -30,7 +30,7 @@ Metrics/ClassLength:
 
 # Offense count: 15
 Metrics/CyclomaticComplexity:
-  Max: 20
+  Max: 22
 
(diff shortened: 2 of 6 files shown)


spec/grape/validations/params_scope_spec.rb
```diff
@@ -227,7 +227,7 @@ def app
     it 'errors with an unsupported type' do
       expect do
         subject.params do
-          group :a, type: Set do
+          group :a, type: Rational do
```
dblock: Could you please explain this one?

dblock:
I would say "yes' if with 2 parameters with identical values you would get an error (or you would get only 1 value). Otherwise this has no chance of behaving any different than an array, no?

spec/grape/validations/params_scope_spec.rb
u2: Sorry, I was napping at that moment.

u2:
Yes.I think it's better to get only 1 value.And I think we need more specs for going on.

dblock:
Bump @u2.

u2:
Mm hmm, I nearly forgot this.I will do it later.But it may be not a easy work.So could you merge it firstly,and I will open another PR for further work.

dblock:
Lets add full Set support so that we don't have to change behavior for clients in the future. I'll leave this open, maybe @u2 or someone else can finish this?

u2:
Sorry about leaving aside the problem until now.
I reconsider about whether it's necessary for us and its rationality.
Firstly, for the params `Array` and `Set` are the same expression for http params.
Secondly, I have search issues about `Set` and I could not find anybody cared about that.

> Set implements a collection of unordered values with no duplicates. This is a hybrid of Array's intuitive inter-operation facilities and Hash's fast lookup.

In ruby, Set is different with `Array`.But, for us, it's just a uniq `Array`,and if it was a `Set`, what about the meaning of `unordered values`?

Before I committed this PR about `Set`, just because I saw something like
https://github.com/ruby-grape/grape/blob/d721c5d7643eea6f2ecfc92ef9dc4c529daf9762/lib/grape/validations/types.rb#L54.

What's your opinion? @dblock

dblock:
I think it could be valuable, a Set is a fairly basic structure. But clearly not many people care :) I'll just leave this open, feel free to finish it.

PR #997, opened 2015-04-21
Add SplitValidator

u2:
https://gist.github.com/jordansexton/260662acf75e66d9bf79


dblock:
This is a very esoteric validator, frankly I am against merging it. Why would we encourage anyone to pass data this way when URI parsing supports arrays?

u2:
Yes, I agree about that.

dm1try:
For me it's not a validator. We already have two "special validators" that can modify the params(`coerce` and `default`) but I don't really like that.

dblock:
I am closing this without merging, thanks for your suggestion @u2!

PR #1000, opened 2015-04-24
Add clarification about minitest 5.6.0 errors

kkirsche:
Add clarification about minitest 5.6.0 errors


kkirsche:
Oops. My mistake, I had loaded 5.5.1 into my grape app accidently

PR #1007, opened 2015-04-30
Fix assert_equal arguments order in README

akicho8:
Fixed order of the arguments of assert_equal that was supposed to reverse.


dblock:
Makes sense.

Issue #920, opened 2015-02-11
desc dsl with a block will ignore given options hash


calfzhou:
grape v0.10.1

see [my inline comment](https://github.com/intridea/grape/commit/0e13e9a06e66c05e270c14da380bd7407a93886b#diff-dfd505897d6f335cd86741ef7fc2c90aL22) on commit 0e13e9a06e66c05e270c14da380bd7407a93886b .

With the following code, i could get `my_options` out of `desc`:
```ruby
desc 'Gets something', my_option: 'blah blah'
```

But when using a block, `my_option` is missing. Is this by-design, or a bug?
```ruby
desc 'Gets something', my_option: 'blah blah' do
    detail 'Detail info'
end
```

calfzhou:
If it's by-design, how can i specify custom options to a desc dsl?

I'm using [grape-swagger](https://github.com/tim-vandecasteele/grape-swagger) to generate swagger document, it has some features need specify custom options to desc dsl (such as [`hidden`](https://github.com/tim-vandecasteele/grape-swagger#hiding-an-endpoint) to hide an endpoint, [`notes`](https://github.com/tim-vandecasteele/grape-swagger#markdown-in-notes) to add more notes to an endpoint. But now these additional options will missing if a block is present.

dblock:
I think we want a strong, extensible DSL, and to deprecate the `options` way of adding things. I think what we need is:

* A documented way to add other DSL here, like `notes ...`.
* It looks like we broke grape-swagger on notes, so I think that needs a clear explanation of how to deal with it in UPGRADING. Probably the best would be to document that notes -> detail, and make that work, but I am not 100% sure.

calfzhou:
Thanks for the quick response!

By the way, since grape-swagger doesn't leverage `detail` dsl, i'm confusing about the difference between these two terms, `detail` and `notes`.

dblock:
I think we should make detail in the block DSL = notes in the options DSL.

calfzhou:
That would be great! :+1:

uppfinnarn:
Any updates on this? The detail -> notes thing sounds like a really quick fix for a rather annoying bug, and getting that alone down for now would work until a more extensible DSL is worked out.

Right now, `detail` can be considered plain ol' broken with `grape-swagger`, which means I have nowhere to put important information such as rate limiting, a warning that one of my endpoints return a bloody huge response unless filtered (?q=...), etc.

calfzhou:
[...]

PR #1008, opened 2015-05-04
make route detail = notes

calfzhou:
ref #920

### DIFF
--- a/lib/grape/route.rb
+++ b/lib/grape/route.rb
@@ -5,6 +5,10 @@ module Grape
       @options = options || {}
     end
 
+    def route_notes
+      route_details
+    end
+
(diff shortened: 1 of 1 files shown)


dblock:
This seems like a hack for grape-swagger, I don't think we want to merge this as is. Maybe this is something that we want in grape-swagger?

If there's a compelling argument to merge it, this needs a CHANGELOG entry and tests.

dblock:
Bump @calfzhou.

rngtng:
:-1: this will fix it instead: https://github.com/tim-vandecasteele/grape-swagger/pull/286

dblock:
:+1: @rngtng I am going to close this without merging.

PR #1016, opened 2015-05-15
Update readme to include content type in case content type validation…

JohnMorales:
If you try to upload a file with the current curl statement, chances are you'll get an error since the latest grape encourages you to enable validation, and one would assume that most will use content type validation.


dblock:
thx

PR #1017, opened 2015-05-19
Fixed validation of invalid params with mutual_exclusive inside hash

lasseebert:
This fixes an issue with `mutual_exclusive` params inside a `type: Hash` param.

With these params:

```ruby
params do
  optional :wine, type: Hash do
    optional :grape
    optional :country
    mutually_exclusive :grape, :country
  end
end
```

And the request params `{ wine: "This is a string, not a hash" }`

I'd expect an error message, but instead get an unhandled error:

    undefined method `stringify_keys' for "This is a string, not a hash":String


dblock:
Can you please squash/rebase these? Thanks.

lasseebert:
Rebased to latest master

lasseebert:
...and squashed :)

dblock:
Merged, thanks.

lasseebert:
Thanks :)

PR #1027, opened 2015-06-03
Pretty-print json by default

patscan:


dblock:
We definitely don't want this in an API, if you need to look at pretty JSON when invoking an API, try a plugin like JSONView in Chrome.

patscan:
Thanks for the reply, didn't mean to open a PR to grape itself but instead
to my own fork.

No worries if you don't have time to reply but wondering:
- Is there a possibility that this could be added as an option for output
in the future?
- Is your main concern performance?

On Wed, Jun 3, 2015 at 10:44 AM, Daniel Doubrovkine (dB.) @dblockdotorg <
notifications@github.com> wrote:

> We definitely don't want this in an API, if you need to look at pretty
> JSON when invoking an API, try a plugin like JSONView in Chrome.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/intridea/grape/pull/1027#issuecomment-108542949>.
>



-- 
patscan@gmail.com
415.724.6575

dblock:
You can write a formatter that does this or just do `formatter :json, ->(object, options) { ... }` in your API, so I don't think we'll want to add it to Grape unless it becomes something everybody wants all the time. 

Performance-wise you're rendering then parsing JSON, that can't be good, mostly I don't see why this is a useful feature, happy to be convinced otherwise.

patscan:
Thanks again for the reply, Daniel. The use case I'm thinking of is an API
that uses the grape-active_model_serializers gem for serialization. When
this is the case, the formatter call in my API looks like so:

formatter :json, Grape::Formatter::ActiveModelSerializers

When I use a custom formatter like you describe, I'm able to pretty-print,
but, I'm no longer invoking serializer, and my API returns every field from
my DB directly. Since grape-active_model_serializers ends up calling
Grape's JSON formatter (
https://github.com/jrhe/grape-active_model_serializers/blob/master/lib/grape-active_model_serializers/formatter.rb#L11),
I thought I could perhaps update Grape's formatter itself and I would get
serialization + pretty_print output.

Maybe I need to look at a patch to  grape-active_model_serializers rather
than Grape itself if I want pretty-printing + serialization.

On Wed, Jun 3, 2015 at 11:06 AM, Daniel Doubrovkine (dB.) @dblockdotorg <
notifications@github.com> wrote:

> You can write a formatter that does this or just do formatter :json,
> ->(object, options) { ... } in your API, so I don't think we'll want to
> add it to Grape unless it becomes something everybody wants all the time.
>
> Performance-wise you're rendering then parsing JSON, that can't be good,
> mostly I don't see why this is a useful feature, happy to be convinced
> otherwise.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/intridea/grape/pull/1027#issuecomment-108550770>.
>



-- 
patscan@gmail.com
415.724.6575

dblock:
Just write a new custom formatter that does everything you want, maybe based on Grape::Formatter::ActiveModelSerializers

patscan:
You're right - that method worked perfectly for me. Appreciate the time and
for pointing me in the right direction.

On Wed, Jun 3, 2015 at 2:09 PM, Daniel Doubrovkine (dB.) @dblockdotorg <
notifications@github.com> wrote:

> Just write a new custom formatter that does everything you want, maybe
> based on Grape::Formatter::ActiveModelSerializers
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/intridea/grape/pull/1027#issuecomment-108616827>.
>



-- 
patscan@gmail.com
415.724.6575

PR #1032, opened 2015-06-13
fixed hashes

askl56:
Updated Hash syntax


dblock:
The spec now fails for a good reason. The custom formatter example just tries to call `to_s` on a hash, therefore producing the correct output that was in the spec. You can rewrite the spec if you want. The descriptions in the specs do make sense, if we're going to be pedantic, add a space after `cascade:[HERE]false`.

dblock:
Closed via https://github.com/ruby-grape/grape/commit/af74bdf2ac7b29aedd6d12733a0d948d213880d6.

Issue #891, opened 2015-01-13
Cloned InheritableSetting seems to clone unnecessary stuff


elado:
#739 creates a `point_in_time_copy` of route configuration. It `deep_dup`s everything in the hash, including classes. Problem is:

```ruby
desc "...", {
  entity: ArticlePresenter
}
```

The `ArticlePresenter` class is also getting `deep_dup`ed, which produces a `#<Class:0x...>` which leads to problems with equality and also not copying all class methods. It doesn't play nice with grape-swagger in some cases.

Currently I solved it by adding

```ruby
class Grape::Entity
  def self.deep_dup
    self
  end
end
```

I think I understand the need to make a current copy, but do you think it's OK to take out classes from the `deep_dup`?

dblock:
Some of the dup'ing introduced in #739 was eventually removed, because we couldn't come up with a single failing test case. You should try with this one, too.

In theory, the not duping classes makes sense to me.

elado:
I'm referring specifically to https://github.com/intridea/grape/blob/master/lib/grape/util/inheritable_setting.rb#L50
It uses ActiveSupport's clone and deep_dup. So to prepare a more efficient copy it needs to copy only instances, hashes, arrays etc.

`grape-swagger` for example, can show hundreds of `models` from all APIs, even when there are only a few unique.

What was the specific case to make a point in time copy?

dblock:
I am not sure, @dspaeth-faber can speak to it. I think a lot of this was theoretical to avoid modifying a parent's namespace's settings down the nesting path.

dspaeth-faber:
Well, it's not easy to explain, I think. You need point in time copies because values can change. 
This happens when configurations are made for a namespace.  Some thing like this:

```ruby

namespace :statuses do
  params do
    requires :user_id, type: Integer, desc: "A user ID."
  end
  namespace ":user_id" do
    desc "Retrieve a user's status."
    params do
      requires :status_id, type: Integer, desc: "A status ID."
    end
    get ":status_id" do
      User.find(params[:user_id]).statuses.find(params[:status_id])
    end
  end

 # now the inherited param should be user_name and not user_id
# to solf this you need point in time copies
  params do
     requires :user_name, type: String, desc: "A user name."
  end
  namespace ":user_name" do
     ... 
  end
end

```

I think there have been other examples too.
The approach to generate the config at "compile" time / when a class is loaded and evaluted
make things hard. A better way would be to create `Commands` and
[...]

PR #1034, opened 2015-06-15
Only dup StackableValues

toddmazierski:
## Summary

To fix #930, settings managed with `StackableValues` (including custom middleware) are now `dup`ed instead of `deep_dup`ed when `mount` is used.

## Impact

Although it's a small change in terms of character count, the impact seems large. One category of affected code appears to be "custom middleware that maintains global state" like [Warden's test hook system](https://github.com/hassox/warden/blob/master/lib/warden/test/warden_helpers.rb), but there may be many others.

Please note that to reduce the impact, _only_ `StackableValues` has been modified. `InheritableValues`, which also persists settings, still uses `deep_dup`.  If this inconsistency is confusing or otherwise undesirable, I'm happy to drop `deep_dup` usage (and support) entirely. Regardless, I'm hoping the authors of #739 can weigh in with any suggestions!

## Background

Here are the events that led to me to this change:

  * [My original bug report](https://github.com/intridea/grape/pull/930) about Warden's `login_as` test helper, where I guess `deep_dup`ing is responsible
  * A [deep dive](https://github.com/intridea/grape/pull/930#issuecomment-76635077) into the effects of `deep_dup` on Warden
  * [@dblock suggests](https://github.com/intridea/grape/pull/930#issuecomment-76754338) we develop a white-list for settings that _need_ to be `deep_dup`ed
  * However, [the tests suggest](https://github.com/intridea/grape/commit/009be00e4f6d149072ef065abeea6c3bc845ea43) that _no_ settings need to be `deep_dup`ed

## Testing

  * I've added a test case to `stackable_values_spec.rb` that closely resembles the root cause of #930
  * I've also confirmed that the bug is fixed in my real Grape project

Beyond this, perhaps we could have a few volunteers that:

  * have complex configurations and/or
  * use a lot of custom middleware
 
&hellip;try this change in real projects?

## Related Issues

  * #739: when `deep_dup`ing settings was introduced, later released in [v0.10.0](https://github.com/intridea/grape/releases/tag/v0.10.0)
  * #930: my original bug report about Warden's `login_as` test helper
  * #891: another reference to negative effects of `deep_dup`ing

### DIFF
--- a/lib/grape/util/stackable_values.rb
+++ b/lib/grape/util/stackable_values.rb
@@ -45,7 +45,7 @@ module Grape
       def initialize_copy(other)
         super
         self.inherited_values = other.inherited_values
-        self.new_values = other.new_values.deep_dup
+        self.new_values = other.new_values.dup
       end
     end
   end
--- a/spec/grape/util/stackable_values_spec.rb
+++ b/spec/grape/util/stackable_values_spec.rb
@@ -109,6 +109,16 @@ module Grape
 
           expect(grandchild.clone.to_hash).to eq(some_thing: [:foo, [:bar, :more], :grand_foo_bar], some_thing_more: [:foo_bar])
         end
+
+        context 'complex (i.e. not primitive) data types (ex. middleware, please see bug #930)' do
+          let(:middleware) { double }
+
+          before { subject[:middleware] = middleware }
+
+          it 'copies values; does not duplicate them' do
+            expect(obj_cloned[:middleware]).to eq [middleware]
+          end
+        end
       end
     end
   end


dblock:
:clap: @toddmazierski

dblock:
I am running this overnight on a failry large battery of tests here at [artsy](https://www.artsy.net), via: 

```
gem 'grape', github: 'generalassembly/grape', branch: 'only-dup-stackable-values'
```

Results tomorrow.

dblock:
My tests passed, I am going to merge this.

Issue #873, opened 2015-01-02
Rubinus: Unable to clone/dup String class (TypeError)


dblock:
See https://github.com/dblock/grape-on-rack/commit/de6c0022d203785fa0a25a6d02e8c85fe12aa374.

This needs a repro + spec, but I think it's as simple as having a class in the parameters of an API, Grape mount will attempt to dup the whole thing and Rubinius refuses to dup classes.

rnubel:
Hello! I did some research into this, and came up with a fix (#1038).

The root cause is that the `duplicable?` method (added by ActiveSupport and/or the backport in Grape) on String returns `true`, but in Rubinius, the String class actually [can't be duped](https://github.com/rubinius/rubinius/blame/60cfdadf0149ca0efebf2b924a0482e4cfa1b404/kernel/common/string.rb#L39).

```ruby
# ruby-2.1.5
irb(main):002:0> String.duplicable?
=> true
irb(main):003:0> { :foo => String }.deep_dup
=> {:foo=>#<Class:0xa61ec70>}

# rbx-2.5.5
irb(main):002:0> String.duplicable?
=> true
irb(main):003:0> { :foo => String }.deep_dup
TypeError: Unable to clone/dup String class
        from kernel/common/string.rb:39:in `dup (clone)'
        from /home/robert/.gem/rbx/2.1.0/gems/activesupport-4.2.2/lib/active_support/core_ext/object/deep_dup.rb:14:in `deep_dup'
        from /home/robert/.gem/rbx/2.1.0/gems/activesupport-4.2.2/lib/active_support/core_ext/object/deep_dup.rb:43:in `deep_dup'
        from kernel/common/enumerable.rb:83:in `each_with_object'
        from kernel/common/hash.rb:342:in `each'
        from kernel/common/enumerable.rb:81:in `each_with_object'
        from /home/robert/.gem/rbx/2.1.0/gems/activesupport-4.2.2/lib/active_support/core_ext/object/deep_dup.rb:42:in `deep_dup'
        from (irb):3
```

This affects InheritableValues' [usage of `deep_dup`](https://github.com/intridea/grape/blob/master/lib/grape/util/inheritable_values.rb#L39), so it blows up whenever a hash including a String-class key or value is duped (e.g., when you have an endpoint with a parameter that you describe with the String-class type).

[...]

PR #1038, opened 2015-06-19
Don't try to dup the String class

rnubel:
This addresses #873 by preventing ActiveSupport#deep_dup from trying to duplicate the String class (note: I mean strictly the String class, not String instances). However, there are a few points to consider:

* Is this change worth making? The issue likely doesn't affect too many people, but the fix would apply to everyone.
* If yes to the above, should it be pushed up to ActiveSupport and/or left as a monkey-patch in Grape?
* Are there other classes in Rubinius with landmines like [this one](https://github.com/rubinius/rubinius/blame/60cfdadf0149ca0efebf2b924a0482e4cfa1b404/kernel/common/string.rb#L39)?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 Next Release
 ============
 
-* Your contribution here.
+#### Features
+
+#### Fixes
+* [#1038](https://github.com/intridea/grape/pull/1038): Avoid dup-ing the String class when used in inherited params - [@rnubel](https://github.com/rnubel).
 
 0.12.0 (6/18/2015)
 ==================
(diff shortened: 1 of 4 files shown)


dblock:
Oh wow, this is interesting. I vote to merge it, but I would like a CHANGELOG entry, please.

rnubel:
Changelog updated. Also, I've submitted an upstream patch to Rails for this: https://github.com/rails/rails/pull/20640

dblock:
Merged squashed via https://github.com/intridea/grape/commit/a37df3e5421f9e5d8a3c600fea5742d7f33e37c0.

Issue #993, opened 2015-04-17
Allow definition of custom types


lpil:
Hello! Thank you for Grape. :)

We would like to be able to do something like this:

```ruby
params do
  requires :id, type: UUID, desc: 'Content id.'
end
```

Rather then set the type as String, and then validate the format of that string. We realise we can achieve the same result using validators, but we think this would be a more elegant, declarative syntax.

Is this possible?

Thanks,
Louis

dblock:
It probably is with a bit of work - you need to dig through the implementation because `type` is something special. Grape's validation is based on Virtus (https://github.com/solnic/virtus), so I think defining a Virtus type for UUID would be sufficient. I don't know exactly how, will leave this open as a feature request.

dblock:
Closed via https://github.com/intridea/grape/pull/1039.

PR #1039, opened 2015-06-19
Support custom types

rnubel:
This implements #993. Originally, I thought it'd be possible to do this simply by having the custom type extend Virtus::Attribute, but a) it turned out to not be that easy and b) I figured it'd be better to define a pattern for this that isn't directly coupled to Virtus.

So, my change allows you to take any class and coerce parameters into it. All that's required is that the class implement a `from_param` class method. This name is arbitrary, but I feel it matches up with the intent of the change... totally open to suggestions, though. As a quick example:

```ruby
class Color
  attr_reader :value
  def initialize(color)
    @value = color
  end
  def self.from_param(value)
    raise "Invalid color" unless %w(blue red green).include?(value)
    new(value)
  end
end

# ...

params do
  required :color, type: Color, default: Color.new('blue')
end
get '/stuff' do
  # params[:color] is already a Color.
  params[:color].value
end

# /bin/sh
$ curl localhost:9292/api/stuff?color=red
"red"
```

Things to consider:
* If coercion fails, it's hard to tell why. The global rescue inside `CoerceValidator#coerce_value` hides any errors inside the coercion method with a "is invalid" message. This is probably acceptable for a first pass, though.
* Should my tests be in `param_scope_spec` or `coerce_spec`? I'm leaning towards the latter, even though they're currently in the former.
* Is there a better/cleaner approach here than checking for the existence of the `from_param` method?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ Next Release
 
 #### Features
 
-* Your contribution here.
+* [#1039](https://github.com/intridea/grape/pull/1039): Added support for custom parameter types - [@rnubel](https://github.com/rnubel).
+* Your contribution here!
 
 #### Fixes
 
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@
   - [Declared](#declared)
   - [Include Missing](#include-missing)
 - [Parameter Validation and Coercion](#parameter-validation-and-coercion)
+  - [Supported Parameter Types](#supported-parameter-types)
+  - [Custom Types](#custom-types)
   - [Built-in Validators](#built-in-validators)
   - [Namespace Validation and Coercion](#namespace-validation-and-coercion)
   - [Custom Validators](#custom-validators)
@@ -727,6 +729,54 @@ The correct implementation is to ensure the default value passes all validations
 ```ruby
 params do
   optional :color, type: String, default: 'blue', values: ['blue', 'red', 'green']
+end
+```
+
+#### Supported Parameter Types
+
+The following are all valid types, supported out of the box by Grape:
+
+* Integer
+* Float
+* BigDecimal
+* Numeric
+* Date
+* DateTime
+* Time
+* Boolean
+* String
+* Symbol
+* Rack::Multipart::UploadedFile
+
+#### Custom Types
+
+Aside from the default set of supported types listed above, any class can be
+used as a type so long as it defines a class-level `parse` method. This method
+must take one string argument and return an instance of the correct type, or
+raise an exception to indicate the value was invalid. E.g.,
+
+```ruby
+class Color
+  attr_reader :value
+  def initialize(color)
+    @value = color
+  end
+
+  def self.parse(value)
+    fail 'Invalid color' unless %w(blue red green).include?(value)
+    new(value)
+  end
+end
+
+# ...
+
+params do
+  requires :color, type: Color, default: Color.new('blue')
+end
+
+get '/stuff' do
+  # params[:color] is already a Color.
+  params[:color].value
 end
 ```
 
--- a/lib/grape/validations/validators/coerce.rb
+++ b/lib/grape/validations/validators/coerce.rb
@@ -56,7 +56,14 @@ module Grape
         return val || Set.new if type == Set
         return val || {}      if type == Hash
 
-        converter = Virtus::Attribute.build(type)
+        # To support custom types that Virtus can't easily coerce, pass in an
+        # explicit coercer. Custom types must implement a `parse` class method.
+        converter_options = {}
+        if ParameterTypes.custom_type?(type)
+          converter_options[:coercer] = type.method(:parse)
+        end
+
+        converter = Virtus::Attribute.build(type, converter_options)
         converter.coerce(val)
 
       # not the prettiest but some invalid coercion can currently trigger
--- a/lib/grape.rb
+++ b/lib/grape.rb
@@ -22,7 +22,6 @@ require 'active_support/core_ext/object/conversions'
 require 'active_support/core_ext/array/extract_options'
 require 'active_support/core_ext/hash/deep_merge'
 require 'active_support/dependencies/autoload'
-require 'grape/util/content_types'
 require 'multi_json'
 require 'multi_xml'
 require 'virtus'
@@ -159,6 +158,9 @@ module Grape
     autoload :Presenter
   end
 end
+
+require 'grape/util/content_types'
+require 'grape/util/parameter_types'
 
 require 'grape/validations/validators/base'
 require 'grape/validations/attributes_iterator'
--- a/lib/grape/util/parameter_types.rb
+++ b/lib/grape/util/parameter_types.rb
@@ -1,0 +1,58 @@
+module Grape
+  module ParameterTypes
+    # Types representing a single value, which are coerced through Virtus
+    # or special logic in Grape.
+    PRIMITIVES = [
+      # Numerical
+      Integer,
+      Float,
+      BigDecimal,
+      Numeric,
+
+      # Date/time
+      Date,
+      DateTime,
+      Time,
+
+      # Misc
+      Virtus::Attribute::Boolean,
+      String,
+      Symbol,
+      Rack::Multipart::UploadedFile
+    ]
+
+    # Types representing data structures.
+    STRUCTURES = [
+      Hash,
+      Array,
+      Set
+    ]
+
+    # @param type [Class] type to check
+    # @returns [Boolean] whether or not the type is known by Grape as a valid
+    #   type for a single value
+    def self.primitive?(type)
+      PRIMITIVES.include?(type)
+    end
+
+    # @param type [Class] type to check
+    # @returns [Boolean] whether or not the type is known by Grape as a valid
+    #   data structure type
+    # @note This method does not yet consider 'complex types', which inherit
+    #   Virtus.model.
+    def self.structure?(type)
+      STRUCTURES.include?(type)
+    end
+
+    # A valid custom type must implement a class-level `parse` method, taking
+    #   one String argument and returning the parsed value in its correct type.
+    # @param type [Class] type to check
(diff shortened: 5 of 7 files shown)


README.md
```diff
@@ -730,6 +731,36 @@ params do
...
+# ...
+
+params do
+  required :color, type: Color, default: Color.new('blue')
+end
+get '/stuff' do
+  # params[:color] is already a Color.
+  params[:color].value
+end
+```
```
dblock: Run this "code" through RuboCop, just to keep things consistent, change " to ', add a space between methods, etc.

README.md
```diff
@@ -730,6 +731,36 @@ params do
...
+  def self.from_param(value)
+    raise "Invalid color" unless %w(blue red green).include?(value)
+    new(value)
+  end
+end
+
+# ...
+
+params do
+  required :color, type: Color, default: Color.new('blue')
```
dblock: It's `requires`.

dblock:
This is great. Needs a CHANGELOG entry, please. 

I am not in love with the name `from_param`. Can we brainstorm better names? Maybe `coerce`?

That said, I think the constructor should also be the coercer. This avoids the whole interface problem and allows you to use pretty much any type, no?

```ruby
class Color
    attr_reader :value
    def initialize(value)
        fail 'Invalid color' unless %w(blue red green).include?(value)
        @value = value
    end
end
```

rnubel:
Thanks for the feedback! Using the initializer would be clean, but also a bit limiting. It's fine if you're creating the custom type from scratch for just this purpose, but it would restrict you if you wanted to use an existing class as a type. Suppose you already had a Color class which initializes with R, G, and B values, for example.

Let me know what you think. If we go with the extra-class-method approach, I think `coerce` is a pretty good name.

dblock:
It could just be Monday, but I don't see why it wouldn't work with an existing class, care to provide an example?  :)

rnubel:
Maybe I'm misunderstanding, but suppose we already have a Color class used by the rest of our API, like so:

```ruby
class Color
  def initialize(r, g, b)
    @r, @g, @b = r, g, b
  end
end
```

The constructor takes three integer arguments, so it couldn't directly be used to coerce the value. However, if we add a `::coerce` method like so:

```ruby
class Color
  def self.coerce(param)
    @r, @g, @b = param.unpack('A2A2A2').map(&:hex)
  end
end
```

Then you could hit endpoints like `/api/pixel?color=FF0000` and it would parse the `color` param into the same thing as `Color.new(255,0,0)`.

dblock:
In the example above you would still have to extend `Color` with a `coerce` method that doesn't respect the original wishes of `Color`, is it worse to rewrite an initializer?

Some library gives you `Color`:

```ruby
class Color
  def initialize(r, g, b)
    @r, @g, @b = r, g, b
  end

  def to_s
    "#{@r}|#{@g}|#{@b}"
  end
end
```

you open it up:

```ruby
class Color
  alias_method :_initialize, :initialize
  def initialize(* args)
    if args.length == 3
      _initialize(*args)
    elsif args.length == 1
      @r, @g, @b = args.first.unpack('A2A2A2').map(&:hex)
    else
      fail 'Oh no.'
    end
  end
end
```

And:

```ruby
2.0.0-p598 :183 >   Color.new(10, 20, 55)
 => #<Color:0x007f8e5595ceb0 @r=10, @g=20, @b=55> 
2.0.0-p598 :184 > Color.new('FFCC11')
 => #<Color:0x007f8e55965718 @r=255, @g=204, @b=17> 
```

Now that I wrote it I think it's worse. 

Ruby actually has a way to deal with this for many times, eg. DateTime. I think the method should be called `parse`, not `coerce`, this is how most classes support parsing a String into an instance of themselves.

rnubel:
Yeah, `parse` definitely expresses the intent of the method, and aligns with existing Ruby standards. My only concern is that [my implementation](https://github.com/rnubel/grape/blob/custom_param_types/lib/grape/validations/validators/coerce.rb#L61) would then directly set the coercer for types that already respond to `parse`, like `DateTime` and `Date`. Granted, [that's all Virtus' coercible gem does anyway](https://github.com/solnic/coercible/blob/master/lib/coercible/coercer/string.rb#L110), so it should be fine. And, I did a quick check and none of the other common types (Integer, Float, Hash, Array) support the `parse` method.

I'll push up a change to make it `parse` instead of `from_param`. Let me know what you think.

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ Next Release
 
 #### Features
 
-* Your contribution here.
```
dblock: Put back this line below yours ;)

lib/grape/validations/validators/coerce.rb
```diff
@@ -54,7 +54,14 @@ def coerce_value(type, val)
         return val || Set.new if type == Set
         return val || {}      if type == Hash
 
-        converter = Virtus::Attribute.build(type)
+        # To support custom types that Virtus can't easily coerce, pass in an
+        # explicit coercer. Custom types must implement a `parse` class method.
+        converter_options = {}
+        if ![DateTime, Date, Time].include?(type) && type.respond_to?(:parse)
```
dblock: I think we want a coercer if this is not a known Virtus type or something like that, no? It feels wrong that we're checking for a list of known types, or at least that responsibility should be elsewhere (in Virtus or an extension of Virtus). I think the logic should be something like this:

```ruby
converter_options = {}
case type
when Virtus::Attribute.something_that_can_be_coerced_or_virtus_knows_about_it?
   # do nothing
when type.respond_to?(:parse)
   convert_options[:coercer] = ...
end
```

What do you think?

dblock: Also if Date types are potentially special, make sure to write a test around them explicitly.

dblock:
See comments above and squash your commits, please.

lib/grape/validations/validators/coerce.rb
rnubel: So I haven't been able to find a "good" way to identify if Virtus knows about a type or not... this works, but I think it relies too much on internal knowledge to be usable:

```ruby
def virtus_type?(type)
  Coercible::Coercer.const_get(type.to_s, false) 
end
```

However, it occurs to me that Grape doesn't have any single place where the parameter types you can use are listed. Nowhere in the README is there a list of what types one can use; it's just shown by example that String, Integer, Float, Date, and DateTime are the useful ones. What if we made the list of "default types" explicit, and simply checked whether or not to coerce by looking at our list of default types?

The added benefit of doing that is that we could detect invalid parameter types at runtime. I initially tried to add an error along the lines of `you're using a custom type but it doesn't respond to :parse` to this PR, but was stymied by the fact most of the default types would trigger that error. If we could immediately identify that a custom type is being used, we could add more helpful error messages and make things easier on developers.

I'm thinking something along the lines of `Grape::ParameterTypes`, ala `Grape::ContentTypes`. What do you think?

dblock: I think it's a good idea, go for it!

rnubel:
@dblock changes made. I'm not entirely sure the `ParameterTypes` module is worth it, as-is, so let me know what you think.

Part of the problem is, Grape delegates coercion directly to Virtus, and Virtus' coercion logic is pretty complex -- for example, if you specify a type which has included `Virtus.model`, Grape will handle it just fine (this is why the "complex type" logic as referenced in #1043 happens to actually work). So, deciding up-front whether or not we'll be able to coerce a given type is almost impossible. It might be a good long-term decision to start enforcing a list of allowed types, in order to decouple Grape from Virtus' inner workings. But that's not for this PR.

lib/grape/validations/validators/coerce.rb
```diff
@@ -56,7 +56,14 @@ def coerce_value(type, val)
         return val || Set.new if type == Set
         return val || {}      if type == Hash
 
-        converter = Virtus::Attribute.build(type)
+        # To support custom types that Virtus can't easily coerce, pass in an
+        # explicit coercer. Custom types must implement a `parse` class method.
+        converter_options = {}
+        if ParameterTypes.custom_type?(type)
+          converter_options[:coercer] = type.method(:parse)
+        end
```
dblock: I am surprised that RuboCop didn't flag this one, the if would go after the line. Maybe we autoignored it, if you don't mind looking into it for a separate PR?

lib/grape/validations/validators/coerce.rb
```diff
@@ -56,7 +56,14 @@ def coerce_value(type, val)
         return val || Set.new if type == Set
         return val || {}      if type == Hash
 
-        converter = Virtus::Attribute.build(type)
+        # To support custom types that Virtus can't easily coerce, pass in an
+        # explicit coercer. Custom types must implement a `parse` class method.
+        converter_options = {}
+        if ParameterTypes.custom_type?(type)
+          converter_options[:coercer] = type.method(:parse)
```
dblock: Nitpick: if ParameterTypes knows `custom_type?` it should also have `ParameterTypes.custom_type_parser_for(type)` or something like that, otherwise there's an assumption that custom_type means parse exists in two places.

dblock:
This is great. If you want to address any of the above in a future PR, please feel free. Merging.

lib/grape/validations/validators/coerce.rb
rnubel: What are you referring to? It looks like [only a few cops are turned off](https://github.com/intridea/grape/blob/master/.rubocop_todo.yml) and I don't think any of them apply here.

lib/grape/validations/validators/coerce.rb
rnubel: Agreed. The responsibility for knowing how to coerce a custom type should live in just one place. I'll put that refactoring into a separate PR.

lib/grape/validations/validators/coerce.rb
dblock: I'd have to dig :) I think it should tell you to do `converter_options[:coercer] = type.method(:parse) if ...`. It's really nbd.

Issue #787, opened 2014-10-21
Named routes for resources?


midu:
I am trying to link to across resources, but I can't find a way to generate the path/url to one without hardcoding it.

For example, I thought I would be able to achieve something like this (loosely based on [GitHub's API](https://api.github.com/users/midu)):

``` json
{
  "login": "midu",
  "name": "Stefano",
  "repos_url": "https://api.github.com/users/midu/repos" // <- reference to other resources
}
```

Here's a quick example of what I am trying to do:

``` rb
module Entities
  class User < Grape::Entity
    expose :id
    expose :login
    expose :name

    expose :repos_url

    private

    def repos_url
      '???'
    end
  end
end

resource :users do
  desc "Return a user."
  params do
    requires :login, type: String, desc: "GitHub login"
  end

  route_param :login do
    get do
      user = User.find(params[:login])
      present user, with: Entities::User
    end

    get :repos do
      User.find(params[:login]).repos
    end
  end
end
```

What would be the best practice to writing `repos_url` here?

If that matters, I am using Grape with Rails and `grape-entities`, I don't think it really changes anything. I am open to other configurations.

PR #1040, opened 2015-06-19
Add support for named routes

hedgesky:
This PR implements support for named routes. It's related to #787 issue. I added possibility to access routes tagged with :as option and compile path from them via substituting parameters. Example:

```ruby
module Twitter
  class API < Grape::API
    prefix :api

    resource :statuses do
      desc 'Route with route_param.'
      params do
        requires :id, type: Integer
      end
      route_param :id do
        get as: :status do
          Twitter::API.get_named_path(:status, id: params[:id], format: 'json')
          # returns "/api/statuses/1.json" when params[:id] == '1'

          Twitter::API.get_named_path(:status, id: params[:id])
          # returns "/api/statuses/1" when params[:id] == '1'

          Twitter::API.get_named_path(:wrong_route, id: params[:id])
          # fails with Grape::NamedRoutes::NamedRouteNotFound

          Twitter::API.get_named_path(:status)
          # fails with Grape::NamedRoutes::MissedRequiredParam

        end
      end
    end
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@
 - [Current Route and Endpoint](#current-route-and-endpoint)
 - [Before and After](#before-and-after)
 - [Anchoring](#anchoring)
+- [Named Routes](#named-routes)
 - [Using Custom Middleware](#using-custom-middleware)
   - [Rails Middleware](#rails-middleware)
     - [Remote IP](#remote-ip)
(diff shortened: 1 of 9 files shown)


dblock:
This has been attempted in https://github.com/intridea/grape/pull/983, too.

I think the DSL that we want is something more Rails-like. For example `status_path`, instead of `get_named_path( ... )`. This is pretty close from enabling that, want to try?

It also looks like this is something that be its own gem, and maybe we can start modularizing Grape further this way?

hedgesky:
I've just added dynamic `path` DSL (forgot to change README but there are corresponding lines in specs). Anyway, #983 looks like a far more solid way to approach required functionality. But maybe mine one has some benefits, too (e.g simplicity, custom names for routes and support for inline parameters).

hedgesky:
About extracting to this code to a separate gem: I think it would be wonderful if there will be something like Grape extensions list in Readme. With it new users will know about different ways to extend default Grape functionality

dblock:
I'll leave this open for comments. Would you like to try and extract this into an extension gem?

hedgesky:
Yes, I'll do it.

dmitry:
> About extracting to this code to a separate gem: I think it would be wonderful if there will be something like Grape extensions list in Readme. With it new users will know about different ways to extend default Grape functionality.

Please join the discussion here: https://github.com/intridea/grape/issues/1018

hedgesky:
Finally I've created an extension: [grape-named_routes](https://github.com/hedgesky/grape-named_routes). Please look at it, maybe you can suggest any improvements?

dblock:
The other variation with automatic names is https://github.com/reprah/grape-route-helpers, also created over the w/e :) 

@reprah and @hedgesky You should look at each-other's code and consider whether there should be two gems or one, my 0.02c is that if one of you feels strongly that the other person's gem is much better in terms of implementation, make a PR with whatever missing features are.

I'll close this.

PR #1046, opened 2015-06-25
Failing spec showing rescue_all block breaking header versioning

jerrod:
The purpose of this pull request is add a spec which demonstrates that adding a rescue_all block to a grape API breaks header versions in an undetermined way.


dblock:
@jerrod I think this may be fixed, care to check?

Issue #958, opened 2015-03-18
Allow parameter only if another parameter is present.


dblock:
For example:

```ruby
requires :product_id, type: String # a product id
optional :shelf_id, type: String # only if the product is on this shelf
optional :shelf_reachable, type: Boolean # only if the shelf is reachable
```

You shouldn't be able to specify `shelf_reachable` unless `:shelf_id` is present.

u2:
That's a good idea.I just have this demand.
```
        params do
          requires :cid, type: Integer
          requires :car_order, type: Hash do
            optional :cars_params, type: Array do
              requires :id, type: Integer
              requires :quantity, type: Integer #
            end
            optional :car_sets_params, type: Array
          end
```
We want at least one of `cars_params` and `car_sets_params` is present.

dblock:
@u2 We already have `at_least_one_of`.

u2:
Thank you.:smile:

rnubel:
After thinking about this for a bit, I think it'd be more useful to be able to say "if `shelf_id` is given, then require `shelf_reachable`" rather than the inverse. The two are closely related, at any rate, and I could imagine there being use cases for both. So along those lines, I've quickly prototyped a couple possible ways to implement such logic:

### Option 1 (DSL style)
https://github.com/rnubel/grape/compare/dependent_parameters_dsl
```ruby
params do
  optional :shelf_id, type: Integer
  if_given :shelf_id do
    requires :shelf_reachable, type: Boolean
  end
end
```

### Option 2 (Validation style)
https://github.com/rnubel/grape/compare/dependent_parameters_validation
```ruby
params do
  optional :shelf_id, type: Integer
  optional :shelf_reachable, type: Boolean
  if_given :shelf_id, require: :shelf_reachable
end
```

What do you think?

dblock:
I like the first one, but I don't love the name. Maybe just `with`?

rnubel:
Hmm... `with` is a bit vague. I think that's the tricky part here -- we want a term that's clear about what it means, but still concise. What about just `given`?

```ruby
params do
  optional :shelf_id, type: Integer
  given :shelf_id do
    requires :shelf_reachable, type: Boolean
  end
end
```

dblock:
I like `given`, go for it!

chrisyoumans:
Should `given` work within a nested parameter? E.g.:

```
requires :document, type: Hash do
    optional :file_contents, type: String
    given :file_contents do
[...]

PR #1047, opened 2015-06-26
Validate params only if another param is present

rnubel:
Implements #958 (sort of). This allows for the use case where you have some parameters to an endpoint which are only relevant or meaningful if some other parameters are given. It can be used like so:

```ruby
params do
  optional :shelf_id, type: Integer
  given :shelf_id do
    requires :bin_id, type: Integer
  end
end
```

Nested use cases work, too:
```ruby
params do
  requires :product_info, type: Hash do
    requires :name, type: String
    optional :shelf_id, type: Integer
    given :shelf_id do
      requires :bin_id, type: Integer
    end
  end
```

One concern: This type of relationship won't show up properly in grape-swagger, or anything which looks at `Route#route_params` to inspect parameter definitions; even though the elements in the `given` block are actually optional unless `shelf_id` is given, `Route#route_params` will indicate that `bin_id` is required. I could probably get that to instead show up as optional, but I don't know if that's ideal.

### DIFF
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -15,13 +15,16 @@ module Grape
       # @option opts :optional [Boolean] whether or not this scope needs to have
       #   any parameters set or not
       # @option opts :type [Class] a type meant to govern this scope (deprecated)
+      # @option opts :dependent_on [Symbol] if present, this scope should only
+      #   validate if this param is present in the parent scope
       # @yield the instance context, open for parameter definitions
       def initialize(opts, &block)
-        @element  = opts[:element]
-        @parent   = opts[:parent]
-        @api      = opts[:api]
-        @optional = opts[:optional] || false
-        @type     = opts[:type]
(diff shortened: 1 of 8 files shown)


lib/grape/validations/params_scope.rb
```diff
@@ -57,7 +71,7 @@ def required?
       protected
 
       # Adds a parameter declaration to our list of validations.
-      # @param attrs [Array] (see Grape::DSL::Parameters#required)
+      # @param attrs [Array] (see Grape::DSL::Parameters#requires)
```
dblock: Thanks for fixing the comment here.

dblock:
I think this is good enough. Swagger doesn't concern be as much, these parameters are closer to required than optional IMO. It needs CHANGELOG and README updates and I will merge.

rnubel:
Updated. I also added a test and fixed a bug where `declared(params)` would explode if used with `given`.

dblock:
Merged, thank you.

PR #1048, opened 2015-06-26
Only dup InheritableValues, remove support for deep_dup

### LINKED ISSUE #891: Cloned InheritableSetting seems to clone unnecessary stuff

toddmazierski:
## Summary

To fix #891, settings managed with `InheritableValues` are now `dup`ed instead of `deep_dup`ed. This is a continuation of the change made in #1034 for `StackableValues`, and the same notes on impact and testing (ex. trying this out in real projects) likely apply!

## Other Changes

Now that `deep_dup` is no longer used, support for it has been removed (`lib/backports`).

## Related Issues

* #873 / #1038


dblock:
Thanks, can you please squash these commits?

toddmazierski:
> Thanks, can you please squash these commits? 

Done (bdb334c3aa73533f6aedba177913e26de616f1f8)! The Travis build appears to be stuck, though.

dblock:
Merged, thank you.

toddmazierski:
You're welcome. Thanks for your help, and all your hard work on Grape!

PR #1049, opened 2015-06-28
Remove deprecated rubyforge_project attribute from gemspec

hedgesky:
Rubyforge.org [closed](https://twitter.com/evanphx/status/399552820380053505) on May 15, 2014, and rubyforge_project attribute [became deprecated](http://rubygems.rubyforge.org/rubygems-update/Gem/Specification.html#attribute-i-rubyforge_project). Remove it from gemspec, then.


dblock:
Merci.

PR #1050, opened 2015-06-29
Nested mount points don't inherit parent declared params

plukevdh:
Not sure if this is documented elsewhere, but I would expect the following to work:

```ruby
class Another < Grape::API
  namespace :another do
    params do
      required :my_id, type: Integer
    end
    route_param :my_id do
    get do
      declared(params)
    end
  end
end

class Something < Grape::API
  format :json
  namespace :something do
    params do
       requires :id
    end
    route_param :id do
      mount Another
    end
  end
end
```

And for the request `get /something/123/another/456`

to return `{ id: 123, my_id: 456 }`, but I only get `{ my_id: 456 }`

I have attached a failing test case to this PR to demonstrate the issue, though I am unsure as of yet how to go about fixing.

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -480,6 +480,43 @@ describe Grape::Endpoint do
     end
   end
 
+  describe '#declared; from a nested mounted endpoint' do
+    before do
+      mounted = Class.new(Grape::API)
+      mounted.namespace :another do
+        params do
+          requires :mount_space, type: Integer
+        end
+        route_param :mount_space do
+          get do
+            {
+              params: params,
+              declared_params: declared(params)
+            }
+          end
+        end
+      end
+
+      subject.format :json
+      subject.namespace :something do
+        params do
+          requires :id, type: Integer
+        end
+        resource ':id' do
+          mount mounted
+        end
+      end
+    end
+
+    it "can access parent attributes" do
+      get '/something/123/mount_space/456'
+      expect(last_response.status).to eq 200
+      json = JSON.parse(last_response.body, symbolize_names: true)
+      expect(json[:declared_params][:id]).to eq 123
+      expect(json[:declared_params][:mount_space]).to eq 456
+    end
+  end
+
   describe '#params' do
     it 'is available to the caller' do
       subject.get('/hey') do
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -508,7 +508,7 @@ describe Grape::Endpoint do
       end
     end
 
-    it "can access parent attributes" do
+    it 'can access parent attributes' do
       get '/something/123/mount_space/456'
       expect(last_response.status).to eq 200
       json = JSON.parse(last_response.body, symbolize_names: true)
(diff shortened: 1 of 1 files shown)


dblock:
If someone is going to fix this, lets carefully document what is being inherited when mounting an endpoint under some scope.

rnubel:
So, technically [this hack](https://github.com/rnubel/grape/commit/8ad4e3b175588e3782f328885a88793d2c6281d8) makes the fixed spec pass. However, I'm pretty sure it's not the right way to do this. I'm having trouble untangling how settings get passed and inherited and stacked during the API definition process. However, I can explain why both lines are necessary:

First, `inheritable_setting.route` on the mounted endpoint isn't updated simply because `route` is only [populated from the parent upon the initial inherit_from call](https://github.com/intridea/grape/blob/master/lib/grape/util/inheritable_setting.rb#L53) and isn't updated when its parent's value is like `namespace_stackable` would be. So, this hack forces the parent's new parent (step-grandfather?)'s values to be propagated downward. It would be possible to make route updates be re-propagated downward whenever a parent changes, but I'm not sure if that's desired or not.

As for the validations line, it's necessary because validations on each endpoint are [only set during initialization](https://github.com/intridea/grape/blob/master/lib/grape/endpoint.rb#L58), which isn't redone when an API is mounted. This might be solvable by having validations be re-pulled from namespace_stackable(:validations) at the time of `endpoint#prepare_routes` (which happens after mounting).

Anyway, maybe someone with more experience in the internals of Grape can weigh in here.

dblock:
@rnubel Thanks! With only a cursory look it's actually possible that this is a viable solution, but I think a useful exercise would be to extend the specs to something that potentially can break.

plukevdh:
Dunno how that path got messed up. Thanks @rnubel for the investigation!

rnubel:
Okay, I've cleaned up the integration a bit: https://github.com/intridea/grape/compare/master...rnubel:declared_and_nested_mount_missing_params

Neither of my suggestions from my original post actually worked out -- `InheritableSetting#route` is just a hash, and can't intelligently be merged with a mounting API's; for validations, we need to persist the state of the stacked validations at the time of endpoint creation.

However, the new `Endpoint#inherit_settings` method (in need of a better name) is fairly straightforward, and makes it pretty clear what's inherited from a mounting API. The recursion allows it to handle multiply-nested APIs (which I updated the test to prove), too. So maybe this is acceptable. I feel like there's a more elegant solution out there, but I don't know what it is.

Arkanain:
Hi guys. Maybe it make sense to merge this pull request? Some of us really need this fix and it will be really cool to see it.

plukevdh:
@Arkanain This PR only had a failing spec. There were some other branches that had fixes but I don't know their states. cc @rnubel

plukevdh:
IIRC, @rnubel's fix did work for my case, but that was almost a year ago and the project that was utilizing this has not been under active development so I can't speak to this.

Arkanain:
I will create new fresh pull request with this fixes because this issue is still available.

Arkanain:
you can close this pull request because all needed changes is already in latest master.

dblock:
Closed via https://github.com/ruby-grape/grape/pull/1422 that is.

PR #1051, opened 2015-06-30
Fix description rewriting

marshall-lee:
Validations DSL writes something to `route_setting(:description)[:params]` but doesn't clean it in `reset_validations!`.

See new spec `spec/grape/api_spec.rb` — this is similar to our case in a real application where we got grape-swagger generating wrong documentation.

```ruby
desc 'global description'
  params do
    requires :param1
    optional :param2
  end
  namespace 'ns1' do
    get do; end
  end
  params do
    optional :param2
  end
  namespace 'ns2' do
    get do; end
  end
end
```

grape-swagger shows that `/ns2/` requires `param1` but in fact it's callable without it.

### DIFF
--- a/lib/grape/dsl/configuration.rb
+++ b/lib/grape/dsl/configuration.rb
@@ -72,6 +72,21 @@ module Grape
           namespace_setting :description, options
           route_setting :description, options
         end
+
+        def description_field(field, value=nil)
+          if value
+            description = route_setting(:description) || route_setting(:description, {})
+            description[field] = value
+          else
+            description = route_setting(:description)
+            description[field] if description
+          end
+        end
+
+        def unset_description_field(field)
+          description = route_setting(:description)
+          description.delete(field) if description
+        end
       end
 
       module_function
--- a/lib/grape/dsl/validations.rb
+++ b/lib/grape/dsl/validations.rb
@@ -13,9 +13,7 @@ module Grape
           unset_namespace_stackable :declared_params
           unset_namespace_stackable :validations
           unset_namespace_stackable :params
-          if route_setting(:description)
-            route_setting(:description)[:params] = nil
-          end
+          unset_description_field :params
         end
 
         # Opens a root-level ParamsScope, defining parameter coercions and
@@ -26,11 +24,7 @@ module Grape
         end
 
         def document_attribute(names, opts)
-          route_setting(:description, {}) unless route_setting(:description)
-
-          route_setting(:description)[:params] ||= {}
-
-          setting = route_setting(:description)[:params]
+          setting = description_field(:params) || description_field(:params, {})
           Array(names).each do |name|
             setting[name[:full_name].to_s] ||= {}
             setting[name[:full_name].to_s].merge!(opts)
--- a/lib/grape/dsl/validations.rb
+++ b/lib/grape/dsl/validations.rb
@@ -13,6 +13,9 @@ module Grape
           unset_namespace_stackable :declared_params
           unset_namespace_stackable :validations
           unset_namespace_stackable :params
+          if route_setting(:description)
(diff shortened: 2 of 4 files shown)


lib/grape/dsl/validations.rb
```diff
@@ -13,6 +13,9 @@ def reset_validations!
           unset_namespace_stackable :declared_params
           unset_namespace_stackable :validations
           unset_namespace_stackable :params
+          if route_setting(:description)
```
dblock: I would add a helper ala `unset_namespace_stackable` for this.

spec/grape/dsl/validations_spec.rb
```diff
@@ -25,6 +31,18 @@ class Dummy
         it 'resets validations' do
           expect(subject.namespace_stackable(:validations)).to eq []
         end
+
+        it 'resets params' do
+          expect(subject.namespace_stackable(:params)).to eq []
+        end
+
+        it 'resets documentation params' do
+          expect(subject.route_setting(:description)[:params]).to eq nil
```
dblock: `to be nil`

lib/grape/dsl/validations.rb
marshall-lee: But it  only deals with `:description` route setting.
Helpers like `description_field` and `unset_description_field` would be good?

There is also `namespace_setting(:description)` what purpose I cannot realize — it's just not used anywhere. Should this new helpers write to this setting too?

dblock: I think your helper suggestion is good, I like it. If something is not used, nuke it.

dblock:
Also squash your commits, please.

marshall-lee:
@dblock what about `unset_` helper you proposed? Implement it or not?

dblock:
I mean "Helpers like description_field and unset_description_field would be good?", my answer is yes :)

marshall-lee:
@dblock I refactored my code with new helpers, please look at it. If it's good for you i'll open new PR with squashed changes.

About `namespace_setting` — it's actually not used anywhere but there are a lot of specs to change after removing it. We should remove all `namespace_setting` related specs and also rewrite many specs in `spec/grape/dsl/settings_spec.rb` that cover generic settings dsl behavior using `namespace_setting` storage — it should now use route settings or namespace_stackable.
It'll be better to do it in a separate PR.

lib/grape/dsl/configuration.rb
```diff
@@ -72,6 +72,21 @@ def desc(description, options = {}, &config_block)
           namespace_setting :description, options
           route_setting :description, options
         end
+
+        def description_field(field, value=nil)
+          if value
+            description = route_setting(:description) || route_setting(:description, {})
```
dblock: This is a bit hard to read, you're having a side-effect assignment on the right-hand. Do this:

```ruby
description = route_setting(:description)
description ||= route_setting(:description, {}) 
```

lib/grape/dsl/validations.rb
```diff
@@ -23,11 +24,7 @@ def params(&block)
         end
 
         def document_attribute(names, opts)
-          route_setting(:description, {}) unless route_setting(:description)
-
-          route_setting(:description)[:params] ||= {}
-
-          setting = route_setting(:description)[:params]
+          setting = description_field(:params) || description_field(:params, {})
```
dblock: Same as above, make the assignment explicit on two lines with an `||=`.

dblock:
This looks good, fix the build, squash please (you can do this in-place I am sure, good git exercise ;)).

marshall-lee:
@dblock I prefer not to do `git push -f` here because it'll erase all the discussions here.

marshall-lee:
@dblock please see #1052

PR #1055, opened 2015-07-03
Remove case insensitive Accept header support

marshall-lee:
Hello!

What was the purpose of finding a `HtTp_AccEPT` case-insensitive way? It also requires traversing all the giant `env` hash :sob:  I cannot find any good reason for doing this and think it's a sort of obsolete stuff.
https://tools.ietf.org/html/rfc3875 says that all the `HTTP_` things are uppercased.

Even if there's a good reason for doing this we should anyway decide what to do with the fact that Grape accesses `Accept` header from another middleware [`Grape::Middleware::Versioner::Header`](https://github.com/intridea/grape/blob/e0b73571388780d46f20cd1fa6bdbb70e5a9b74f/lib/grape/middleware/versioner/header.rb#L87) using uppercased `'HTTP_ACCEPT'`.

### DIFF
--- a/lib/grape/http/headers.rb
+++ b/lib/grape/http/headers.rb
@@ -20,7 +20,6 @@ module Grape
       HTTP_TRANSFER_ENCODING = 'HTTP_TRANSFER_ENCODING'.freeze
       HTTP_ACCEPT            = 'HTTP_ACCEPT'.freeze
 
-      ACCEPT                 = 'accept'.freeze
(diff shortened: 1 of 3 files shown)


dblock:
I think this goes all the way back to https://github.com/intridea/grape/commit/f7728b88c8cb33e5a783f7a779c761196d697343, I tried to find any good reason for this and failed. I think you're totally right. Merging.

Issue #1056, opened 2015-07-07
Using `route :any` causes OPTIONS and HEAD requests to fail


ekampp:
According to the documentation, any GET endpoint also serves OPTIONS and HEAD requests. 

I would expect these to be hit before the `route :any` at the bottom of my api. But this isn't the case.

Is this a bug or intentional? If it's intentional, what's the right way to catch stray requests without overriding OPTIONS and HEAD for the real endpoints? 

```ruby
route :any, '*path' do
  raise NotFound, 'Path not NotFound'
end
```

[Here is a demo api](https://github.com/ekampp/grape-options-demo). I would expect this request to return the `Allow` header for the `/names` endpoint. But instead it return `I'm not here`, which means it hits the `route :any` fallback.

```http
OPTIONS /names
```

dblock:
You should turn it into a test, do GETs behave differently from HEADs?

ekampp:
@dblock any suggestions as to where to start adding a test? 

If I understand your question right, then yes: HEAD behaves differently from GET. HEAD doesn't return any response body - only headers. It's so to speak a compacted request. Useful for checking if a cache is busted, and a new one should be fetched.

dblock:
For now just add a standalone test in https://github.com/intridea/grape/tree/master/spec/grape/integration.

arempe93:
@ekampp @dblock closed by #1263?

dblock:
Yes.

PR #1057, opened 2015-07-07
Add test for OPTIONS and HEAD requests with catch-all

ekampp:
This illustrates the issue described in #1056 by adding a options and
head request when the catch-all route is enabled.

This fails intentionally. It illustrates my confusion about the HEAD and OPTIONS requests failing when a catch-all route is enabled.


dblock:
This is possibly fixed in https://github.com/ruby-grape/grape/pull/1276, care to try this again @ekampp? Thanks.

dblock:
Verified fixed in https://github.com/ruby-grape/grape/pull/1326.

PR #1067, opened 2015-07-13
Remove unwanted test

balasankarc:
Remove the unwanted version checking test. Tests run fine with Ruby 1.5.2 and 1.6.4 perfectly. So the test is unnecessary.

### DIFF
--- a/spec/grape/integration/rack_spec.rb
+++ b/spec/grape/integration/rack_spec.rb
@@ -21,7 +21,7 @@ describe Rack do
 
       unless RUBY_PLATFORM == 'java'
         major, minor, release = Rack.release.split('.').map(&:to_i)
-        pending 'Rack 1.5.3 or 1.6.1 required' unless major >= 1 && ((minor == 5 && release >= 3) || (minor >= 6))
+        pending 'Rack >= 1.5 required' unless ((major == 1 && minor >= 5) || (major > 1))
       end
 
       expect(JSON.parse(app.call(env)[2].body.first)['params_keys']).to match_array('test')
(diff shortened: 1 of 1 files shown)


dblock:
This build failed, it was here for a good reason :)

balasankarc:
But that doesn't make any sense. The commit just removed a test. Altered nothing in the library code. Then how can it cause build failure? AFAIK that test was an isolated one. It doesn't affect or influence the `expect` statement.

I am confused.

rnubel:
The change removes the `pending` invocation, which would stop running the example and return it as "pending" (which doesn't count as a failure, hence why the build previously wasn't broken). Since you removed that `pending` call, the test continues to run and, [because Rack was at 1.4.2](https://travis-ci.org/intridea/grape/jobs/70785647#L106), fails.

That said, the check seems a little odd (it's okay for this not to work, sometimes?). I think it would be more reasonable to require Rack 1.5.3+ explicitly. However, it seems a bit test-specific, so maybe it really is optional... in which case, the test as-is is probably best.

balasankarc:
OK. Since the test works fine for Rack>=1.5.2 (from my experience), I am going to enforce Rack>=1.5.2 requirement in Debian and remove this test from the Debian package. So it will pass, for sure.

About the upstream code, I think we can do version >= 1.5. That is, major>=1 and minor>=5. And remove the release part. Won't that work?

dblock:
Lets backup. I am ok if Grape _requires_ Rack 1.5 or newer explicitly. And then we should adjust the test as such or remove it.

dblock:
I am still interested in this provided it's squashed, builds pass, etcetera.

dblock:
Bump @balasankarc, want to finish this?

dblock:
Closing this, please feel free to reopen if you want to finish it.

Issue #1060, opened 2015-07-09
Support for the HTTP TRACE verb


ekampp:
Hi. 

I thought I would start a discussion here, about how to approach implementing the [HTTP TRACE verb](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html). 

This would be a great tool for developers integrating to an API in a test enviroinment. And also for developers developing new clients or APIs consuming a TRACEable API. 

Since Grape already responds with HEAD and OPTIONS on all endpoints, I assume that we could find some way to implement the TRACE verb in a similar fashion? 

All input is appreciated. 

Thanks.

dblock:
:+1: would like to see this in Grape

ekampp:
I suppose doing [something like this](https://github.com/intridea/grape/blob/master/lib/grape/api.rb#L160-L162) could work?

dblock:
We replaced the router in https://github.com/ruby-grape/grape/pull/1276. Maybe we can do something about this now? /cc: @namusyaka

namusyaka:
Sure, we can do provide the `TRACE` verb. I'm going to implement the new verb if @ekampp is not interested in the feature now.

PR #1068, opened 2015-07-15
Enable trace verb

ekampp:
Initial idea for enabling the [TRACE verb](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html) as discussed in #1060

### DIFF
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -172,7 +172,18 @@ module Grape
               end
             end
 
-            not_allowed_methods = %w(GET PUT POST DELETE PATCH HEAD) - allowed_methods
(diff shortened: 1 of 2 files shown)


dblock:
cc: @namusyaka, I think with the new router we can do this nicely. Maybe @ekampp wants to finish this?

ekampp:
Hi @dblock I have not had time to investigate the new router in detail, so I don't have any real basis for implementing this in the new router. If someone else feels like rebasing or retracing my work on the new router it would be awesome! :)

namusyaka:
Okay, I'm going to dig into the `TRACE` verb and implement it in the new router.
Thanks!

PR #1070, opened 2015-07-16
Header support for middleware before and after

### LINKED ISSUE #1069: Middleware unable to write headers in before blocks
Reproduction:
```ruby
class API < Grape::API
  before do
    header "X-Test-A", "Success"
  end
  
  class HeaderTest < Grape::Middleware::Base
    def before
      endpoint = env['api.endpoint']
      endpoint.header "X-Test-B", "Success"
    end
  end

  use HeaderTest

  get '/' do
    present ""
  end
end
```

Expected result:

```
  HTTP/1.1 200 OK
  Content-Type: text/plain
  Content-Length: 0
  Cache-Control: max-age=0, private, must-revalidate
  Connection: keep-alive
  X-Test-A: Success
  X-Test-B: Success
  Server: thin
```
Actual result:

```
  HTTP/1.1 500 Internal Server Error
  Content-Type: text/plain
  Content-Length: 39
  Cache-Control: max-age=0, private, must-revalidate
  Connection: keep-alive
  Server: thin

  undefined method `[]=' for nil:NilClass
```

Cause: ~~Headers are sent before any `before` calls in middleware~~ Before is called before any response is created.

Workaround: Set headers in `after` (`@app_response[1]`)

yukimono:
Fixes #1069

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Next Release
 * [#1039](https://github.com/intridea/grape/pull/1039): Added support for custom parameter types - [@rnubel](https://github.com/rnubel).
 * [#1047](https://github.com/intridea/grape/pull/1047): Adds `given` to DSL::Parameters, allowing for dependent params - [@rnubel](https://github.com/rnubel).
 * [#1064](https://github.com/intridea/grape/pull/1064): Add public `Grape::Exception::ValidationErrors#full_messages` - [@romanlehnert](https://github.com/romanlehnert).
+* [#1070](https://github.com/intridea/grape/pull/1070): Added header support for Grape::Middleware::Base before and after - [@yukimono](https://github.com/yukimono).
 * Your contribution here!
 
 #### Fixes
--- a/lib/grape/middleware/base.rb
+++ b/lib/grape/middleware/base.rb
@@ -21,8 +21,39 @@ module Grape
       def call!(env)
         @env = env
         before
+
         @app_response = @app.call(@env)
-        after || @app_response
+
+        # Merge Headers from API into @header
+        unless @header.nil?
+          if @app_response.is_a?(Array) && @app_response[1].is_a?(Hash)
+            @header.merge! @app_response[1]
+          elsif @app_response.is_a?(Rack::Response)
+            @header.merge! @app_response.headers
+          end
+        end
+
+        res = after || @app_response
+
+        # Merge Headers from After into return
+        unless @header.nil?
+          if res.is_a?(Rack::Response)
+            @header.each do |k, v| res.headers[k] = v end
+          elsif res.is_a?(Array) && res[1].is_a?(Hash)
+            res[1].merge!(@header)
+          end
+        end
+
+        res
+      end
+
+      def header(key = nil, val = nil)
+        @header ||= {}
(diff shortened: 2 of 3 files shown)


dblock:
This needs tests and CHANGELOG entry, please.

dblock:
Also please make sure to write tests about order, before headers should be overwritten by api headers should be overwritten by after headers.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Next Release
 * [#1039](https://github.com/intridea/grape/pull/1039): Added support for custom parameter types - [@rnubel](https://github.com/rnubel).
 * [#1047](https://github.com/intridea/grape/pull/1047): Adds `given` to DSL::Parameters, allowing for dependent params - [@rnubel](https://github.com/rnubel).
 * [#1064](https://github.com/intridea/grape/pull/1064): Add public `Grape::Exception::ValidationErrors#full_messages` - [@romanlehnert](https://github.com/romanlehnert).
+* [#1070](https://github.com/intridea/grape/pull/1070): Added header support for Grape::Middleware::Base before and after
```
dblock: Make this look like the other lines, it needs a - [@yourname](...) at the end with a period :)

lib/grape/middleware/base.rb
```diff
@@ -21,8 +22,38 @@ def call(env)
       def call!(env)
         @env = env
         before
+
         @app_response = @app.call(@env)
-        after || @app_response
+
+        # Merge Headers from API into @header
```
dblock: Lets optimize this a bit. Default `@header` to `nil`, then extract this part into a method and `return unless @header`. Assign `@header` the first time inside `header` with `@header ||= {}`. 

Also lets clear the ambiguity between `header` and `headers`. I don't think they should be aliased, maybe `header` should only be a function that sets a header instead of returning `@header`?

dblock: This also duplicates some functionality in `InsideRoute`, maybe this can be a module?

spec/grape/middleware/base_spec.rb
```diff
@@ -76,6 +76,100 @@
     end
   end
 
+  describe '#header' do
+    class HeaderExampleWare < Grape::Middleware::Base
```
dblock: These classes are globally scoped so it's a bit annoying in specs. You can fix this by doing 

```ruby
let(:header_example_ware) { Class.new(Grape::Middleware::base) do ... }
```

dblock:
This is good work. See my comments above, this is getting close to what we want.

lib/grape/middleware/base.rb
yukimono: Can you call InsideRoute methods from middleware without `env['api.endpoint']`?

dblock: Probably not, I trust you to dig this a bit and only do it if it makes any sense :)

dblock:
Sorry I am late to this one (Github doesn't send notifications on updates). Would you mind rebasing? Is this good to merge?

lib/grape/middleware/base.rb
```diff
@@ -21,8 +21,39 @@ def call(env)
...
+        unless @header.nil?
```
dblock: Should be able to refactor this into methods. These similarly manipulate `@app_response`.

PR #1073, opened 2015-07-20
No longer resetting validations twice

Charlesetc:
In /lib/grape/dsl/settings.rb there was a call to `reset_validations!` that seemed necessary. All the tests are passing.


dblock:
I'll buy that.

PR #1079, opened 2015-07-30
Add `stream` method to InsideRoute to leverage Rack::Chunked

zbelzer:
Also necessary was the introduction of an object (`FileResponse`) so the formatting middleware would leave files and other enumerables alone. Seemed better than relying on methods on the object as enumerables have `each` defined. I've been writing scala recently so I think I have case classes on the brain.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -35,7 +35,7 @@ Metrics/MethodLength:
 # Offense count: 8
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 243
+  Max: 271
 
 # Offense count: 17
 Metrics/PerceivedComplexity:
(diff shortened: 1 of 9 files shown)


dblock:
Very good @zbelzer! Will merge.

dblock:
The build is failing on Rubocop you have some fixes to do, please (via `--amend`).

zbelzer:
Oh rubocop...

zbelzer:
There a preferable way to update that Module length?

dblock:
I run `rubocop --auto-gen-config`.

dblock:
Merged

zbelzer:
Thanks!

I ran that but it looked like it threw a fit. Maybe if I had run again it would have been happy.

PR #1092, opened 2015-08-06
Changes Intridea name to Grape.

towanda:


dblock:
Thanks! Missed this one.

Issue #1091, opened 2015-08-05
Cannot restrict boolean param


Fryie:
Basically what I want to to is this:

```
params do
  optional :password, allow_blank: false
  optional :generate_password, type: Boolean, values: [true]
  exactly_one_of :password, :generate_password
end
```

However, that won't even start up the app, because Grape / Virtus is complaining that ```true``` is not a valid value (which it truly isn't - ```true``` is not an instance of ```Virtus::Attribute::Boolean```). Using ```values: [Virtus::Attribute.build(true)]``` runs the app, but does not seem to run the proper validation.

The only I was able to do this, was with
```
class Truthy < Grape::Validations::Base
  def validate_param!(attr_name, params)
    unless params[attr_name]
      raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: 'must be true'
    end
  end
end

params do
  optional :generate_password, type: Boolean, truthy: true
end
```

but I'm not sure why this doesn't work the way one would expect it to work.

dblock:
Try writing a spec that reproduces this, I agree we should fix it. I also find the case where the only option for a boolean is true quite interesting.

PR #1096, opened 2015-08-08
Fix boolean type

towanda:
This PR fixes https://github.com/ruby-grape/grape/issues/1091, also tested with an example grape app to reproduce the bug (see https://github.com/towanda/grape-boolean-example/blob/master/example.rb)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,10 @@ Next Release
 * [#1038](https://github.com/ruby-grape/grape/pull/1038): Avoid dup-ing the String class when used in inherited params - [@rnubel](https://github.com/rnubel).
 * [#1042](https://github.com/ruby-grape/grape/issues/1042): Fix coercion of complex arrays - [@dim](https://github.com/dim).
 * [#1045](https://github.com/ruby-grape/grape/pull/1045): Do not convert `Rack::Response` to `Rack::Response` in middleware - [@dmitry](https://github.com/dmitry).
-* [#1048](https://github.com/ruby-grape/grape/pull/1048): Only dup `InheritableValues`, remove support for `deep_dup` - [@toddmazierski](https://github.com/toddmazierski/)
+* [#1048](https://github.com/ruby-grape/grape/pull/1048): Only dup `InheritableValues`, remove support for `deep_dup` - [@toddmazierski](https://github.com/toddmazierski/).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Next Release
 * [#1048](https://github.com/ruby-grape/grape/pull/1048): Only dup `InheritableValues`, remove support for `deep_dup` - [@toddmazierski](https://github.com/toddmazierski/)
 * [#1052](https://github.com/ruby-grape/grape/pull/1052): Reset `description[:params]` when resetting validations - [@marshall-lee](https://github.com/marshall-lee).
 * [#1088](https://github.com/ruby-grape/grape/pull/1088): Support ActiveSupport 3.x by explicitly requiring Hash#except - [@wagenet](https://github.com/wagenet)
+* [#1096](https://github.com/ruby-grape/grape/pull/1096): Fix coercion on booleans - [@towanda](https://github.com/towanda)
```
dblock: Missing a period at the end. Some of the lines above as well, if you don't mind fixing please.

lib/grape/validations/params_scope.rb
```diff
@@ -254,10 +254,15 @@ def validate_value_coercion(coerce_type, values)
         return if values.is_a?(Proc)
         coerce_type = coerce_type.first if coerce_type.is_a?(Array)
         value_types = values.is_a?(Range) ? [values.begin, values.end] : values
+        value_types = convert_booleans(value_types) if coerce_type == Virtus::Attribute::Boolean
         if value_types.any? { |v| !v.is_a?(coerce_type) }
           fail Grape::Exceptions::IncompatibleOptionValues.new(:type, coerce_type, :values, values)
         end
       end
+
+      def convert_booleans(values)
```
dblock: Since this is not reused and is a 1-liner I think we shouldn't make it a method. I know it feels long, but either roll it in, or make coerce_xyz for other types?

dblock:
Good, see my comments above and squash your commits.

CHANGELOG.md
towanda: Sure :)

lib/grape/validations/params_scope.rb
towanda: Changed, thanks.

dblock:
Merging thanks.

PR #1098, opened 2015-08-10
Add Skylight to Performance section

wagenet:


wagenet:
@dblock can this be merged? The Grape stuff is now in a full release (0.8.0) of the skylight gem.

dblock:
I'll merge this. Generally I think we need to start taking links like these out of here and into ruby-grape.org.

dblock:
So @wagenet you definitely want this on http://www.ruby-grape.org/projects/ as well. And maybe you can remove this list from Grape's README and link to a section on the website? Something similar would be useful for entities and such.

wagenet:
@dblock are you thinking of this as a top-level section on the website? or a sub-section?

dblock:
In projects, not top level.

PR #1101, opened 2015-08-11
incorrect version accept header does not return a 406 response

elliotlarson:
If you've configured Grape to use accept header versioning with strict
equal to true, Grape should return a 406 response when making a request
with an incorrect version header.  However, the system was not
responding with a 406 when using the header 'Accept:application/xml'.

### DIFF
--- a/lib/grape/middleware/versioner/header.rb
+++ b/lib/grape/middleware/versioner/header.rb
@@ -22,54 +22,78 @@ module Grape
       # X-Cascade header to alert Rack::Mount to attempt the next matched
       # route.
       class Header < Base
+        VENDOR_VERSION_HEADER_REGEX =
+          /\Avnd\.([a-z0-9*.]+)(?:-([a-z0-9*\-.]+))?(?:\+([a-z0-9*\-.+]+))?\z/
+
         def before
-          header = rack_accept_header
-
-          if strict?
-            # If no Accept header:
-            if header.qvalues.empty?
-              fail Grape::Exceptions::InvalidAcceptHeader.new('Accept header must be set.', error_headers)
-            end
-            # Remove any acceptable content types with ranges.
-            header.qvalues.reject! do |media_type, _|
-              Rack::Accept::Header.parse_media_type(media_type).find { |s| s == '*' }
-            end
-            # If all Accept headers included a range:
-            if header.qvalues.empty?
-              fail Grape::Exceptions::InvalidAcceptHeader.new('Accept header must not contain ranges ("*").',
-                                                              error_headers)
-            end
-          end
-
-          media_type = header.best_of available_media_types
+          strict_header_checks if strict?
 
           if media_type
-            type, subtype = Rack::Accept::Header.parse_media_type media_type
-            env['api.type']    = type
-            env['api.subtype'] = subtype
-
-            if /\Avnd\.([a-z0-9*.]+)(?:-([a-z0-9*\-.]+))?(?:\+([a-z0-9*\-.+]+))?\z/ =~ subtype
-              env['api.vendor']  = Regexp.last_match[1]
-              env['api.version'] = Regexp.last_match[2]
-              env['api.format']  = Regexp.last_match[3]  # weird that Grape::Middleware::Formatter also does this
-            end
-          # If none of the available content types are acceptable:
-          elsif strict?
-            fail Grape::Exceptions::InvalidAcceptHeader.new('406 Not Acceptable', error_headers)
-          # If all acceptable content types specify a vendor or version that doesn't exist:
-          elsif header.values.all? { |header_value| has_vendor?(header_value) || version?(header_value) }
-            fail Grape::Exceptions::InvalidAcceptHeader.new('API vendor or version not found.', error_headers)
+            media_type_header_handler
+          elsif headers_contain_wrong_vendor_or_version?
+            fail_with_invalid_accept_header!('API vendor or version not found.')
           end
         end
 
         private
+
+        def strict_header_checks
+          strict_accept_header_presence_check
+          strict_verion_vendor_accept_header_presence_check
+        end
+
+        def strict_accept_header_presence_check
+          return unless header.qvalues.empty?
+          fail_with_invalid_accept_header!('Accept header must be set.')
+        end
+
+        def strict_verion_vendor_accept_header_presence_check
+          return unless versions.present?
+          return if an_accept_header_with_version_and_vendor_is_present?
+          fail_with_invalid_accept_header!('API vendor or version not found.')
+        end
+
+        def an_accept_header_with_version_and_vendor_is_present?
+          header.qvalues.keys.any? do |h|
+            VENDOR_VERSION_HEADER_REGEX =~ h.sub('application/', '')
+          end
+        end
+
+        def header
+          @header ||= rack_accept_header
+        end
+
+        def media_type
+          @media_type ||= header.best_of(available_media_types)
+        end
+
+        def media_type_header_handler
+          type, subtype = Rack::Accept::Header.parse_media_type(media_type)
+          env['api.type'] = type
+          env['api.subtype'] = subtype
+
+          if VENDOR_VERSION_HEADER_REGEX =~ subtype
+            env['api.vendor'] = Regexp.last_match[1]
+            env['api.version'] = Regexp.last_match[2]
+            # weird that Grape::Middleware::Formatter also does this
+            env['api.format'] = Regexp.last_match[3]
+          end
+        end
+
+        def fail_with_invalid_accept_header!(message)
+          fail Grape::Exceptions::InvalidAcceptHeader
+            .new(message, error_headers)
+        end
 
         def available_media_types
           available_media_types = []
 
           content_types.each do |extension, _media_type|
             versions.reverse_each do |version|
-              available_media_types += ["application/vnd.#{vendor}-#{version}+#{extension}", "application/vnd.#{vendor}-#{version}"]
+              available_media_types += [
+                "application/vnd.#{vendor}-#{version}+#{extension}",
+                "application/vnd.#{vendor}-#{version}"
+              ]
             end
             available_media_types << "application/vnd.#{vendor}+#{extension}"
           end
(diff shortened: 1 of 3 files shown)


dblock:
Add to CHANGELOG and squash the commits please?

lib/grape/middleware/versioner/header.rb
```diff
@@ -23,53 +23,83 @@ module Versioner
...
+            )
+          end
+        end
+
+        def header
+          @header ||= rack_accept_header
+        end
+
+        def media_type
+          @media_type ||= header.best_of available_media_types
```
dblock: I would add a pair of `(` `)`, so `header.best_of(available_media_types)`, just more readable.

lib/grape/middleware/versioner/header.rb
```diff
@@ -23,53 +23,83 @@ module Versioner
...
+
+        def media_type_header_handler
+          type, subtype = Rack::Accept::Header.parse_media_type media_type
+          env['api.type']    = type
+          env['api.subtype'] = subtype
+
+          if vendor_version_header_regex =~ subtype
+            env['api.vendor']  = Regexp.last_match[1]
+            env['api.version'] = Regexp.last_match[2]
+            # weird that Grape::Middleware::Formatter also does this
```
dblock: Open an issue with a link to this code to figure out why Grape::Middleware::Formatter does this instead.

lib/grape/middleware/versioner/header.rb
```diff
@@ -23,53 +23,83 @@ module Versioner
...
+            env['api.format']  = Regexp.last_match[3]
+          elsif versions.size > 0
+            fail_with_invalid_accept_header(
+              'API vendor or version not found.'
+            )
+          end
+        end
+
+        def vendor_version_header_regex
+          /\Avnd\.([a-z0-9*.]+)(?:-([a-z0-9*\-.]+))?(?:\+([a-z0-9*\-.+]+))?\z/
```
dblock: This can be a constant, `VENDOR_VERSION_HEADER_REGEX`, right?

lib/grape/middleware/versioner/header.rb
```diff
@@ -23,53 +23,83 @@ module Versioner
...
+            media_type_header_handler
           # If none of the available content types are acceptable:
           elsif strict?
-            fail Grape::Exceptions::InvalidAcceptHeader.new('406 Not Acceptable', error_headers)
-          # If all acceptable content types specify a vendor or version that doesn't exist:
-          elsif header.values.all? { |header_value| has_vendor?(header_value) || version?(header_value) }
-            fail Grape::Exceptions::InvalidAcceptHeader.new('API vendor or version not found.', error_headers)
+            fail_with_invalid_accept_header('406 Not Acceptable')
+          elsif headers_contain_wrong_vendor_or_version?
+            fail_with_invalid_accept_header('API vendor or version not found.')
```
dblock: We should probably rename `fail_with_invalid_accept_header` to be a bang method, `fail_with_invalid_accept_header!`.

lib/grape/middleware/versioner/header.rb
```diff
@@ -23,53 +23,83 @@ module Versioner
...
+
+        def vendor_version_header_regex
+          /\Avnd\.([a-z0-9*.]+)(?:-([a-z0-9*\-.]+))?(?:\+([a-z0-9*\-.+]+))?\z/
+        end
+
+        def fail_with_invalid_accept_header(message)
+          fail(
+            Grape::Exceptions::InvalidAcceptHeader.new(message, error_headers)
+          )
+        end
```
dblock: Super minor, here I would just `fail Grape::Exceptions ...`, no parenthesis.

elliotlarson:
Thanks for your code review, dblock!  I've added these updates in.

dblock:
This looks good. Can I have someone else confirm that I am not missing anything here, cause this bug is a bit big to be true :) Maybe @dm1try? Thx.

elliotlarson:
Yeah, that makes sense.  Thanks for your time on this.

lib/grape/middleware/versioner/header.rb
```diff
@@ -22,54 +22,82 @@ module Versioner
...
+          type, subtype = Rack::Accept::Header.parse_media_type media_type
+          env['api.type']    = type
+          env['api.subtype'] = subtype
+
+          if VENDOR_VERSION_HEADER_REGEX =~ subtype
+            env['api.vendor']  = Regexp.last_match[1]
+            env['api.version'] = Regexp.last_match[2]
+            # weird that Grape::Middleware::Formatter also does this
+            env['api.format']  = Regexp.last_match[3]
+          elsif versions.size > 0
```
dm1try: this condition and expectation seems strange to me.
if **size of predefined versions  more than zero** then **fail with parsed "vendor or version not found"**

as we can see the `versions` it's a predefined value: options[:versions] => [:v1]
so if you remove in the context of your spec  [version value :v1](https://github.com/elliotlarson/grape/blob/master/spec/grape/middleware/versioner/header_spec.rb#L165) then it'll fail again.
I mean it's not expected behaviour.

dm1try:
Anyway original behaviour was implemented [here](https://github.com/ruby-grape/grape/pull/259/files#diff-fc867ab5863c687f310c0e69c11fc432R71)

see the `#available_media_types`, it returns sum of "prepared vendor media-types `["application/vnd.vendor-v2+xml", "application/vnd.vendor-v2", "application/vnd.vendor-v1+xml", "application/vnd.vendor-v1" ...]`
and defaut media types`["application/xml", "application/json", "application/json", "application/octet-stream", "text/plain"..]`
[this spec describes](https://github.com/ruby-grape/grape/blob/master/spec/grape/middleware/versioner/header_spec.rb#L16-L37) why default media types are returned too, seems like it's a desired behaviour (too much responsibility for the "versioner-header" middleware btw)

so ATM it's can be  easily fixed by adding `unless strict?` [in this line](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/versioner/header.rb#L81) (use default content types only in "non-strict" mode)

maybe a better way is removing the logic which allow to work with default content types from this middleware at all but it's a breaking change. (or move this to `base` middleware)

dblock:
@elliotlarson Can you please see @dm1try's comments above?

@dm1try What do you think of the code change itself, better? worse? no different?

dm1try:
@dblock definitely better!

lib/grape/middleware/versioner/header.rb
elliotlarson: Agreed.  This was a bit of a kludgy fix on my part.   I tried to make the smallest possible change to get the desired result.  Now that I've gone down the road of doing a bit of refactoring, I feel like this is the wrong place for this logic.  I've updated the code a bit to reflect what I think is probably a better approach.

elliotlarson:
@dm1try Thanks for taking the time to review and provide feedback.  I agree that there is a bit of muddled responsibility in this middleware.  This could use some more time and refactoring.  My hope is to fix this bug with a reasonable, incremental improvement to the code, and then perhaps dig in a bit further after that's out of the way and see if there is a good way to untangle the responsibilities.

I've refactored my fix a bit.  I feel like returning a `406` if there's no version or vendor in the accept header belongs further up in the code than the place where I stuck it.  If you have `strict` set, and you're not passing in a version header, the very first `strict_header_checks` method should field that.  So, I've removed my original fix and I've added some checking code into here instead.

Also, the original code had some logic for media type header checking if `strict` was set.  I feel like these are sort of moot checks if `strict` is applied.  If it's applied and we're passing in a header like `Accept: application/*`, we don't need to return an error about media type ranges not being allowed since the bigger issue is that there isn't a valid version header.  The update I've added in reflects that opinion.  

Anyway, I hope this is a better solution that meets with your approval.  I look forward to your feedback.

spec/grape/middleware/versioner/header_spec.rb
```diff
@@ -184,24 +184,6 @@
...
-
-    it 'fails with 406 Not Acceptable if subtype is a range' do
-      expect { subject.call('HTTP_ACCEPT' => 'application/*').last }.to raise_exception do |exception|
-        expect(exception).to be_a(Grape::Exceptions::InvalidAcceptHeader)
-        expect(exception.headers).to eql('X-Cascade' => 'pass')
-        expect(exception.status).to eql 406
-        expect(exception.message).to include('Accept header must not contain ranges ("*").')
-      end
-    end
-
```
dblock: Would you be so kind to explain why this is no longer something we need?

dblock:
@dm1try Would you mind making calls around merging this? I don't feel like I'm qualified :) Thanks.

spec/grape/middleware/versioner/header_spec.rb
elliotlarson: Yeah, sorry, I tried to explain this in the third paragraph above, but this may need a bit more explanation.

The existing code is doing some checking to make sure that you're not passing in media type ranges.  However, these checks are only happening in the case when `strict` is set to true.  If `strict` is set to true and you're passing in an accept header like `Accept:application/*`, it's my contention that returning an error about not allowing media ranges is less important that returning an error about not passing in a correct version/vendor header.  

As I was refactoring the code, I placed the version/vender header checking mechanism before the media range checks.  After I did this, it became clear that these media range checks were never getting exercised because the version vendor checks were always returning the `406` error first.  So, pulling out the media range checks and the related specs seemed prudent.

Of course, there may be some untested/un-documented outcome that I'm not seeing here related to media type accept headers.  Beyond the README and the specs, is there some kind of written spec or standard that you're intending to follow that might provide more guidance here?

dblock: This makes perfect sense, thanks.

dm1try: :+1:

dm1try:
@elliotlarson this is great, thanks!
@dblock feel free to merge this

dblock:
Merged via https://github.com/ruby-grape/grape/commit/d7718f0b0d0a00fda256e2a0b5a7e1f087ceb89e, thank you.

elliotlarson:
No problem.  Thanks again @dblock and @dm1try for your time reviewing and providing feedback!

dblock:
Thanks for hanging along @elliotlarson, great job!

Issue #1066, opened 2015-07-13
Rack version mismatch


balasankarc:
Hi,
The gemspec file (of version 0.12.0) specifies that the rack version requirement is ">=1.3.0". However, spec/grape/integration/rack_spec.rb checks whether the Rack version is 1.5.3 or 1.6.1. This induces ambiguity. Please fix it.

I am packaging grape for Debian (as part of GitLab packaging. Grape is a member of the dependency chain), where the rack version is still 1.5.2. But all the tests (other than the one mentioned above) seems to run perfectly with Rack v1.5.2. So, is it necessary for grape to require the above mentioned exact rack versions? If not, can you remove that spec I mentioned above?

Thanks.

dblock:
I think we should make it >= 1.5.2 in gemspec. Would appreciate a PR.

balasankarc:
I'll send a PR ASAP. 

BTW, Rack 1.5.2 sets it's release as "1.5", not "1.5.2" ( as seen [here](https://github.com/rack/rack/blob/ac590d055c936bb9a618e955a690dc836c625211/lib/rack.rb#L23) ). So, the split on period to `Rack.release` will assign `nil` to the variable 'release' which will cause the mentioned test to fail.

If Rack versions >= 1.5.2 work succesfully, can we disable the mentioned test? All the tests run fine with 1.5.2 (and 1.6.4 - I checked). So, is that test really needed?

dblock:
I am fine with anything that makes the test match what's in Gemspec or removes any unnecessary tests. Make a call.

dm1try:
>I think we should make it >= 1.5.2

rails 3 locked on `rack ~> 1.4.5`, do you wanna drop support of rails 3?)

dblock:
I would be open to dropping support for Rails 3 en general, but only if there's a good reason.

dm1try:
The mentioned test is fixed in #1104.

> but only if there's a good reason.
The reason above(updating old dependencies) is not good one?) Anyway it's another issue and can be fixed later, I agree.

PR #1104, opened 2015-08-14
fix spec for rack 1.5.2 dependency

dm1try:
see #1066


dblock:
More specs more compat optins, more better. Merging.

PR #1105, opened 2015-08-17
Minor clarifications to declared parameters section

ddzz:


dblock:
This is good, thanks.

PR #1108, opened 2015-08-18
Raise a warning when `desc` is called with options hash and block 

### LINKED ISSUE #920: desc dsl with a block will ignore given options hash

rngtng:
Raise a warning when `desc` is called with options hash and block.
relates to: #920

### DIFF
--- a/lib/grape/dsl/configuration.rb
+++ b/lib/grape/dsl/configuration.rb
@@ -64,6 +64,9 @@ module Grape
             end
 
             config_class.configure(&config_block)
+            unless options.empty?
+              warn '[DEPRECATION] Passing a options hash and a block to `desc` is deprecated. Move all hash options to block.'
+            end
             options = config_class.settings
           else
             options = options.merge(description: description)
--- a/spec/grape/dsl/configuration_spec.rb
+++ b/spec/grape/dsl/configuration_spec.rb
@@ -70,6 +70,19 @@ module Grape
           expect(subject.namespace_setting(:description)).to eq(expected_options)
           expect(subject.route_setting(:description)).to eq(expected_options)
         end
+
+        it 'can be set with options and a block' do
+          expect(subject).to receive(:warn).with('[DEPRECATION] Passing a options hash and a block to `desc` is deprecated. Move all hash options to block.')
+
+          desc_text = 'The description'
+          detail_text = 'more details'
+          options = { message: 'none' }
+          subject.desc desc_text, options do
+            detail detail_text
+          end
+          expect(subject.namespace_setting(:description)).to eq(description: desc_text, detail: detail_text)
+          expect(subject.route_setting(:description)).to eq(description: desc_text, detail: detail_text)
+        end
       end
     end
   end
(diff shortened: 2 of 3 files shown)


dblock:
I feel like this has been brought up before and we didn't want this in favor of a 1st class DSL.

rngtng:
I agree, but it should at least rise a deprecation warning or similar...

dblock:
Maybe, but this was introduced a long time ago and is well documented in README. I'd be up for a deprecation note at runtime or similar.

rngtng:
ok updated, but I still wonder how #920 can be resolved then. I wouldn't prefer the monkey patch. 
In my opinion, the option hash should not be ignored by now but the deprecation warning given. in future releases the option hash should be removed eventually. This way you give people who update or read outdated docs a chance to figure out what's wrong...

dblock:
Well, it *has* been removed for multiple releases, so at this point it would be introducing a regression.

Which version are you upgrading from?

lib/grape/dsl/configuration.rb
```diff
@@ -64,6 +64,9 @@ def desc(description, options = {}, &config_block)
             end
 
             config_class.configure(&config_block)
+            unless options.empty?
+              warn "[DEPRECATION] passing a options hash and a block to `desc` is deprecated. Move all hash options to block"
```
dblock: This is good, although I think the message should be a full phrase, so "Passing ... " and with a period at the end of "Move ...".

spec/grape/dsl/configuration_spec.rb
```diff
@@ -70,6 +70,16 @@ class Dummy
           expect(subject.namespace_setting(:description)).to eq(expected_options)
           expect(subject.route_setting(:description)).to eq(expected_options)
         end
+
+        it 'can be set with options and a block' do
+          expect(subject).to receive(:warn)
```
dblock: Add a match here to some text, just in case some other future `warn` is triggered.

spec/grape/dsl/configuration_spec.rb
```diff
@@ -70,6 +70,16 @@ class Dummy
...
+
+          desc_text = 'The description'
+          options = { message: 'none' }
+          subject.desc desc_text, options do
+            detail 'more details'
+          end
```
dblock: It would be great to ensure explicitly that the `options` hash is *not* merged here.

rngtng:
ok all fixed. I didn't upgrade, but started a new project with latest version. I must have ran in an outdated/undocumented issue with details/notes and had to reverse engineer the issue.
Well as long as the method interface allows me to pass in an options hash and a block and its not removed. Any ideas when the options has will be removed finally?

dblock:
You have to rebase this.

rngtng:
done

dblock:
Perfect, thanks.

rngtng:
great thanks

PR #1111, opened 2015-08-19
Optimize construction of headers.

marshall-lee:
...And remove some cryptic regexp stuff from headers construction.
New algorithm is easier to read and also faster.

### DIFF
--- a/lib/grape.rb
+++ b/lib/grape.rb
@@ -38,7 +38,7 @@ module Grape
 
     autoload :Cookies
     autoload :Validations
-    autoload :Request, 'grape/http/request'
+    autoload :Request
     autoload :Env, 'grape/util/env'
   end
 
--- a/lib/grape/request.rb
+++ b/lib/grape/request.rb
@@ -1,8 +1,6 @@
 module Grape
   class Request < Rack::Request
-    HTTP_PREFIX  = 'HTTP_'.freeze
-    UNDERSCORE   = '_'.freeze
-    MINUS        = '-'.freeze
+    HTTP_PREFIX = 'HTTP_'.freeze
 
     def params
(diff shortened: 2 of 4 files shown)


dblock:
I think this is a good opportunity to introduce a basic set of tests for this class, doesn't look like we have any.

marshall-lee:
@dblock https://github.com/ruby-grape/grape/blob/5154b36ea924e7e17805d905c8e8e072f1cb1299/spec/grape/endpoint_spec.rb#L123

dblock:
I mean a standalone test for this class alone, not in context of a request.

marshall-lee:
@dblock ok, i'll write some.

marshall-lee:
@dblock 

I added some basic tests, please look at it!

> is not case-sensitive

I don't think it's necessay. Remember #1055?

dblock:
This is great. Since we have lib/grape/http/request.rb lets move the spec into spec/grape/http/request_spec.rb?

marshall-lee:
@dblock 

I suggest to move `lib/grape/http/request.rb` to `lib/grape/request.rb` because class is named `Grape::Request` :)

dblock:
I am with you. Do it.

marshall-lee:
Done! :)

lib/grape.rb
```diff
@@ -38,7 +38,7 @@ module Grape
 
     autoload :Cookies
     autoload :Validations
-    autoload :Request, 'grape/http/request'
+    autoload :Request
     autoload :Env, 'grape/util/env'
```
dblock: You have your work cut out for you @marshall-lee right here :)

dblock:
Merged, thanks!

PR #1113, opened 2015-08-20
Set sudo: false for running tests in containers

marshall-lee:
More info here: http://docs.travis-ci.com/user/migrating-from-legacy/


dm1try:
@marshall-lee thanks

dblock:
I thought that was already the default on Travis-CI, but it looks like repos after 1/2015 only are.

Issue #968, opened 2015-03-24
endpoints with same URL but different versions throws InvalidAcceptHeader on 0.11.0 w/ rescue_all


jpaas:
I am mounting two versions v1 and v2, and I'm using the :header option. Two of the endpoints have the same URL /api/version.

In 0.10.1 I could select the correct version using the Accept header.
In 0.11.0 I can access v1(which is mounted first), but I can no longer select v2 (mounted second). It throws Grape::Exceptions::InvalidAcceptHeader.

dblock:
Could you try building a test that reproduces this?

mattsnyder:
Confirmed with 2 separate apps. When upgrading from 0.9 to 0.11 we receive this error on endpoints with multiple versions.

mattsnyder:
Interesting, took one my projects where this was failing and stripped out all but the bare minimum and versioning works with both 0.9 and 0.11 using the header. See repo: https://github.com/mattsnyder/grape-versioning-example/tree/master

0.9 is the master and 0.11 is on a branch. Will try to investigate further.

mattsnyder:
@dblock @jpaas I found it!! 

I began adding in behavior from my other apps that experienced this problem into this sample app and found it was when I added a `rescue_from :all` that the versioning broke.

So if you run the specs on the 0.11.0 branch a test will fail with:

```
1) GrapeVersioningExample::Endpoints::PingV1 GET /ping(version 1)
     Failure/Error: expect( async.response ).to include("version 1")
       expected "[:error, \"\\nProblem:\\n  Invalid accept header\\nResolution:\\n  API vendor or version not found.\"]" to include "version 1"
     # ./spec/api/ping_spec.rb:7:in `block (4 levels) in <top (required)>'
     # ./spec/api/ping_spec.rb:6:in `block (3 levels) in <top (required)>'
```

I'm guessing error handling is used for version handling??

jpaas:
Yes I also have a `rescue_from :all`. Nice find!

mattsnyder:
@jpaas Also found an interesting bug in https://github.com/mjackson/rack-accept that prohibits the use of `_` in the hypermedia type. And underscore is allowed per http://tools.ietf.org/html/rfc6838#page-4 section 4.2.

We're working on a PR for rack-accept to fix it.

dblock:
Interesting, care to link a bug in rack-accept?

dblock:
Also, again, we *want* a test in Grape for this.

[...]

PR #1114, opened 2015-08-20
Fix rescue all header versioning regression

suan:
A suggested fix for https://github.com/ruby-grape/grape/issues/968

There are two failing specs which show that unfortunately there is no way to `rescue_from` an invalid version, since it's possibly valid and needs to propogate. This was also the case before the regression was introduced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 #### Fixes
 
+* [#1114](https://github.com/ruby-grape/grape/pull/1114): Fix regression which broke identical endpoints with different versions - [@suan](https://github.com/suan).
 * [#1109](https://github.com/ruby-grape/grape/pull/1109): Memoize Virtus attribute and fix memory leak - [@marshall-lee](https://github.com/marshall-lee).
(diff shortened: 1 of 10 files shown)


dblock:
The code looks good, but I am confused to what the spec failures mean - are the specs wrong? If so they need to change. 

Also please update CHANGELOG and squash the commits.

dblock:
I am curious what @dm1try thinks about this.

Also, this definitely needs to be documented in both README and the behavior change in UPGRADING.

Finally, maybe we need to make this behavior configurable?

suan:
@dblock The spec failures show that with my changes it's (unfortunately) not possible to rescue any `Grape::Exceptions::InvalidHeaderVersion`s - as we'll _always_ pass those through to allow multi-versioned endpoints to work

croeck:
The changes seem good to me. All failing tests are invalid and did not regard the scenario as defined in the related issue.

dblock:
@suan remove the bad tests, squash the commits please, update README, UPGRADING and CHANGELOG.

suan:
@dblock I've made the suggested changes - let me know if there's anything you'd need, thanks!

dblock:
This is good, merging.

suan:
Awesome, appreciate it @dblock!

PR #1119, opened 2015-08-23
Fix `do_not_route_options!` handling: do not respond to OPTIONS if disabled!

jf:
FYI: so apparently this wasn't caught by `spec/grape/dsl/routing_spec.rb`. Not too sure how to fix that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#1109](https://github.com/ruby-grape/grape/pull/1109): Memoize Virtus attribute and fix memory leak - [@marshall-lee](https://github.com/marshall-lee).
 * [#1101](https://github.com/intridea/grape/pull/1101): Fix: Incorrect media-type `Accept` header now correctly returns 406 with `strict: true` - [@elliotlarson](https://github.com/elliotlarson).
 * [#1108](https://github.com/ruby-grape/grape/pull/1039): Raise a warning when `desc` is called with options hash and block - [@rngtng](https://github.com/rngtng).
+* [#1119](https://github.com/ruby-grape/grape/pull/1119): Fix `do_not_route_options!` handling: do not respond to OPTIONS if disabled! - [@jf](https://github.com/jf).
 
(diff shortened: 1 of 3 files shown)


dblock:
First, there should be a test for the fixed behavior, please.

dblock:
If this is caused by #1120 then the CHANGELOG should be a bit clearer.

jf:
No, it's not caused by #1120. #1120 is a separate issue.

dblock:
Thanks @jf. The only think I am missing in this PR is a test for #1119 being fixed.

jf:
Can I get some clarification on that?

"Test for #1119 being fixed":
- would that refer to my 2nd commit for this PR which fixed the Travis build (https://github.com/jf/grape/commit/cb44238645e1179fd67a3bf597a795bb3a9b4c69)?
- are you looking for something less brittle and not reliant on a fixed ordering of the options as per now? (Because that's what broke Travis. A change in the order)
- if you're talking about putting something into `spec/grape/dsl/routing_spec.rb`, now that I've had the chance to see it, would you say that Travis has highlighted that I should be looking to fix `spec/grape/api_spec.rb` instead (which I have sort of done; but am happy to make the test little brittle as mentioned) ?

dblock:
The order of headers isn't the problem in the bug. There should be a test (in any spec appropriate) that reproduces the actual problem, aka that the API responds to OPTIONS when it shouldn't. That test should fail without the fix. Does this help?

arempe93:
@jf Is the objective of the PR to remove `OPTIONS` from the Allow header when `do_not_route_options!` is active?

jf:
Yup. That was the spirit of https://github.com/jf/grape/commit/3f3c0bab660fcf035bbe86d86f7224a292cfdaa6. OPTIONS should not be be advertised as an allowed method when `do_not_route_options!` is set. Thanks for completing this!

dblock:
So this can be closed in favor of #1266. @jf reopen if I got it wrong.

Issue #1123, opened 2015-08-25
Requiring Grape appears to break the ruby YAML parser!


atomless:
```
require 'yaml'

my_yaml = '!!omap 
  - doc: 0'
yml = YAML::load(my_yaml)
puts "#{yml}"

require 'grape'

my_yaml2 = '!!omap 
  - doc: 0'
yml2 = YAML::load(my_yaml2)
puts "#{yml2}"
```
Output from the first puts is : ```{"doc"=>0}```
Output from the second puts is : ```{}``` and the following error is thrown by the line YAML::load(my_yaml2) :

```
/Library/Ruby/Gems/2.0.0/gems/activesupport-4.2.3/lib/active_support/ordered_hash.rb:4: warning: wrong element type String at 0 (expected array)
/Library/Ruby/Gems/2.0.0/gems/activesupport-4.2.3/lib/active_support/ordered_hash.rb:4: warning: ignoring wrong elements is deprecated, remove them explicitly
/Library/Ruby/Gems/2.0.0/gems/activesupport-4.2.3/lib/active_support/ordered_hash.rb:4: warning: this causes ArgumentError in the next release
```

As if this were not strange enough, if I change ```!!omap``` to ```!omap``` all works fine and no errors.

Any idea what's going on here?

marshall-lee:
Interesting...

dblock:
This is nuts.

marshall-lee:
So many related issues in one day!
This is again `ActiveSupport`'s quirk.
`ordered_hash.rb` doesn't simply define `OrderedHash`, it adds some extension to `YAML`
https://github.com/rails/rails/blob/7779e63371d1248bce5f30ebbae43e13ee7abb6a/activesupport/lib/active_support/ordered_hash.rb#L3

I think we don't need `OrderedHash` in `Grape` at all. I'll open PR soon.

marshall-lee:
@atomless 

I guess you maybe use `omap` wrong. But I agree with you, `Grape` should not affect `YAML` behavior.

atomless:
@marshall-lee good catch! The yaml I need to parse is coming in from an external source and appears to be valid according to http://www.yamllint.com/

marshall-lee:
@atomless please report it to rails!

atomless:
[...]

PR #1122, opened 2015-08-25
require core_ext/object/blank

marshall-lee:
This one is for #1121

`active_support/core_ext/object/conversions` was the most heavy dependency that slows everything **twice**. 
What do you think about cutting `ActiveSupport` usage to minimum? Ideally, not depend on it at all.

PR #1124, opened 2015-08-25
Remove `ActiveSupport::OrderedHash` usage.

marshall-lee:
Minus one ActiveSupport dependency!
`ActiveSupport::OrderedHash` is a hash that preserves insertion order. It's a legacy stuff since Ruby >= 1.9 already preserves insertion order.

Related to #1123, #1122


marshall-lee:
If you want to see the real motivation behind `OrderedHash`, do this thing:

```
$ git clone git@github.com:rails/rails.git
$ git log -p activesupport/lib/active_support/ordered_hash.rb
```

We just don't need this thing.

marshall-lee:
The very first implementation of `OrderedHash` from the 2008:

```ruby
module ActiveSupport
  # Hash is ordered in Ruby 1.9!
  if RUBY_VERSION >= '1.9'
    OrderedHash = ::Hash
  else
    class OrderedHash < Array #:nodoc:
      def []=(key, value)
        if pair = assoc(key)
          pair.pop
          pair << value
        else
          self << [key, value]
        end
      end

      def [](key)
        pair = assoc(key)
        pair ? pair.last : nil
      end

      def delete(key)
        pair = assoc(key)
        pair ? array_index = index(pair) : nil
        array_index ? delete_at(array_index).last : nil
      end

      def keys
        collect { |key, value| key }
      end

      def values
        collect { |key, value| value }
      end

      def to_hash
        returning({}) do |hash|
          each { |array| hash[array[0]] = array[1] }
        end
      end
    end
  end
end
```

2008 was good a time :)

dblock:
Oh yeah image)

PR #1125, opened 2015-08-26
Use extract_options!

marshall-lee:
This one, on the contrary to #1122, #1124, utilizes ActiveSupport feature.
`extract_options!` is harmless and not heavy, also `Grape` already uses it in some place.

It should not impact performance since the fixed methods  are `DSL` methods (they're called just once). Here we better have clarity.


marshall-lee:
If we can somehow get rid of `ActiveSupport`'s `autload` feature then only dependency on `core_ext/...` thing remains. I wonder if there's a way to not use them for Ruby >= 2.0.0 and implement them as refinements? It won't pollute standard classes.

dblock:
:+1: 

eventually Ruby 1.9 won't be a thing.

PR #1130, opened 2015-08-26
Add built-in New Relic support to Monitoring options

kwugirl:
We released built-in support for Grape instrumentation a few months ago and realized we should probably update the list here as well!


dblock:
Thanks. I think we should really take this out of this README at some point soon and add it to http://ruby-grape.org.

PR #1133, opened 2015-08-27
Use reverse_merge to set default options.

towanda:


dblock:
Reading the old (and the new) code I am wondering whether we really wanted to modify the incoming options hash? Maybe we should be dup-ing it first?

towanda:
@dblock Changed, this way It doesn't modify the original options hash.

dblock:
Thanks, merged.

Issue #1074, opened 2015-07-20
declared(params) doesn't coerce types in before blocks


iangreenleaf:
In `before` blocks, `declared(params)` doesn't seem to perform the usual type coercion. For example:

```ruby
      params do
        requires :id, type: Integer
      end
      group "foo/:id" do
        before do
          puts declared(params)[:id].inspect # this is a string
       end
     end
```

Strangely, it does still seem to validate in the blocks, as undeclared params are still filtered out. It just doesn't convert the types.

rnubel:
I believe this is the expected behavior, actually. Type coercion is really a [type of validation](https://github.com/intridea/grape/blob/b5c83a4cd75fb85f67fa4741e567d17aa15739f1/lib/grape/validations/validators/coerce.rb), so if you want a given block to run after coercion has happened, you'd want it to be in the [after_validations hook](https://github.com/intridea/grape/blob/b474f5e40f8c06490fba0790e57e438c630ec23f/lib/grape/dsl/callbacks.rb#L13).

Granted, that may violate the Principle of Least Surprise a bit, but changing this behavior would be a breaking change.

iangreenleaf:
Thanks for the tip, I'll change my callbacks. It's reasonable to consider this as part of validation, but there's a couple things about this I'd rate as especially surprising:

1. That `declared` still runs in this context, instead of erroring out because I'm trying something that doesn't make sense.
2. That the validations are incompletely applied at this point - undeclared params are already filtered out, but types are not yet coerced.

dblock:
@iangreenleaf I think you are onto something here, maybe this should be fixed?

iangreenleaf:
Agreed! Either make the method unavailable in this scope, or have it raise an error if run.

jrforrest:
I ran into this issue a bit ago and put together this somewhat sloppy solution for my own fork since I hit this pitfall enough for it to annoy me (I am not a smart guy.)

```
diff --git a/lib/grape/endpoint.rb b/lib/grape/endpoint.rb
index 2b8ee61..4f8fc9b 100644
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -73,6 +73,7 @@ module Grape

       @options[:method] = Array(options[:method])
       @options[:route_options] ||= {}
+      @run_filters = Hash.new

       return unless block_given?

@@ -319,6 +320,23 @@ module Grape
           instance_eval(&filter)
         end
       end
[...]

PR #1142, opened 2015-09-01
Makes #declared unavailable to before filters

jrforrest:
In response to issue #1074

This adds a mechanism by which Grape::DSL::InsideRoute methods may be
overriden after certain filters are run.

The first problem case of a filter being utilized before it should that
necessitated this enhancement is the `#declared` method, which was
returning un-coerced params in `before` filters.

There may be other problem cases, which may be rectified using this same
mechanism.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,6 +534,10 @@ The returned hash is a `Hashie::Mash` instance, allowing you to access parameter
   declared(params).user == declared(params)["user"]
 ```
 
+
+The `#declared` method is not available to `before` filters, as those are evaluated prior
+to parameter coercion.
+
 ### Include missing
 
 By default `declared(params)` includes parameters that have `nil` values. If you want to return only the parameters that are not `nil`, you can use the `include_missing` option. By default, `include_missing` is set to `true`. Consider the following API:
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -6,44 +6,69 @@ module Grape
       extend ActiveSupport::Concern
       include Grape::DSL::Settings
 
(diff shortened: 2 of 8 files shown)


lib/grape/dsl/inside_route.rb
```diff
@@ -6,46 +6,73 @@ module InsideRoute
...
+                                   else
+                                     params[parent]
+                                   end
+              end
             end
           end
         end
       end
 
+      # A filtering method that will return a hash
```
jrforrest: It's a little weird to have the docs on the method stub rather than the full version in the `PostBeforeFilter` module.  Anyone more savvy with YARD know if this is the best way to be doing this?

README.md
```diff
@@ -534,6 +534,10 @@ The returned hash is a `Hashie::Mash` instance, allowing you to access parameter
   declared(params).user == declared(params)["user"]
 ```
 
+
+`#declared` is not available to `before` filters, as those are evaluated prior
```
dblock: Maybe make this a full phrase, so The `#declared` method ...`.

dblock:
Please do fix the build, thx.

jrforrest:
Sorry about that.  Didn't realize Rubocop was doing its thing.

Should be good now.

dblock:
This looks good, mind squashing the commits please?

dblock:
Needs a CHANGELOG entry and I think something in UPGRADING that says if you used this method before it will now raise ...

jrforrest:
Squashed, and updated CHANGELOG and UPGRADING docs.

dblock:
Merged, thanks.

jrforrest:
Sure thing.  Thank you for the thorough review.

Issue #1144, opened 2015-09-03
Using type: Boolean in a param validation throws an error when placed in a helper file.


atomless:
This works fine:
```
params do
  requires :toggle_prm, type: Boolean
end
post 'endpoint' do
  # ...
end
```
However, when placed in an external helper file like so: 
```
module ParamValidation
  extend Grape::API::Helpers

  params :requires_toggle_prm do
    requires :toggle_prm, type: Boolean
  end

end
```
it throws the error shown below:
```
/Users/jamestindall/Documents/soasta/DOC/source/g-speak/Baton/BatonSite/dist/server/api/v1/param_validation.rb:26:in `block in <module:ParamValidation>': uninitialized constant ParamValidation::Boolean (NameError)
	from /Users/jamestindall/.rvm/gems/ruby-2.0.0-p643/gems/grape-0.13.1/lib/grape/dsl/parameters.rb:39:in `instance_exec'
	from /Users/jamestindall/.rvm/gems/ruby-2.0.0-p643/gems/grape-0.13.1/lib/grape/dsl/parameters.rb:39:in `block in use'
	from /Users/jamestindall/.rvm/gems/ruby-2.0.0-p643/gems/grape-0.13.1/lib/grape/dsl/parameters.rb:35:in `each'
	from /Users/jamestindall/.rvm/gems/ruby-2.0.0-p643/gems/grape-0.13.1/lib/grape/dsl/parameters.rb:35:in `use'
	from /Users/jamestindall/Documents/soasta/DOC/source/g-speak/Baton/BatonSite/dist/server/api/v1/v1.rb:72:in `block (5 levels) in <class:V1>'
	from /Users/jamestindall/.rvm/gems/ruby-2.0.0-p643/gems/grape-0.13.1/lib/grape/validations/params_scope.rb:30:in `instance_eval'
	from /Users/jamestindall/.rvm/gems/ruby-2.0.0-p643/gems/grape-0.13.1/lib/grape/validations/params_scope.rb:30:in `initialize'
	from /Users/jamestindall/.rvm/gems/ruby-2.0.0-p643/gems/grape-0.13.1/lib/grape/dsl/validations.rb:23:in `new'
	from /Users/jamestindall/.rvm/gems/ruby-2.0.0-p643/gems/grape-0.13.1/lib/grape/dsl/validations.rb:23:in `params'
[...]

PR #1153, opened 2015-09-13
Fixes boolean declaration in an external helper file.

towanda:
See https://github.com/ruby-grape/grape/issues/1144.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


dblock:
Do you think you can write a test for this?

dblock:
I also wonder if we need to check whether Boolean is already defined here. I would move the `const_set` into a function, just to make it cleaner, ala `inject_api_helpers_to_mod` and check.

towanda:
@dblock added spec and now checks if it's defined in a separate function.

dblock:
Great, merging.

PR #1156, opened 2015-09-17
Fixed edge case regarding Hash with inner param

lasseebert:
This fixes an edge case where the API would blow up with a `"no implicit conversion of Symbol into Integer"`.

To trigger the error, make a param nested inside a `Hash` param and add `values` validations to the param. Then send something that is not a hash to the outer param. (see spec)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 #### Fixes
 
+* [#1156](https://github.com/ruby-grape/grape/pull/1156): Fixed `no implicit conversion of Symbol into Integer` with nested `values` validation - [@quickpay](https://github.com/quickpay).
 * [#1153](https://github.com/ruby-grape/grape/pull/1153): Fixes boolean declaration in an external file - [@towanda](https://github.com/towanda).
 * [#1142](https://github.com/ruby-grape/grape/pull/1142): Makes #declared unavailable to before filters - [@jrforrest](https://github.com/jrforrest).
 * [#1114](https://github.com/ruby-grape/grape/pull/1114): Fix regression which broke identical endpoints with different versions - [@suan](https://github.com/suan).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 #### Fixes
 
+* [#1156](https://github.com/ruby-grape/grape/pull/1156): Fixed `no implicit conversion of Symbol into Integer` with nested `values` validation - [@quickpay](https://github.com/quickpay).
```
dblock: This is missing a period at the end :)

dblock: Whoops ignore me

lib/grape/validations/validators/values.rb
```diff
@@ -7,6 +7,7 @@ def initialize(attrs, options, required, scope)
       end
 
       def validate_param!(attr_name, params)
+        return unless params.is_a?(Hash)
```
dblock: We are skipping the error here, shouldn't it be a fail?

dblock: Actually I take it back. The parent is the problem, we don't get deep enough to care about the values validator.

dblock:
Merged, thanks.

lib/grape/validations/validators/values.rb
lasseebert: This just skips the validation of the inner param if the outer param does not have a valid value.

The "real" error happens in the `coerce` validator that correctly invalidates a `Hash` param with string value.

This is the same thing happening in the `allow_blank` validator here: https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/allow_blank.rb#L5

lasseebert:
Thanks :)

lib/grape/validations/validators/values.rb
dblock: Yes I should first have coffee and then review code :)

lasseebert: :) Always coffee first.

PR #1157, opened 2015-09-18
Add text content type to redirects

franzliedke:
The redirect helper returned an empty response, which confused the XML and JSON helpers (e.g. when rendering in a browser).

This sets the content type explicitly, which solves that problem. :)

### DIFF
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -103,6 +103,7 @@ module Grape
           end
         end
         header 'Location', url
+        content_type 'text/plain'
         body ''
       end
 


dblock:
At the very least this needs tests and a CHANGELOG entry.

This has been discussed before though and the consensus was that a redirect should not set a content type because there's no content. If you find an RFC that says otherwise happy to consider. I'll leave this open for now.

franzliedke:
Well, redirects are unusable for me with `content_type :xml`, `content_type :json` as well as `default_format :json` in my base API class. As I said, the formatters try to interpret the empty body as JSON or XML and fail.

If you give me the go-ahead, I'll gladly add a test.

dblock:
I am just confused to why they are not usable given that it's the "correct" behavior?

franzliedke:
When I redirect, I obviously don't want to render any JSON or XML, so the formatters shouldn't even try.

dblock:
Maybe I am missing the point, can you explain "which confused the XML and JSON helpers" please? I am not trying to be difficult, just trying to help :)

franzliedke:
Here you go:

~~~ruby
get 'me' do
  redirect "users/#{current_user.id}"
end
~~~

With default_format set to XML, I get this output:

~~~xml
<error>
  <message>cannot convert String to xml</message>
</error>
~~~

For JSON, it works now. Weird.

dblock:
Ok this is what I'd expect. I think we need defined behavior for XML here, which is not an error. It's the same problem as discussed in https://github.com/ruby-grape/grape/pull/1159, can we merge these two discussions?

dblock:
I am going to close this in favor of https://github.com/ruby-grape/grape/issues/1162, without merging. Lets discuss what the "right fix" would be.

Issue #1085, opened 2015-08-03
namespace param validation doesn't propagate to nested Grape modules


dantethegrey:
Not sure if it is a bug or a feature, but I'd welcome if I could declare parameter validation even for nested API modules like below.

For the former code sample the UserId validator doesn't get called, while for the latter it does.
Behaves the same for builtin validators.
```
params do
  requires :user_id, type: String, user_id: true
end
resource 'user/:user_id' do 
  mount UserProfile
end
```

```
params do
  requires :user_id, type: String, user_id: true
end
resource 'user/:user_id' do 
  get '/' do 
  ...
  end
end
```
Thanks

dblock:
I think it's an interesting corner case, if you can fix it without breaking anything else I would take it.

rnubel:
I believe this fails for basically the same reason as https://github.com/ruby-grape/grape/pull/1050. I'd be curious if [my patch from that issue](https://github.com/intridea/grape/compare/master...rnubel:declared_and_nested_mount_missing_params) also fixes this issue.

namusyaka:
Closed via #1422 
Thanks for reporting the issue, and thanks @rnubel and @plukevdh for your efforts.
Please reopen this issue if you still have a problem.

Issue #810, opened 2014-11-12
Explicit parameters only


dblock:
We need a way to say: only the parameters declared are allowed. For a new API the whitelist of `params` could really be super useful for typos and wondering why `filter_by_foobbar_id=....` isn't working when you mispelled `foobar`.

whatasunnyday:
@dblock Can I take this on? Also, if I don't respond within 30 days anyone else should feel free to take this on.

dblock:
Of course @whatasunnyday, no need to ask, just make pull requests!

Fryie:
IMHO declared(params) should be the default. Otherwise, it's just too easy to inject a mass assignment vulnerability in a Grape API. This lesson has been learned by Rails, that's why we first had attribute_accessible and now strong_params.

23tux:
@Fryie +1

wrtsprt:
+1!

senhalil:
Any news on this feature? Currently we are handling this issue with a check along the lines of `params.to_a - declared(params).to_a` in dev/test but it is too crude.

I checked the existing PR's but it went a bit over my head unfortunately so don't think I can attack this ticket but I wanted to say that I found both of the following constructs very useful.

```
params(undeclared: :ignore / :raise) do
end
```
```
class MyAPI < Grape::API
   include Grape::Params::AllowAny::False / True
end
```

PR #1158, opened 2015-09-18
added undeclared params validator #810

hobofan:
Only allow declared parameters as proposed in #810.

Works like this:

```
params do
  optional :beer
  optional :wine
  optional :juice
  declared_only
end
```

Now if there is a parameter present that is not `beer`, `wine` or `juice` a validation exception will be thrown.
Also tested for nested parameters.

It is not quite finished yet:
- I plan on adding a `except` option to skip the validator for some parameters. Especially in combination with the bug described in #1050 / #1085 , a param that is accepted on every endpoint would have to either be declared in every params block or will throw an exception with this validator
- Not tested/tried it for Arrays yet.
- Some feedback on the name of `declared_only` would be nice. I not sure if it fits in well.


dblock:
I like it. I think that the DSL could be clearer and be on `params`, maybe turn it around, for example `params undeclared: :error do ...`. I think that if a param is accepted with every endpoint it should be in a `helpers ...` and reused via `use`, introducing `except` will defeat the purpose of this IMO. 

I think I would like to be able to turn this globally on for my entire API. 

I'd like a way to turn this on gradually, we have 400 endpoints in a project and we would love to make params declared only, so maybe something like `params undeclared: :warn`?

dblock:
Something that's not declared is smuggled? Should `undeclared` be `smuggled`? j/k

hobofan:
Great, I would also prefer to have it on `params`, but decided to follow the conventions of the existing validator (even though it works somewhat different than the other validators).

hobofan:
@dblock 

> I'd like a way to turn this on gradually, we have 400 endpoints in a project and we would love to make params declared only, so maybe something like params undeclared: :warn?

I've put it in there now, but it isn't easily testable and I haven't really found any similar behaviour in grape yet.

dblock:
I really like this. I think it needs README and CHANGELOG and it would be good to go. Testing via expecting a warning is good AFAIK.

hobofan:
I had to change a bit more, to get the validation for generic hashes working in an expected way.

E.g.:
```ruby
params undeclared: :error do
  optional :config, type: Hash
end
```

If you only specify the type Hash, you probably want to allow any Hash.

In contrast, if you declared you Hash with a block like
```ruby
params undeclared: :error do
  optional :config, type: Hash do
    requires :language
  end
end
```
you have a strict expectation of how the Hash should look like.

As for the warning test, I tried my best (even ugly stuff like expect_any_instance_of) but didn't manage to get a correct expectation running.

hobofan:
When trying to integrate it into our own API I noticed that it failed when a Hash had more than one nested parameter defined.
This is now fixed but requires more involved Hash inclusion checking for the declared params. On the other hand this opens up the possibility to expand `given` blocks to work with nested blocks like:

```ruby
params do
  optional :a, type: Hash do
    optional :b
  end
  given :a => :b do
    requires :c
  end
end
```

lib/grape/dsl/parameters.rb
```diff
@@ -163,9 +163,38 @@ def given(attr, &block)
...
+        convert_hash = lambda do |obj|
+          return [obj, true] unless obj.is_a?(Array) || obj.is_a?(Hash)
+          if obj.is_a? Array
+            return obj.map { |e| convert_hash.call(e) }
+          elsif obj.is_a? Hash
+            obj.map do |k, v|
+              return k, convert_hash.call(v).to_h
+            end
+          end
+        end
```
dblock: Should be refactored out of here.

lib/grape/dsl/parameters.rb
```diff
@@ -163,9 +163,38 @@ def given(attr, &block)
...
+          key = smaller.keys.first
+
+          return included.call(bigger[key], smaller[key])
+        end
+
+        conv_declared_params = convert_hash.call(@declared_params).to_h
+        conv_param = convert_hash.call([param]).to_h
+
+        included.call(conv_declared_params, conv_param)
+      end
```
dblock: What if `declared_param?` took a path (an array) instead? Would it make anything simpler?

dblock:
Maybe there's a simpler implementation here that calculates a hash difference (only caring about keys), I haven't dug into things like https://github.com/liufengyun/hashdiff or others, but it sounds like a good way to abstract the problem, no?

hobofan:
Yes, having `declared_param?` accept a path as array of keys might make things simpler. I'll look into it, but might be a week until I have the time.

wrtsprt:
What is the status on this? Is there anything we can do to help? 

This seems to be a quite important feature to me. Especially with being able to turn that on by default ('strict mode') for the whole API.

hobofan:
I sadly won't have time to work on it in the forseeable future, sorry.

I have no problem if anyone picks up where I left off and submits a new PR.

dblock:
I am still very interested in this, it's a super useful feature. Was thinking maybe we should have `strict: true` on `params` as a more rigid DSL, `undeclared: ` still feels odd to me. Looking for someone to finish this PR!

hobofan:
Closing this, as I'm currently clearing out old/outdated PRs of mine.

Issue #1135, opened 2015-08-28
Support JSON as a first class group parameter type


dblock:
Today Grape supports Hash and Array as a group parameter type and [raises UnsupportedGroupTypeError](https://github.com/ruby-grape/grape/blob/cc1e46f218b8d0029c6533c81716289c44335192/lib/grape/dsl/parameters.rb#L114) for everything else. We would like to be able to do `type: JSON`, like so: 

```ruby
   resource :reticulated_splines do
      params do
        requires :splines, type: JSON do
          requires :id, type: Integer
          requires :reticulated, type: Boolean
        end
      end
      get do
        params[:splines].map do |spline|
          spline.merge(reticulated: !spline[:reticulated])
        end
      end
    end
```

Today you have to workaround this, see https://github.com/ruby-grape/grape-on-rack/commit/0f904bfe7f17603ce605a3458d6bcbbaad1a96ea:

```ruby
   resource :reticulated_splines do
      before do
        params.each_pair do |k, v|
          params[k] = JSON.parse(v) rescue v
        end
      end
      params do
        requires :splines, type: Array do
          requires :id, type: Integer
          requires :reticulated, type: Boolean
        end
      end
      get do
        params[:splines].map do |spline|
          spline.merge(reticulated: !spline[:reticulated])
        end
      end
    end
```

Note that the JSON itself can be an array or a hash. So maybe we can just generalize:

```ruby
requires :splines, type: Array, coerce_type: JSON do

requires :splines, type: Array, coerce_method: ->(val) { JSON.parse(val) }

requires :splines, type: Array, coerce_method: JSON.method(:parse) 
```

dslh:
Here's my attempt. I've gone with a generalized solution to support both hashes and arrays, but it seemed natural to support primitive types as well so things like `type: Time, coerce_using: Chronic` are also possible.

At the moment it cannot be supplied without a `type` or `coerce` parameter. You could drop this constraint by either implicitly setting type to `String`, which the documentation already gives as the default type. Or with some adjustments it could bypass type checking altogether, but I don't know if that's good flexible API design or an invitation to lazy programmers.

I called it `coerce_using` to cover both classes and methods. Maybe that's confusing given the `using` parameter? `coerce_with` maybe?

dblock:
Very nice work @dslh! I just commented on `coerce_with` vs. `coerce_using` in the PR, so sounds like we both think it's more natural to do `coerce_with`. LMK when you have updated it.

[...]

PR #1161, opened 2015-09-21
Custom parameter coercion using `coerce_with` #1135

dslh:
Addresses issue #1135. Adds `coerce_using` option to
`Grape::DSL::Parameters::requires` and `::optional`,
allowing implementors arbitrary coercion uncoupled
from the parameter type.

See README.md#custom-types-and-coercions for usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 * Your contribution here.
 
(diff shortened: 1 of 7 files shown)


dblock:
This is very good. Before I merge it, should it be `coerce_with:`? We have `using:` for header type versioning but we have `with:` for `present`. Any strong feelings?

dblock:
Merged.

PR #1163, opened 2015-09-22
First-class JSON parameter type #1135

dslh:
Addresses issue #1135. Allows `type: JSON` and `type: Array[JSON]`
as options for `Grape::DSL::Parameters::requires` and `::optional`,
allowing convenient handling of parameters supplied as JSON-encoded
strings.

See README.md#first-class-json-types for usage.

### DIFF
--- a/lib/grape/validations/validators/coerce.rb
+++ b/lib/grape/validations/validators/coerce.rb
@@ -43,6 +43,13 @@ module Grape
       def valid_type?(val)
         if val.instance_of?(InvalidValue)
           false
+        elsif type == JSON
+          # Special JSON type is ambiguously defined.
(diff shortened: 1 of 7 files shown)


lib/grape/validations/validators/coerce.rb
```diff
@@ -51,6 +58,14 @@ def valid_type?(val)
       end
 
       def coerce_value(val)
+        # JSON is not a type as Virtus understands it,
```
dblock: As a future improvement maybe we should write a custom Virtus type for this, to be consistent and avoid special handling?

dblock:
This is perfect. I am merging it, it's an +:100: feature.

Issue #1164, opened 2015-09-23
Param type Rack::Multipart::UploadedFile does not result in object of that type


alexandergitter:
Declaring a parameter as Rack::Multipart::UploadedFile doesn't seem to actually yield a parameter of the same type. Instead I'm getting instances of Hashie::Mash.

The endpoint i use to test this looks as follows:

```ruby
params do
  requires :avatar, type: Rack::Multipart::UploadedFile
end
post '/' do
  ...
end
```

dslh:
That's correct. This hash is created by `Rack::Request` when processing multipart file data. Grape wraps it with Hashie for convenience.

```ruby
post '/' do
  params.avatar.filename # => 'avatar.png'
  params.avatar.type # => 'image/png'
  params.avatar.tempfile # => #<File>
end
```

Not sure if this is documented anywhere in either library.

dblock:
I'll leave this open, @alexandergitter feel free to suggest a doc improvement?

alexandergitter:
Sure, I'll take a look at it later this week.

dslh:
This is now documented in the readme. Close?

dblock:
Closing.

Issue #693, opened 2014-07-24
Allow multiple parameter types


dblock:
I have the following API: `/api/v1/artworks/tagged?tag[]=`, I'd like to be able to do `/api/v1/artworks/tagged?tag=painting` as well as `/api/v1/artworks/tagged?tag[]=painting&tag[]=sculpture`. So something like this:

``` ruby
params do
  requires :tag, types: [String, Array[String]], desc: "Tag or tags."
end
```

Maybe instead I could write this:

```
params do
  optional :tag, type: String, desc: "Tag or tags."
  optional :tag, type: Array[String], desc: "Tag or tags."
  exactly_one_of :tag
end
```

That would also solve the problem of allowing a parameter like `size` to be a number (eg. 10) or a string, (eg. "all").

viranch:
This can also be approached with making `tag` only an `Array[String]`:

```ruby
params do
  optional :tag, type: Array[String], desc: "Tag or tags."
end
get :tagged do
  p params[:tag] #=> this is always an array
end
```

This will accept both `/tagged?tag=painting` and `/tagged?tag[]=painting&tag[]=sculpture`, except that grape validator will also convert the former to an array with single element. So you can expect the value of `tag` to always be an array in your models.

Issue #690, opened 2014-07-24
Expand parameter validation to type File


dblock:
It's unclear how to require an uploaded multipart file parameter. We could also expand this to file formats.

dnesteryuk:
@dblock I think this [one is covered](https://github.com/ruby-grape/grape/blob/c5cd3326fcbc896e958bf521e1ed35b0b0021452/spec/grape/validations/validators/coerce_spec.rb#L361-L376). Isn't it?

dblock:
Yup. Also if you're sure just close it @dnesteryuk, what do I know? :)

Issue #689, opened 2014-07-24
Expand parameter validation to type DateTime


dblock:
Dates may be posted as Integer, ISO strings, Date, DateTime, etc.

pragmaticivan:
Can you give more details? +1

u2:
More details?

ngouy:
up

samotarnik:
@dblock 
I've been playing with this for the past couple of days. From what I understand, there are two separate issues here, namely coercion and validation.

Coercion is currently done via Virtus which brings along some interesting consequences. For example
```ruby
Virtus::Attribute.build(DateTime).coerce(1234) #=> #<DateTime: 1970-01-01T00:20:34+00:00 ((2440588j,1234s,0n),+0s,2299161j)> 
Virtus::Attribute.build(Date).coerce(1234) #=> 1234
Virtus::Attribute.build(Time).coerce(1234) #=> 1234
```
and
```ruby
Virtus::Attribute.build(DateTime).coerce('1234') #=> "1234"
Virtus::Attribute.build(Date).coerce('1234') #=> "1234"
Virtus::Attribute.build(Time).coerce('1234') #=> "1234"
```
There are several options here. One is to augment Grape to handle such cases. Another is to fix Virtus to handle these better. In any case, in my opinion, Grape should somehow be aware that coercion actually failed, i.e. that you demanded a `DateTime` but got a `String` instead. I suspect that with dates/datetimes, though, there will always be one more format to cover, so the most appropriate thing would probably be to define a custom parameter type for your custom datetime format and handle the parsing there.

Validations are a different beast. The `:include?` method (used in the `ValuesValidator`) is not really appropriate to test continuous types/variables. For example, this works
```ruby
a = DateTime.new(2015,1,1,0,0,0); b = DateTime.new(2015,2,1,0,0,0); c = DateTime.new(2015,1,10,0,0,0)
(a..b).include?(c) #=> true
```
but these do not
```ruby
a = DateTime.new(2015,1,1,0,0,0); b = DateTime.new(2015,1,1,10,0,0); c = DateTime.new(2015,1,1,5,0,0)
(a..b).include?(c) #=> false

a = Time.new(2015,1,1,0,0,0); b = Time.new(2015,2,1,0,0,0); c = Time.new(2015,1,10,0,0,0)
(a..b).include?(c) #=> TypeError
```
One possible thing to do here would be to use the `:between?` method implemented in the `Comparable`.
```ruby
[...]

PR #1167, opened 2015-09-28
Refactor and extend coercion and type validation

dslh:
Addresses #1164, #690, #689, #693.
Depends on https://github.com/solnic/virtus/pull/343

`Grape::ParameterTypes` is renamed `Grape::Validations::Types`
to reflect that it should probably be bundled with an eventual
`grape-validations` gem. It is expanded to include two new
categories of types, 'special' and 'recognized' (see
'lib/grape/validations/types.rb'). `CoerceValidator` now makes
use of `Virtus::Attribute::value_coerced?`, simplifying its
internals.

`CustomTypeCoercer` is introduced, attempting to standardize
support for custom types by decoupling coercion and type-checking
logic from the `type` class supplied to
`Grape::Dsl::Parameters::requires`.

`JSON`, `Array[JSON]` and `Rack::Multipart::UploadedFile` (a.k.a
`File`) are designated 'special' types, for which special
implementations of `Virtus::Attribute` are provided.

Instances of `Virtus::Attribute` built with `Virtus::Attribute.build`
may now also be passed as the `type` parameter for `requires`.
A number of pre-rolled attributes are available providing coercion
for `Date` and `DateTime` objects from various formats in
`lib/grape/validations/types/date.rb` and `date_time.rb`.


dslh:
I've had a go at making better use of virtus to make type validation a bit more flexible. I would say this PR needs few more things done before it should be accepted:
 - The pending PR to virtus needs to be accepted.
 - Is `Grape::Validations::Types` a good namespace name?
 - See `CustomTypeCoercer#infer_type_check`. There's support there for more involved type checking than just `value.class == primitive`, but I'm unsure what the conventions should be. This should be agreed on and documented under `Grape::Dsl::Parameters#requires`. Alternately arbitrary type-check control can be achieved by supplying a `Virtus::Attribute` subtype overriding `value_coerce?`, see `lib/grape/validations/types/json.rb` and `file.rb`. This could also perhaps be better documented.
 - The pre-rolled attributes in `lib/grape/validations/types/date.rb` and `date_time.rb` are as-yet undocumented. That's because these 'convenience' types must currently be accessed as `type: Grape::Validations::Types::Date::Iso8601`... I'm unsure of the best way to make these more readily available to `ParamsScope`.
 - I'm not 100% sure that coercion was tested to the point where I could confidently say no regressions have been introduced around validating nulls, blank strings, and the like. Removing the special type check for `Rack::Multipart::UploadedFile` for example, did not cause any failures until I added to the spec. This might need to be reviewed.
 - I do not have a good understanding of how some of the more exotic uses of the `type` parameter interact with the automatic documentation system. If anyone can point me in the right direction I would be happy to revise the code as needed.

I think what's missing now is a `coerce_from: :param or %w{array of params}` and you have a type system that can be adapted to most requirements.

lib/grape/validations/types.rb
```diff
@@ -0,0 +1,110 @@
...
+      # @return [Boolean] +true+ if special routines are available
+      def self.special?(type)
+        SPECIAL.key? type
+      end
+
+      # A valid custom type must implement a class-level `parse` method, taking
+      #   one String argument and returning the parsed value in its correct type.
+      # @param type [Class] type to check
+      # @return [Boolean] whether or not the type can be used as a custom type
+      def self.custom_type?(type)
```
dblock: Should be just `custom?` to match the other names.

lib/grape/validations/validators/coerce.rb
```diff
@@ -1,3 +1,5 @@
+require 'grape/validations/types'
```
dblock: Shouldn't need this.

lib/grape/validations/validators/coerce.rb
```diff
@@ -87,46 +53,29 @@ def type
...
-                          @option[:method].method(:parse)
-                        else
-                          # Otherwise expect a lambda function or similar
-                          @option[:method]
-                        end
+          if type.is_a? Virtus::Attribute
+            # Accept pre-rolled virtus attributes without interference
+            type
+          else
+            converter_options = {
```
dblock: Feels like this should be in a different module.

dblock:
I really like this code and I think it's important to get some or all of it in. I would vote to even monkey-patch Virtus in Grape until there's a version out there with the PR, or implementing a work-around in Grape that explicitly does what that PR does. Either way, bring this PR to a green state and I'll merge it trusting our specs not to introduce any regressions.

dslh:
Tidied up a bit:
 - `lib/grape/validations/types/virtus_collection_patch.rb` duplicates the virtus PR, works with both fixed and unfixed virtus versions so we can remove it after a fixed virtus gem is published.
 - Logic from `CoerceValidator#converter` moved to module function `Types.build_coercer`. Not crazy about the name, `coercer_for` maybe?
 - I've removed `Types::Date` and `Types::DateTime`, the specs seem to run foul of bugs in certain ruby versions and it's not really core grape stuff anyway. If there's any interest I could put together a 'grape-formats' gem with these and a few other more or less common formats.

Most of the inference that `CustomTypeCoercer` does on the `type` parameter still isn't documented in the README. I don't have any strong opinion as to how it should work, maybe leaving it buried will encourage proposals.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 * Your contribution here.
 
+* [#1167](https://github.com/ruby-grape/grape/pull/1167): Refactor and extend coercion and type validation system - [@dslh](https://github.com/dslh).
```
dblock: You've added `type: File`, right?

dblock:
I'm merging this, great work. Lets iterate on top of it.

CHANGELOG.md
dslh: Yep, doc'd and spec'd.

dblock: Maybe should be added to changelog as a line, like that we support `File` now out of the box.

PR #1170, opened 2015-10-02
allow dashes, periods, and other RFC6838-compliant characters in header vendor

suan:
The current accept header parsing logic treats everything after the first `-` as the version, whereas it's not uncommon to have dashes in more complex vendor/resource media type definitions (an IANA-registered example is `application/vnd.ms-office.activeX+xml`)

This change correctly handles dashes in the vendor part, while still assuming what's after the last dash is the version.

The current logic also doesn't allow underscores, ampersands, etc, which are allowed media type characters per [RFC 6838](http://tools.ietf.org/html/rfc6838#section-4.2). This change makes Grape itself handle these characters correctly, however for them to actually work is contingent on https://github.com/mjackson/rack-accept/pull/17 which unfortunately has been open for more than 6 months (I'll ping maintainer again, but any weight you can throw around there would also be appreciated :smile:)

_P/S: It should be noted that the current header versioning logic is very limited and doesn't support formats such as `application/vnd.github.v3+json` (as what [github uses now](https://developer.github.com/v3/media/#request-specific-version)) I'll try to add a caveat in the README while I add the CHANGELOG entry for this_

### DIFF
--- a/lib/grape/middleware/versioner/header.rb
+++ b/lib/grape/middleware/versioner/header.rb
@@ -8,13 +8,13 @@ module Grape
       # application/vnd.:vendor-:version+:format
       #
       # Example: For request header
-      #    Accept: application/vnd.mycompany-v1+json
+      #    Accept: application/vnd.mycompany.a-cool-resource-v1+json
       #
       # The following rack env variables are set:
       #
       #    env['api.type']    => 'application'
-      #    env['api.subtype'] => 'vnd.mycompany-v1+json'
-      #    env['api.vendor]   => 'mycompany'
+      #    env['api.subtype'] => 'vnd.mycompany.a-cool-resource-v1+json'
+      #    env['api.vendor]   => 'mycompany.a-cool-resource'
       #    env['api.version]  => 'v1'
       #    env['api.format]   => 'json'
       #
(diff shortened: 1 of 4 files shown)


dblock:
This is good. It needs a CHANGELOG entry and possibly a README update, please.

I would be down with a monkey-patch of rack-accept and a corresponding test. We have some precedent for patches in https://github.com/ruby-grape/grape/pull/1167 where we monkey-patched virtus.

lib/grape/middleware/versioner/header.rb
```diff
@@ -177,11 +177,16 @@ def request_vendor(media_type)
...
+          _, subtype = Rack::Accept::Header.parse_media_type(media_type)
+          subtype.match(VENDOR_VERSION_HEADER_REGEX)[2]
+        end
+
         # @param [String] media_type a content type
         # @return [Boolean] whether the content type sets an API version
         def version?(media_type)
           _, subtype = Rack::Accept::Header.parse_media_type(media_type)
-          subtype[/\Avnd\.[a-z0-9*.]+-[a-z0-9*\-.]+/]
+          subtype[/\Avnd\.([a-z0-9.\-_!#\$&\^]+?)(?:-([a-z0-9*.]+))+/]
```
aselder: Should we extract this to a constant to avoid recreating the regex every time?

suan:
@dblock I've updated the CHANGELOG, added a README section about Grape's limitations with the structure of versioned mediatypes, and converted those regexes to constants like recommended by @aselder 

I'm thinking I can make a followup PR where the appropriate rack-accept call is monkey-patched to allow all the RFC-compliant characters, once this PR is merged.

dblock:
Merging.

PR #1172, opened 2015-10-04
Removes redundant nil? check.

towanda:
If we are calling ```.blank?```, calling ```.nil?``` first is redundant.


dblock:
:+1:

marshall-lee:
Maybe replace `return nil` with just `return`?

dblock:
@marshall-lee :+1:

towanda:
@dblock @marshall-lee You are right, fixed here: https://github.com/ruby-grape/grape/pull/1176

PR #1179, opened 2015-10-07
Patch rack-accept to support all valid RFC6838 characters in media types

suan:
Continuation and completion of https://github.com/ruby-grape/grape/pull/1170

This patches rack-accept in a manner similar to how virtus was patched (as discussed)


dblock:
For reference this has been PRed into rack-accept as https://github.com/mjackson/rack-accept/pull/17.

suan:
Getting `Gem::InstallError: rack-cache requires Ruby version >= 2.0.0` errors on the Rails3 Gemfile jruby build in Travis... but I don't see how these changes would cause that :confused:

dblock:
Maybe there was a new release of rack-cache or something like that? Would you try to hunt it down please?

dblock:
Yep, rack-cache was released yesterday, https://rubygems.org/gems/rack-cache. So something broke there, lets fix that first?

dblock:
https://github.com/rtomayko/rack-cache/issues/120

suan:
Yeah, this seems tricky. The way I'm reading it is that Rails itself on Ruby1.9 will break now and will need to be updated, as actionpack 3.x will try to install rack-cache v1.3: https://github.com/rails/rails/blob/v3.2.22/actionpack/actionpack.gemspec

dblock:
I would be ok to force an older version of rack-cache in Grape's gemfile. Want to try?

suan:
Sure, on it

rnubel:
First of all, hi @suan! So from what I can tell, only Rails 3 requires rack-cache, so I'd suggest we put this into the `Appraisals` file for `rails-3`.

suan:
Hey @rnubel !! :smile: So I tried putting it in `Appraisals` instead of the Gemfile but it doesn't seem to have the desired effect, whereas having it in the Gemfile itself worked... could u take a look and see whether my changes were what you had in mind?

rnubel:
@suan: you have to run `appraisal install`, to re-generate the `gemfiles/*` files. That could be part of the Travis build, maybe, but right now it's manual.

suan:
Thanks @rnubel - done!

@dblock Everything outside of the "Allowed Failures" section has passed. The very last specs in that section has been running for almost an hour tho and I think it's just TravisCI wonkiness?

suan:
Force-pushed again and Travis is :green_heart:! :)

dblock:
Merged.

PR #1181, opened 2015-10-13
Update describing method headers example

huynhquancam:
Should we change the example to this as `[foo: :bar]` will actually return an array that contains a hash?


dblock:
Which one works?

huynhquancam:
I believe they both work but it looks weird for header to be an array having a hash? It's harder to iterate as well.

dblock:
Fair enough, I'll merge this.

Issue #1178, opened 2015-10-07
Grape::Exceptions::UnknownValidator


gottfrois:
I have the following files:

```ruby
# app/controllers/api/v2/base.rb
module Api
  module V2
    class Base < Grape::API
      mount V2::Jobs
    end
  end
end

# app/controllers/api/v2/jobs.rb
module Api
  module V2
    class Jobs < Grape::API
      resources :jobs do
        params do
          requires :id, type: String, uuid: true
        end
        get ':id' do
        end
      end
    end
  end
end

# app/controllers/api/v2/validators/uuid.rb
module Api
  module V2
    module Validators
      class Uuid < ::Grape::Validations::Base

        REGEX = %r{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}}

        def validate_param!(attr_name, params)
          unless (params[attr_name] =~ REGEX).zero?
            fail ::Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: 'must be a valid UUID'
          end
        end

      end
    end
  end
end
```

If I want the above to work, I have to manually require the validator class in `base.rb` file, like so:

```ruby
require_relative './validators/uuid'
```

which result in throwing warnings:

```
app/controllers/api/v2/validators/uuid.rb:6: warning: already initialized constant Api::V2::Validators::Uuid::REGEX
app/controllers/api/v2/validators/uuid.rb:6: warning: previous definition of REGEX was here
```

since rails will automatically require the validator as well. Any idea why grape is not able to pick up the custom validator? Looks like a require order issue or something. Any clue?

Thanks

towanda:
Yes, It seems like a require order issue, how are you loading grape in rails?

gottfrois:
nothing fancy, i have grape in the gemfile and have this in my route:

```
mount Api::V2::Base => '/'
```

dm1try:
> Any idea why grape is not able to pick up the custom validator?

`grape` doesn't have an autoloading mechanism for custom validators (PR are welcomed).

In your case you can use [require-dependency] (http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#require-dependency)
```ruby
# app/controllers/api/v2/base.rb
require_dependency 'api/v2/validators/uuid'
```

But I should explain why this happens.
First, how does `grape` load buit-in validators?
[...]

PR #1186, opened 2015-10-16
add possibility to automatically load user-defined validators(a.k.a custom)

dm1try:
the solution relies on ActiveSupport::Dependencies, so it works for Grape app mounted on Rails by default.

``` bash
 --app
   --controllers
     --api
       --validators
         custom.rb
```

``` ruby
params do
   requires :my_param, custom: true
end
```

on the dev enviroment the code of custom validator is refreshed as any other dependency

"proof-of-concept" that we discussed in #1178

### DIFF
--- a/lib/grape/validations/autoload.rb
+++ b/lib/grape/validations/autoload.rb
@@ -1,0 +1,52 @@
+module Grape
+  module Validations
+    class Autoload
+      DEFAULT_PATH = %w(api validators)
+
+      def initialize(api)
+        @api = api
+        @loaded_validators = []
+      end
+
+      def try_load(name)
+        return nil unless defined?(::ActiveSupport::Dependencies)
+
+        name = name.to_s
+        inner_path   = File.join(@api.to_s.deconstantize.underscore, 'validators', name).to_s
+        default_path = File.join(DEFAULT_PATH, name).to_s
+
+        found_path = [inner_path, default_path].find do |path|
+          begin
+            require_dependency(path)
+            mark_as_loaded(name)
+            update_before_remove_const_callback(path)
+
+            true
+          rescue LoadError
+            nil
+          end
+        end
+        (found_path && found_path.classify.constantize) || nil
+      end
+
+      private
+
+      def mark_as_loaded(name)
+        @loaded_validators << name
+      end
+
+      def update_before_remove_const_callback(path)
+        parent_const = path.classify.split('::').first.constantize
+
+        # see ActiveSupport::Dependencies.remove_unloadable_constants!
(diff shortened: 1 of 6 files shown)


lib/grape/validations/autoload.rb
```diff
@@ -0,0 +1,52 @@
+module Grape
+  module Validations
+    class Autoload
+      DEFAULT_PATH = %w(api validators)
```
dblock: Having a default path `validators` makes sense to me, but `api` feels too specific.

lib/grape/validations/autoload.rb
```diff
@@ -0,0 +1,52 @@
...
+            require_dependency(path)
+            mark_as_loaded(name)
+            update_before_remove_const_callback(path)
+
+            true
+          rescue LoadError
+            nil
+          end
+        end
+        (found_path && found_path.classify.constantize) || nil
```
dblock: `|| nil`?

lib/grape/validations/autoload.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      private
+
+      def mark_as_loaded(name)
+        @loaded_validators << name
+      end
+
+      def update_before_remove_const_callback(path)
+        parent_const = path.classify.split('::').first.constantize
```
dblock: I think this is incorrect for A::B::C::D.

dblock:
I like it, but I would try to simplify it. Maybe with Rails we can use a Railtie to "enable" it or set paths, but otherwise not have autoload available or something like that?

lib/grape/validations/autoload.rb
dm1try: oh, it's redundant )

dm1try:
@dblock I agree with your comments above. It's just a start point.

We need more tests, an implementation for "non-rails" apps(maybe using straitforward `require_relative`) 
Also it would be good if we extract registration logic from `Validatation#inherited` callback(as I mentioned in #1178) to make things a little bit cleaner.

lib/grape/validations/autoload.rb
gottfrois: I agree. What about the autoload tries to find the constant in grape validators directory `lib/grape/validations/validators/custom.rb`?

Client would need to put their custom validators in `lib/grape/validations/validators/` which sounds reasonable to me.

dm1try: @gottfrois sounds reasonable, I agree.

dblock:
:+1:

dm1try:
closed in favor of #2204

PR #1187, opened 2015-10-17
Routes helpers

saveav:
Add Grape::Routes to make it easy to refer to the routes as a procedure call


dblock:
Nice work, however:

* there's already support for gathering routes in Grape, you can just do `API.routes`, how is this different?
* have you seen https://github.com/reprah/grape-route-helpers? 

We've had quite a debate about the helpers side in https://github.com/ruby-grape/grape/pull/983 and and decided that it didn't belong in Grape proper.

dblock:
@saveav Still curious to hear your thoughts on the above.

saveav:
Please feel free to close :)

Sorry long reply, haven't check this email address for a while.

PR #1188, opened 2015-10-19
Allow parameters with more than one type.

### LINKED ISSUE #693: Allow multiple parameter types

dslh:
Fixes #693.

Adds `types` option for `requires` and `optional` endpoint parameter
declarations, allowing for parameters that have more than one allowed
type. See README.md for usage.

I have been thinking about the best way to expose some of the more
interesting possibilities for arbitrary type validation that are hidden in Grape,
but in the end this is the only use case I have seen that makes sense to me.

### DIFF
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -177,10 +177,7 @@ module Grape
       def validates(attrs, validations)
         doc_attrs = { required: validations.keys.include?(:presence) }
 
-        # special case (type = coerce)
-        validations[:coerce] = validations.delete(:type) if validations.key?(:type)
-
-        coerce_type = validations[:coerce]
+        coerce_type = infer_coercion(validations)
 
         doc_attrs[:type] = coerce_type.to_s if coerce_type
 
@@ -212,7 +209,7 @@ module Grape
           validations.delete(:presence)
         end
 
-        # Before we run the rest of the validators, lets handle
+        # Before we run the rest of the validators, let's handle
         # whatever coercion so that we are working with correctly
         # type casted values
         coerce_type validations, attrs, doc_attrs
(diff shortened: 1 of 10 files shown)


lib/grape/validations/params_scope.rb
```diff
@@ -178,12 +178,23 @@ def validates(attrs, validations)
         doc_attrs = { required: validations.keys.include?(:presence) }
 
         # special case (type = coerce)
-        validations[:coerce] = validations.delete(:type) if validations.key?(:type)
+        validations[:coerce] = validations[:type] if validations.key?(:type)
+        validations[:coerce] = validations.delete(:types) if validations.key?(:types)
```
dblock: Maybe do something here when you have both? Should probably be an exception.

dblock:
I think this is really good. I am down to merging this unless you think I should hold? Anyone else wants to comment?

dslh:
There's the suggested guard and a quick stab at refactoring params_scope.rb a bit. Bundling some of this coercion stuff out could be a good move but I won't get any more time to work on this until at least the weekend. I can either do a supplemental PR or keep working on this one when I get the chance.

dblock:
I will merge this and we can iterate on top.

Issue #1162, opened 2015-09-21
XML formatter can raise errors in scenarios where it shouldn't possibly even be invoked


dblock:
The scenario in https://github.com/ruby-grape/grape/pull/1159 is that OPTIONS returns an empty string that the XML formatter tries to serialize and fails.

The scenario in https://github.com/ruby-grape/grape/pull/1157 is that a redirect causes an empty body and fails.

dblock:
@urkle would love a spec that fails with an accept header which included xml but not JSON

@franzliedke would love a spec where the redirect causes the same error

I'll take suggestions on what the "right" fix would be here.

urkle:
What about adding a method that can be called by the "redirect / options" handlers to disable the use of the formatters?

urkle:
Here is a spec that fails with the current code. This is using request headers that firefox sends in the OPTIOPNS request that it sends before a POST request.  It sends this Accept header regardless of whether the POST has it's Accept header set to JSON or XML!

```ruby
    it 'options fails with a 500 error with an Accepts header that includes xml' do
      subject.content_type :xml, 'application/xml'
      subject.default_format :xml
      subject.post 'example' do
        {message: 'example'}
      end
      options '/example', {}, {
        Origin: 'http://example.com',
        Accept: 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
        'Acccess-Control-Request-Method' => 'POST'
      }
      expect(last_response.status).to eql 204
      expect(last_response.body).to eql ''
    end
```

Also I noticed that the OPTIONS request by default is sending back an Allow header..  the Allow header is only supposed to be sent back with a 406 response.   OPTIONS uses the Access-Control-Allow-Methods to return that information as to what methods are allowed on this path..

urkle:
So, a cleaner proposed change.   the redirect and option essentially have NO content.. (thus should not even have the Content-Type header).   So we need a way for the handler to denote this to the grape core.  either via a method  e.g. ```no_content```  or by returning some fixed value (e.g. a class) that can be tested against and the code would bypass the formatters in that scenario and omit the content-type header.

dblock:
Why can't we make the XML formatter behave like the JSON formatter?

urkle:
because an empty string is not valid XML.. but an empty string "" IS valid json..   JSON can be simply 
```json
""
```

[...]

PR #1190, opened 2015-10-21
Bypass formatting for statuses with no entity-body

tylerdooling:
## tl;dr;
Allow the formatter middleware to bypass serialization of response bodies when the status indicates there is no content.

## Background
#1162 describes a scenario where `OPTIONS` requests and redirects can raise exceptions when content types are limited to those that don't consider `''` as valid or are not prepared to handle it.  The specific example given involved content types being limited (or set by default) to XML. `''` is not consider to be valid XML and as such results in an error when serialization is attempted. 

A resolution was proposed to handle this in the formatters by checking the request method.  It seems - and was suggested - that this opens up the scope of concern for the formatters too much and I agree. Additionally, it places a burden on all custom formatters to account for responses intended to have no content - like the OPTIONS request in question.

## Process
I initially referenced the RFC to determine the best course of action, and found this:
> A 200 response SHOULD include any header fields that indicate optional features implemented by the server and applicable to that resource (e.g., Allow), possibly including extensions not defined by this specification. The response body, if any, SHOULD also include information about the communication options. The format for such a body is not defined by this specification, but might be defined by future extensions to HTTP. Content negotiation MAY be used to select the appropriate response format. If no response body is included, the response MUST include a Content-Length field with a field-value of "0".

This indicates that while the specification for a response body is not defined in the RFC, it is most certainly allowed given that the status code is `200` and should ideally be accounted for. This led me to begin ensuring that `OPTIONS` requests could be serialized appropriately by the formatter middleware, which led to ideas involving a special representation of no content that could be checked conditioned there.

When thinking through this and beginning to write some tests, I found that I was unable to get the content type to persist in the response.  This led me to [Rack](https://github.com/rack/rack) where I found [this](https://github.com/rack/rack/blob/master/lib/rack/response.rb#L63):

```ruby
 if [204, 205, 304].include?(status.to_i)
   header.delete CONTENT_TYPE
   header.delete CONTENT_LENGTH
   close
   [status.to_i, header, []]
 else
   [status.to_i, header, BodyProxy.new(self){}]
 end
```
After some additional digging I found that `Rack::Lint` takes it a step further and does not consider a response to be valid if it has content information and it's status code is included in `Rack::Utils::STATUS_WITH_NO_ENTITY_BODY`.
- https://github.com/rack/rack/blob/master/lib/rack/lint.rb#L659
- https://github.com/rack/rack/blob/master/lib/rack/lint.rb#L673

## Proposal
Using these observations, my proposal here is to condition the formatter middleware to look for response statuses that indicate a content body in not appropriate and make the decision to bypass serialization.

This allows for both `OPTIONS` routes that return a `200` with a serialized body based on normal content negotiation, and for the default `OPTIONS` route that is auto-generated by the framework which returns a `204` and no content and does add additional concerns to each default and custom formatter.

Additionally, I believe this addresses the second issue in #1162 involving a similar issue with redirects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 * Your contribution here.
 
+* [#1190](https://github.com/ruby-grape/grape/putt/1190): Bypass formatting for statuses with no entity-body - [@tylerdooling](https://github.com/tylerdooling).
 * [#1188](https://github.com/ruby-grape/grape/putt/1188): Allow parameters with more than one type - [@dslh](https://github.com/dslh).
 * [#1179](https://github.com/ruby-grape/grape/pull/1179): Allow all RFC6838 valid characters in header vendor - [@suan](https://github.com/suan).
 * [#1170](https://github.com/ruby-grape/grape/pull/1170): Allow dashes and periods in header vendor - [@suan](https://github.com/suan).
--- a/README.md
+++ b/README.md
@@ -1976,6 +1976,11 @@ Built-in formatters are the following.
 * `:serializable_hash`: use object's `serializable_hash` when available, otherwise fallback to `:json`
 * `:binary`: data will be returned "as is"
 
+Response statuses that indicate no content as defined by [Rack](https://github.com/rack)
+[here](https://github.com/rack/rack/blob/master/lib/rack/utils.rb#L567)
+will bypass serialization and the body entity - though there should be none -
+will not be modified.
+
 ### JSONP
 
 Grape supports JSONP via [Rack::JSONP](https://github.com/rack/rack-contrib), part of the
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -15,6 +15,14 @@ Identical endpoints with different versions now work correctly. A regression int
 
 See [#1114](https://github.com/ruby-grape/grape/pull/1114) for more information.
 
+#### Bypasses formatters when status code indicates no content
+
+To be consistent with rack and it's handling of standard responses
+associated with no content, both default and custom formatters will now
+be bypassed when processing responses for status codes defined [by rack](https://github.com/rack/rack/blob/master/lib/rack/utils.rb#L567)
+
+See [#1190](https://github.com/ruby-grape/grape/pull/1190) for more information.
+
 ### Upgrading to >= 0.12.0
 
 #### Changes in middleware
--- a/lib/grape/middleware/formatter.rb
+++ b/lib/grape/middleware/formatter.rb
@@ -20,6 +20,7 @@ module Grape
 
       def after
         status, headers, bodies = *@app_response
+        return @app_response if Rack::Utils::STATUS_WITH_NO_ENTITY_BODY.include?(status)
 
         if bodies.is_a?(Grape::Util::FileResponse)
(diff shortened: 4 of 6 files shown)


tylerdooling:
I left my refactoring commits separate as to not add noise when reviewing the feature change.  Please let me know if you'd prefer them squashed.

dblock:
This makes a lot of sense. It definitely needs an entry in UPGRADING.

I would be down with just merging this "as is", fix the build though please ;)

tylerdooling:
Got caught by the cop!

Thanks for the reminder regarding UPGRADING.  I added a section there.  Please let me know if it is adequate.

dm1try:
As for me, it makes more sense if we could skip the formatter explicity outside the formatter scope. I mean
```ruby
# endpoint
def status(code)
   skip_formatter_middleware if no_body_status_code(code)
   ...
end

def redirect
   skip_formatter_middleware
   ...
end

def skip_formatter_middleware
   env['grape.formatter.skip'] = true
end

# formatter
def call(env)
   return response if  env['grape.formatter.skip']
end
```

dblock:
@dm1try I think it makes sense to do this by default wrt `STATUS_WITH_NO_ENTITY_BODY`, no? An `ENV` variation that could go one way or another could be an addon I think.

Tell us how strongly you feel about this or something else before I merge?

dm1try:
@dblock PR looks good, you can merge it "as is")

dm1try:
My point was about having possibility to skip `formatter` middleware for some reason(the STATUS_WITH_NO_ENTITY_BODY is one of them).

dblock:
Merged

dblock:
Can we close #1162?

tylerdooling:
#1159 (OPTIONS) is definitely addressed here, but I recant my suggestion regarding #1157 (Redirects).

[The RFC](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3) suggests for entity bodies for `301` and `302` responses:

> the entity of the response SHOULD contain a short hypertext note with a hyperlink to the new URI(s).

Do you think it would make sense for `#redirect` to set the body to a Location-like object that supports standard serialization? - i.e. responds to `#to_json`, `#to_xml`, and `#to_s`.  It seems like that would be consistent with the theme of the spec.  In addition, an override could be provided in `options` or someplace else if desired by the user.

tylerdooling:
Another approach could be to set the content type explicitly inside `#redirect` and allow users to override it via an optional argument:

```ruby
def redirect(url, options = {})
  # ...
  header 'Location', url
  header Grape::Http::Headers::CONTENT_TYPE, options.fetch(:content_type) { env[Grape::Env::API_FORMAT] }
  body ''
end
```
The downside is that of course that the burden is on the user if they support multiple serialization formats.  In that case though, it could be as simple as:
```ruby
redirect(some_url, content_type: env[Grape::Env::API_FORMAT]
```
In that case it might make sense for them to be able to override the body as well to ensure it is serializable.

@dblock - do either of those suggestions strike you as appropriate?

dblock:
I am not sure. Redirect with content type seems strange to me. 

I think this discussion should be re-raised in a new issue either way.

PR #1193, opened 2015-10-23
Fixed regexp that match unintended string in README.md.

ionis111:
1. In ruby, regexp default mode is multiline mode.
2. [[:alnum:]] don't match only ascii character. it match also multibyte character.

for example.

about 1.
```ruby
$ irb
2.2.1 :001 > "aaaa\nbbbb\nccc" =~ /^b+$/
 => 5
2.2.1 :002 > "aaaa\nbbbb\nccc" =~ /\Ab+\z/
 => nil
```

about 2.
```ruby
$ irb
2.2.1 :001 > "あ" =~ /\A[[:alnum:]]+\z/
 => 0
2.2.1 :002 > "A" =~ /\A[[:alnum:]]+\z/
 => 0
```

I think that it may become vulnerability.
so, it is better to quit as a sample.

### DIFF
--- a/README.md
+++ b/README.md
@@ -684,7 +684,7 @@ You can define validations and coercion options for your parameters using a `par
 ```ruby
 params do
   requires :id, type: Integer
-  optional :text, type: String, regexp: /^[a-z]+$/
+  optional :text, type: String, regexp: /\A[a-z]+\z/
   group :media do
     requires :url
   end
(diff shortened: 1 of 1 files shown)


dblock:
But I think we want to be able to pass non-US-ASCII text there, don't we? It's an AlphaNumeric validator and I suppose the example should work for non-English text. What do you think?

ionis111:
Thank you for your reply.

I think that AlphaNumeric validator should pass only alphabet and number.
But, if the validator that want to be able to pass non-US-ASCII text there is your policy, I think it's good too.

What do you think about using multiline mode?

dblock:
You mean English alphabet? :) I think it should be able to pass non-US-ASCII text, so lets leave it alone. I agree on multiline mode though, it shouldn't allow multiline anything. Care to amend this to reflect that? Thanks.

ionis111:
> You mean English alphabet? :)
Yes.

> I think it should be able to pass non-US-ASCII text, so lets leave it alone. I agree on multiline mode though, it shouldn't allow multiline anything. Care to amend this to reflect that?
I understand.

I amend commit.
Please check it.

dblock:
Merged, thanks!

PR #1201, opened 2015-11-11
Don't call builder on every request.

marshall-lee:
Hello!

This is an attempt to fix a most wrong thing in Grape: middleware stack is initializing *on every request* and even more Grape `dup` them when they're called.

I don't know should it be merged or not. It might break someone's code even when specs are green.

### DIFF
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -188,6 +188,7 @@ module Grape
     end
 
     def call(env)
+      build_app
       dup.call!(env)
     end
 
@@ -195,13 +196,8 @@ module Grape
       extend helpers
 
       env[Grape::Env::API_ENDPOINT] = self
-      if options[:app]
-        options[:app].call(env)
-      else
-        builder = build_middleware
-        builder.run ->(arg) { run(arg) }
-        builder.call(env)
-      end
+      @env = env
+      @app.call(env)
     end
 
     # Return the collection of endpoints within this endpoint.
@@ -216,9 +212,8 @@ module Grape
 
     protected
 
-    def run(env)
+    def run
       ActiveSupport::Notifications.instrument('endpoint_run.grape', endpoint: self, env: env) do
-        @env = env
         @header = {}
 
         @request = Grape::Request.new(env)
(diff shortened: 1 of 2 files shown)


marshall-lee:
The most scary thing here is that if I put `build_app` call to `initialize` there is exactly one failing spec: `rspec spec/grape/api_spec.rb:2431` that was caught by accident, I think.

I need someone to carefully review this change.

dblock:
I actually think this is absolutely right, it's amongst the oldest pieces of code that nobody dared touch. It also used to be much more involved when we didn't have all those stackable settings. I read the code and ran your branch against my large API test suite and everything but [grape-newrelic](https://github.com/xinminlabs/newrelic-grape) worked, too.

I'd like to merge this, as "in specs we trust". I'll debug that one unless someone beats me to it.

It would be also great to get some benchmarks wrt what this change means to performance. Then I think you should send an email to the mailing list asking people to try HEAD, maybe someone will be brave enough to even try it in production :)

dm1try:
Such a nice PR,  
> everything but grape-newrelic worked

it [overrides the original behaviour](https://github.com/xinminlabs/newrelic-grape/blob/master/lib/newrelic-grape/instrument.rb#L67-L75) 

Looks good to me!

dblock:
Right on @dm1try. I merged this and opened https://github.com/xinminlabs/newrelic-grape/issues/31, lets come up with a better interface to register middleware other than a monkey patch?

dm1try:
@dblock I agree.

lib/grape/endpoint.rb
```diff
@@ -188,6 +188,7 @@ def compile_path(prepared_path, anchor = true, requirements = {})
     end
 
     def call(env)
+      build_app
```
etehtsea: is it threadsafe to do `@app` lazy initialization here?

dm1try: On `YARV` - yes.
On 'JRuby' - theoretically `#build_stack` can be called multiple times but I don't see any side effect here( see the implementation of `#build_stack`); just extra an extra calculation, if many threads will try to set `@app`.
Anyway

dm1try: @etehtsea , I updated the comment sorry :)

marshall-lee: @dm1try @etehtsea personally i didn't want lazy initialization but when i put it in `initialize` there're failing test :(

PR #1203, opened 2015-11-14
Allow custom coercion failure messages

zuk:
Currently if something goes wrong inside a custom param type's `.parse` method, the only feedback in the resulting validation failure response is a generic `"____ is invalid"` message. This pull request allows custom coercion failure messages by having the `.parse` method return an `InvalidValue` initialized with the failure message.

For example:

``` ruby
class Color
  attr_reader :value
  def initialize(color)
    @value = color
  end

  def self.parse(value)
    if  %w(blue red green).include?(value)
      new(value)
    else
      Grape::Validations::Types::InvalidValue.new "is not a valid color"
    end
  end
end

# ...

params do
  requires :color, type: Color, default: Color.new('blue')
end

get '/stuff' do
  # params[:color] is already a Color.
  params[:color].value
end
```

### DIFF
--- a/lib/grape/validations/validators/coerce.rb
+++ b/lib/grape/validations/validators/coerce.rb
@@ -13,8 +13,16 @@ module Grape
       def validate_param!(attr_name, params)
         raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: message(:coerce) unless params.is_a? Hash
         new_value = coerce_value(params[attr_name])
-        raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: message(:coerce) unless valid_type?(new_value)
-        params[attr_name] = new_value
+        if valid_type?(new_value)
+          params[attr_name] = new_value
+        else
+          bad_value = new_value
+          if bad_value.is_a?(Types::InvalidValue) && !bad_value.message.nil?
+            fail Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: bad_value.message.to_s
+          else
+            fail Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message_key: :coerce
+          end
+        end
       end
 
       private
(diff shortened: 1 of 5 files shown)


dblock:
Great, can you squash these commits please?

zuk:
Ok done. Not sure what's up with those failed Rubinius build in Travis.

lib/grape/validations/validators/coerce.rb
```diff
@@ -11,7 +11,12 @@ def validate_param!(attr_name, params)
         if valid_type?(new_value)
           params[attr_name] = new_value
         else
-          fail Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message_key: :coerce
+          bad_value = new_value
+          if bad_value.is_a?(Types::InvalidValue) && !bad_value.message.nil?
```
dblock: I am not in love with this. We're now explicitly checking what kind of exception we got. Any better suggestions?

zuk:
I think there's a bigger problem here -- coercion and validation take separate paths but are really converging towards the same thing (i.e. we'll eventually end up need to do full-fledged validation inside `.parse`) -- but this feels too monumental for a casual contributor to take on. Another way to do this would be to give custom types an optional `.validate` method that behaves like a regular custom validator, and when present, gets called prior to coercion. I considered doing it this way but thought this way would be more likely to get pulled since it leverages the exiting `BadValue` type, which already seems a bit out of place (it feels like an exception, but isn't...)

dblock:
I encountered something related in #1675. I think our validation + coercion needs some serious revisiting. Just thinking out loud.

zuk:
I agree :) We've been using our own fork of Grape with the above change, periodically rebasing to origin. It works, but it's pretty hacky. There's probably a bigger architectural issue starting to surface here, but like I said before, it feels like too much for me to try to take on.

PR #1204, opened 2015-11-15
Ditch each_with_object

marshall-lee:
Remove `each_with_object` where it's slow.
    
Remember https://github.com/ruby-grape/grape/pull/1128?
Unfortunately, calling `each_with_object` on `Hash` is not a good idea (`inject` has the same problem btw). It allocates an array for each `(k,v)` pair which is not good.
    
More details here: https://github.com/rails/rails/commit/960de47f0eef79d234eb3cfc47fabb470fef1529
@tenderlove exellently covered it all.

### DIFF
--- a/lib/grape/request.rb
+++ b/lib/grape/request.rb
@@ -2,27 +2,39 @@ module Grape
   class Request < Rack::Request
     HTTP_PREFIX = 'HTTP_'.freeze
 
+    alias_method :rack_params, :params
+
     def params
-      @params ||= begin
-        params = Hashie::Mash.new(super)
-        if env[Grape::Env::RACK_ROUTING_ARGS]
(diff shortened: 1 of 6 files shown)


lib/grape/request.rb
```diff
@@ -2,27 +2,39 @@ module Grape
...
     end
 
     def headers
-      @headers ||= env.each_with_object({}) do |(k, v), h|
+      @headers ||= _build_headers
+    end
+
+    private
+
+    def _build_params
```
dblock: Why these underscore methods? I think to be consistent they should just be `build_params` and `build_headers`? Not to introduce another convention ...

marshall-lee: @dblock I agree

marshall-lee:
@dblock fixed

PR #1205, opened 2015-11-15
Memoize endpoint helpers.

marshall-lee:
This PR follows the same logic as https://github.com/ruby-grape/grape/pull/1201.


dblock:
RuboCop! ;)

marshall-lee:
@dblock should be fixed

PR #1208, opened 2015-11-19
Lazy initialization of Endpoint and trick with helpers.

marshall-lee:
Now I'm working on reducing calls to `extend` in `Grape` because they invalidate Ruby's method cache slowing everything down. I applied a *clever* trick with subclassing `Endpoint` and including helpers into this subclass.

It also forced me to fix a thread safety issue reported by @etehtsea here: https://github.com/ruby-grape/grape/pull/1201#discussion_r44900471 because calling `include` twice would be really inconsistent.

Also I'm sorry, 06930da does not actually fix anything, I overlooked it :)


dblock:
:+1:

PR #1211, opened 2015-11-27
mime-types-3 is OK with grape

halostatue:


dblock:
Not on JRuby.

halostatue:
It is, actually. The problem is in the appraisals that need further refinement to restrict mime-types to < 3.0 *only for those versions that do not support Ruby 2.0 syntax*. You’re also missing matrices for JRuby 9000.

I’ll look at adding those, but this is going to make it a bigger change for a project I don’t use.

dblock:
I am still interested in this. I just want a passing build :)

namusyaka:
Thanks for pointing this out.
However, mime-types has been moved from gemspec to Gemfile since [this commit](https://github.com/ruby-grape/grape/commit/e9c6885a8b11bd4c86277423724a16c9113ac491).
I'm going to close this.

Issue #1037, opened 2015-06-18
error! helper: symbol to JSON conversion


loto:
Hi,

In 0.11.0:

``` ruby
error! :unauthorized, 401
```

Was returned as:

``` json
{ "error": "unauthorized" }
```

In 0.12.0 I get:

``` json
""unauthorized""
```

Does anyone else experience this?

dblock:
Something is going on with either the error format or with what is being returned in a handled error. Read through https://github.com/intridea/grape/blob/master/UPGRADING.md carefully in case there's a clear answer (there're some things around errors), and if that doesn't help put up a project where this is reproduced?

L-Oto:
I found nothing relevant in the upgrading documentation. I'll try to reproduce it in a small project.

dm1try:
@dblock see https://github.com/ruby-grape/grape/blob/master/lib/grape/error_formatter/json.rb#L8 
```ruby
result = message.is_a?(String) ? { error: message } : message
```

dblock:
I believe this is resolved via #1216 and some left-overs in #1309. Closing. Please reopen or add comments to #1309.

PR #1216, opened 2015-12-03
Corrects strings in json error responses

jrforrest:
Addresses #1037

### DIFF
--- a/lib/grape/error_formatter/json.rb
+++ b/lib/grape/error_formatter/json.rb
@@ -3,13 +3,22 @@ module Grape
     module Json
       class << self
         def call(message, backtrace, options = {}, env = nil)
-          message = Grape::ErrorFormatter::Base.present(message, env)
+          result = wrap_message(Grape::ErrorFormatter::Base.present(message, env))
 
-          result = message.is_a?(String) ? { error: message } : message
           if (options[:rescue_options] || {})[:backtrace] && backtrace && !backtrace.empty?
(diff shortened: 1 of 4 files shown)


jrforrest:
Looks like this is breaking on some platforms.  Investigating...

lib/grape/error_formatter/json.rb
```diff
@@ -3,14 +3,23 @@ module ErrorFormatter
...
             result = result.merge(backtrace: backtrace)
           end
           MultiJson.dump(result)
         end
+
+        private
+
+        def wrap_message(message)
+          if message.is_a?(Exceptions::ValidationErrors)
+            return message
```
dblock: Remove `return`, that's implied.  Expand the condition to be consistent:

```ruby
if message.is_a?(...)

elsif message.respond_to?(:to_h)

else

end
```

Or better

```
case
when ...
else
...

I am not in love with the check for `to_h`, I think you want to check `is_a?(Hash)` - or are you seeing a Mash or something? In which case I think it should still be checked explicitly.

jrforrest: Yeah you're right that conditional was sloppy.  Corrected.  Will squash if this looks better.

dblock:
It looks better, squash it.

dblock:
Merged via https://github.com/ruby-grape/grape/commit/844b6ace4f2298db8892bd79954d1ac64d9d59da.

Issue #1131, opened 2015-08-27
Context around array validation failures


lukestephenson:
Given a request like:
```
{ "orders": [ {customerId: 'invalid', salesPersonId: 5}, {customerId: 5, salesPersonId: 'invalid'} ] }
```
And the following parameter definitions:
```
    params do
      requires :orders, type: Array do
        requires :customerId, type: Integer
        requires :salesPersonId, type: Integer
      end
    end
```

Grape correctly detects that the input is invalid.  However there is no context in the errors collection about which item in the array was invalid.  Currently the error is reported against the key `orders[customerId]` but I need to be able to pair the errors back to the original element in the array like `orders[0][customerId]`.

Is this supported by grape?

dblock:
Grape doesn't provide array index context, and I think it should. Labelling this as a feature request.

towanda:
@dblock could we close this issue, since It's already implemented? Thanks.

dblock:
:+1:

akeuk:
In Grape version 0.14.1, Array index context only return `[0]`

#### Parameter definitions:

```ruby
params do
  requires :employees, type: Array do
    requires :name, type: String
    requires :email, regexp: /.+@.+/
  end
end
```

#### 1. Failure in _first_ element

Request (invalid email):

```javascript
{
  "employees": [
    {
      "name": "Jovani Kiehn",
      "email": "invalid_email"
    },
    {
      "name": "Eliezer Torphy",
      "email": "eliezer_torphy@gmail.com"
    }
  ]
}
```

Response:

```javascript
employees[0][email] is invalid
```

#### 2. Failure in _second_ element

Request (missing name):

```javascript
{
  "employees": [
    {
      "name": "Jovani Kiehn",
      "email": "jovani_kiehn@yahoo.com"
    },
    {
       "email": "eliezer_torphy@gmail.com"
    }
  ]
}
```

Response:

```javascript
employees[0][name] is missing
```

Thanks!

towanda:
@akeuk I'll take a look tonight, thanks!

akeuk:
@towanda Thank you! :)

How was it now?

galkon:
With this change now Hashes also return indexes so if for example I have
`
params do
  requires :employees, type: Hash do
    requires :name, type: String
    requires :email, type: Hash
  end
end
`

[...]

PR #1218, opened 2015-12-06
Provides Array index context in errors.

towanda:
This PR implements https://github.com/ruby-grape/grape/issues/1131


dblock:
Very nice, merging.

Issue #1206, opened 2015-11-16
Completely leaving out the version number in Accept header versioning schemes with `strict: true` does not result in a 406


suan:
When having an endpoint with:

```
version 'v1', vendor: 'foo.resource', using: header, cascade: false, strict: true
content_type :v1_resource, 'application/vnd.foo.resource-v1+json'
format :v1_resource
```

Any request with both "Accept: application/json" and "Accept: application/vnd.foo.resource-v1+json" headers present does not result in a 406, whereas the expected behavior is that it _would_ be a 406, because of the `strict` option.

(When only the "Accept: application/vnd.foo.resource-v1+json" header is present it correctly responds with a 406)

dblock:
That would be a problem. Could you please try this on HEAD? Some related issues may have been fixed.

If you have a moment, try to replicate this in a Grape test?

dblock:
And I mean specifically https://github.com/ruby-grape/grape/pull/1101.

stjhimy:
This can be closed since #1222 confirmed it was fixed.

dblock:
Closing.

PR #1222, opened 2015-12-09
Add spec for strict request with multiple accept headers

stjhimy:
As discussed on #1206 it was fixed on #1101 but there was not spec for multiple accept headers.


dblock:
:+1:

PR #1224, opened 2015-12-09
Fix build warnings

stjhimy:
Fix hash duplicated keys and raise_error specs avoiding warnings during the build.

### DIFF
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -1289,7 +1289,7 @@ describe Grape::API do
       subject.get '/exception' do
         fail 'rain!'
       end
-      expect { get '/exception' }.to raise_error
+      expect { get '/exception' }.to raise_error(RuntimeError, 'rain!')
     end
 
     it 'rescues all errors if rescue_from :all is called' do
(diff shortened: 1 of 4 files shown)


spec/grape/api_spec.rb
```diff
@@ -1289,7 +1289,7 @@ def three
       subject.get '/exception' do
         fail 'rain!'
       end
-      expect { get '/exception' }.to raise_error
+      expect { get '/exception' }.to raise_error(RuntimeError)
```
dblock: Since we're at it, you should explicitly check that the error raised is the one we threw, so not just `RuntimeError`, but the same message.

```ruby
expect { get '/exception' }.to raise_error RuntimeError, 'rail!'
```

Adjust the others too please?

spec/grape/validations_spec.rb
```diff
@@ -366,7 +366,7 @@ def define_requires_none
       end
 
       it "doesn't throw a missing param when param is present" do
-        get '/required', items: [key: 'hello', key: 'world']
```
dblock: Thanks for this.

stjhimy:
Specified the error messages.

dblock:
Thanks, merged.

PR #1225, opened 2015-12-10
Fix Given with nested params

### LINKED ISSUE #1219: Given does not use proper nesting in declared(params)
When using `given` block inside a nested requires block the validation works properly. But when trying to access these required params they are not nested in the requires block. Following spec (spec/grape/validations/params_scope_spec.rb) illustrate the problem

```
 it 'includes the nested parameter within #declared(params)' do
      subject.params do
        requires :bar, type: Hash do
          optional :a
          given :a do
            requires :b
          end
        end
      end
      subject.get('/nested') { declared(params).to_json }

      get '/nested', bar: { a: true, b: 'yes' }
      expect(JSON.parse(last_response.body)).to eq('bar' => { 'a' => true, 'b' => 'yes'})
    end
```

This fails with following message:

```
expected: {"bar"=>{"a"=>true, "b"=>"yes"}}
     got: {"b"=>nil, "bar"=>{"a"=>"true"}}

(compared using ==)

Diff:
@@ -1,2 +1,3 @@
-"bar" => {"a"=>true, "b"=>"yes"},
+"b" => nil,
+"bar" => {"a"=>"true"},
```

I tried different options (repeating the requires block in the given block) but nothing seems to work. I also cannot find the root of the problem but I guess it has to do with the following MR: https://github.com/ruby-grape/grape/pull/1047

JanStevens:
When a lateral scope is defined push the declared params on the parent and not itself,

Fixes #1219 

I added extra test case. Another issue is still not fixed, described in the following spec:

```ruby
    it 'includes level 2 nested parameter within #declared(params)' do
      subject.params do
        requires :bar, type: Hash do
          optional :a

          optional :c, type: Hash do
            given :a do
              requires :b
            end
          end
        end
      end
      subject.get('/nested') { declared(params).to_json }
      get '/nested', bar: { a: true, c: { b: 'yes' } }
      expect(JSON.parse(last_response.body)).to eq('bar' => { 'a' => "true", 'b' => 'yes' })
    end
```

For this the validation of the given DSL should be moved somewhere else. Right now `given` only works for parameters defined in the same block. I highly doubt that somebody would need this level of complexity for defining depending params.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 =============
 
 * [#1216](https://github.com/ruby-grape/grape/pull/1142): Fix JSON error response when calling `error!` with non-Strings - [@jrforrest](https://github.com/jrforrest).
+* [#1225](https://github.com/ruby-grape/grape/pull/1225): Fix `given` with nested params not returning correct declared params - [@JanStevens](https://github.com/JanStevens).
 * Your contribution here.
 
 0.14.0 (12/07/2015)
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -90,7 +90,11 @@ module Grape
       # Adds a parameter declaration to our list of validations.
       # @param attrs [Array] (see Grape::DSL::Parameters#requires)
       def push_declared_params(attrs)
-        @declared_params.concat attrs
+        if lateral?
+          @parent.push_declared_params(attrs)
+        else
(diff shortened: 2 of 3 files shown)


dblock:
This looks good, can you squash the two commits please? Open a separate bug for the remaining case.

JanStevens:
Posted as: https://github.com/ruby-grape/grape/issues/1226

Will look into that later on

dblock:
:+1:

PR #1234, opened 2016-01-02
Use Ruby 2.3.0 on Travis

namusyaka:
Ruby 2.3 is now available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: ruby
 sudo: false
 
 rvm:
+  - 2.3.0
   - 2.2
   - 2.1
   - 2.0.0
--- a/gemfiles/rails_3.gemfile
+++ b/gemfiles/rails_3.gemfile
@@ -2,7 +2,7 @@
 
 source 'https://rubygems.org'
 
-gem 'rails', '3.2.19'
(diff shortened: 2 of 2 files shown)


namusyaka:
The errors can be resolved by [this commit](https://github.com/rails/rails/commit/0599bfa2bc49c885fd4ec3ab7ecce37dc5c03ce5)

nonmadden:
@namusyaka travis ci not complete

namusyaka:
The errors has been occured by stopping to rescue exceptions of `<=>`.
So I fixed a problem by [modifing a code in cookiejar](https://github.com/dwaite/cookiejar/pull/20), but the gem has not been maintained for quite a while. 

@nonmadden Really? That seems to be completed, but failed.

nonmadden:
not clearly you say completed but failed ???

dblock:
@namusyaka Why is it only failing on Rails 3? If there's a reasonably explanation, then maybe you can update the spec to check for that `nil` or `expired?` for example?

namusyaka:
@dblock I noticed I overlooked rails-3.2.22 release.
So I tried to update the dependency and tests seem to be working well. How about this one?

dblock:
Merged, thanks!

PR #1236, opened 2016-01-05
Don't use URI.escape as it has been obsoluted

namusyaka:
I guess we can use `CGI.escape` instead of `URI.escape`.
Ref: https://github.com/ruby/ruby/commit/238b979f1789f95262a267d8df6239806f2859cc


dblock:
:+1:

Issue #1231, opened 2015-12-21
How to validate dependent parameters 


brahamshakti:
In you documentation you said that dependent parameter can be validated using

    params do
        optional :shelf_id, type: Integer
        given :shelf_id do
            requires :bin_id, type: Integer
        end
    end

but how can i perform validation on basis of two fields like

    params do
        optional :first_name, type: String
        optional :last_name, type: String
        given :first_name, :last_name do
            requires :ssn, type: String
        end
    end

Is it possible to do that I tried it but swagger is not validating ssn.

dblock:
A nested `given` _might_ work, but I think this is a feature request, it seems fairly straightforward. Want to try to implement it?

ochagata:
This can be closed

dblock:
Closed via #1237.

PR #1237, opened 2016-01-05
Changed the input of `given`

ochagata:
It will now expect any amount of inputs and will iterate on the input

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 * [#1227](https://github.com/ruby-grape/grape/pull/1227): Store `message_key` on Grape::Exceptions::Validation - [@stjhimy](https://github.com/sthimy).
 * [#1232](https://github.com/ruby-grape/grape/pull/1232): Helpers are now available inside `rescue_from` - [@namusyaka](https://github.com/namusyaka).
+* [#1237](https://github.com/ruby-grape/grape/pull/1237): Allow multiple parameters in `given`, which behaves as if the scopes were nested in the inputted order - [@ochagata](https://github.com/ochagata).
 * Your contribution here.
 
 #### Fixes
--- a/lib/grape/dsl/parameters.rb
+++ b/lib/grape/dsl/parameters.rb
@@ -165,9 +165,11 @@ module Grape
       # @raise Grape::Exceptions::UnknownParameter if `attr` has not been
       #   defined in this scope yet
       # @yield a parameter definition DSL
-      def given(attr, &block)
-        fail Grape::Exceptions::UnknownParameter.new(attr) unless declared_param?(attr)
-        new_lateral_scope(dependent_on: attr, &block)
+      def given(*attrs, &block)
+        attrs.each do |attr|
+          fail Grape::Exceptions::UnknownParameter.new(attr) unless declared_param?(attr)
+        end
+        new_lateral_scope(dependent_on: attrs, &block)
       end
 
       # Test for whether a certain parameter has been defined in this params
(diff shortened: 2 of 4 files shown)


ochagata:
In response to #1231

lib/grape/dsl/parameters.rb
```diff
@@ -165,9 +165,11 @@ def all_or_none_of(*attrs)
       # @raise Grape::Exceptions::UnknownParameter if `attr` has not been
       #   defined in this scope yet
       # @yield a parameter definition DSL
-      def given(attr, &block)
-        fail Grape::Exceptions::UnknownParameter.new(attr) unless declared_param?(attr)
-        new_lateral_scope(dependent_on: attr, &block)
+      def given(*attrs, &block)
+        attrs.each do |attr|
+          fail Grape::Exceptions::UnknownParameter.new(attr) unless declared_param?(attr)
+          new_lateral_scope(dependent_on: attr, &block)
```
rnubel: Hey Ozzy! This is a good start. But I think this is going to open up one scope per each attribute listed and repeat the nested definitions each time. I haven't tested it, so this is a guess, but I believe the behavior it will actually cause is more like if you'd written:

```
given :a do
  requires :c
end
given :b do
  requires :c
end
```

Instead, I think you either want to update how `dependent_on` is parsed to allow *it* to be an array, or create these scopes such that they're nested. That is, emulate:

```
given :a do
  given :b do
    requires :c
  end
end
```

ochagata: @rnubel 
I made a change to fix this. Now the behavior is this:
`given :a, :b` will now only do the validation if both `:a` and `:b` are both given, which would mimic the nested structure as well

rnubel: Yep, looks good to me now.

dblock:
This looks great. You might want to split up the specs a bit, one for a single parameter, one for multiple. Please squash the commits and update CHANGELOG.

ochagata:
@dblock should be good to go once travis finishes

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 * [#1227](https://github.com/ruby-grape/grape/pull/1227): Store `message_key` on Grape::Exceptions::Validation - [@stjhimy](https://github.com/sthimy).
 * [#1232](https://github.com/ruby-grape/grape/pull/1232): Helpers are now available inside `rescue_from` - [@namusyaka](https://github.com/namusyaka).
+* [#1237](https://github.com/ruby-grape/grape/pull/1237): Given now takes multiple inputs and behaves as though the scopes were nested in the inputted order - [@ochagata](https://github.com/ochagata)
```
dblock: Given is a keyword, so maybe "Allow multiple parameters in `given`, which behaves as if the scopes were nested in the inputted order".
 
Missing a trailing period, please.

lib/grape/validations/params_scope.rb
```diff
@@ -36,7 +36,11 @@ def initialize(opts, &block)
       #   validated
       def should_validate?(parameters)
         return false if @optional && params(parameters).respond_to?(:all?) && params(parameters).all?(&:blank?)
-        return false if @dependent_on && params(parameters).try(:[], @dependent_on).blank?
+        if @dependent_on
+          @dependent_on.each do |dependency|
+            return false if params(parameters).try(:[], dependency).blank?
+          end
```
dblock: You can avoid the nesting, move `if @dependent_on` after the `end`?

dblock:
I just have two small nitpicks on CHANGELOG and stuff ... sorry to be a pain.

ochagata:
All notes addressed and changes made

dblock:
Merged, :+1:

PR #1240, opened 2016-01-07
Call #after of middleware on error

### LINKED ISSUE #1147: Call After block on error!
Not sure if this was resolved, but I've been trying to log api calls from the middleware and on an error! the after call doesn't get executed. Is there a good way around this issue? Thanks!

namusyaka:
This might be able to close #1147.
Could you review this PR?

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,25 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 0.15.0
+
+#### Changes to behavior of `after` method of middleware on error
+
+The `after` method of the middleware is now called on error.
+The following code would work correctly.
+
+```ruby
+class ErrorMiddleware < Grape::Middleware::Base
+  def after
+    if @api_response[0] == 500
+      env['rack.logger'].debug("Raised error on #{env['PATH_INFO']}")
+    end
+  end
+end
+```
+
+See [#1147](https://github.com/ruby-grape/grape/issues/1147) and [#1240](https://github.com/ruby-grape/grape/issues/1240) for discussion of the issues.
+
 ### Upgrading to >= 0.14.0
 
 #### Changes to availability of DSL methods in filters
--- a/lib/grape/middleware/base.rb
+++ b/lib/grape/middleware/base.rb
@@ -22,8 +22,12 @@ module Grape
       def call!(env)
         @env = env
         before
-        @app_response = @app.call(@env)
-        after || @app_response
+        begin
+          @app_response = @app.call(@env)
+        ensure
+          after_response = after
+        end
+        after_response || @app_response
       end
 
       # @abstract
--- a/lib/grape/middleware/formatter.rb
+++ b/lib/grape/middleware/formatter.rb
@@ -19,6 +19,7 @@ module Grape
       end
 
       def after
+        return unless @app_response
         status, headers, bodies = *@app_response
 
         if Rack::Utils::STATUS_WITH_NO_ENTITY_BODY.include?(status)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [#1227](https://github.com/ruby-grape/grape/pull/1227): Store `message_key` on Grape::Exceptions::Validation - [@stjhimy](https://github.com/sthimy).
 * [#1232](https://github.com/ruby-grape/grape/pull/1232): Helpers are now available inside `rescue_from` - [@namusyaka](https://github.com/namusyaka).
 * [#1237](https://github.com/ruby-grape/grape/pull/1237): Allow multiple parameters in `given`, which behaves as if the scopes were nested in the inputted order - [@ochagata](https://github.com/ochagata).
+* [#1238](https://github.com/ruby-grape/grape/pull/1238): Call `after` of middleware on error - [@namusyaka](https://github.com/namusyaka).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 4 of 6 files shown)


lib/grape/middleware/base.rb
```diff
@@ -21,9 +21,17 @@ def call(env)
...
+        begin
+          @app_response = @app.call(@env)
+        rescue => e
+          error_class = e
+        ensure
+          after_response = after
+        end
+
+        after_response || (error_class ? (fail error_class) : @app_response)
       end
```
dblock: Either all the paths aren't covered by specs or this can be:

```ruby
begin
ensure
  after || @app_response
end
```
Causes the after middleware to execute on error, and raises the exception as before. 

What am I missing?

lib/grape/middleware/formatter.rb
```diff
@@ -19,6 +19,7 @@ def before
       end
 
       def after
+        return unless @app_response
```
dblock: I think the semantics of `after` is to always execute, with or without `@app_response`.

lib/grape/middleware/base.rb
namusyaka: Well, If `after` method returns rack response format on error, the response should be used forcibly, I think.
If not reasonable, I'm going to fix it.

lib/grape/middleware/formatter.rb
namusyaka: Raised an exception if your suggestion is reflect.
Should the formatter's after method fully be executed in this case?

lib/grape/middleware/base.rb
namusyaka: Also, the `ensure` section isn't used as retured value. Thoughts?

dblock: I think we shouldn't be expecting the `after` block to return a response. Otherwise we need to change the behavior of `before` as well and this will definitely have impact on API implementations out there. These are callbacks, so I think they should be called, but their results discarded, at least as the next increment.

lib/grape/middleware/formatter.rb
dblock: Oh I see, this is right, leave it here.

lib/grape/middleware/base.rb
namusyaka: Makes sense.
So should I remove Response from [this line](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/base.rb#L36)?

dblock: Hmmm... wait. Looks like the old behavior *was* to use the response. So let me take this back, we should preserve the response if it was there. I guess it should be closer to this:

```ruby
begin
ensure
  @app_response = after || @app_response
end
```

namusyaka: Okay, thanks.

namusyaka:
I pushed new changes into this branch.
Currently, we raise the exception even if an exception is raised on `@app.call`.

If there is no problem, I'm going to squash a few commits and force push.

dblock:
This needs a spec for the behavior of `after` returning a Rack response and overwriting the app response, so we don't break this in the future. Otherwise it's good to go.

namusyaka:
Added.

dblock:
Forgot one more thing. The behavior is different from previous versions (even if it was undocumented), so this needs an entry in UPGRADING, please. Squash everything and I'll merge.

dblock:
Thanks for hanging in with me!

namusyaka:
Yeah, squashed and force pushed.
Thank you, too!

UPGRADING.md
```diff
@@ -3,6 +3,30 @@ Upgrading Grape
 
 ### Upgrading to >= 0.14.0
 
+#### Changes to behavior of `after` method of middleware on error
```
dblock: This should be under >= 0.15.0 and increment the version in version.rb to 0.15.0 as well, now that we have a major change.

UPGRADING.md
```diff
@@ -3,6 +3,30 @@ Upgrading Grape
...
+  end
+end
+
+class API < Grape::API
+  use CustomMiddleware
+
+  get { raise StandardError }
+end
+
+mock = Rack::MockRequest.env_for(?/)
```
dblock: This is very spec-y, remove. Rewrite the example to be more about a real world API, and maybe return a proper rack response in `after`. In this example it's unclear what happen.

On this note we also need a README update about the behavior of `after`. Shoot, I keep forgetting things.... lmk when you had enough of my comments and I'll do the rest ;)

namusyaka: Rewriting is ok. However, the rack response thing should be separated with the error thing because the rack response in `after` is not used in error case.

namusyaka:
Force pushed. Also I added a little Grape middleware section.

dblock:
Merged. Thanks!

namusyaka:
Thank you for your polite response.

rngtng:
continue discussion from #1255: this change breaks https://github.com/aserafin/grape_logging (and all other middlewares which assumes `@app_response` isn't `nil`) on application error. 

While updates on the middleware are pending, I'd propose to  at least rescue any `after` call errors silently:

```ruby
begin
ensure
    after_response = after rescue nil
end
```

... or would setting `@app_response` to a default value a better solution?

dblock:
@rngtng Could you open a new issue please? I really don't like hiding errors like this. Should we just take it all in and make grape_logging more resilient?

rngtng:
jepp, updating grape_logging is the right way to go, I only guess there are more middleware out there which probably break.  Instead of hiding, we should communicate the error at least. (a warning?) - I'll open a issue for that...

Issue #986, opened 2015-04-15
Access request in custom validators.


nalbeza:
Is it possible to access request context in custom validators ?

It would be very useful, for example to check for uniqueness or use helpers such as current_user.

dblock:
It certainly is. See https://github.com/intridea/grape/blob/master/lib/grape/validations/params_scope.rb#L8 for how it plucks various things at construction time - I think this is what you're looking for. 

Use the mailing list for questions, if the above doesn't work post some code of what you'd like to accomplish.

suan:
@dblock Would you mind providing a more specific example of how we can get at the request from within a validator? I see that I can access the `@scope` variable from within my validator, but I don't see the params_scope containing any info about the actual request

(In my case, I'd like to get at the request path and headers)

dblock:
Right @suan, I reopened it because I don't actually see how. I'd welcome someone solving it!

glennpratt:
@dblock locally I've changed the signature of Validator#validate! (with backwards compatibility using arity) to:

```ruby
def validate!(params, env)
```

Is env too broad?

glennpratt:
Oh, I see `@request` includes `env` anyway. I'll just do that instead.

glennpratt:
So, should we replace params with request (breaking backwards compatibility) or only send request if the method arity != 1?

glennpratt:
See #1242

dnesteryuk:
@dblock The attached PR already [implemented this task](https://github.com/ruby-grape/grape/blob/master/spec/grape/api/custom_validations_spec.rb#L52-L59) :wink:

dblock:
Closing! Thanks @dnesteryuk

PR #1242, opened 2016-01-08
Pass request to validators along with params.

glennpratt:
Addresses #986


dblock:
I think we want to build `request` into the base class of `Validator` and not change the method signature. Something like [in formatter](https://github.com/ruby-grape/grape/blob/925a3fbaa63709f72ab4d6060c98ede5cdb4dd45/lib/grape/middleware/formatter.rb#L67).

glennpratt:
@dblock the validator is per Endpoint, not per request, so making it a method of the Validator class seems odd to me.

dblock:
I see your point, but I politely disagree :) The API is all singletons, and there're methods all over the place such as `request`, so my suggestion just keeps things consistent and easier to use. Furthermore, the fact that we need to check for `arity` and what not is definitely telling us something.

glennpratt:
Either we change the method args, the constructor args or add a setter. All the same to me.

Is there no need for thread safety here? I wrote a test to check thread safety with `@request` in the Validator class and it immediately failed. It's quite possible I don't understand how this is supposed to work.

dblock:
I think you shouldn't memoize anything _or_ it should be thread-safe. I vote for the former to start.

glennpratt:
Huh, I'm not sure how to do this without attaching something to the instance - memoizing or whatever.

This is one way I got it to work that passed my thread safe test - duplicating the validator objects:

```ruby
        route_setting(:saved_validations).map(&:dup).each do |validator|
          req = @request
          validator.define_singleton_method(:request) { req }
          begin
            validator.validate!(params)
          rescue Grape::Exceptions::Validation => e
            validation_errors << e
          end
        end
```

`define_singleton_method` could just as well be a setter - this was just quicker.

It's not clear to me how `@request` and other instance vars in Endpoint instances are thread safe - perhaps some magic with the binding (`generate_api_method`?). It passes my threading tests, so it seems they are. Whatever magic that is doesn't happen when calling the Validator.

dblock:
Not sure why/what/etc., I am speaking without looking at the code in depth, so I'll take a look soon. 

I think all I would want here is to be able to have `request` inside the validator context without having to change any function interfaces. Hope that makes sense?

glennpratt:
@dblock updated.

lib/grape/validations/validators/coerce.rb
```diff
@@ -5,6 +5,11 @@ class API
 
   module Validations
     class CoerceValidator < Base
+      def initialize(*_args)
+        super
+        @converter = Types.build_coercer(type, @option[:method])
```
glennpratt: This change is to make sure `@converter` is only created once, rather than in each validator dup.

`coerce_spec.rb` is expecting this behavior.

dblock:
Calling `dup` so many times makes me nervous. @dm1try I could use another opinion on this thread, please

Is https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/multiple_params_base.rb#L7 saying that we shouldn't worry about thread-safety in any of these and therefore don't need a `dup` but can just make `@request` appear in there by either implementing another `validate` (like you did) or possibly deprecating the signature like [here](https://github.com/dblock/grape/commit/be935ebc07d8adca8ad798675565beb731bcc2e2)?

glennpratt:
If I use threads to make requests to the same app, that `@request` instance variable in dblock@7815ad0 won't be thread safe - `@request` can be overwritten in another thread.

I haven't tested it in a common threaded app server, but I could if that would help.

dm1try:
>It's not clear to me how @request and other instance vars in Endpoint instances are thread safe - perhaps some magic with the binding (generate_api_method?).

No magic. Just `dup` on `call`: [1](https://github.com/ruby-grape/grape/blob/master/lib/grape/endpoint.rb#L209), [2](https://github.com/ruby-grape/grape/blob/master/lib/grape/endpoint.rb#L314), [3](https://github.com/ruby-grape/grape/blob/master/lib/grape/endpoint.rb#L230).

@glennpratt has the good point: all validators for a specific endpoint share the same state; they are [saved in `route_setting` on the initialization]((https://github.com/ruby-grape/grape/blob/master/lib/grape/endpoint.rb#L80)) and they are not cloned on `call/run`. 
>Is https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/multiple_params_base.rb#L7 saying that we shouldn't worry about thread-safety in any of these and therefore

This line seems unsafe (multiple requests for the same endpoint with shared state :worried:)

In short, don't provide any "per-request" state to validators because the state is shared across all endpoint instances(add to README) or copy them. 

I'll have a deeper look at this later.

glennpratt:
> No magic. Just dup on call: 1, 2, 3.

@dm1try doh, thank you. When I was poking at it I could have sworn multiple requests were in the same Endpoint object, but I must have been wrong.

I'm thinking `Validator#dup` probably isn't that expensive and probably prevents future mistakes. Would probably require benchmarks to say.

If `#dup` doesn't work, then we could use `#validate(request)` as the new method with `#validate!(params)` for backwards compatibility with NO access to the request.

dblock:
Ok so we're saying that https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/multiple_params_base.rb#L7 is a bug.

@dm1try Should we merge this PR as is giving validators a `dup`? I would want to simplify the interface and not pass both `request` and `params` and do what @glennpratt suggests above for backwards compat.

dm1try:
>@dm1try Should we merge this PR as is giving validators a dup? 

As mentioned before, we might do some benchmarks.
I agree on the extending a scope of a validation object(`params` => `request`) without providing any additional state to validators and do not `dup` for now. It will be work as before.

>I would want to simplify the interface and not pass both request and params and do what @glennpratt suggests above for backwards compat.

Sounds good to me!

glennpratt:
Super Scientific Benchmarks

```
#validate(request)
                        20.840k i/100ms
#validate_no_attr(request)
                        20.409k i/100ms
#validate_no_dup(request)
                        22.787k i/100ms
#validate!(params)
                        23.616k i/100ms
-------------------------------------------------
#validate(request)
                        245.524k (± 7.2%) i/s -      2.459M
#validate_no_attr(request)
                        245.349k (± 7.4%) i/s -      2.449M
#validate_no_dup(request)
                        293.073k (± 7.9%) i/s -      2.917M
#validate!(params)
                        318.294k (± 7.1%) i/s -      3.165M
```

dblock:
Looks like `dup` isn't that big of a deal, but we still don't have to  if we just do https://github.com/dblock/grape/commit/7815ad0515a2018a5913d23e42dad512e5546858 (minus the `@request` assignment).

To summarize:

* Add `validate` that takes a `request` instead of `params` (not both)
* Call `validate(request.params)` by default

This will let you override `validate` in a custom validator.

We still need to fix the existing problem in `MultipleParams::Base`.

dblock:
@glennpratt Want to give above a try? I am interested in getting this merged, I think we're close.

glennpratt:
@dblock sure, I have a moment now.

glennpratt:
@dblock The cleanest, backwards compatible fix for `#scoped_params` I see is:

```ruby
      def validate(request)
        dup.validate(request)
      end
```

Assuming external users could be depending on the `#scoped_params` method.

glennpratt:
@dblock changes pushed. Do you want me to push that fix for `MultipleParams::Base ` now or add a todo?

dblock:
Ok, this change looks good, merging. Make a separate PR for the `MultipleParams` fix, please, and much much appreciated!

magni-:
#986 mentioned being able to "use helpers such as `current_user`."

How does the changes from this PR let us do that?

dblock:
This PR passes `request` along with params, which is a step closer, but I think more work is needed for #986

PR #1243, opened 2016-01-13
Support more kindly for after method

namusyaka:
ref #1240

The current implementation is going to try to return `after_response` if `after_response` is not false.
I guess the behavior is not expected, and users will be confusing.
Thoughts?

### DIFF
--- a/lib/grape/middleware/base.rb
+++ b/lib/grape/middleware/base.rb
@@ -27,7 +27,7 @@ module Grape
         ensure
           after_response = after
         end
-        after_response || @app_response
+        response_valid?(after_response) ? after_response : @app_response
       end
 
       # @abstract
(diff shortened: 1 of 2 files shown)


dblock:
I think we would be trying to do too much for the user here. There should be a contract since this is an API. Right now it says you need to return a Rack compatible response from `after`, and I think we shouldn't be having code at runtime trying to "help" API authors work around that. 

I vote to close this without merging, lets leave it here for a bit see if others think strongly otherwise.

namusyaka:
Now, we must avoid returning value except false or nil on filters. I guess the constraint is very confusing. Thoughts?

dblock:
So if in `after` you return anything, that's what's returned from the API. I think that's OK, otherwise it feels too magical. Curious what others think.

namusyaka:
@dblock Yeah I see. I came to conclusion that your opinion is reasonable, but I'd like to hear other opinions.

PR #1244, opened 2016-01-14
Add header support for middleware

namusyaka:
ref #1069 
This adds the `header` method for activating it in our middleware.
Could you review the pull request?


lib/grape/dsl/inside_route.rb
```diff
@@ -138,16 +138,20 @@ def status(status = nil)
...
       end
 
-      # Set an individual header or retrieve
-      # all headers that have been set.
-      def header(key = nil, val = nil)
-        if key
-          val ? @header[key.to_s] = val : @header.delete(key.to_s)
-        else
-          @header
+      module HeaderMethods
```
dblock: Should be in a separate file probably. And I don't love the name "HeaderMethods", maybe just "Headers"?

lib/grape/middleware/base.rb
```diff
@@ -27,7 +29,21 @@ def call!(env)
...
         end
-        after_response || @app_response
+
+        response = after_response || @app_response
+        reflect_headers response
+        response
+      end
+
+      def reflect_headers(response)
+        return unless header.is_a?(Hash)
```
dblock: What else can `header` be? I think a check for `nil` is more appropriate, so `unless header`. Also alias `header` to `headers`, when we check on a collection of headers we should be doing `headers`.

lib/grape/middleware/base.rb
```diff
@@ -27,7 +29,21 @@ def call!(env)
...
+        response_header =
+          case response
+          when Rack::Response then response.headers
+          when Array          then response[1]
+          end
+        response_header.merge!(header)
+        header.clear
```
dblock: Why is this necessary?

lib/grape/dsl/inside_route.rb
namusyaka: Agreed. I'll fix it.

lib/grape/middleware/base.rb
namusyaka: Nope, there is a possibility of inserting `Rack::Accept::MediaType` into `@headers`.
See https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/versioner/header.rb#L69

However, aliasing seems reasonable to me. So I'd like to fix it.

lib/grape/middleware/base.rb
namusyaka: Right. My Bad. Thanks!

namusyaka:
Fixed and force pushed.

lib/grape/middleware/base.rb
```diff
@@ -27,7 +31,20 @@ def call!(env)
...
           after_response = after
         end
-        after_response || @app_response
+
+        response = after_response || @app_response
+        reflect_headers response
+        response
+      end
+
+      def reflect_headers(response)
```
dblock: Should this be private?

Maybe rename it to `merge_headers`?

lib/grape/middleware/base.rb
```diff
@@ -27,7 +31,20 @@ def call!(env)
...
+        return unless headers.is_a?(Hash)
+        response_headers =
+          case response
+          when Rack::Response then response.headers
+          when Array          then response[1]
+          end
+        response_headers.merge!(headers)
```
dblock: If the API developer made a mistake and `response` is neither, we get a `nil`, I would stay defensive and do `merge!() ... if response_headers`.

lib/grape/middleware/base.rb
```diff
@@ -27,7 +31,20 @@ def call!(env)
...
-        after_response || @app_response
+
+        response = after_response || @app_response
+        reflect_headers response
+        response
+      end
+
+      def reflect_headers(response)
+        return unless headers.is_a?(Hash)
+        response_headers =
```
dblock: I am nitpicking but I would avoid a temporary `response_headers` and just do `when .... .merge!(...)`.

dblock:
Sorry more minor comments, please.

namusyaka:
Thanks for polite reviewing! Pushed.

dblock:
Merged, thanks!

PR #1252, opened 2016-01-22
do not raise an IncompatibleOptionValues For a couple of potentially valid use cases

jeradphelps:
I believe the following two params blocks should be valid:

~~~
# The default is the entire range of allowed values
params { optional :numbers, type: Array[Integer], values: 0..2, default: 0..2 }
~~~

~~~
# The default is a subset of the entire range of allowed values
subject.params { optional :numbers, type: Array[Integer], values: [0, 1, 2], default: [1, 0] }
~~~

The current implementation doesn't account for these use cases and raises `Grape::Exceptions::IncompatibleOptionValues`.  This PR will allow the two cases above.


dblock:
Looks good! :+1:

PR #1254, opened 2016-01-22
Fix bug at array index context in errors 

totothink:
```ruby 
subject.params do
    group :children, type: Array do
       requires :name
       group :parents, type: Array do
          requires :name
       end
    end
end
```

if params is:
```ruby
children: [
          { name: 'Job', parents: [{ name: 'Joy' }, {name: 'Bob'}] },
          { name: 'Jim', parents: [{}] }
]
```

expect raise 'children[1][parents][0][name] is missing'，got: "children[0][parents][0][name] is missing"

### DIFF
--- a/lib/grape/dsl/parameters.rb
+++ b/lib/grape/dsl/parameters.rb
@@ -184,9 +184,28 @@ module Grape
       alias_method :group, :requires
 
       # @param params [Hash] initial hash of parameters
-      # @return hash of parameters relevant for the current scope
-      # @api private
+      # @return hash of parameters with index information relevant for the current scope
+      # example:
+      # params: [{"name"=>"Job", "parents"=>[{"name"=>"Joy"}, {"name"=>"Bob"}], "_param_index": {"children_index": 0}}, {"name"=>"Jim", "parents"=>[{}], "_param_index": {"children_index": 1}}]
+      # element: parents
+      # result: [{"name"=>"Joy", "_param_index": {children[parents]: 0, "children": 0}}, {"name"=>"Bob", "_param_index": {children[parents]: 1, "children": 0}}, { "_param_index": {children[parents]: 0, "children": 1}}]
       def params(params)
+        params = @parent.params(params) if @parent
+        if @element
+          if params.is_a?(Array)
+            params = params.map.with_index do |el, index|
+              el[@element] ? convert_to_param_index(el[@element], el['_param_index'], index) : {}
+            end.flatten
+          elsif params.is_a?(Hash)
+            params = params[@element] ? convert_to_param_index(params[@element], params['_param_index']) : {}
+          else
+            params = {}
+          end
+        end
+        params
+      end
+
+      def params_without_index(params)
         params = @parent.params(params) if @parent
         if @element
           if params.is_a?(Array)
@@ -198,6 +217,17 @@ module Grape
           end
         end
         params
+      end
+
+      def convert_to_param_index(item, param_index, index = nil)
+        param_index ||= {}
+        param_index[@parent.full_name(@element)] = index if index && @parent
+        if item.is_a?(Array)
+          item.map.with_index{|e, i| e.merge!('_param_index' => param_index.merge(@parent.full_name(@element) =>i)) if e.respond_to?(:merge!) }
+        else
+          item.merge!('_param_index' => param_index) if item.respond_to?(:merge!)
+        end
+        item
       end
     end
   end
--- a/lib/grape/dsl/parameters.rb
+++ b/lib/grape/dsl/parameters.rb
@@ -183,21 +183,13 @@ module Grape
 
       alias_method :group, :requires
 
-      # @param params [Hash] initial hash of parameters
-      # @return hash of parameters with index information relevant for the current scope
-      # example:
-      # params: [{"name"=>"Job", "parents"=>[{"name"=>"Joy"}, {"name"=>"Bob"}], "_param_index": {"children_index": 0}}, {"name"=>"Jim", "parents"=>[{}], "_param_index": {"children_index": 1}}]
-      # element: parents
-      # result: [{"name"=>"Joy", "_param_index": {children[parents]: 0, "children": 0}}, {"name"=>"Bob", "_param_index": {children[parents]: 1, "children": 0}}, { "_param_index": {children[parents]: 0, "children": 1}}]
       def params(params)
         params = @parent.params(params) if @parent
         if @element
           if params.is_a?(Array)
-            params = params.map.with_index do |el, index|
-              el[@element] ? convert_to_param_index(el[@element], el['_param_index'], index) : {}
-            end.flatten
+            params = params.flat_map { |el| el[@element] || {} }
           elsif params.is_a?(Hash)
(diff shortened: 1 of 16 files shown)


lib/grape/dsl/parameters.rb
```diff
@@ -199,6 +196,42 @@ def params(params)
...
+      # element: parents
+      # result: [{"name"=>"Joy", "_param_index": {children[parents]: 0, "children": 0}},
+      #   {"name"=>"Bob", "_param_index": {children[parents]: 1, "children": 0}},
+      #   { "_param_index": {children[parents]: 0, "children": 1}}]
+      def params_with_index(params)
+        params = @parent.params_with_index(params) if @parent
+        if @element
+          if params.is_a?(Array)
+            params = params.map.with_index do |el, index|
+              el[@element] ? convert_to_param_index(el[@element], el['_param_index'], index) : {}
```
dblock: Should `'_param_index'` be `:param_index`?

lib/grape/dsl/parameters.rb
```diff
@@ -199,6 +196,42 @@ def params(params)
...
+      # example:
+      # params: [{"name"=>"Job", "parents"=>[{"name"=>"Joy"}, {"name"=>"Bob"}], "_param_index": {"children_index": 0}},
+      #   {"name"=>"Jim", "parents"=>[{}], "_param_index": {"children_index": 1}}]
+      # element: parents
+      # result: [{"name"=>"Joy", "_param_index": {children[parents]: 0, "children": 0}},
+      #   {"name"=>"Bob", "_param_index": {children[parents]: 1, "children": 0}},
+      #   { "_param_index": {children[parents]: 0, "children": 1}}]
+      def params_with_index(params)
+        params = @parent.params_with_index(params) if @parent
+        if @element
```
dblock: Minor: `return params unless @element` or move the `if` to the `end if @element`, less nesting.

lib/grape/dsl/parameters.rb
```diff
@@ -199,6 +196,42 @@ def params(params)
...
+          if params.is_a?(Array)
+            params = params.map.with_index do |el, index|
+              el[@element] ? convert_to_param_index(el[@element], el['_param_index'], index) : {}
+            end.flatten
+          elsif params.is_a?(Hash)
+            params = params[@element] ? convert_to_param_index(params[@element], params['_param_index']) : {}
+          else
+            params = {}
```
dblock: When does this happen?

lib/grape/validations/validate_service.rb
```diff
@@ -0,0 +1,51 @@
+module Grape
+  module Validations
+    class ValidateService
```
dblock: I think service is very ambitious :) Maybe `ValidateService` should be just `Validator`, what do you think?

lib/grape/validations/validate_service.rb
```diff
@@ -0,0 +1,51 @@
...
+      def initialize(validators, params)
+        @validators = validators
+        @params = params
+      end
+
+      # api
+      def validate!
```
dblock: This method implies that an exception is going to be raised (it has a bang), but instead it returns a boolean. So it should either be renamed to `validate` or it should `fail ...`.

spec/grape/validations_spec.rb
```diff
@@ -647,6 +647,22 @@ def validate_param!(attr_name, params)
         expect(last_response.body).to eq('children[0][parents][0][name] is missing')
       end
 
+      it 'should prompt index in array when a parameter is not present' do
```
dblock: should prompt => prompts, been trying to weed these out

dblock:
This is great, excellent job. 

See my comments above, the most important ones are missing tests for what's now `ValidatorService` and a bang method confusion. Also please update CHANGELOG and squash commits.

lib/grape/dsl/parameters.rb
totothink: 'params_with_index' purpose is to keep index information of array params. Because it is implemented by means of the new field in hash, new field should avoid conflict with parameter name. The prefix "_" indicates internal use.

lib/grape/dsl/parameters.rb
totothink: yes,it's right

lib/grape/dsl/parameters.rb
totothink: params method return hash of parameters relevant for the current scope. Because index information of array will not be retained, when the validation fails, error information is not complete. params_with_index will retain index information of array, when the validation fails to get the complete message by full_message method.

example：	
I defined the following parameters：
```ruby
subject.params do
    group :children, type: Array do
       requires :name
       group :parents, type: Array do
          requires :name
       end
    end
end
```
if param is:
```ruby
children: [
          { name: 'Job', parents: [{ name: 'Joy' }, {name: 'Bob'}] },
          { name: 'Jim', parents: [{}] }
]
```
expect raise 'children[1][parents][0][name] is missing'，got: "children[0][parents][0][name] is missing"

spec/grape/validations_spec.rb
totothink: ok

lib/grape/validations/validate_service.rb
totothink: Yes, you're right.

lib/grape/validations/validate_service.rb
totothink: Yes, ValidationService named inappropriate. But it is not Validator. It does not assume responsibility, specific validation performed by specific Validator. ValidateService class purpose is move the relevant code of verify from Endpoint to Validations module. 

Another idea is to assume that responsibility by the Validations module object.

example：
1. lib/grape/validations.rb
```ruby
module Grape
  # Registry to store and locate known Validators.
  module Validations
    class << self
      attr_accessor :validators

      def validate!(endpoint, validators, params)
        errors = []
        validators.each do |validator|
          begin
            validator.validate!(params)
          rescue Grape::Exceptions::Validation => e
            errors << e
          end
        end
        fail Grape::Exceptions::ValidationErrors, errors: errors, headers: endpoint.header if errors.any?

        clean_params_index(params)
      end

      def clean_params_index(params)
        case params
        when Array
          params = params.map do |item|
            clean_params_index(item)
          end
        when Hash
          params.delete("_param_index")
          params.each do |key, value|
            params[key] = clean_params_index(value)
          end
        end
        params
      end
    end

    self.validators = {}

    # Register a new validator, so it can be used to validate parameters.
    # @param short_name [String] all lower-case, no spaces
    # @param klass [Class] the validator class. Should inherit from
    #   Validations::Base.
    def self.register_validator(short_name, klass)
      validators[short_name] = klass
    end

  end
end
```

2. lib/grape/endpoint.rb
```ruby
module Grape
  class Endpoint
    ...

    def run_validators(validators, params)
      @params = Validations.validate!(self, validators, params)
    end

    ...
  end
end
```

totothink:
Thank you for your comment and I will correct your errors pointed out. My have poor English, is working to improve，please bear with me.

dblock:
@totothink You're doing awesome, your English is perfectly fine!

dblock:
@totothink Did you mean to close this?

Issue #1009, opened 2015-05-05
Allow param definitions in route_param


ctembreull:
It seems like it would make sense to at least allow the 'type' modifier in a route param call. Instead of doing:

    route_param :foo do
        params do
            requires :foo, type: String
        end
        #...
    end

it would make a lot more sense to simply say:

    route_param :foo, type: String do
        #...
    end

dblock:
:+1:

kuraga:
No ambigiouty or conflicts?

ctembreull:
I wouldn't believe so, but I haven't started looking at what it would take to do this, either. It should function the same as `required :param_name, type: Type, {...}` in other params definitions, but it would have the extra power to declare a new route section as it does now. `route_param` already requires that parameter to be in place to access any of its nested routes.

namusyaka:
@dblock This issue can be closed by #1255

dblock:
:+1:

PR #1255, opened 2016-01-25
Allow param type definition in route_param

namusyaka:
ref #1009 
I'd like to use type definition in `route_param`, so I implemented it by using `requires` DSL.


dblock:
Very nice thanks.

rngtng:
Great feature, unfortunately this breaks with `grape_logging` middleware, due to `nil` value for `Response` object which is required [here]( https://github.com/aserafin/grape_logging/blob/master/lib/grape_logging/middleware/request_logger.rb#L36). Nevertheless, the fact that response is `nil`, should be solved in grape. Any idea why this happens?

In a very simplified version, my migration looks like:

```ruby
params do
  requires :id, type: Integer
end
get '/:id' do
  params.to_json
end
```

```ruby
route_param :id, type: Integer do
  get do
    params.to_json
  end
end
```

namusyaka:
@rngtng Could you provide a minimal failing project with Gemfile.lock?

rngtng:
@namusyaka thx for you quick response. while creating an example project i figured it actually hasn't anything todo with the refactoring to `route_param`, but with upgrading to `grape HEAD`. This pulled in 
https://github.com/ruby-grape/grape/pull/1240 which call `after` on logging `https://github.com/aserafin/grape_logging/blob/master/lib/grape_logging/middleware/request_logger.rb#L18`. let's continue discussion on #1240

PR #1256, opened 2016-01-25
Don't use hash rocket unless there is any particular reason [ci skip]

namusyaka:
I think we should not use hash rocket unless there is any particular reason.


dblock:
:+1:

PR #1257, opened 2016-01-26
Allow to pass method name to :with option of rescue_from

### LINKED ISSUE #790: Extend rescue_from to take a with: parameter
It would be nice to define a module with error handling methods, then pass the error handler to the rescue_from method in the following fashion.

    module API
      module ErrorHandler
         def record_not_found
            rack_response API::Utils::ApiErrors.new({type: e.class.name, message: 'Record not found'}).to_json, 404
         end
      end
    end

Then in the GRAPE::API base class one could call rescue_from like this...

    module API
      class Root < Grape::API
        prefix 'api'
        format :json

        helpers API::ErrorHandler

        rescue_from ActiveRecord::RecordNotFound, with: :record_not_found # Use the helper method as the handler for this error
      end
    end

namusyaka:
Closes #790

I think the feature is useful for writing error handling without block.
Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#1243](https://github.com/ruby-grape/grape/pull/1243): Add `header` support for middleware - [@namusyaka](https://github.com/namusyaka).
 * [#1252](https://github.com/ruby-grape/grape/pull/1252): Allow default to be a subset or equal to allowed values without raising IncompatibleOptionValues - [@jeradphelps](https://github.com/jeradphelps).
 * [#1255](https://github.com/ruby-grape/grape/pull/1255): Allow param type definition in route_param - [@namusyaka](https://github.com/namusyaka)
+* [#1257](https://github.com/ruby-grape/grape/pull/1257): Allow to pass method name to :with option of rescue_from - [@namusyaka](https://github.com/namusyaka)
 * Your contribution here.
 
 #### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [#1252](https://github.com/ruby-grape/grape/pull/1252): Allow default to be a subset or equal to allowed values without raising IncompatibleOptionValues - [@jeradphelps](https://github.com/jeradphelps).
 * [#1255](https://github.com/ruby-grape/grape/pull/1255): Allow param type definition in route_param - [@namusyaka](https://github.com/namusyaka)
 * [#1257](https://github.com/ruby-grape/grape/pull/1257): Allow to pass method name to :with option of rescue_from - [@namusyaka](https://github.com/namusyaka)
+* [#1257](https://github.com/ruby-grape/grape/pull/1257): Deny to pass values except Proc, Symbol or String to :with option of rescue_from - [@namusyaka](https://github.com/namusyaka)
 * Your contribution here.
 
 #### Fixes
--- a/README.md
+++ b/README.md
@@ -1771,6 +1771,22 @@ end
 
 The `rescue_from` block must return a `Rack::Response` object, call `error!` or re-raise an exception.
 
+The `with` keyword is available as `rescue_from` options, it can be passed method name or `Rack::Response` object.
+
+```ruby
+class Twitter::API < Grape::API
+  format :json
+  helpers do
+    def server_error!
+      error!({ error: 'Server error.' }, 500, { 'Content-Type' => 'text/error' })
+    end
+  end
+
+  rescue_from :all,          with: :server_error!
+  rescue_from ArgumentError, with: Rack::Response.new('rescued with a method', 400)
+end
+```
+
 #### Unrescuable Exceptions
 
 `Grape::Exceptions::InvalidVersionHeader`, which is raised when the version in the request header doesn't match the currently evaluated version for the endpoint, will _never_ be rescued from a `rescue_from` block (even a `rescue_from :all`) This is because Grape relies on Rack to catch that error and try the next versioned-route for cases where there exist identical Grape endpoints with different versions.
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -1771,6 +1771,22 @@ end
...
+class Twitter::API < Grape::API
+  format :json
+  helpers do
+    def server_error!
+      error!({ error: 'Server error.' }, 500, { 'Content-Type' => 'text/error' })
+    end
+  end
+
+  rescue_from :all,          with: :server_error!
+  rescue_from ArgumentError, with: Rack::Response.new('rescued with a method', 400)
```
dblock: I think `with: :method` is very useful, however a rescue with a `Rack::Response` is probably dangerous, because the instance of the `Rack::Response` is constructed at load time. This tends to cause all kinds of wonderful behavior like memory leaks, people are reckless with them. I think the above should not be allowed, you should do a `-> () { Rack...}` when you want to return something like this.

lib/grape/middleware/error.rb
```diff
@@ -45,6 +45,10 @@ def find_handler(klass)
         handler = options[:rescue_handlers].find(-> { [] }) { |error, _| klass <= error }[1]
         handler ||= options[:base_only_rescue_handlers][klass]
         handler ||= options[:all_rescue_handler]
+        with_option = options[:rescue_options][:with]
+        if with_option.instance_of?(Symbol) && respond_to?(with_option)
```
dblock: I think we need an exception here, if you do `with: :foobar` and it doesn't exist you want to fail loudly at the time the API is loaded, not at runtime. I am not sure though, it could also happen at runtime. This either way needs a different behavior and tests, we can't just ignore an invalid `with` handler.

README.md
namusyaka: Yeah, you're right. But the dangerous behaviour is [expected](https://github.com/ruby-grape/grape/blob/master/spec/grape/api_spec.rb#L1506-L1519).
Should I remove that and introduce lambda approach?

lib/grape/middleware/error.rb
namusyaka: I agreed. I'll implement exception feature. Btw, if specifed helper is not included yet, the exception will be raised on loading the middleware.

README.md
dblock: I think that yes. You should either use a symbol or a proc here.

lib/grape/middleware/error.rb
dblock: I see. I vote for the simplest and least error prone implementation and for catching bugs "early".

namusyaka:
Force pushed. 
@dblock Can you take a look at the changes?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 * [#1243](https://github.com/ruby-grape/grape/pull/1243): Add `header` support for middleware - [@namusyaka](https://github.com/namusyaka).
 * [#1252](https://github.com/ruby-grape/grape/pull/1252): Allow default to be a subset or equal to allowed values without raising IncompatibleOptionValues - [@jeradphelps](https://github.com/jeradphelps).
 * [#1255](https://github.com/ruby-grape/grape/pull/1255): Allow param type definition in route_param - [@namusyaka](https://github.com/namusyaka)
+* [#1257](https://github.com/ruby-grape/grape/pull/1257): Allow to pass method name to :with option of rescue_from - [@namusyaka](https://github.com/namusyaka)
```
dblock: We didn't have `with: ` before, right? So both of these are the same change, which is "Added `rescue_from with: ...`" that can take a method name or a proc.

dblock: Nevermind, we did, it's early in the AM ;)

lib/grape/dsl/request_response.rb
```diff
@@ -105,7 +105,13 @@ def rescue_from(*args, &block)
           end
 
           options = args.extract_options!
-          handler ||= proc { options[:with] } if options.key?(:with)
+          if options.key?(:with)
```
dblock: Maybe refactor this into a method?

```
handler ||= extract_with(options)
```

lib/grape/middleware/error.rb
```diff
@@ -45,6 +45,14 @@ def find_handler(klass)
         handler = options[:rescue_handlers].find(-> { [] }) { |error, _| klass <= error }[1]
         handler ||= options[:base_only_rescue_handlers][klass]
         handler ||= options[:all_rescue_handler]
+        with_option = options[:rescue_options][:with]
+        if with_option.instance_of?(Symbol)
+          if respond_to?(with_option)
+            handler ||= self.class.instance_method(with_option).bind(self)
+          else
+            fail NoMethodError, "undefined method `#{with_option}' for your application"
```
dblock: I would remove "for your application"

lib/grape/dsl/request_response.rb
```diff
@@ -105,7 +105,13 @@ def rescue_from(*args, &block)
...
+            if options[:with].instance_of?(Proc)
+              handler ||= options[:with]
+            elsif !(options[:with].instance_of?(Symbol) || options[:with].instance_of?(String))
+              fail ArgumentError, ":with option expected Symbol, String or Proc, got #{options[:with].class}"
```
dblock: I would turn this around, "with: #{options[:with].class}, expected Symbol, String or Proc".

dblock:
One last thing we need a test (and possibly a fix for) is when you mix a block and `with:`. I suppose it should raise an exception as well.

```ruby
rescue_from ArgumentError, with: error_handler do | ... |
  # is this called?
end
```

namusyaka:
Done. Thank you for your review!
I think both :with option and block should not be passed, too.

namusyaka:
@dblock Let me know if you have any problems.

dblock:
merged via https://github.com/ruby-grape/grape/commit/110c4f17aaa27a517b33927d62b3e5bcaca51e4b. I just edited the CHANGELOG a bit. Nice work @namusyaka, thanks!

namusyaka:
Thanks!

PR #1259, opened 2016-01-27
ValidationErrors initialize check args[:errors] is present

wuxiaoyi:
ValidationErrors will throw NoMethodError (undefined method `each' for nil:NilClass) when I manually throw a ValidationErrors exception without params.
This argument appears to be necessary, so it should detect the presence or absence of this key. Or give it a default value.
And the fail word should let us know which key is missing.

### DIFF
--- a/lib/grape/exceptions/validation.rb
+++ b/lib/grape/exceptions/validation.rb
@@ -7,7 +7,7 @@ module Grape
       attr_accessor :message_key
 
       def initialize(args = {})
-        fail 'Params are missing:' unless args.key? :params
+        fail 'Params are missing: params' unless args.key? :params
         @params = args[:params]
         @message_key = args[:message_key]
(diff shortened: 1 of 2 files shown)


dblock:
Thanks!

The first change broke a legit spec. I think it already says "Params are missing", so it's confusing to say "Params are missing: params", no?

The second one looks legit, but in which context do you get it? At least needs a test.

wuxiaoyi:
@dblock , thank you for your review :-)
I added test and fixes description.

dblock:
Re-reading all this I have more questions, sorry :)

First, for this implementation. The error should be an `ArgumentError`. We should also standardize the message matching the rest of Ruby implementations in the core library, so `ArgumentError: missing args: errors`. If you can please make those changes and squash commits.

Finally, maybe it just makes sense to treat `args` without `errors` as an empty array? I am not sure.

PR #1262, opened 2016-01-31
Don't use single quotation if string interpolation exists

namusyaka:


dm1try:
@namusyaka :thumbsup:

Issue #1260, opened 2016-01-29
Recommended 404 handling breaks OPTIONS requests


arempe93:
When using the 404 handling recommended by the README, it breaks all options requests in the API.

Spec

``` ruby
it 'allows options with 404 handling' do

  subject.post :hello do
    'Hello World'
  end

  subject.route :any, '*path' do
    error! :not_found, 404
  end

  options '/hello'

  expect(last_response.status).to eql 204
  expect(last_response.body).to be_blank
end
```

Output

```
Failure/Error: expect(last_response.status).to eql 204

    expected: 204
    got: 404
```

Changing the method to `route [:get, :post, :put, :delete], '*path'` is a workaround. But it was definitely unexpected behavior for me. The README seems to suggest that an OPTIONS route is created along with the declared route, leading one to believe it would also take precedence over the 404 handling route as well.

dblock:
I think this is a bug.

arempe93:
Closing - fixed by #1263

dblock:
@arempe93 I have reverted the fix committed in #1263 via https://github.com/ruby-grape/grape/commit/5aa655406398e547f16e32f3f1327a3f8248c820. See https://github.com/ruby-grape/grape/issues/1307 for details. Now we have specs on master for behavior that was regressed.

dblock:
The solution is probably not to create dynamic routes for other HTTP verbs, but to be able to say "the API has a set of matching routes", "do any match the verb", if yes, execute, otherwise generate a 405. Maybe we can do that part of #1276?

dblock:
This is likely fixed in https://github.com/ruby-grape/grape/pull/1276, care to followup on this @arempe93 ? Thanks.

arempe93:
Yes I can take a look on the edge version soon

dblock:
Bump @arempe93.

namusyaka:
Fixed in https://github.com/ruby-grape/grape/commit/9f4ba67d4aca90f04a65c5ff921b1687942eb99c.  I've just confirmed that.

PR #1263, opened 2016-02-01
Place generated OPTIONS ahead of user defined routes

arempe93:
Fix for #1260 

This change places the generated OPTIONS routes ahead of user defined routes, so that 404 handling does not interfere.

I also noticed a problem when doing this, declaring a route with `:any` as the http verb would declare an OPTIONS route and a 405 Not Allowed route for ALL methods. It would just be below the user defined route in the stack, and therefore never called. This, however, interfered with my change so I fixed it to map to all methods properly in `Grape::API#add_head_not_allowed_methods_and_options_methods`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [#1216](https://github.com/ruby-grape/grape/pull/1142): Fix JSON error response when calling `error!` with non-Strings - [@jrforrest](https://github.com/jrforrest).
 * [#1225](https://github.com/ruby-grape/grape/pull/1225): Fix `given` with nested params not returning correct declared params - [@JanStevens](https://github.com/JanStevens).
 * [#1249](https://github.com/ruby-grape/grape/pull/1249): Don't fail even if invalid type value is passed to default validator - [@namusyaka](https://github.com/namusyaka).
+* [#1263](https://github.com/ruby-grape/grape/pull/1263): Fix `route :any, '*path'` breaking generated OPTIONS, Not Allowed routes - [@arempe93](https://github.com/arempe93)
 
 0.14.0 (12/07/2015)
 ===================
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -555,6 +555,9 @@ describe Grape::API do
         subject.get 'example' do
           'example'
         end
+        subject.route :any, '*path' do
+          error! :not_found, 404
+        end
         options '/example'
       end
 
@@ -583,13 +586,53 @@ describe Grape::API do
       end
     end
 
-    it 'allows HEAD on a GET request' do
-      subject.get 'example' do
-        'example'
+    describe 'adds a 405 Not Allowed route that' do
+      before do
+        subject.before { header 'X-Custom-Header', 'foo' }
+        subject.post :example do
+          'example'
+        end
+        subject.route :any, '*path' do
+          error! :not_found, 404
+        end
+        get '/example'
       end
-      head '/example'
-      expect(last_response.status).to eql 200
-      expect(last_response.body).to eql ''
+
+      it 'returns a 405' do
+        expect(last_response.status).to eql 405
+      end
+
+      it 'has an empty body' do
+        expect(last_response.body).to be_blank
+      end
+
+      it 'has an Allow header' do
+        expect(last_response.headers['Allow']).to eql 'OPTIONS, POST'
+      end
+
+      it 'has a X-Custom-Header' do
+        expect(last_response.headers['X-Custom-Header']).to eql 'foo'
+      end
+    end
+
+    context 'allows HEAD on a GET request that' do
+      before do
+        subject.get 'example' do
+          'example'
+        end
+        subject.route :any, '*path' do
+          error! :not_found, 404
+        end
+        head '/example'
+      end
+
+      it 'returns a 200' do
+        expect(last_response.status).to eql 200
+      end
+
+      it 'has an empty body' do
+        expect(last_response.body).to eql ''
+      end
     end
 
     it 'overwrites the default HEAD request' do
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -32,6 +32,10 @@ end
 
 See [#1147](https://github.com/ruby-grape/grape/issues/1147) and [#1240](https://github.com/ruby-grape/grape/issues/1240) for discussion of the issues.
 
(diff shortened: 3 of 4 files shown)


arempe93:
I think there might actually be a slightly better way to do this. I will revisit tomorrow

dblock:
@arempe93 Thanks. Let us know how you feel about this PR. Either way please do squash the commits and update CHANGELOG. Possibly also update README.

arempe93:
Ok - I think this is ready for review now

spec/grape/api_spec.rb
```diff
@@ -583,13 +586,53 @@ def subject.enable_root_route!
       end
     end
 
-    it 'allows HEAD on a GET request' do
-      subject.get 'example' do
-        'example'
+    describe 'adds a 405 Not Allowed route that', focus: true do
```
dblock: There'a  `focus: true` still here, remove.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [#1216](https://github.com/ruby-grape/grape/pull/1142): Fix JSON error response when calling `error!` with non-Strings - [@jrforrest](https://github.com/jrforrest).
 * [#1225](https://github.com/ruby-grape/grape/pull/1225): Fix `given` with nested params not returning correct declared params - [@JanStevens](https://github.com/JanStevens).
 * [#1249](https://github.com/ruby-grape/grape/pull/1249): Don't fail even if invalid type value is passed to default validator - [@namusyaka](https://github.com/namusyaka).
+* [#1263](https://github.com/ruby-grape/grape/pull/1263): Refactor generated OPTIONS and Not Allowed routes to fix 404 handling - [@arempe93](https://github.com/arempe93)
```
dblock: I think the important part is that this fixes handling of 404 with OPTIONS, so I would just say that, eg. "Fix: using route :any breaks OPTIONS" or something like that.

dblock:
There's a behavior change here when you use `route :any` with `OPTIONS`, it's possible that some people accidentally relied on broken behavior, I think you should also add a paragraph into UPGRADING, thanks!

dblock:
Code looks good FYI, nice work.

arempe93:
I believe this change should address your comments. I left in the part about Not Allowed methods because the following spec would also fail before my changes.

```ruby
it 'should route 405' do
  subject.get :hello do
    'Hello World'
  end
  subject.route :any, '*path' do
    error! :not_found, 404
  end

  post '/hello'

  expect(last_response.status).to eql 405
end
```

Let me know if there needs to be any other documentation

dblock:
Merged via https://github.com/ruby-grape/grape/commit/e54b8b0e40d34b64df65c2002f0dcc8be7edfcf3 (I added a period in the CHANGELOG, OCD ;)).

dblock:
This introduced a regression, see https://github.com/ruby-grape/grape/issues/1307. I am not sure what the right fix is (yet).

dblock:
@arempe93 I have reverted this fix in https://github.com/ruby-grape/grape/commit/5aa655406398e547f16e32f3f1327a3f8248c820. Couldn't fix the last spec in that commit. Reopening #1260 as well.

arempe93:
@dblock the problem is (using the example from `spec/grape/api/namespace_parameters_in_route_spec`) is that `me/:id` should get a 405 for a `GET`, but that can also match `me/pending`. And it does match that first because of how this PR re-ordered the route stack - like this

```ruby
# PR 1263 ordering
version=, method=PUT, path=/me/pending(.:format)
version=, method=POST, path=/me/pending(.:format)
version=, method=DELETE, path=/me/pending(.:format)
version=, method=PATCH, path=/me/pending(.:format)
version=, method=OPTIONS, path=/me/pending(.:format)
version=, method=GET, path=/me/:x(.:format)    # first match, incorrect (405)
version=, method=POST, path=/me/:x(.:format)
version=, method=DELETE, path=/me/:x(.:format)
version=, method=PATCH, path=/me/:x(.:format)
version=, method=HEAD, path=/me/:x(.:format)
version=, method=OPTIONS, path=/me/:x(.:format)
version=, method=PUT, path=/me/:id(.:format)
version=, method=GET, path=/me/pending(.:format)
```

compared to `master` ordering

```ruby
# original ordering
version=, method=PUT, path=/me/:id(.:format)
version=, method=GET, path=/me/pending(.:format)   # first match, correct (200)
version=, method=OPTIONS, path=/me/:x(.:format)
version=, method=GET, path=/me/:x(.:format)
version=, method=POST, path=/me/:x(.:format)
version=, method=DELETE, path=/me/:x(.:format)
version=, method=PATCH, path=/me/:x(.:format)
version=, method=HEAD, path=/me/:x(.:format)
version=, method=OPTIONS, path=/me/pending(.:format)
version=, method=PUT, path=/me/pending(.:format)
version=, method=POST, path=/me/pending(.:format)
version=, method=DELETE, path=/me/pending(.:format)
version=, method=PATCH, path=/me/pending(.:format)
```

But this is complicated, because even if we do some complicated ordering where the 405/options are placed after each route, this can still be incorrect

```ruby
# concept complicated ordering
version=, method=PUT, path=/me/:id(.:format)
version=, method=OPTIONS, path=/me/:x(.:format)
version=, method=GET, path=/me/:x(.:format)   # still first match, incorrect (405)
version=, method=POST, path=/me/:x(.:format)
version=, method=DELETE, path=/me/:x(.:format)
version=, method=PATCH, path=/me/:x(.:format)
version=, method=HEAD, path=/me/:x(.:format)
version=, method=GET, path=/me/pending(.:format)
version=, method=OPTIONS, path=/me/pending(.:format)
version=, method=PUT, path=/me/pending(.:format)
version=, method=POST, path=/me/pending(.:format)
version=, method=DELETE, path=/me/pending(.:format)
version=, method=PATCH, path=/me/pending(.:format)
```

The only solution I see to make 405 work is to dump them all after the routes...which was the point of this PR to fix 404 handling. So I am unsure of how to fix it, short of re-ordering your routes in your grape project and having some more complicated route ordering.

dblock:
@arempe93 I totally understand the problem (and your explanation is correct). Of the two evils I chose the lesser, not to have a regression. I think we need a much smarter fix here that doesn't create these other routes, but inspects the routing table to figure out if there's a match with a different verb and generates a 405 accordingly.

arempe93:
ok I will look into a solution more in that direction when I finally have the free time again

Issue #1177, opened 2015-10-07
default_error_status does not inherit from its mounted parent


antulik:
``` ruby
class Api::Base < Grape::API
  default_error_status 400

  mount Api::Status
end
```

`error!` inside `Api::Status` would respond with 500

namusyaka:
This is a timing problem.  Look at the following code.

```ruby
# successful
  it 'inherits default_error_status' do
    subject.version 'v1', using: :path
    subject.default_error_status 400
    subject.get('/erro') { error!('erro') }
    api = Class.new(Grape::API)
    subject.mount api
    api.get('/error') do
      error!('error!')
    end

    get '/v1/error'
    expect(last_response.status).to eql 400
    expect(last_response.body).to eq('error!')
  end

# failure
  it 'inherits default_error_status' do
    subject.version 'v1', using: :path
    subject.default_error_status 400
    subject.get('/erro') { error!('erro') }
    api = Class.new(Grape::API)
    api.get('/error') do
      error!('error!')
    end
    subject.mount api

    get '/v1/error'
    expect(last_response.status).to eql 400
    expect(last_response.body).to eq('error!')
  end
```

So we can fix by force setting the value on mount, it's something like this:
```diff
diff --git lib/grape/dsl/routing.rb lib/grape/dsl/routing.rb
index cdfe761..3962946 100644
--- lib/grape/dsl/routing.rb
+++ lib/grape/dsl/routing.rb
@@ -87,6 +87,11 @@ module Grape
               change!
             end

+            app.endpoints.each do |endpoint|
+              current_error_status = namespace_inheritable(:default_error_status) || 500
+              endpoint.namespace_inheritable(:default_error_status, current_error_status)
+            end if !app.is_a?(Proc) && app.ancestors.include?(Grape::API)
+
             endpoints << Grape::Endpoint.new(
               in_setting,
               method: :any,
```
The patch seems working well and all tests are passed.

However, if we should not force set the value, I guess we should remove [this line](https://github.com/ruby-grape/grape/blob/master/lib/grape/endpoint.rb#L83) and introduce other approach.

dblock:
I believe we should think twice about inheriting anything from the mounting parent. Today we don't, you mount the entire API with its options and conditions, not with its parent's options. Some people want one behavior and others want another, so I think we want both.

Maybe we should add `include` or `extend` or a similar verb next to `mount` that has the suggested behavior that delegates all settings to the parent?

namusyaka:
[...]

PR #1264, opened 2016-02-01
Unify behavior of default_error_status, ref #1177

namusyaka:
ref #1177
I think we should unify behavior of default_error_status before adding useful verbs.


dblock:
If this change is to be merged we need an UPGRADING entry, a README doc update and CHANGELOG. I am very worried though. It would be the first thing that carries into what's being `mount`ed and that can go both ways.

namusyaka:
@dblock Force pushed. Yes, this is first step for providing that.

dblock:
Thanks. Leaving this open for others to comment.

PR #1266, opened 2016-02-02
Fix Allow header including OPTIONS when do_not_route_options! set

arempe93:
I believe this may be a fix for the intention of #1119, or maybe just its own small bug


dblock:
Yep, merging. Thanks.

Issue #1239, opened 2016-01-06
A param of type Hash is converted to Array in declared(params)


arempe93:
Spec describing the issue

``` ruby
it 'declared Hash groups should be a Hash' do

    app.params do
        optional :group, type: Hash do
            optional :foo, type: String
        end
    end
    app.get :group do
        declared(params).to_json
    end

    get '/group'
    json = JSON.parse(last_response.body)

    puts json['group'].class.name
    expect(json['group']).to be_kind_of(Hash)
end
```

This produces the following output

```
Array
Failure/Error: expect(json['group']).to be_kind_of(Hash)
       expected [] to be a kind of Hash
```

Setting a default on the param is a workaround

``` ruby
it 'declared Hash groups should be a Hash' do

    app.params do
        optional :group, type: Hash, default: {} do
            optional :foo, type: String
        end
    end
    app.get :group do
        declared(params).to_json
    end

    get '/group'
    json = JSON.parse(last_response.body)

    puts json['group']
    expect(json['group']).to be_kind_of(Hash)
end
```

This spec passes without error with the expected output

```
{ "foo" => nil }
```

Using Grape v0.14.0, running specs on master

dblock:
Looks like a definite bug, I would expect `group` not to be in the response at all without a `default`, right?

arempe93:
That could be an expectation, yes. Personally I like having it still be there and showing the values of the nested parameters because of the way I use / think about them as groups. But in either case, I don't think an Array is an expected value

ochagata:
I believe that this is due to this specific line:

https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/inside_route.rb#L48

arempe93:
Is this an intended behavior, as set forth by [this spec in `specs/grape/endpoint_spec.rb`?](https://github.com/ruby-grape/grape/blob/fbf633dbfec2b4a1586d60340d7517aecb5662fb/spec/grape/endpoint_spec.rb#L343). This tests pretty much the exact opposite of the spec from the issue description

dblock:
I think that spec ignores the fact that there's an explicit `type`.

dnesteryuk:
@kadotami fixed this issue in https://github.com/ruby-grape/grape/pull/2043 :fireworks:

PR #1267, opened 2016-02-03
Adds failing spec for ruby-grape/grape#1239

arempe93:
Reference to #1239 
Not sure how to fix this...but I added a few more specs to `declared` that demonstrate the issue.

The problem lies in this code in [`lib/grape/dsl/inside_route.rb#50`](https://github.com/ruby-grape/grape/blob/86e3cd1f63c58d78c257b043dcf259954a594661/lib/grape/dsl/inside_route.rb#L50)

``` ruby
children.is_a?(Array) ? [] : {}
```

`children` is always an Array so it always gives `[]`. Replacing that code with a plain `[]` gives the same results. Replacing it with `{}` passes the nested Hash spec, but then fails the nested Array spec.

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -255,8 +255,11 @@ describe Grape::Endpoint do
         requires :first
         optional :second
         optional :third, default: 'third-default'
-        optional :nested, type: Hash do
+        optional :hash, type: Hash do
           optional :fourth
+        end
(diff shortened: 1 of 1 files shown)


dnesteryuk:
@arempe93 thanks for your contribution, similar tests were added in https://github.com/ruby-grape/grape/pull/2043, it should be sufficient :wink:

Issue #1253, opened 2016-01-22
Document how to return data/terminate the request in a before block


dblock:
I believe is by calling `error!`, but I realize that's not documented, at least clearly. Also it's not clear whether the `after` block would be called anyway in this case. Needs tests for all these cases.

Issue #1251, opened 2016-01-21
Support present in before


yoonwaiyan:
From what I've understand from Issue #39, I've create a before filter for a block of my API classes, which authenticates user for every API call within that block. The API should return a JSON with status and error message that notifies user about failed API calls. Currently I've overwritten the error block of my API  base file which returns status of 1 and the error message whenever an error occured in the API call. 

There's no documentation on how to call the `present` syntax in before_block and end the request which behaves like Rails `before_filter`. For authentication, I wanted to return another status other than 1, maybe a status of 9, but the only way to end the request is to raise an error which return the default status of 1. I tried use `present` and put `return` to end the API call but it raised up an error. Any workaround on this?

The code below is my API base file.

```ruby
require 'grape-swagger'

module Api  
  class Base < Grape::API
    rescue_from :all do |e|
      Rails.logger.error "#{e.message}\n\n#{e.backtrace.join("\n")}"
      Rollbar.error(e)
      # Call Grape error
      error!(e)
    end

    prefix 'app_name.api'
    version 'v1', using: :param
    format :json
	  error_formatter :json, ->(message, backtrace, options, env) {
      { status: 1, message: "error: #{message}" }.to_json
	  }

    before do
      header['Access-Control-Allow-Origin'] = '*'
      header['Access-Control-Request-Method'] = '*'
    end

    # API that needs authentication
    group do
      before do
        if params[:auth_token]
          unless Customer.where(auth_token: params[:auth_token]).count >= 1
            # how to return an API using present in this before block?
            error!("Authentication failed.")
          end
        end
      end

      params do
        optional :auth_token, type: String, desc: "Authentication Token"
      end
      mount Api::Categories
      mount Api::Coupons
      mount Api::Customers
    end

    # API that does not need authentication
    group do
      mount Api::UnauthenticatedCustomers
    end
    
    add_swagger_documentation api_documentation: { desc: 'Reticulated splines API swagger-compatible documentation.', hidden: true }
  end
end
```

dblock:
In this case, I think what you want is an error handler that returns a certain type of data, and not to use present in the `before` block.

```ruby
rescue_from :all, backtrace: true do |e|
  error = { type: 'other_error', message: e.message }
  error[:backtrace] = e.backtrace unless Rails.env.production?
  rack_response(error.to_json, 400)
end
```

We could support `present` in the `before` and `after` blocks, as we now support things like `header`, so I'll leave this open as a feature request.

[...]

PR #1268, opened 2016-02-03
Adds documentation and specs for callback response behaviors

arempe93:
References #1251, #1253 

I found that there is already support for using `present` in a `before` filter. Any filter actually, along with `error!`. Using `error!` will act as though the response stops there completely. Any filter or api call that would have happened afterwords is not called (including `after` to be specific to #1253).

Wrote specs and documentation for current behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -2453,7 +2453,9 @@ Before and after callbacks execute in the following order:
 
 Steps 4, 5 and 6 only happen if validation succeeds.
 
-E.g. using `before`:
+#### Examples
+
+Using a simple `before` block to set a header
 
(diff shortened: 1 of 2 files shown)


dblock:
There's some issue with Ruby 1.9.3 rubocop, if you could please take a look. Rubocop has a setting that targets a certain Ruby version somewhere - I don't think we should drop 1.9.3 just yet, although I am open to talking about it.

arempe93:
Ah, I forgot `%i` is 2.0 syntax. Fixed now

dblock:
Merged.

Issue #1270, opened 2016-02-08
Param versioning param key isn't set using version options


wshatch:
As stated in the docs, setting parameter for the param strategy like so 
```ruby
version 'v1', using: :param, parameter: 'v'
```
Won't work since the param strategy is looking in ```@options``` instead of ```@options[:version_options]```

mneumark:
Just tried the same thing and had the same issue.

dblock:
Closed via https://github.com/ruby-grape/grape/pull/1271.

PR #1271, opened 2016-02-08
Param versioning will now correctly set parameter key

wshatch:
Issue: https://github.com/ruby-grape/grape/issues/1270

Currently, we're looking for the parameter in ```@options``` instead of ```@options[:version_options]```. This should fix the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * [#1249](https://github.com/ruby-grape/grape/pull/1249): Don't fail even if invalid type value is passed to default validator - [@namusyaka](https://github.com/namusyaka).
 * [#1263](https://github.com/ruby-grape/grape/pull/1263): Fix `route :any, '*path'` breaking generated `OPTIONS`, Method Not Allowed routes - [@arempe93](https://github.com/arempe93).
 * [#1266](https://github.com/ruby-grape/grape/pull/1266): Fix `Allow` header including `OPTIONS` when `do_not_route_options!` is active - [@arempe93](https://github.com/arempe93).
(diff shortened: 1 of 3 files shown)


dblock:
Did this break in an older version? This definitely needs an integration test that demonstrates the bug, please. And a CHANGELOG entry.

mneumark:
Thanks for fixing this though. This bug is pretty easy to replicate. I was ready submit a PR myself before I saw this.

wshatch:
Integration tests and changelog entry added. I Don't know when this broke.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * [#1249](https://github.com/ruby-grape/grape/pull/1249): Don't fail even if invalid type value is passed to default validator - [@namusyaka](https://github.com/namusyaka).
 * [#1263](https://github.com/ruby-grape/grape/pull/1263): Fix `route :any, '*path'` breaking generated `OPTIONS`, Method Not Allowed routes - [@arempe93](https://github.com/arempe93).
 * [#1266](https://github.com/ruby-grape/grape/pull/1266): Fix `Allow` header including `OPTIONS` when `do_not_route_options!` is active - [@arempe93](https://github.com/arempe93).
+* [#1270](https://github.com/ruby-grape/grape/pull/1270): Fixed an issue where the param versioning strategy wouldn't work with a custom parameter - [@wshatch](https://github.com/wshatch).
```
dblock: Just say "Fix `param` versioning with a custom parameter."

dblock:
Looks like there're some legit build failures.

wshatch:
Ok, issues should be fixed.

dblock:
:+1:

Issue #1272, opened 2016-02-11
rails rspec est :I got error  uninitialized constant RSpec::Rails (NameError)


rajcybage:
I have got error after in rspec rails app.

`````
C:\Sites\api_product>bundle exec rake spec
DL is deprecated, please use Fiddle
C:/RailsInstaller/Ruby2.1.0/bin/ruby.exe -I'C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/lib';'C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-support-3.4.1/lib' 'C
:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/exe/rspec' --pattern 'spec/**{,/*/**}/*_spec.rb'
C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/lib/rspec/core.rb:178:in `block in const_missing': uninitialized constant RSpec::Rails (NameError)
        from C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/lib/rspec/core.rb:178:in `fetch'
        from C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/lib/rspec/core.rb:178:in `const_missing'
        from C:/Sites/api_product/spec/spec_helper.rb:44:in `block in <top (required)>'
        from C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/lib/rspec/core.rb:97:in `configure'
        from C:/Sites/api_product/spec/spec_helper.rb:20:in `<top (required)>'
        from C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/lib/rspec/core/configuration.rb:1295:in `require'
        from C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/lib/rspec/core/configuration.rb:1295:in `block in requires='
        from C:/RailsInstaller/Ruby2.1.0/lib/ruby/gems/2.1.0/gems/rspec-core-3.4.2/lib/rspec/core/configuration.rb:1295:in `each'
[...]

PR #1273, opened 2016-02-11
[ci skip] RSpec::Rails should be in rails_helper instead spec_helper …

rajcybage:
…for rails app

#1272


dblock:
Yes , rails has changed how this works in 4.x, so lets just go with it.

rajcybage:
Thanks you :)

PR #1274, opened 2016-02-12
kill multi_json Dependencies

u2:
https://github.com/ruby-grape/grape-entity/pull/209


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 
 #### Fixes
 
+* [#1274](https://github.com/ruby-grape/grape/pull/1274): Kill multi_json Dependencies - [@u2](https://github.com/u2).
```
dblock: This should say "Removed `multi_json` dependencies", lets not kill anything ;)

dblock:
If this is to be merged, we need an UPGRADING entry that says that we no longer use MultiJson and provides a Json formatter example of how to put it back.

I am torn though. I have a project that has heavily benchmarked itself using `oj`, so I will probably want to use `MultiJson` to get that performance. Or maybe I'll want to benchmark it again, I am not sure. Right now Grape basically says "you can use any JSON library without any work", so effectively we're removing a feature. 

Thoughts?

wshatch:
What exactly is the advantage of this? It seems that it can kill OJ support which, from my understanding, performs better than the standard json module.

dblock:
I think it comes from http://www.mikeperham.com/2016/02/09/kill-your-dependencies/.

screen shot 2016-02-12 at 5 53 02 pm

u2:
Yes, it comes from mikeperham blog [Kill Your Dependencies](http://www.mikeperham.com/2016/02/09/kill-your-dependencies/).
>
Every dependency in your application has the potential to bloat your app, to destabilize your app, to inject odd behavior via monkeypatching or buggy native code. When you are considering adding a dependency to your Rails app, it’s a good idea to do a quick sanity check, in order of preference:

> Do I really need this at all? Kill it.
Can I implement the required minimal functionality myself? Own it.
If you need a gem:

> Does the gem have a native extension? Look for pure ruby alternatives.
Does the gem transitively pull in a lot of other gems? Look for simpler alternatives.

Less gem means less memory, less dependencies, and less bugs.And the promoting of performance of `oj` or other json gems is very tiny for grape.

And 
> All modern Rubies ship JSON as part of stdlib. Using the gem actually hurts multi-platform support due to build difficulties on Windows.

dblock:
I am on the side of removing the `multi_json` dependency because the number people using something like OJ is on the decline. I do want to make it very easy to bring back the old behavior.

First, this definitely needs UPGRADING describing how to re-introduce something like OJ (a custom formatter for example).

Maybe we can add configuration to the formatter where you can give it the engine to use for formatting and parsing? Or maybe it can detect the presence of multi_json and use that where available? Not sure what's best. Another option would be a `grape-multi-json` gem that brings the old behavior back. What do you think?

u2:
Yes, what about adding a  configuration option like `multi_json`? The default value is false,the JSON codes will be the native JSON, if it's true and existing `multi_json` gem, it will act like before.

dblock:
Maybe `format :json, processor: JSON` or `format :json, processor: MultiJson`?

u2:
:+1:  It looks workable.

dblock:
Replacing in favor of https://github.com/ruby-grape/grape/pull/1623.

PR #1276, opened 2016-02-14
Replace rack-mount with new router

namusyaka:
This is a first step for replcaing rack-mount with new router.
I'm going to improve performance to realize faster new router.

The new router has a few imcompatibility with old one.
The following cases:

* `route_*` attribute should be deprecated because the feature is implemented by using `method_missing`. So it's very slow. (However, reviving the attribute is possible. )
* Changed `rack.routing_args` to `api.routing_args`. The new router is for the Grape. I think the namespace is incorrect. 

Thoughts?


dblock:
This is excellent work. I am excited for getting this onto Grape master soon. It's really important to build some benchmarks and see how much improvement this is if at all, I think we owe this to the Grape community for switching something as essential as the router.

namusyaka:
All required tests has been successful now.

arempe93:
:+1: excited to see benchmarks

namusyaka:
I've been tring to improve performance of the new router, and then I'd like to report a few points and hear your opinions.

# Benchmark

First of all, I want you to know that the performance of rack-mount is excellent.
So I focused reducing memory utilization, and it was successful.

Please look at the following benchmark code and its results.

## Sample code

Environment: Mac OS X, Ruby2.3

```ruby
require 'rack'
require 'grape'
require 'benchmark'
require 'objspace'
require 'memprof2'

def memory_utilization
  Memprof2.start
  yield
  profiler = Memprof2.new
  profiler.configure
  mem = profiler.collect_info.values.inject(&:+)
  Memprof2.stop
  puts "Memory Utilization: #{mem/(1024.0 * 1024.0)}MB"
end

req1 = Rack::MockRequest.env_for('/100')
req2 = Rack::MockRequest.env_for('/users/1234', method: 'PUT')
req3 = Rack::MockRequest.env_for('/sources/hoge/aiueohogefuga', method: 'POST')
req4 = Rack::MockRequest.env_for('/sources/hoge/aiueohogefuga', method: 'PUT')

app = Class.new(Grape::API)
memory_utilization do

  500.times do |n|
    app.get("/#{n}") { n.to_s }
  end

  app.get('/') { 'hello' }
  app.put('/users/:id') { 'hello' }
  app.post('/sources', anchor: false) { 'hello' }

  app.call(req1)
  app.call(req2)
  app.call(req3)
  app.call(req4)
end

Benchmark.bm do |x|
  x.report do
    app.call(req1)
  end

  x.report do
    app.call(req2)
  end

  x.report do
    app.call(req3)
  end

  x.report do
    app.call(req4)
  end
end
```

## rack-mount

```
Memory Utilization: 32.22776699066162MB
       user     system      total        real
   0.000000   0.000000   0.000000 (  0.000638)
   0.000000   0.000000   0.000000 (  0.000393)
   0.000000   0.000000   0.000000 (  0.000249)
   0.000000   0.000000   0.000000 (  0.000061)
```

## new router

```
Memory Utilization: 14.685898780822754MB
       user     system      total        real
   0.000000   0.000000   0.000000 (  0.000495)
   0.000000   0.000000   0.000000 (  0.000263)
   0.000000   0.000000   0.000000 (  0.000231)
   0.000000   0.000000   0.000000 (  0.000534)
```

@dblock @arempe93 What do you think? Any ideas?

dblock:
Do I read this correctly that the new router consumes half the memory and performs 30% faster (except for that last row which seems like an outlier for both?).

I think you want to make a lot of those "http" calls, at the very least those routers could have a cold start.

namusyaka:
The new router consumes half the memory, but I think its performance is ambiguous.
Maybe the performance will be on the same level with rack-mount.

Okay, I tried to calculate benchmark for calling a lot of requests.
Please look at this sample code.

```ruby
require 'rack'
require 'grape'
require 'benchmark'
require 'benchmark/ips'
require 'objspace'
require 'memprof2'

def memory_utilization
  Memprof2.start
  yield
  profiler = Memprof2.new
  profiler.configure
  mem = profiler.collect_info.values.inject(&:+)
  Memprof2.stop
  puts "Memory Utilization: #{mem/(1024.0 * 1024.0)}MB"
end

req1 = Rack::MockRequest.env_for('/100')
req2 = Rack::MockRequest.env_for('/users/1234', method: 'PUT')
req3 = Rack::MockRequest.env_for('/sources/hoge/aiueohogefuga', method: 'POST')
req4 = Rack::MockRequest.env_for('/sources/hoge/aiueohogefuga', method: 'PUT')

app = Class.new(Grape::API)
memory_utilization do

  500.times do |n|
    app.get("/#{n}") { n.to_s }
  end

  app.get('/') { 'hello' }
  app.put('/users/:id') { 'hello' }
  app.post('/sources', anchor: false) { 'hello' }

  app.call(req1)
  app.call(req2)
  app.call(req3)
  app.call(req4)
end

Benchmark.ips do |x|
  [req1, req2, req3, req4].each do |req|
    x.report { app.call(req) }
  end
end
```
Also, the results are:

## rack-mount

```
Memory Utilization: 32.254828453063965MB
                       531.000  i/100ms
                       505.000  i/100ms
                       478.000  i/100ms
                         2.558k i/100ms
-------------------------------------------------
                          5.870k (±12.8%) i/s -     28.674k
                          5.507k (± 7.9%) i/s -     27.775k
                          5.594k (± 7.1%) i/s -     28.202k
                         46.889k (± 9.6%) i/s -    232.778k
```

## new router

```
Memory Utilization: 14.55474853515625MB
                       396.000  i/100ms
                       552.000  i/100ms
                       561.000  i/100ms
                       259.000  i/100ms
-------------------------------------------------
                          4.008k (± 7.4%) i/s -     20.196k
                          5.638k (± 4.2%) i/s -     28.704k
                          5.417k (± 4.5%) i/s -     27.489k
                          2.650k (± 4.2%) i/s -     13.468k
```

spec/grape/path_spec.rb
```diff
@@ -244,7 +244,7 @@ module Grape
           allow(path).to receive(:uses_specific_format?) { true }
           allow(path).to receive(:settings) { { format: :json } }
 
-          expect(path.path_with_suffix).to eql('/the/path(.json)')
+          expect(path.path_with_suffix).to eql('/the/path(.:format)')
```
dblock: Can you please explain this one?

dblock:
I will run this against a production codebase to see if anything fails. Generally I am down with merging this, properly READMEed, CHANGELOGed, squashed, etc., please. We need to replace rack-mount since it's a dead project.

dblock:
Haven't debugged, but the first failure in a real world project is with grape-swagger.

```
/Users/dblock/.rvm/gems/ruby-2.2.2/bundler/gems/grape-swagger-2f9aa49fc2e4/lib/grape-swagger.rb:27:in `block in add_swagger_documentation': undefined method `split' for nil:NilClass (NoMethodError)
	from /Users/dblock/.rvm/gems/ruby-2.2.2/bundler/gems/grape-swagger-2f9aa49fc2e4/lib/grape-swagger.rb:25:in `each'
	from /Users/dblock/.rvm/gems/ruby-2.2.2/bundler/gems/grape-swagger-2f9aa49fc2e4/lib/grape-swagger.rb:25:in `add_swagger_documentation'
	from /Users/dblock/source/artsy/gravity/dblock/app/api/v1/root_endpoint.rb:237:in `<class:RootEndpoint>'
	from /Users/dblock/source/artsy/gravity/dblock/app/api/v1/root_endpoint.rb:3:in `<module:V1>'
	from /Users/dblock/source/artsy/gravity/dblock/app/api/v1/root_endpoint.rb:2:in `<module:Api>'
	from /Users/dblock/source/artsy/gravity/dblock/app/api/v1/root_endpoint.rb:1:in `<top (required)>'
```

It's missing `routes`.

spec/grape/path_spec.rb
namusyaka: Oops, this is my bad. Fixed.

namusyaka:
@dblock Sorry for the failing with grape-swagger.
This failure is caused by breaking comatibility of `route_*` attrs. So I fixed the problems. 
Maybe you can use it now.

lib/grape/api.rb
```diff
@@ -126,16 +129,22 @@ def cascade?
...
-      methods_per_path = {}
+      routes_map = {}
 
       self.class.endpoints.each do |endpoint|
         routes = endpoint.routes
         routes.each do |route|
-          methods_per_path[route.route_path] ||= []
-          methods_per_path[route.route_path] << route.route_method
+          # using the :any shorthand produces [nil] for route methods, substitute all manually
+          route_settings = (routes_map[route.pattern.to_regexp] ||= {})
```
dblock: I am not in love with this `||=` assignment inside the right hand side. It's just a little difficult to read.

lib/grape/api.rb
```diff
@@ -155,18 +168,21 @@ def add_head_not_allowed_methods_and_options_methods
             allow_header = (self.class.namespace_inheritable(:do_not_route_options) ? allowed_methods : [Grape::Http::Headers::OPTIONS] | allowed_methods).join(', ')
 
             unless self.class.namespace_inheritable(:do_not_route_options)
-              generate_options_method(path, allow_header) unless allowed_methods.include?(Grape::Http::Headers::OPTIONS)
+              generate_options_method(path, allow_header, config[:requirements]) unless allowed_methods.include?(Grape::Http::Headers::OPTIONS)
```
dblock: Just make the third parameter `options`, a hash.

lib/grape/api.rb
```diff
@@ -155,18 +168,21 @@ def add_head_not_allowed_methods_and_options_methods
...
+            self.class.namespace_stackable(:format, [])
           end
         end
       end
     end
 
     # Generate an 'OPTIONS' route for a pre-exisiting user defined route
-    def generate_options_method(path, allow_header)
-      self.class.options(path, {}) do
+    def generate_options_method(path, allow_header, requirements = nil)
```
dblock: If instead of `requirements` you have `options` (a Hash), you have no work to do here.

lib/grape/api.rb
```diff
@@ -178,18 +194,15 @@ def generate_options_method(path, allow_header)
 
     # Generate a route that returns an HTTP 405 response for a user defined
     # path on methods not specified
-    def generate_not_allowed_method(path, allowed_methods, allow_header)
+    def generate_not_allowed_method(path, allowed_methods, allow_header, requirements = nil, endpoint = nil)
```
dblock: Same as above, just make the `options` a pass-through.

lib/grape/dsl/routing.rb
```diff
@@ -98,6 +98,8 @@ def mount(mounts)
               method: :any,
               path: path,
               app: app,
+              format: app.is_a?(Grape::API) ? app.class.namespace_inheritable(:format) : nil,
```
dblock: What can `app` be here otherwise?

lib/grape/error_formatter/base.rb
```diff
@@ -35,10 +35,10 @@ def present(message, env)
 
         presenter = env[Grape::Env::API_ENDPOINT].entity_class_for_obj(message, present_options)
 
-        unless presenter || env[Grape::Env::RACK_ROUTING_ARGS].nil?
+        unless presenter || env[Grape::Env::API_ROUTING_ARGS].nil?
```
dblock: Are we sure we need to change RACK_ROUTING_ARGS into API_ROUTING_ARGS? If we're going to change this maybe they should just be ROUTING_ARGS or GRAPE_ROUTING_ARGS?

namusyaka:
Done, thanks for pointing out!

dblock:
First regression, or at least a behavior change. Clearly we need specs on master for this - feel free to contribute that separately.

```ruby
require 'spec_helper'

describe Grape::API do
  subject do
    Class.new(Grape::API) do
      format :json

      get ':id' do
        params
      end
    end
  end

  def app
    subject
  end

  it 'does not include extension in id' do
    get '/foo.bar'
    expect(last_response.status).to eq 404
  end
end
```

This succeeds on master, fails on your branch with a 200, and returns a value of `foo.bar` for `id`, which is definitely wrong since in this case `.bar` should be interpreted as an extension causing a format selection.

dblock:
Do you think we can preserve `route_xyz` methods, possibly with a deprecation warning?

lib/grape/router.rb
```diff
@@ -0,0 +1,144 @@
...
+          response = yield(route)
+          break unless cascade?(response)
+        end
+        response
+      end
+    end
+
+    def transaction(env)
+      input, method, routing_args = *extract_required_args(env)
+      response = yield(input, method, ROUTING_ARGS_SETTER.curry[env, routing_args])
```
dblock: Could you explain what this magic is? :) This is beyond my Ruby-fu.

lib/grape/router/pattern.rb
```diff
@@ -0,0 +1,55 @@
...
+      extend Forwardable
+      def_delegators :pattern, :named_captures, :params
+      def_delegators :@regexp, :===
+      alias_method :match?, :===
+
+      def initialize(pattern, options = {})
+        @origin  = pattern
+        @path    = build_path(pattern, options)
+        @capture = extract_capture(options)
+        @pattern = Mustermann::Rails.new(@path, pattern_options)
```
dblock: Maybe we should contribute Mustermann::Grape? Do we really always want Rails semantics here?

dblock: Actually I think we explicitly *don't* want Rails semantics to be inherited because of https://github.com/rkh/mustermann/blob/master/mustermann-rails/lib/mustermann/rails.rb#L38, which is very Rails-version specific. I think you can safely commit a copy of that rails.rb minus everything we don't need into Grape.

namusyaka:
@dblock 
Thank you for the review.

>>First regression. We need specs on master for this since it wasn't caught in the Grape specs and I got it from my real world project, would appreciate a separate PR just for this story on master.

Okay, I've opened a pull request.
See https://github.com/ruby-grape/grape/pull/1296

>>This succeeds on master, fails on your branch with a 200, and returns a value of foo.bar for id, which is definitely wrong since in this case .bar should be interpreted as an extension causing a format selection. 

Yes, you're right.
[Mustermann::Grape](https://github.com/namusyaka/mustermann19/blob/19support/lib/mustermann/grape.rb) fixes the problem.

>>Do you think we can preserve route_xyz methods, possibly with a deprecation warning?

Yeah, the route_xxx methods have been deprecated.
See https://github.com/namusyaka/grape/commit/0453efed2b559c3f7434f08a9cc198dddc4c7fea

>>Maybe we should contribute Mustermann::Grape? Do we really always want Rails semantics here?

I've implemented [Mustermann::Grape](https://github.com/namusyaka/mustermann19/blob/19support/lib/mustermann/grape.rb), and also, I've opened a [pull request](https://github.com/rkh/mustermann/pull/46) to the central repository.

lib/grape/router.rb
namusyaka: Okay, I removed the tricky code ;)

namusyaka:
@dblock [The pull request](https://github.com/rkh/mustermann/pull/46) has been merged. 
Should I rebase and update CHANGELOG?

dblock:
Definitely, yes please! We should move forward with this for the next major release of Grape. I'll re-run my big test suite once you tell me and then maybe you can ask the mailing list to try this as well?

namusyaka:
I squashed commits and edit a few points in README.

Also I've opend [a topic](https://groups.google.com/forum/#!topic/ruby-grape/fb60Elvt6ao) in grape ML.

namusyaka:
@dblock Could you check it out, please?

dblock:
@namusyaka I am on it. Give me a couple of days to try this against my live project, then I'll probably do a Grape release, then merge this onto master.

namusyaka:
@dblock Great! I'm look forward to your feedback!

lib/grape/router/route.rb
```diff
@@ -0,0 +1,93 @@
...
+        else
+          parsed = pattern.params(input)
+          parsed ? parsed.delete_if { |_, value| value.nil? }.symbolize_keys : {}
+        end
+      end
+
+      private
+
+      def warn_route_methods(name)
+        warn "route_xxx methods such as route_#{name} have been deprecated, Please use #{name}"
```
dblock: make `please` lowercase or maybe this is better: `The method route_#{name} has been deprecated, please use #{name}.".

namusyaka: Yeah, right.
I've just implemented that. See [these lines](https://github.com/namusyaka/grape/blob/implement-new-router/lib/grape/router/route.rb#L90-L94).
The message will be something like this:
```
/Users/numb/dev/ruby/grape/test.rb:15: The route_xxx methods such as route_foo have been deprecated, please use foo.
```

dblock:
I've been tracking regressions on master for the next Grape release so far. Getting closer.

dblock:
I am wondering whether this new router can help us fix https://github.com/ruby-grape/grape/issues/1260 as well in a nice(r) way?

namusyaka:
The new router will fix the problem by not adding other routes. I've add specs for #1263.
Could you please check?

arempe93:
The PR that keeps on giving. Thanks a ton!

dblock:
:+1: 

I ran this against a production codebase and have about 20 failing specs out of maybe 1200 to look at tomorrow, stay tuned.

dblock:
Finally got to this. Here's the regression that is responsible for most of the failures I am seeing ATM.

```ruby
require 'spec_helper'

describe Grape::Endpoint do
  subject { Class.new(Grape::API) }

  def app
    subject
  end

  before do
    subject.namespace do
      format :json
      params do
        requires :id, desc: 'Identifier.'
      end
      get ':id' do
        {
          id: params[:id],
          format: params[:format]
        }
      end
    end
  end

  context 'get' do
    it 'no format' do
      get '/foo'
      expect(last_response.status).to eq 200
      expect(last_response.body).to eq(MultiJson.dump({ id: 'foo', format: nil }))
    end
    it 'json format' do
      get '/foo.json'
      expect(last_response.status).to eq 200
      expect(last_response.body).to eq(MultiJson.dump({ id: 'foo', format: 'json' }))
    end
    it 'invalid format' do
      get '/foo.invalid'
      expect(last_response.status).to eq 200
      expect(last_response.body).to eq(MultiJson.dump({ id: 'foo', format: 'invalid' }))
    end
  end
end
```

Introducing `format :json` causes the problem here, with default format handling we seem to do better.

The above passes on master. Failures with the new router:

```ruby
  1) Grape::Endpoint get json format
     Failure/Error: expect(last_response.body).to eq(MultiJson.dump({ id: 'foo', format: 'json' }))
     
       expected: "{\"id\":\"foo\",\"format\":\"json\"}"
            got: "{\"id\":\"foo\",\"format\":null}"
     
       (compared using ==)
     # ./spec/grape/api/invalid_format_spec.rb:34:in `block (3 levels) in <top (required)>'

  2) Grape::Endpoint get invalid format
     Failure/Error: expect(last_response.status).to eq 200
     
       expected: 200
            got: 404
     
       (compared using ==)
     # ./spec/grape/api/invalid_format_spec.rb:38:in `block (3 levels) in <top (required)>'
```

It doesn't correctly default to the format we want it to and ignores requests for another format which should default to the same one as well.

dblock:
@namusyaka Do commit the above spec to master like the last time too please.

namusyaka:
@dblock Thanks for your spec, I've just fixed the behavior and squashed.
Could you please check it out?

dblock:
Running my battery of tests ... you can write UPGRADING in the meantime maybe. Thanks.

README.md
```diff
@@ -2568,11 +2568,15 @@ Examine the routes at runtime.
...
 TwitterAPI::routes # yields an array of Grape::Route objects
-TwitterAPI::routes[0].route_version # => 'v1'
-TwitterAPI::routes[0].route_description # => 'Includes custom settings.'
-TwitterAPI::routes[0].route_settings[:custom] # => { key: 'value' }
+TwitterAPI::routes[0].version # => 'v1'
+TwitterAPI::routes[0].description # => 'Includes custom settings.'
+TwitterAPI::routes[0].settings[:custom] # => { key: 'value' }
 ```
 
+Note that `Route#route_xyz` methods have been deprecated since 0.15.0.
```
dblock: This should be in UPGRADING.

dblock:
From my tests ... this should be in UPGRADING: `route.route_method` was replaced by `route.request_method`, not `route.method` as per the deprecation.

dblock:
Another tricky one, https://github.com/namusyaka/grape/pull/2.

lib/grape/router/route.rb
```diff
@@ -0,0 +1,99 @@
...
+        end
+      end
+
+      private
+
+      def warn_route_methods(name, location)
+        path, line = *location.scan(SOURCE_LOCATION_REGEXP).first
+        path = File.realpath(path) if Pathname.new(path).relative?
+        warn <<-EOS
+#{path}:#{line}: The route_xxx methods such as route_#{name} have been deprecated, please use #{name}.
```
dblock: I think this will say the wrong thing for `route_method` ;)

namusyaka:
@dblock Thank you for polite review! Fixed.

namusyaka:
If the changes have no problem, I'm going to squashed them.

dblock:
```
Finished in 46 minutes 33 seconds (files took 49.5 seconds to load)
4712 examples, 0 failures, 28 pending
```

Lets squash it and get it merged!

namusyaka:
Wow, great! I'm so happy! Thank you!

dblock:
Merged.

arempe93:
:tada:

kuraga:
Is this pendragon or... ?..

dblock:
https://github.com/rkh/mustermann

PR #1280, opened 2016-02-19
Support Rack::Sendfile middleware

lfidnl:
Hello.

I've added support of `Rack::Sendfile` middleware. Before this fix` Rack::Sendfile` does not apply strategy of adding X-Accel-Redirect (or X-Sendfile) header, because Grape returns simple `Rack::Response` which doesn't respond to `to_path` method. So I've implemented simple `SendfileReponse` class which responds to `to_path`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2328,6 +2328,30 @@ class API < Grape::API
 end
 ```
 
+If you want to take advantage of `Rack::Sendfile`, which intercepts responses whose body is
+being served from a file and replaces it with a server specific X-Sendfile header, specify `to_path`
+method in your file streamer class which returns path of served file:
+
+```ruby
+class FileStreamer
+  # ...
+
+  def to_path
+    @file_path
+  end
+
+  # ...
+end
+```
+
+Note: don't forget turn on `Rack::Sendfile` middleware in your API:
+
+```ruby
+class API < Grape::API
+  use Rack::Sendfile
+end
+```
+
 ## Authentication
 
 ### Basic and Digest Auth
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [#1252](https://github.com/ruby-grape/grape/pull/1252): Allow default to be a subset or equal to allowed values without raising IncompatibleOptionValues - [@jeradphelps](https://github.com/jeradphelps).
 * [#1255](https://github.com/ruby-grape/grape/pull/1255): Allow param type definition in `route_param` - [@namusyaka](https://github.com/namusyaka).
 * [#1257](https://github.com/ruby-grape/grape/pull/1257): Allow Proc, Symbol or String in `rescue_from with: ...` - [@namusyaka](https://github.com/namusyaka).
+* [#1280](https://github.com/ruby-grape/grape/pull/1280): Support `Rack::Sendfile` middleware - [@lfidnl](https://github.com/lfidnl).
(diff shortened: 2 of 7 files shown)


dblock:
Interesting. Lets start with these:

* The test should use `Rack::Sendfile`, otherwise it's just expecting some contract to be fulfilled.
* Fix build, add to CHANGELOG.

lfidnl:
Hi! Sorry, for long response.

I wrote test `spec/grape/middleware/formatter_spec.rb:286` because condition with `Grape::Util::FileResponse` wasn't covered. Additionaly, as I understood, I should write some integration test with using `Rake::Sendfile` middleware, right?

dblock:
Yes, please and thank you @lfidnl. I am glad this gets cleaned up and tested.

lfidnl:
@dblock, I've added test, update CHANGELOG and fix build. I hope, I done everything well :)

README.md
```diff
@@ -2328,6 +2328,21 @@ class API < Grape::API
 end
 ```
 
+For using Rack::Sendfile middleware, specify `to_path` method in your
```
dblock: But this is not true, no? The code change makes it that I don't have to worry about file streamers or any other thing, I can just plop `Rack::Sendfile` in there and it will "just work", right? Maybe just say that this is compatible with Rack::Sendfile and add a note on what that does?

dblock:
Tests are good. See my above note on README, squash commits please.

README.md
lfidnl: @dblock , thanks for comment! I mean that agrument of `file` dsl method should respond to `to_path` method (of couse `use Rack::Sendfile` shoule be applied in API class).
So it can be `FileSteamer` with realized `to_path` method or other object with `to_path` method. I uses "file streamer" notion for supporting article in the same style. 
I think that adding information about `Rack::Sendfile` will be overkill. I can replace information with this:

For using `Rack::Sendfile` middleware, specify `to_path` method in your file streamer class which returns path of served file:

```ruby
class FileStreamer
  # ...

  def to_path
    @file_path
  end

  # ...
end
```

Note: don't forget turn on `Rack::Sendfile` middleware in your API:

```ruby
class API < Grape::API
  use Rack::Sendfile
end
```

If you want to just add `Rack::Sendfile` and I starts to work see my comment on below :)

lfidnl:
I think we must rewrite Grape's api for serving file like in Rails framework:

```ruby
get '/file' do
  file '/path/to/file'
end
```

`file` method will create "file streamer" object and if `Rack::Sendfile` middleware turns on, it will works throurg middleware, otherwise it will stream file through Grape.
It will simplify using of Grape.

Lets close this pull request and create a new issue, or I'll try to provide propose with Pull Request.

dblock:
:+1: on extending `file` to take a path

README.md
dblock: Ok I get it. I think it should look like the text above, which is _If you want a file-like object to be streamed using Rack::Chunked, use stream._

Maybe 

```
If you want to take advantage of Rack::Sendfile, which intercepts responses whose body is being served from a file and replaces it with a server specific X-Sendfile header, ...
```

Update it as you see fit and I'll merge.

lfidnl: Ok. I'll update readme during for couple days

lfidnl:
README updated

dblock:
Merged, thanks.

PR #1281, opened 2016-02-21
remove API class @route_set instance variable

totothink:


dblock:
This is great, merging. Also I think https://github.com/ruby-grape/grape/blob/7d1640ec6e0dfc050ca1850cdd88ee8ef0578405/lib/grape/endpoint.rb#L118 should be `.each(&)` and not `map`, maybe you want to fix that too?

totothink:
@dblock ok

PR #1282, opened 2016-02-21
Fix specs circular dependency

304:
The whole test suit is green but specs below are failing if you execute them one by one:
```
rspec spec/grape/api_spec.rb
rspec spec/grape/entity_spec.rb
rspec spec/grape/middleware/formatter_spec.rb
```

The reason is `require 'active_support/core_ext/hash'` which is defined
in exception_spec.rb. It hides the real problem that Grape does not
require active_support library which provides hash to xml convertation.


dblock:
:+1:

PR #1283, opened 2016-02-21
Fix 500 error for xml format when method is not allowed

304:
Problem:

All requests to Not Allowed methods are returning 500 error instead of
405 when the format is xml.

Cause:

Route definitions for not allowed methods are returning an empty string
as a response body. But the empty string fails to convert to xml and
raises InvalidFormatter error.

Solution:

Define an exception for "method is not allowed" case.
Return an error message "405 Not Allowed" instead of empty string.


dblock:
This is much better. Can you squash these two and rebase please?

304:
@dblock, done.

dblock:
Merged thanks.

dblock:
This introduced a behavior change that I am documenting in UPGRADING, because `Grape::Exceptions::MethodNotAllowed` is now raised the exception is being rescued in `rescue_from :all` and therefore needs to be handled explicitly. I think the following restores old behavior:

```ruby
        rescue_from Grape::Exceptions::Base do |e|
          error! e.message, e.status, e.headers
        end
```

dblock:
https://github.com/ruby-grape/grape/commit/252bfd27c320466ec3c0751812cf44245e97e5de

Issue #1265, opened 2016-02-01
Communicate `#after` hook errors in Middleware


rngtng:
As of https://github.com/ruby-grape/grape/pull/1240 the `#after` hook is called on middlewares even when exception is raised. This breaks middleware which assume `response`, (`@app_response`) is set. (e.g. [grape_logging](https://github.com/aserafin/grape_logging)). 

Firstly, any middleware should be migrated to handle this case, secondly grape should give a warning at least, to help debugging and prevent the application from hard failure.

See discussion at: https://github.com/ruby-grape/grape/pull/1240#issuecomment-177977684

dblock:
:+1:

jpaas:
I am also running into this issue. My logging middleware is unable to log the response in the after hook.

gregormelhorn:
I am currently trying to fix this issue. One thing I stumbled upon is that the formatter also uses error handling to communicate issues in the after callback.

Fortunately, from what I can see now, catch(:error) and rethrowing while swallowing any StandardError from the after callback would fix the issue - but it also alters the behaviour of the formatter. I am not sure if this should really be implemented I have to say.

I would vote for a setting to switch on and off the "rescue StandardError behaviour" in the after callback (while still keeping and rethrowing any explicitly thrown error). 

What do you think?

gregormelhorn:
The more I think about it, the more it would make  sense to add a backwards compatibility switch (do not run after callbacks on error) instead of magically catching exceptions...

dblock:
@gregormelhorn Maybe. That seems to undo the change we made, I'd rather not have both behaviors and issue a warning as described in this issue. But do try some code and show us what happens before/after.

gregormelhorn:
The changes mean that any StandardError thrown gets transformed into a warning.

It also means that catching standard exceptions inside formatters is not possible, so I removed the corresponding tests. After all, the code does not care if the issue is in the after callback of a logging middleware or the after callback of a formatter.

I am still not sure if making this fail hard with exceptions is not the better way - it will be very easy to miss coding errors.

rngtng:
Ok, true a hard fail makes sense, but for debugging reason It would help a lot to know that the error happened in middleware  so what about a warning  + re-raise the original error?

gregormelhorn:
ok, I changed it accordingly :+1:

dblock:
Closed via https://github.com/ruby-grape/grape/pull/1285.

rngtng:
nice, thx

PR #1285, opened 2016-02-22
After callback warning

gregormelhorn:
https://github.com/ruby-grape/grape/issues/1265

See discussion in the issue...

### DIFF
--- a/lib/grape/middleware/base.rb
+++ b/lib/grape/middleware/base.rb
@@ -29,7 +29,12 @@ module Grape
         begin
           @app_response = @app.call(@env)
         ensure
-          after_response = after
+          begin
+            after_response = after
+          rescue StandardError => e
+            warn "caught error of type #{e.class} in after callback inside #{self.class.name} : #{e.message}"
+            raise e
+          end
         end
 
         response = after_response || @app_response
--- a/spec/grape/middleware/base_spec.rb
+++ b/spec/grape/middleware/base_spec.rb
@@ -48,6 +48,25 @@ describe Grape::Middleware::Base do
     end
   end
 
+  context 'after callback with errors' do
+    it 'does not overwrite the application response' do
+      expect(subject.call({})).to eq([200, {}, 'Hi there.'])
+    end
+
+    context 'with patched warnings' do
+      before do
+        @warnings = warnings = []
+        allow_any_instance_of(Grape::Middleware::Base).to receive(:warn) { |m| warnings << m }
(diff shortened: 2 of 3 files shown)


lib/grape/middleware/base.rb
```diff
@@ -25,11 +25,20 @@ def call(env)
...
       def call!(env)
         @env = env
+        after_response, error = nil
         before
         begin
           @app_response = @app.call(@env)
         ensure
-          after_response = after
+          begin
+            error = catch(:error) do
```
rngtng: is the `catch(:error) do` rack /grape specific? Otherwise I don't see why to use this instead of re-raising the error?

gregormelhorn: you're right, throw is not rescued, so this is not necessary. I removed the extra catch.

rngtng:
:+1:  pls squash

gregormelhorn:
Done.

spec/grape/middleware/base_spec.rb
```diff
@@ -48,6 +48,31 @@
...
+  context 'after callback with errors' do
+    it 'does not overwrite the application response' do
+      expect(subject.call({})).to eq([200, {}, 'Hi there.'])
+    end
+
+    context 'with patched warnings' do
+      before do
+        @warnings = warnings = []
+        Grape::Middleware::Base.class_eval do
+          define_method(:warn) { |m| warnings << m }
```
dblock: This should be 

```ruby
allow_any_instance_of(Grape::Middleware::Base).to receive(:warn) { |m| warnings << m }
```

dblock:
Does need a CHANGELOG entry too, please.

gregormelhorn:
Alright, done :+1:  I really should have thought of `allow_any_instance_of`, thanks for the hint!

spec/grape/middleware/base_spec.rb
```diff
@@ -48,6 +48,25 @@
...
+    context 'with patched warnings' do
+      before do
+        @warnings = warnings = []
+        allow_any_instance_of(Grape::Middleware::Base).to receive(:warn) { |m| warnings << m }
+        allow(subject).to receive(:after).and_raise(StandardError)
+      end
+
+      it 'does show a warning' do
+        expect { subject.call({}) }.to raise_error(StandardError)
+        expect(@warnings).not_to be_empty
```
dblock: No need to change this, but it reads more English to say `to_not` :)

dblock:
Merged.

PR #1286, opened 2016-02-22
clean Endpoint

totothink:


dblock:
Nice and clean work @totothink!

totothink:
@dblock thx!

PR #1287, opened 2016-02-23
clean DSL module

totothink:


dblock:
:+1:

totothink:
https://github.com/totothink/grape/blob/6a49d6c2086884770f9a322fa01d4fb3fd14c256/lib/grape/dsl/configuration.rb
```ruby 
      # Merge multiple layers of settings into one hash.
       def stacked_hash_to_hash(settings)
          return if settings.blank?
          settings.each_with_object({}) { |value, result| result.deep_merge!(value) }
        end
```

stacked_hash_to_hash method like procedural programming. It should serve as settings object method，you think?

dblock:
@totothink Maybe ;)

PR #1292, opened 2016-02-24
remove multi_json dependency

u2:
#1274

### DIFF
--- a/README.md
+++ b/README.md
@@ -2023,11 +2023,25 @@ end
 
 Built-in formatters are the following.
 
-* `:json`: use object's `to_json` when available, otherwise call `MultiJson.dump`
+* `:json`: use object's `to_json` when available, otherwise call `JSON.dump`
 * `:xml`: use object's `to_xml` when available, usually via `MultiXml`, otherwise call `to_s`
 * `:txt`: use object's `to_txt` when available, otherwise `to_s`
 * `:serializable_hash`: use object's `serializable_hash` when available, otherwise fallback to `:json`
 * `:binary`: data will be returned "as is"
+
+Grape now not longer use MultiJson as JSON processor.If you want to continue to use MultiJson, you can use it by below
+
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -2023,12 +2023,26 @@ end
...
 Built-in formatters are the following.
 
-* `:json`: use object's `to_json` when available, otherwise call `MultiJson.dump`
+* `:json`: use object's `to_json` when available, otherwise call `JSON.dump`
 * `:xml`: use object's `to_xml` when available, usually via `MultiXml`, otherwise call `to_s`
 * `:txt`: use object's `to_txt` when available, otherwise `to_s`
 * `:serializable_hash`: use object's `serializable_hash` when available, otherwise fallback to `:json`
 * `:binary`: data will be returned "as is"
 
+Grape now not longer use MultiJson as JSON processor.If you want to continue to use MultiJson, you can use it by below
```
dblock: This should be moved to UPGRADING. Some minor "better" English: "Grape no longer uses MultiJson as default JSON processor. To use MultiJson add `multi_json` to your Gemfile and set `config.json_processor`.

dblock: README should contain an explanation of how to switch JSON processors though.

dblock:
First, thanks for keeping at it. It's an important and sensitive change given that it will affect everybody.

I don't like the global configuration, that prevents you from shipping two versions of the API where old uses multi_json and new uses json, basically we don't provide a path for deprecating an old JSON processor, we force you to remove it.

I think what we want is a per-api configuration, like `format :json, processor: MultiJson`. 

I am not against another global option, but I don't think it should be Grape global, maybe just formatter global, like `Grape::Formatter::JSON.configure do ...`. 

What do you think?

dblock:
Replacing in favor of https://github.com/ruby-grape/grape/pull/1623.

Issue #621, opened 2014-04-07
Allow Setting Custom Error Message with :requires 


jackdesert:
Would you mind allowing a custom error message to be set through the 'requires' method? For example, I would like to set a custom error message like this:

```
params do
    requires :name, regexp: /^[a-z]+$/, message: "Name may only contain letters a-z"
end
```

But this doesn't work. And creating a custom validator just to set a custom error message seems like a lot of overhead.

MSadir:
Hi, has anyone taken this on?

dm1try:
@MSadir pretty sure, nobody

If anyone will try to implement this
do not forget about DSL for multiple validations on the single param
```ruby
params do
    optional :name, values:      { value: 1..10, message: 'not in range from 1 to 10' }
                    allow_blank: { value: false, message: 'must present' }
end

dnesteryuk:
@dblock This one is [already implemented](https://github.com/ruby-grape/grape#custom-validation-messages).

PR #1295, opened 2016-02-26
Custom Validation error messages

railsmith:
Provides settings for custom validation error messages. Motivated by requests to build this feature on here [Grape Custom Validation error message](http://stackoverflow.com/questions/35454925/ruby-grape-custom-validation-error-message/35455246) and here #621

If custom error message options are not provided error messages fall backs to defaults specified in `grape/locale/en.yml`. If you want i18n for your custom error messages passing locale symbol keys will translate the messages.

**presence, allow_blank, values, regexp**

    params do
       requires :name, values: { value: 1..10, message: 'not in range from 1 to 10' },
                                allow_blank: { value: false, message: 'cannot be blank' },
                                regexp: { value: /^[a-z]+$/, message: 'format is invalid' }, 
                                message: 'is required'
    end

**all_or_none_of**
    
    params do
        optional :beer
        optional :wine
        optional :juice
        all_or_none_of :beer, :wine, :juice, message: "all params are required or none is required"
    end

**mutually_exclusive**

    params do
        optional :beer
        optional :wine
        optional :juice
        mutually_exclusive :beer, :wine, :juice, message: "are mutually exclusive cannot pass both params"
     end

**exactly_one_of**

     params do
         optional :beer
         optional :wine
         optional :juice
         exactly_one_of :beer, :wine, :juice, message: {exactly_one: "are missing, exactly one parameter is required", mutual_exclusion: "are mutually exclusive, exactly one parameter is required"}
     end

**at_least_one_of**

    params do
         optional :beer
         optional :wine
         optional :juice
         at_least_one_of :beer, :wine, :juice, message: "are missing, please specify at least one param"
     end

**Coerce**

     params do
        requires :int, type: {value: Integer, message: "type cast is invalid" }
      end

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * [#1255](https://github.com/ruby-grape/grape/pull/1255): Allow param type definition in `route_param` - [@namusyaka](https://github.com/namusyaka).
 * [#1257](https://github.com/ruby-grape/grape/pull/1257): Allow Proc, Symbol or String in `rescue_from with: ...` - [@namusyaka](https://github.com/namusyaka).
 * [#1285](https://github.com/ruby-grape/grape/pull/1285): Add a warning for errors appearing in `after` callbacks - [@gregormelhorn](https://github.com/gregormelhorn).
-* Your contribution here.
+* [#1295](https://github.com/ruby-grape/grape/pull/1295): Add custom validation messages for parameter exceptions - [@railsmith](https://github.com/railsmith).
 
 #### Fixes
 
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@
   - [Custom Validators](#custom-validators)
   - [Validation Errors](#validation-errors)
   - [I18n](#i18n)
+  - [Custom Validation Messages](#custom-validation-messages)
 - [Headers](#headers)
 - [Routes](#routes)
 - [Helpers](#helpers)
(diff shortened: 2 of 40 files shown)


dblock:
Thanks! Looking forward to code reviewing this - can you please rebase/squash it first? There're a lot of unrelated changes here to grok.

railsmith:
I have squashed it. There are still one or two unrelated changes which have penetrated while rebasing. Hope it helps to review the code.

dblock:
Well those two changes shouldn't be there :) This should fix it:

```
git checkout master
git pull upstream master
git checkout custom-messages
git rebase master
git push origin custom-messages -f
```

dblock:
The change itself looks OK to me. It needs a README update and CHANGELOG. I am glad to see a very full test coverage too, wonder if there's a way to simplify the implementation a bit so that you don't have to copy paste the thing again and again. Like maybe enumerating something in the messages YML?

railsmith:
Originally I had a similar idea in my mind. We can certainly enumerate messages from a YML file and extract the key based on the current route description. We can first check for the message in the `message` settings DSL if it doesn't exist there then we can check in the `messages.yml` and if it doesn't exist there too we can safely fall back to message keys in `en.yml`. What do you propose?
     
    # messages.yml
    
    grape:
       errors:
          messages:
             endpoint:
                 /custom_messages:
                     verb:
                         get:
                            email:
                                presence: 'is required'
                                blank: 'has no value'
                 /:
                     verb:
                         post:
                              number:
                                 values: 'not in range 1 to 10'

dblock:
I don't have any great suggestions wrt your question :)

Please update README and CHANGELOG and this is good to go.

railsmith:
Thanks! I am done with this.

dblock:
Merged via https://github.com/ruby-grape/grape/commit/e6a6c9c86b2e89004497f83f8a5c0007654d6150 (I re-added a CHANGELOG line "Your contribution here"). Thanks!

dblock:
This bit me on a real life project, returning empty messages when a `Grape::Exceptions::Validation` was raised explicitly. I added a note in UPGRADING in https://github.com/ruby-grape/grape/commit/77ec1cd509dd27395ce9a3f6ecc37a422ee544f9 to rename `message_key` to `message`. We could have deprecated `message_key` first with a warning, but I think it's OK to get rid of it here since it's rare to be implementing your own validators and it's easily fixed.

railsmith:
The spec for Custom Validators with an explicit `message_key` was indeed failing and I updated the spec by changing `message_key` to `message` but forgot to deprecate `message_key` and update the README for Custom Validators as I thought it was trivial but it looks OK and updated now.

PR #1296, opened 2016-02-27
Don't include extension in capture if format is specified

namusyaka:
ref #1276 
/cc @dblock


dblock:
I merged this. We do need another spec that without `format :json` responds to this API and makes sure the ID strips out the extension.

namusyaka:
Okay, wait a sec.

PR #1298, opened 2016-02-28
clean error_formatter

totothink:


dblock:
I am a little concerned by the removal of autoload. Half the time you don't want all these formatters, no? Is it possible to put them back?

totothink:
yeah, you're right.

dblock:
Ok, this is much better. Thanks.

PR #1304, opened 2016-03-04
Params validation when Array has different objects

ne0h12:
Hello all! 
I have a problem parameters validation, when :items Array has different objects containing optional Hash. The my test 'handles validation within arrays with different objects' does failed. When will be changed :nested field from requires to optional my test will be passed, but i want require this field :nested because of parent field is optional. I think this is bug of grape. Help me guys. Thanks.

### DIFF
--- a/spec/grape/validations_spec.rb
+++ b/spec/grape/validations_spec.rb
@@ -770,6 +770,39 @@ describe Grape::Validations do
       end
     end
 
+    context 'nested optional Array blocks with optional Hash blocks' do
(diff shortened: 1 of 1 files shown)


dnesteryuk:
@ne0h12 I got good news for you :wink: I run your test suit with Grape 1.3.2, it passed :tada: The issue doesn't exist anymore.

ne0h12:
@dnesteryuk Thank you!

PR #1306, opened 2016-03-07
Add example for admin only custom validation to both readme and spec

ashkan18:
# Problem
Wanted to be able to have custom validation based on request and parameter. As an example wanted to be able to have _Admin Only_ fields that are only available to endpoint if user is admin.

# Solution
With new `validate(request)` method added for custom validation (https://github.com/ruby-grape/grape/pull/1242) we can now support these kinds of attributes. I only added a new sample in spec and also a sample in README.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 #### Features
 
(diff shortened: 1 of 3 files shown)


ashkan18:
closing this and opening https://github.com/ruby-grape/grape/pull/1308 after resolving conflicts.

PR #1308, opened 2016-03-07
Added a sample spec and example in README.md for admin only fields

ashkan18:
# Problem
Wanted to be able to have custom validation based on request and parameter. As an example wanted to be able to have _Admin Only_ fields that are only available to endpoint if user is admin.

# Solution
With new `validate(request)` method added for custom validation (https://github.com/ruby-grape/grape/pull/1242) we can now support these kinds of attributes. I only added a new sample in spec and also a sample in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1198,6 +1198,35 @@ params do
 end
 ```
 
+You can also create custom validation that use request to validate the attribute. For example if you want to have parameters that are available to only admins, you can do the following.
+
+```ruby
+class Admin < Grape::Validations::Base
+  def validate(request)
+    # return if the param we are checking was not in request
+    # @attrs is a list containing the attribute we are currently validating
+    # in our sample case this method once will get called with 
+    # @attrs being [:admin_field] and once with @attrs being [:admin_false_field]
+    return unless request.params.key? @attrs.first
+    # check if admin flag is set to true
+    return unless @option
+    # check if user is admin or not
+    # as an example get a token from request and check if it's admin or not
+    fail Grape::Exceptions::Validation, params: @attrs, message: 'Can not set admin-only field.' unless request.headers['X-Access-Token'] == 'admin'
+  end
+end
+```
+
+And use it in your endpoint definition as:
+
+```ruby
+params do
+  optional :admin_field, type: String, admin: true
+  optional :non_admin_field, type: String
+  optional :admin_false_field, type: String, admin: false
+end
+```
+
 ### Validation Errors
 
 Validation and coercion errors are collected and an exception of type `Grape::Exceptions::ValidationErrors` is raised. If the exception goes uncaught it will respond with a status of 400 and an error message. The validation errors are grouped by parameter name and can be accessed via `Grape::Exceptions::ValidationErrors#errors`.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 #### Features
 
+* [#1308](https://github.com/ruby-grape/grape/pull/1308): Add admin only custom validation example to doc and spec - [@ashkan18](https://github.com/ashkan18)
```
dblock: Missing a period at the end. But we don't really need a CHANGELOG line for docs + spec really, though we absolutely do appreciate the contribution!

README.md
```diff
@@ -1198,6 +1198,28 @@ params do
 end
 ```
 
+You can also create custom validation that use request to validate the attribute. For example if you want to have parameters that are available to only admins, you can do this
+```ruby
```
dblock: Add a space above. Change this to a full phrase, eg " admins, you can do the following."

README.md
```diff
@@ -1198,6 +1198,28 @@ params do
...
+class Admin < Grape::Validations::Base
+  def validate(request)
+    # return if the param we are checking was not in request
+    # @attrs is a list containing the attribute we are currently validating
+    return unless request.params.key? @attrs.first
+    # check if admin flag is set to true
+    return unless @option
+    # check if user is admin or not
+    # as an example get a token from request and check if it's admin or not
+    fail Grape::Exceptions::Validation, params: @attrs, message: 'Can not set Admin only field.' unless request.headers['X-Access-Token'] == 'admin'
```
dblock: Make Admin lowercase, I think it should be "admin-only" with a dash to be correct.

README.md
```diff
@@ -1198,6 +1198,28 @@ params do
...
+  end
+end
+```
+And use it in your endpoint definition as:
+```ruby
```
dblock: Make it a full API example so it's clear that this is inside `params`.

dblock:
Thanks. I only have minor English suggestions. Please also squash your commits.

README.md
```diff
@@ -1198,6 +1198,28 @@ params do
 end
 ```
 
+You can also create custom validation that use request to validate the attribute. For example if you want to have parameters that are available to only admins, you can do this
+```ruby
+class Admin < Grape::Validations::Base
+  def validate(request)
+    # return if the param we are checking was not in request
+    # @attrs is a list containing the attribute we are currently validating
```
dblock: This is unclear. Is it an attribute? A list?

spec/grape/api/custom_validations_spec.rb
```diff
@@ -113,4 +113,71 @@ def app
       expect(last_response.body).to eq 'text is missing'
     end
   end
+
+  context 'using a custome request/param validator' do
```
dblock: customE -> custom (typo)

CHANGELOG.md
ashkan18: Yes I wasn't sure about changelog here too, thanks :)

dblock:
Merged, thank you.

PR #1314, opened 2016-03-10
Test for checking multiple items in an array

treacher:
Should this test be failing? It currently is I would of thought this should pass if array indexes was working as intended. Thoughts?


dblock:
On Travis-CI this is failing for some other reason than the one you expect no?

galkon:
Yeah it's failing because it's using the new version of rake to run the tests.

dblock:
Lock it at `~> 10`.

galkon:
Yeah I'll do that is a seperate PR as it will be affecting any new builds.

galkon:
https://github.com/ruby-grape/grape/pull/1315

dblock:
:+1:

PR #1320, opened 2016-03-14
Spec for #1276.

dblock:
cc: @namusyaka


namusyaka:
Great! Thanks for your spec! I've just fixed it!

PR #1321, opened 2016-03-14
Serve files without using FileStreamer-like object

lfidnl:
Hello!

I've implemented serving file without using FileStremer-like object. Now user can use only path to file and Grape send file directly or will use `Rack::Sendfile` middleware if it turned on.

I've saved backward compatibility with FileStreamer-like object with displaying deprecation warning.


.rubocop_todo.yml
```diff
@@ -42,6 +42,8 @@ Metrics/MethodLength:
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
   Max: 272
+  Exclude:
```
dblock: You might want to just run `rubocop --auto-gen-config` for this.

lib/grape/serve_file/file_body.rb
```diff
@@ -0,0 +1,32 @@
...
+      # Need for Rack::Sendfile middleware
+      #
+      # @return [String]
+      def to_path
+        path
+      end
+
+      def each
+        File.open(path, 'rb') do |file|
+          while (chunk = file.read(16_384))
```
dblock: Should probably be extracted as a constant, just in case someone wants to update it at their runtime.

dblock:
Looks great. Needs a CHANGELOG entry please.

lfidnl:
Thanks for response! Ohhh sorry, forgot about CHANGELOG

lib/grape/serve_file/file_body.rb
lfidnl: As I understood, you mean create constant for 16384 number, is it?

dblock: Yep

lfidnl:
Updated according to your comments

lib/grape/dsl/inside_route.rb
```diff
@@ -191,8 +191,12 @@ def body(value = nil)
       #
       #   GET /file # => "contents of file"
       def file(value = nil)
-        if value
-          @file = Grape::Util::FileResponse.new(value)
+        if value.is_a?(String)
+          file_body = Grape::ServeFile::FileBody.new(value)
+          @file = Grape::ServeFile::FileResponse.new(file_body)
+        elsif !value.is_a?(NilClass)
+          warn '[DEPRECATION] Argument as FileStreamer-like object will be deprecated. Use path to file instead'
```
dblock: In other places we say "is deprecated", not "will be" and usually make this a full sentence with a period in the end. Like this: https://github.com/ruby-grape/grape/blob/4032ea8f7747483b84663c260d4aba0491bd7777/lib/grape/dsl/desc.rb#L50.

dblock:
One last thing this needs is an entry in UPGRADING.md, please.

lib/grape/dsl/inside_route.rb
lfidnl: I mean that support of argument as FileSteam-like object will be removed from Grape in next version after 0.15.1 or 0.16.0. After releasing 0.15.1 or 0.16.0 I would like to remove it at all. Sorry for my English :)
OK, I'll replace it with:

```
[DEPRECATION] Argument as FileStreamer-like object is deprecated. Use path to file instead.
```

dblock: Thanks. If you're rubbing the bext version the old code is already deprecated but it does work with a warning. I think present tense is correct :) thanks for making the changes!

dblock:
Last thought, don't feel that strongly about it, but maybe ServeFile should just be File?

lfidnl:
I chose `ServeFile` to avoid confusion with the regular ruby `File` class. I know that we have namespaces and etc, but I think it will be better :)

dblock:
:+1:

UPGRADING.md
```diff
@@ -23,6 +23,12 @@ TwitterAPI::routes[0].settings[:custom] # => { key: 'value' }
 TwitterAPI::routes[0].request_method # => 'GET'
 ```
 
+#### `file` method accepts path to file
+
+Now to serve files via Grape just pass the path to the file. Functionality with FileStreamer-like objects is deprecated.
+
+Please, replace your FileStreamer-like objects with paths of served files.
```
dblock: Care to provide an example here? Thanks.

lfidnl: No problem :)

dblock:
Merged, thanks.

PR #1326, opened 2016-03-16
Add test for OPTIONS and HEAD requests with catch-all

namusyaka:
ref #1057
I've cherry-picked @ekampp's commit and fixed a little bit.
Thoughts?

### DIFF
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -3047,7 +3047,7 @@ XML
         expect(last_response.status).to eq(200)
         expect(last_response.body).to eq('v2')
         options '/v2/hello'
-        expect(last_response.status).to eq(200)
+        expect(last_response.status).to eq(204)
         expect(last_response.body).to be_blank
         head '/v2/hello'
         expect(last_response.status).to eq(200)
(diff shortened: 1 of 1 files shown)


ekampp:
[OPTIONS requests should return 200 OK](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)

namusyaka:
Well, `OPTIONS` seems not to have the restriction of status codes.
Our default `OPTIONS` routes are defined at [this line](https://github.com/ruby-grape/grape/blob/master/lib/grape/api.rb#L180-L186), and this means that a response of the route will be empty. 
So I guess `204` status is reasonable, right?

ekampp:
204 is reasonable for an empty response. But I'm reading that it should return 200 OK. But either way getting this to pass would be an improvement :+1:

dblock:
Very good, the passing spec is a merge. Lets close the related bug and review whether we want to change the status code here (open new issues as appropriate).

dblock:
This possibly needs a CHANGELOG entry since I believe this was a known bug now fixed, @namusyaka please.

Issue #1329, opened 2016-03-17
uninitialized constant Grape::Formatter::Base (NameError)


scalp42:
I'm not sure where I should open the issue (see linked issue: https://github.com/ruby-grape/grape/issues/new) but running into the following with `berkshelf-api` (depending on `grape-msgpack`):

```
/opt/berkshelf-api # berks-api /opt/berkshelf-api/config.json
/usr/lib/ruby/gems/2.2.0/gems/grape-msgpack-0.1.2/lib/grape/msgpack.rb:39:in `<top (required)>': uninitialized constant Grape::Formatter::Base (NameError)
	from /usr/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/lib/ruby/gems/2.2.0/gems/grape-msgpack-0.1.2/lib/grape-msgpack.rb:1:in `<top (required)>'
	from /usr/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/lib/ruby/gems/2.2.0/gems/berkshelf-api-2.1.3/lib/berkshelf/api/endpoint.rb:2:in `<top (required)>'
	from /usr/lib/ruby/gems/2.2.0/gems/berkshelf-api-2.1.3/lib/berkshelf/api.rb:23:in `require_relative'
	from /usr/lib/ruby/gems/2.2.0/gems/berkshelf-api-2.1.3/lib/berkshelf/api.rb:23:in `<module:API>'
	from /usr/lib/ruby/gems/2.2.0/gems/berkshelf-api-2.1.3/lib/berkshelf/api.rb:10:in `<module:Berkshelf>'
	from /usr/lib/ruby/gems/2.2.0/gems/berkshelf-api-2.1.3/lib/berkshelf/api.rb:9:in `<top (required)>'
	from /usr/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/lib/ruby/gems/2.2.0/gems/berkshelf-api-2.1.3/lib/berkshelf/api/srv_ctl.rb:72:in `start'
[...]

PR #1330, opened 2016-03-17
Fix built-in parsers and formatters

namusyaka:
ref 93e699829f7b4a2fe152370eccdf9b2712f4329f #1329 
I'd like to use grape-msgpack without tricky hacking. Currently, our implementation does not support inheritable interface, so I fixed a few points.

Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * [#1325](https://github.com/ruby-grape/grape/pull/1325): Params: Fix coerce_with helper with Array types - [@ngonzalez](https://github.com/ngonzalez).
 * [#1326](https://github.com/ruby-grape/grape/pull/1326): Fix wrong behavior for OPTIONS and HEAD requests with catch-all - [@ekampp](https://github.com/ekampp), [@namusyaka](https://github.com/namusyaka).
+* [#1330](https://github.com/ruby-grape/grape/pull/1330): Fix built-in parsers and formatters - [@namusyaka](https://github.com/namusyaka).
 
 0.15.0 (3/8/2016)
 =================
--- a/lib/grape/error_formatter.rb
+++ b/lib/grape/error_formatter.rb
@@ -1,8 +1,11 @@
+require 'grape/util/registrable'
+
 module Grape
   module ErrorFormatter
+    extend Registrable
     class << self
       def builtin_formatters
-        {
+        @builtin_formatters ||= {
           serializable_hash: Grape::ErrorFormatter::Json,
           json: Grape::ErrorFormatter::Json,
           jsonapi: Grape::ErrorFormatter::Json,
@@ -12,7 +15,7 @@ module Grape
       end
 
       def formatters(options)
-        builtin_formatters.merge(options[:error_formatters] || {})
+        builtin_formatters.merge(default_elements).merge(options[:error_formatters] || {})
       end
 
       def formatter_for(api_format, options = {})
--- a/lib/grape/formatter.rb
+++ b/lib/grape/formatter.rb
@@ -1,8 +1,12 @@
+require 'grape/util/registrable'
+
 module Grape
   module Formatter
+    extend Registrable
+
     class << self
       def builtin_formmaters
-        {
+        @builtin_formatters ||= {
           json: Grape::Formatter::Json,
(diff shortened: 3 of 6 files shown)


dblock:
I see how a require fixes this, but we now load formatters we don't use. Why doesn't autoload work with grape-msgpack?

namusyaka:
Nope, I removed `autoload`s to declare `FORMATTERS` and `PARSERS` constants.

dblock:
Sorry, I am a little slow, and clearly I am missing some Ruby-fu understanding :) 

Why doesn't it work with grape-msgpack?

namusyaka:
At first, `grape-msgpack` didn't consider [the commit](https://github.com/ruby-grape/grape/commit/93e699829f7b4a2fe152370eccdf9b2712f4329f).
However, something like [this line](https://github.com/ruby-grape/grape/blob/93e699829f7b4a2fe152370eccdf9b2712f4329f/lib/grape/error_formatter.rb#L5-L11) has broke [inheritance interface](https://github.com/rosylilly/grape-msgpack/blob/master/lib/grape/msgpack.rb#L44).
Our provided [`options`](https://github.com/ruby-grape/grape/blob/93e699829f7b4a2fe152370eccdf9b2712f4329f/lib/grape/error_formatter.rb#L15) cannot be inherited, so if we want to implement customized grape formatter, we must require the tricky code.

I think we should be able to provide more developer-friendly interface.
If this PR is merged, maybe `grape-msgpack` can resolve by [a little bit changes](https://github.com/rosylilly/grape-msgpack/compare/master...namusyaka:remove-base-constants).

dblock:
What if we just added a global interface to register formatters, instead of inheriting anything? Like so:

```ruby
Grape.configure do |config|
   config.formatters.add
end
```

dblock:
Or

```ruby
class MyFormatter << Grape::Formatter
end

Grape::Formatter.register MyFormatter, ...
```

namusyaka:
Yeah, it's so cool and developer-friendly. :)
Btw, should we contain the configuration feature in 0.16.0?
Also, can this PR be used as a 0.15-backported patch?


2016年3月17日木曜日、Daniel Doubrovkine (dB.) @dblockdotorg<
notifications@github.com>さんは書きました:

namusyaka:
Hmm, right. It's reasonable.

2016年3月17日木曜日、Daniel Doubrovkine (dB.) @dblockdotorg<
notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>>さんは書きました:

dblock:
I don't think we should backport anything and just move forward. If you want to give one of the above a short, go for it! I'd rather not remove autoload. I prefer the second version btw, not the global configuration one because that always bites us.

Vsandhya11:
I am planning to write micro service using ruby-grape, So I am thinking 0.15.0 is stable version to use. Plz lemme know your suggestions.

namusyaka:
I've tried to implement `register` keyword on those by using the second feature as base approach.
@dblock Could you confirm the changes?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * [#1325](https://github.com/ruby-grape/grape/pull/1325): Params: Fix coerce_with helper with Array types - [@ngonzalez](https://github.com/ngonzalez).
 * [#1326](https://github.com/ruby-grape/grape/pull/1326): Fix wrong behavior for OPTIONS and HEAD requests with catch-all - [@ekampp](https://github.com/ekampp), [@namusyaka](https://github.com/namusyaka).
+* [#1330](https://github.com/ruby-grape/grape/pull/1330): Fix built-in parsers and formatters - [@namusyaka](https://github.com/namusyaka).
```
dblock: It's not clear what's being fixed. The built in parsers and formatters weren't broken really.

lib/grape/error_formatter.rb
```diff
@@ -1,8 +1,11 @@
+require 'grape/util/registrable'
```
dblock: Autoload this?

dblock:
I'll merge this as is, feel free to make further changes later. Also I think we don't need to document this publicly for now.

dblock:
Add a similar spec for parsers.

PR #1336, opened 2016-03-23
Do not modify a Hash argument to error!

tjwp:
This change allows an immutable hash, for example a frozen constant, to be passed as the message to `error!`.

If a `:with` value needs to removed from the specified message then a new hash is created without that key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [#1325](https://github.com/ruby-grape/grape/pull/1325): Params: Fix coerce_with helper with Array types - [@ngonzalez](https://github.com/ngonzalez).
 * [#1326](https://github.com/ruby-grape/grape/pull/1326): Fix wrong behavior for OPTIONS and HEAD requests with catch-all - [@ekampp](https://github.com/ekampp), [@namusyaka](https://github.com/namusyaka).
 * [#1330](https://github.com/ruby-grape/grape/pull/1330): Add `register` keyword for adding customized parsers and formatters - [@namusyaka](https://github.com/namusyaka).
+* [#1336](https://github.com/ruby-grape/grape/pull/1336): Do not modify Hash argument to `error!` - [@tjwp](https://github.com/tjwp).
 
 0.15.0 (3/8/2016)
 =================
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [#1325](https://github.com/ruby-grape/grape/pull/1325): Params: Fix coerce_with helper with Array types - [@ngonzalez](https://github.com/ngonzalez).
 * [#1326](https://github.com/ruby-grape/grape/pull/1326): Fix wrong behavior for OPTIONS and HEAD requests with catch-all - [@ekampp](https://github.com/ekampp), [@namusyaka](https://github.com/namusyaka).
 * [#1330](https://github.com/ruby-grape/grape/pull/1330): Add `register` keyword for adding customized parsers and formatters - [@namusyaka](https://github.com/namusyaka).
+* [#1336](https://github.com/ruby-grape/grape/pull/1336): Do not modify Hash argument to `error!` - [@tjwp](https://github.com/tjwp)
```
dblock: Add a period to the end.

lib/grape/error_formatter/base.rb
```diff
@@ -3,9 +3,13 @@ module ErrorFormatter
     module Base
       def present(message, env)
         present_options = {}
-        present_options[:with] = message.delete(:with) if message.is_a?(Hash)
+        presented_message = message
+        if presented_message.is_a?(Hash) && presented_message.key?(:with)
+          present_options[:with] = presented_message[:with]
+          presented_message = presented_message.reject { |k, _| k == :with }
+        end
```
dblock: Seems a little convoluted, why don't you just `message.dup`? It's probably a good idea anyway because users may want to use the response from `present` and modify it further.

dblock:
This makes sense. See comments and please also squash your commits.

lib/grape/error_formatter/base.rb
tjwp: I did it this way to avoid creating the extra hash when it is unnecessary. But I think you're right, since the argument here may also be the result of this method, it's probably better to always dupe.

tjwp:
Updated to use `dup` and commits squashed.

dblock:
:+1:  merged

PR #1337, opened 2016-03-23
Update README.md

glaucocustodio:
When using something like `expose :updated_at {|instance, options| instance.updated_at.to_s}`, the following exception is raise:

```
SyntaxError:
/Users/x/app/api/entities/foo.rb:11: syntax error, unexpected '{', expecting keyword_end
expose :updated_at {|instance, options| instance.updated_at.to_s}
```
See: ruby-grape/grape-entity#12.


namusyaka:
@glaucocustodio Right, thanks!

Issue #1335, opened 2016-03-22
Grape::DSL::Helpers::ClassMethods#helpers modifies helper module


vsorlov:
Hi, `Grape::DSL::Helpers::ClassMethods#helpers` evaluates helper methods in context of a given helper module, that leads to problems if that helper module is used anywhere else.

Example

```ruby
helpers MyHelperModule do
  def resource
    ...
  end
end
```

now if I include `MyHelperModule` anywhere else, I'll have resource method, that I don't want to include

dblock:
But that's the whole point of `helpers`, no? What do you suggest?

vsorlov:
@dblock, you see, we want to add helper module to several endpoints and also we want to add unique `resource` method to each of these endpoints

it is possible with
```ruby
helpers do
  include MyHelperModule

  def resource
    ...
  end
end
```

but grape allows us to do it with
```ruby
helpers MyHelperModule do
  def resource
    ...
  end
end
```

and with that code `resource` method is added to MyHelperModule twice and the last added `resource` method is used in both endpoints

I suggest to always create anonymous module and include module passed in `new_mod` in that anonymous module. https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/helpers.rb#L31

I can create a request if you prefer.

dblock:
Ok that makes total sense @vsorlov. I wonder whether this is a feature through, in Ruby if you write `module MyHelperModule` you'd expect it to reopen a previous one. On the other hand, we could consider modules namespaces, and maybe what you suggest has less of a surprise. I think you should make a pull request either way.

PR #1338, opened 2016-03-25
fix Grape helpers modified given helper module

vsorlov:
fixes https://github.com/ruby-grape/grape/issues/1335


dblock:
I am torn whether we want this. If you think "this is a Ruby module", then in Ruby if you write module `MyHelperModule` you'd expect it to reopen a previous one. On the other hand, we could consider modules namespaces, and maybe what you suggest has less of a surprise. Please add your :+1: or :-1:!

Issue #1072, opened 2015-07-20
How grape recognizes a path?


vasilakisfil:
In rails there is Rails.application.routes.recognize_path: http://www.rubydoc.info/docs/rails/ActionDispatch/Routing/RouteSet#recognize_path-instance_method

Is there anything similar for Grape?

dblock:
Grape uses Rack::Mount, which is now yanked from Github. I actually would like to use this opportunity to figure out what the recommended replacement is and maybe find out a bit about the history of why it was yanked? /cc: @joshpeek

vasilakisfil:
anyone?

dblock:
@vasilakisfil The "how" part is rack-mount, and you can look at the source via `bundle show rack-mount` in any project that has a Grape dependency. But the question of what we should do with it remains, I now [asked rack-devel](https://groups.google.com/forum/?fromgroups#!topic/rack-devel/nNjuNIEZJis).

vasilakisfil:
I hate it when such vital gems don't have documentation......... and why on earth they removed it? it's free to have it somewhere on github...

After 2 hours looking on the rack-mount code, I couldn't figure out what object accepts recognize method of RouteSet class...

Any help is really appreciated.. I have a route string ('/api/v1/foobars') and some request options (like {method: :get} etc) and I need somehow to get the path from grape api.....

dm1try:
>After 2 hours looking on the rack-mount code, I couldn't figure out what object accepts recognize method of RouteSet class...

seems it's just an instance of Rack::Request(or Grape)

```ruby
@request_class = options.delete(:request_class) || Rack::Request
...
req = @request_class.new(env)
recognize(req) do |route, matches, params|
...
end
```

so you can prepare `env` using Rack::MockRequest
then build an instance of Rack::Request
then call `recognize`
```ruby
mocked_env = Rack::MockRequest.env_for('/api/v1/foobars', {:method => 'GET', :params => { foo: 'bar' }})
request = Rack::Request.new(mocked_env)
routes = route_set.recognize(request)
```
or smth similar

>why it was yanked?
really interesting..

vasilakisfil:
Thanks!

dblock:
Nice work @dm1try. Should Grape provide a helper? Or maybe an extension or a separate gem?

dm1try:
@dblock I honestly don't have a good use-case of using this feature in my flow(debugging? seems we need  to make `grape-console` first :) )
[...]

PR #1339, opened 2016-03-26
Implement Grape::API.recognize_path

namusyaka:
ref #1072 
I've tried to implement `Grape::DSL::InsideRoute#recognize_path`.
It returns value conforming the format like `[original pattern, expanded params]`.

This is a first step. I'd like to hear your opinions.
Thanks.

### DIFF
--- a/spec/grape/dsl/inside_route_spec.rb
+++ b/spec/grape/dsl/inside_route_spec.rb
@@ -15,348 +15,348 @@ module Grape
         end
       end
     end
+  end
+end
 
-    describe Endpoint do
-      subject { InsideRouteSpec::Dummy.new }
+describe Grape::Endpoint do
+  subject { Grape::DSL::InsideRouteSpec::Dummy.new }
 
-      describe '#version' do
-        it 'defaults to nil' do
-          expect(subject.version).to be nil
-        end
+  describe '#version' do
+    it 'defaults to nil' do
+      expect(subject.version).to be nil
+    end
 
-        it 'returns env[api.version]' do
-          subject.env['api.version'] = 'dummy'
-          expect(subject.version).to eq 'dummy'
-        end
+    it 'returns env[api.version]' do
+      subject.env['api.version'] = 'dummy'
+      expect(subject.version).to eq 'dummy'
+    end
+  end
+
+  describe '#error!' do
+    it 'throws :error' do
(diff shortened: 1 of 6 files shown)


lib/grape/endpoint.rb
```diff
@@ -120,6 +120,7 @@ def reset_routes!
     end
 
     def mount_in(router)
+      @router ||= router
```
dblock: We didn't address mounting an endpoint in multiple places, but eventually we will. So maybe we don't want `@router`?

spec/grape/dsl/inside_route_spec.rb
```diff
@@ -357,6 +357,33 @@ def initialize
...
+          env = Rack::MockRequest.env_for('/')
+          expect(subject.call(env)[2].body[0]).to eq(['/test/:id', { id: '1234', format: 'json' }].to_json)
+        end
+
+        describe 'returned params' do
+          it 'is able to access params by using string or symbol as a key' do
+            subject.get('/test/:id'){}
+            subject.get do
+              _, params = recognize_path('/test/1234')
+              "string: #{params['id']}, symbol: #{params[:id]}"
```
dblock: Just for the sake of sanity in the test, return json.

dblock:
It's on the right track, but shouldn't `recognize_path` return an endpoint?

lib/grape/endpoint.rb
namusyaka: Yeah, right. Removed.

namusyaka:
@dblock Now `recognize_path` returns an instance of `Grape::Endpoint`. How about this?

dblock:
Very good. I think the last thing we need is a README entry that explains how you can use `recognize_path` from the outside of an API, so given a `API < Grape::API`, calling `recognize_path` to get a endpoint that would match.

namusyaka:
>> I think the last thing we need is a README entry that explains how you can use recognize_path from the outside of an API, so given a API < Grape::API, calling recognize_path to get a endpoint that would match.

Hmm, makes sense. If I understand thing correctly, `recognize_path` should be implemented as a class method of `Grape::API`.
Maybe we can't call `recognize_path` outside of an API, without tricky code.
So I moved the method place from `InsideRoute` to `API`.
Could you please check this?

dblock:
Perfect, merging.

PR #1345, opened 2016-04-04
Move inclusion of namespace stackable middleware to after formatter

ridiculous:
The reason for moving the inclusion of user middleware to after the version and [specifically] the formatter middleware is because useful environment vars are set by them.

Any reason not to do this?


dblock:
I believe there're good reasons around when errors are rendered, see failed specs.

ridiculous:
Yikes ... okay

PR #1349, opened 2016-04-06
Update Readme with Mounting Multiple Versions

przbadu:
improve documentation following (http://code.dblock.org/2013/07/19/evolving-apis-using-grape-api-versioning.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@
   - [Header](#header)
   - [Accept-Version Header](#accept-version-header)
   - [Param](#param)
+  - [Mounting Multiple Versions](#mounting-multiple-versions)
 - [Describing Methods](#describing-methods)
 - [Parameters](#parameters)
   - [Declared](#declared)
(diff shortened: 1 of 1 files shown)


przbadu:
@dblock: I just locked version of grape-entity to `0.5.0` and made a PR. Your suggestion for https://github.com/ruby-grape/grape/issues/1350

dblock:
FYI I did that on master so we can have a build in https://github.com/ruby-grape/grape/pull/1353.

przbadu:
Is it already merged in master??

przbadu:
Great, thank you, let me pull it and merge it

przbadu:
@dblock : please check README and let me know for further improvements.

Thanks

dblock:
It doesn't quite read as documentation, more like a blog post or example. Specifically, I'm not in love with `curl` examples and I definitely don't think we need to say what it took to make it happen (the grape fix). There should be a way to reduce the code, too, and also needs to be `rubocop`-ed.

It's a good start though. I can take a stab at this later or you can try iterating on it.

README.md
```diff
@@ -25,6 +25,7 @@
   - [Header](#header)
   - [Accept-Version Header](#accept-version-header)
   - [Param](#param)
+  - [Mounting Multiple Versions](#mounting-multiple-versions)
```
dblock: I think this belongs in Mounting more than Versioning, because Versioning is all about kinds of versioning schemes.

README.md
```diff
@@ -392,6 +393,73 @@ version 'v1', using: :param, parameter: 'v'
...
+class API < Grape::API
+  mount V2::API
+  mount V1::API
+end
+```
+
+Request and Response
+--------------------
+
+    curl http://localhost:3000/api/v1/test
```
dblock: This assumes path versioning which is just a subset of the problem.

dblock:
I appreciate the effort here, but I think it's a bit far from reading like a doc and wasn't active after my comments. I'll close it for now, please feel free to reopen with an update.

PR #1351, opened 2016-04-06
Fix #1348: global functions called during mount.

### LINKED ISSUE #1348: Global function definitions pollute Grape::API scope on Ruby 2.3.0
Global function definitions pollute Grape::API scope on Ruby 2.3.0 

```ruby
require 'spec_helper'

def namespace
  raise 'should never be called'
end

describe Grape::API do
  subject do
    Class.new(Grape::API) do
      format :json
      get do
        { ok: true }
      end
    end
  end

  def app
    subject
  end

  context 'with Rake::DSL' do
    it 'works' do
      get '/'
      expect(last_response.status).to eq 200
    end
  end
end
```

This fails on Ruby 2.3.0 with

```
  1) Grape::API with Rake::DSL works
     Failure/Error: raise 'should never be called'
     
     RuntimeError:
       should never be called
     # ./spec/grape/integration/global_namespace_function_spec.rb:4:in `namespace'
     # ./lib/grape/router/attribute_translator.rb:29:in `block in to_h'
     # ./lib/grape/router/attribute_translator.rb:28:in `each'
     # ./lib/grape/router/attribute_translator.rb:28:in `each_with_object'
     # ./lib/grape/router/attribute_translator.rb:28:in `to_h'
     # ./lib/grape/endpoint.rb:134:in `block (2 levels) in mount_in'
     # ./lib/grape/endpoint.rb:132:in `each'
     # ./lib/grape/endpoint.rb:132:in `block in mount_in'
     # ./lib/grape/endpoint.rb:127:in `each'
     # ./lib/grape/endpoint.rb:127:in `mount_in'
     # ./lib/grape/api.rb:103:in `block in initialize'
     # ./lib/grape/api.rb:102:in `each'
     # ./lib/grape/api.rb:102:in `initialize'
     # ./lib/grape/api.rb:26:in `new'
     # ./lib/grape/api.rb:26:in `compile'
     # ./lib/grape/api.rb:39:in `block in call'
     # ./lib/grape/api.rb:39:in `synchronize'
     # ./lib/grape/api.rb:39:in `call'
     # /Users/dblock/.rvm/gems/ruby-2.3.0/gems/rack-test-0.6.3/lib/rack/mock_session.rb:30:in `request'
     # /Users/dblock/.rvm/gems/ruby-2.3.0/gems/rack-test-0.6.3/lib/rack/test.rb:244:in `process_request'
     # /Users/dblock/.rvm/gems/ruby-2.3.0/gems/rack-test-0.6.3/lib/rack/test.rb:58:in `get'
     # ./spec/grape/integration/global_namespace_function_spec.rb:23:in `block (3 levels) in <top (required)>'
```

Thanks to @dylanfareed for the repro, took a while to narrow this down.

The `namespace` function is also defined by Rake, so the most common way of seeing this is with `include Rake::DSL` which is something that [fabricator does](https://github.com/paulelliott/fabrication/blob/master/lib/fabrication.rb#L5), for example.

dblock:
Fixes https://github.com/ruby-grape/grape/issues/1348.

A globally defined `def namespace` in Ruby 2.3.0 pollutes OStruct instance methods. This is described in https://bugs.ruby-lang.org/issues/12136 and https://bugs.ruby-lang.org/issues/12251. One way to reproduce this is to `include Rake::DSL` in the global scope, which defines a global `namespace` method.

This PR gets rid of the OStruct. We cannot use a `Hash` here because we want to respond to methods like `.index` which are natively defined on a `Hash`, so we need a dummy structure that responds to anything. Because it behaves like a Hash though we don't need to register parameters or such, cc: @namusyaka in case I missed anything.

### DIFF
--- a/lib/grape/router/attribute_translator.rb
+++ b/lib/grape/router/attribute_translator.rb
@@ -1,39 +1,29 @@
-require 'delegate'
-require 'ostruct'
-
 module Grape
   class Router
-    class AttributeTranslator < DelegateClass(OpenStruct)
-      def self.register(*attributes)
-        AttributeTranslator.supported_attributes.concat(attributes)
-      end
-
-      def self.supported_attributes
-        @supported_attributes ||= []
-      end
-
+    # this could be an OpenStruct, but doesn't work in Ruby 2.3.0, see https://bugs.ruby-lang.org/issues/12251
+    class AttributeTranslator
       def initialize(attributes = {})
-        ostruct = OpenStruct.new(attributes)
-        super ostruct
         @attributes = attributes
-        self.class.supported_attributes.each do |name|
-          ostruct.send(:"#{name}=", nil) unless ostruct.respond_to?(name)
-          self.class.instance_eval do
-            define_method(name) { instance_variable_get(:"@#{name}") }
-          end if name == :format
-        end
       end
 
       def to_h
-        @attributes.each_with_object({}) do |(key, _), attributes|
-          attributes[key.to_sym] = send(:"#{key}")
+        @attributes
+      end
+
+      def method_missing(m, *args)
+        if m[-1] == '='
+          @attributes[m[0..-1]] = *args
+        else
+          @attributes[m]
         end
       end
 
-      private
-
-      def accessor_available?(name)
-        respond_to?(name) && respond_to?(:"#{name}=")
+      def respond_to_missing?(method_name, _include_private = false)
+        if method_name[-1] == '='
+          true
+        else
+          @attributes.key?(method_name)
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


dblock:
Not a correct repro, needs to have a globally defined `def namespace` which pollutes everything. Maybe to make sure this breaks/is fixed we should just include `Rake::DSL` in spec_helper?

lib/grape/router/route.rb
```diff
@@ -48,6 +47,11 @@ def method_missing(method_id, *arguments)
         end
       end
 
+      # workaround Ruby 2.3.0, https://bugs.ruby-lang.org/issues/12251
+      def namespace
+        attributes.namespace
+      end
+
```
dblock: Seems that a different implementation takes precedence over `def_delegators`.

lib/grape/router/attribute_translator.rb
```diff
@@ -1,39 +1,29 @@
...
+      end
 
-      def accessor_available?(name)
-        respond_to?(name) && respond_to?(:"#{name}=")
+      def respond_to?(m, _include_all = false)
+        if m[-1] == '='
+          true
+        else
+          @attributes.key?(m)
+        end
```
ridiculous: You may want to define `respond_to_missing?` for this class (https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding)

dblock: Yep, fixed.

Issue #1355, opened 2016-04-07
Regression: PATCH behavior along with Header version matching


dblock:
```ruby
require 'spec_helper'

describe Grape::API::Helpers do
  module PatchHelpersSpec
    class Patch_Public < Grape::API
      format :json
      version 'public-v1', using: :header, vendor: 'grape'

      get do
        { ok: 'public' }
      end
    end

    module AuthMethods
      def authenticate!

      end
    end

    class Patch_Private < Grape::API
      format :json
      version 'private-v1', using: :header, vendor: 'grape'

      helpers AuthMethods

      before do
        authenticate!
      end

      get do
        { ok: 'private' }
      end
    end

    class Main < Grape::API
      mount Patch_Public
      mount Patch_Private
    end
  end

  def app
    PatchHelpersSpec::Main
  end

  context 'default' do
    it 'public' do
      get '/', {}, 'HTTP_ACCEPT' => 'application/vnd.grape-public-v1+json'
      expect(last_response.status).to eq 200
      expect(last_response.body).to eq({ ok: 'public' }.to_json)
    end

    it 'private' do
      get '/', {}, 'HTTP_ACCEPT' => 'application/vnd.grape-private-v1+json'
      expect(last_response.status).to eq 200
      expect(last_response.body).to eq({ ok: 'private' }.to_json)
    end

    it 'default' do
      get '/'
      expect(last_response.status).to eq 200
      expect(last_response.body).to eq({ ok: 'public' }.to_json)
    end
  end

  context 'patch' do
    it 'public' do
      patch '/', {}, 'HTTP_ACCEPT' => 'application/vnd.grape-public-v1+json'
      expect(last_response.status).to eq 405
    end

    it 'private' do
      patch '/', {}, 'HTTP_ACCEPT' => 'application/vnd.grape-private-v1+json'
      expect(last_response.status).to eq 405
    end

    it 'default' do
      patch '/'
      expect(last_response.status).to eq 405
    end
  end
end
```

This passes on 0.15.0.

In 0.16.1 this fails in interesting ways.

```
~/source/grape/grape/dblock (patch-method-helpers)$ rspec spec/grape/api/patch_method_helpers_spec.rb 

Grape::API::Helpers
  default
    public
    private
    default
  patch
    public (FAILED - 1)
    private
    default

Failures:

  1) Grape::API::Helpers patch public
     Failure/Error: expect(last_response.status).to eq 405
     
       expected: 405
            got: 406
     
[...]

PR #1356, opened 2016-04-07
Avoid evaluating the same route twice.

dblock:
The exact matching route may be evaluated multiple times if it yields a 406 or something like that. Maybe we can do this more elegantly?

### DIFF
--- a/lib/grape/router.rb
+++ b/lib/grape/router.rb
@@ -41,7 +41,7 @@ module Grape
 
     def call(env)
       with_optimization do
-        identity(env) || rotation(env) { |route| route.exec(env) }
+        rotation(env)
       end
     end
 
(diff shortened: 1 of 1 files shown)


namusyaka:
How about this patch?
```diff
diff --git lib/grape/router.rb lib/grape/router.rb
index bb6f830..98426ee 100644
--- lib/grape/router.rb
+++ lib/grape/router.rb
@@ -41,7 +41,8 @@ module Grape

     def call(env)
       with_optimization do
-        identity(env) || rotation(env) { |route| route.exec(env) }
+        response, route = identity(env)
+        response || rotation(env, route)
       end
     end

@@ -54,26 +55,28 @@ module Grape
     private

     def identity(env)
-      transaction(env) do |input, method, routing_args|
+      route = nil
+      response = transaction(env) do |input, method, routing_args|
         route = match?(input, method)
         if route
           env[Grape::Env::GRAPE_ROUTING_ARGS] = make_routing_args(routing_args, route, input)
           route.exec(env)
         end
       end
+      [response, route]
     end

-    def rotation(env)
-      transaction(env) do |input, method, routing_args|
-        response = nil
-        routes_for(method).each do |route|
-          next unless route.match?(input)
-          env[Grape::Env::GRAPE_ROUTING_ARGS] = make_routing_args(routing_args, route, input)
-          response = yield(route)
-          break unless cascade?(response)
-        end
-        response
+    def rotation(env, exact_route = nil)
+      response = nil
+      input, method, routing_args = *extract_required_args(env)
+      routes_for(method).each do |route|
+        next if exact_route == route
+        next unless route.match?(input)
+        env[Grape::Env::GRAPE_ROUTING_ARGS] = make_routing_args(routing_args, route, input)
+        response = route.exec(env)
+        break unless cascade?(response)
       end
+      response
     end

     def transaction(env)
```
I believe we should two search methods should be put separately.

dblock:
That is totally fine too. Go ahead and replace my PR with that or just do it part of https://github.com/ruby-grape/grape/issues/1355.

dblock:
Closing in favor of https://github.com/ruby-grape/grape/pull/1359.

Issue #1344, opened 2016-04-04
Deprecated route_xxx methods for route_params and route_? declared.


dblock:
I've PRed https://github.com/ruby-grape/grape-swagger/pull/368, but it has two problems, access to `route_params` and `route_hidden` in a dummy API declared as `GET '/foobar', hidden: true`. 

The `route_xxx` deprecated code reaches into `@options`, which isn't `@settings`, and that is no longer exposed. So `.params` is not the same thing as the old `route_params`. What should we do about this one?

dblock:
cc: @namusyaka, maybe you have a suggestion?

namusyaka:
@dblock Hmm, I think the `route_xxx` can explore the `@options` and `@settings` in a cascading style.
I'm going to investigate the problem in a few days.

dblock:
If you have time @namusyaka, the simplest thing to look at is https://github.com/dblock/grape-swagger/tree/fix-deprecations, try to fix the remaining deprecations. I think the simplest would be to just expose `options`, but maybe the specs are wrong and it's not the right thing to do.

dblock:
For now I am monkey-patching the deprecation to avoid too many warnings being spewed.

```ruby
module Grape
  class Router
    class Route
      private

      # TODO: remove when https://github.com/ruby-grape/grape-swagger/pull/368 and https://github.com/newrelic/rpm/pull/235 are finished
      def warn_route_methods(name, location, expected = nil)
      end
    end
  end
end
```

dblock:
Closing via #1357.

dblock:
https://github.com/newrelic/rpm/pull/282

PR #1357, opened 2016-04-09
Fix incorrect deprecations

namusyaka:
ref #1344

### DIFF
--- a/lib/grape/router/route.rb
+++ b/lib/grape/router/route.rb
@@ -8,8 +8,9 @@ module Grape
     class Route
       ROUTE_ATTRIBUTE_REGEXP = /route_([_a-zA-Z]\w*)/.freeze
       SOURCE_LOCATION_REGEXP = /^(.*?):(\d+?)(?::in `.+?')?$/.freeze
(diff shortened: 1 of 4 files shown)


lib/grape/router/route.rb
```diff
@@ -56,6 +57,15 @@ def route_path
         pattern.path
       end
 
+      def hidden
```
dblock: Are we really just trying to make grape-swagger happy here? The whole `hidden` thing is not a Grape feature, just happens to be an option that grape-swagger "extends". Maybe it's better to expose `options` instead and let grape-swagger deal with it?

namusyaka:
@dblock Thanks for your opinion. I've just tried to reflect it to my changes.

dblock:
I am happy with this as is, squashed and CHANGELOGed.

dblock:
I also think the code needs comments of what `settings` vs. `options` is.

namusyaka:
@dblock Could you review this updates? I also squashed :)

dblock:
Looks good to me, feel free to merge on green if I don't get to it.

PR #1359, opened 2016-04-11
Fix router bugs

namusyaka:
ref #1355 #1356 
The `PATCH` bug is caused by calling `Grape::Middleware::Versioner::Header`.
Now, grape router calls it because you can not eject route except first route of all routes having same paths.
But maybe it has no problem essentially.
So I think this problem can be solved by not call the middleware.

/cc @dblock

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * [#1357](https://github.com/ruby-grape/grape/pull/1357): Don't include fixed named captures as route params - [@namusyaka](https://github.com/namusyaka).
 * [#1359](https://github.com/ruby-grape/grape/pull/1359): Return 404 even if version is using as header - [@namusyaka](https://github.com/namusyaka).
+* [#1359](https://github.com/ruby-grape/grape/pull/1359): Avoid evaluating the same route twice - [@namusyaka](https://github.com/namusyaka), [@dblock](https://github.com/dblock).
 
 0.16.1 (4/3/2016)
 =================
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 #### Fixes
 
 * [#1357](https://github.com/ruby-grape/grape/pull/1357): Don't include fixed named captures as route params - [@namusyaka](https://github.com/namusyaka).
+* [#1359](https://github.com/ruby-grape/grape/pull/1359): Return 404 even if version is using as header - [@namusyaka](https://github.com/namusyaka).
+* [#1359](https://github.com/ruby-grape/grape/pull/1359): Avoid evaluating the same route twice - [@namusyaka](https://github.com/namusyaka), [@dblock](https://github.com/dblock).
```
dblock: Thanks for giving me credit ;)

lib/grape/router.rb
```diff
@@ -124,6 +127,8 @@ def greedy_match?(input)
     def method_not_allowed(env, methods, endpoint)
       current = endpoint.dup
       current.instance_eval do
+        namespace_inheritable_to_nil(:version)
+        @lazy_initialized = false
```
dblock: This is fairly straightforward in hindsight, I almost had it ;)

dblock:
@namusyaka I am not sure I understand what you're saying with regards to `PATCH` though. Is there're something left to address that you're working on?

PR #1361, opened 2016-04-11
Repro for #1355.

dblock:


dblock:
Merged via https://github.com/ruby-grape/grape/pull/1362.

Issue #1360, opened 2016-04-11
Release 0.16.2


dblock:
I think we can release 0.16.2 fixing the regressions introduced around 0.16.0.

@namusyaka Anything from you? Feel free to make the release otherwise or I can do it if you don't have time.

dblock:
Nevermind, I reopened https://github.com/ruby-grape/grape/issues/1355 with a repro, https://github.com/dblock/grape/blob/patch-method-helpers/spec/grape/api/patch_method_helpers_spec.rb

dblock:
Can do https://github.com/ruby-grape/grape-swagger/pull/375 after this is released.

namusyaka:
Just released [0.16.2](https://rubygems.org/gems/grape/versions/0.16.2) and pushed [the tag](https://github.com/ruby-grape/grape/releases/tag/v0.16.2) :tada: 
Also, I've posted [a topic for 0.16.2](https://groups.google.com/forum/#!topic/ruby-grape/Q-dNFG8uk04).
Thank you!

PR #1362, opened 2016-04-11
Return 405 correctly even if version is using as header and wrong request method

namusyaka:
This reverts commit a0166e77468093601975e7005ec2e2a46e3305ae.

I'll try to fix this problem drastically.


namusyaka:
@dblock Could you take a look this pull request?
I think this approach can fix potential bugs and issues.
If ok, we can release 0.16.2 version ;)

lib/grape/endpoint.rb
```diff
@@ -231,7 +231,7 @@ def run
 
         run_filters before_validations, :before_validation
 
-        run_validators validations, request unless @method_not_allowed
+        run_validators validations, request unless env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED]
```
dblock: I see it now :) Better.

dblock: You might want to audit the use if `@variable` elsewhere, we have at least an `@app_response` in a formatter middleware, `@named_params` in a DSL helper and `@versions` in routing.

dblock:
This worked everywhere where I saw problems, feel free to cut 0.16.2, which is https://github.com/ruby-grape/grape/issues/1360. I would review the use of `@...` first though.

namusyaka:
The `@method_not_allowed` variable usage is wrong. It returns `true` eternally if method_not_allowed is set  even once.
Other `@~` vars looks like no problem.

dblock:
I should have caught `@` in the code reviews too, my bad.

namusyaka:
I've mistaken the coding. Sorry for the bad patch. Btw, we can release the v.0.16.2 now. I'll prepare it soon.

PR #1363, opened 2016-04-14
README-FIX: Grape is not a micro-framework

bugaiov:
Inspired by https://twitter.com/buka_gor/status/689036018708799492

PS. It's really hard to call grape a micro-framework (no matter how hard you try).

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@
 
 ## What is Grape?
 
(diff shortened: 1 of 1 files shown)


dblock:
I agree. Lets do this.

dblock:
Would you also make pull requests into https://github.com/ruby-grape/ruby-grape.github.io please?

bugaiov:
ok. will do tomorrow

Best,
Nick

PR #1365, opened 2016-04-18
Fixed finding handler in error middleware. Fixes #1364

### LINKED ISSUE #1364: Specific exception handlers are not called when using rescue_from with option 'with:'
I ran into the strange behaviour while refactoring my api and moving error handling into separate file. So I wrote this minimal example to reproduce the issue.

Here I use blocks to handle specific exceptions and call specific helper method:
```ruby
class API < Grape::API
  helpers do 
    def argument_error e
      rack_response({ error: "argument_error_block" }.to_json)
    end

    def all_errors e
      rack_response({ error: "all_errors_block" }.to_json)
    end
  end

  rescue_from ArgumentError do |e| argument_error(e); end
  rescue_from :all do |e| all_errors(e); end

  get :argument_error do
    raise ArgumentError
  end
end
```

When I request ```/argument_error``` API responds, as intended, with ```{"error":"argument_error_block"}```. But if I use ```with:``` in rescue_from
```ruby
  rescue_from ArgumentError, with: :argument_error
  rescue_from :all, with: :all_errors
```
I start to receive ```{"error":"all_errors_block"}``` and never get to ```argument_error``` helper.

ktimothy:


dblock:
Need CHANGELOG and a test, please.

ktimothy:
Sure, will do that tomorrow.

ktimothy:
Looks like the issue is a little bit deeper. Last call to rescue_from using with: overwrites previously set with: value.

For now option 'with:' is stored now in rescue_options in middleware/error, which makes it common for all exceptions.

I think the option 'with:' should be stored for each exception type separately.

lib/grape/dsl/request_response.rb
```diff
@@ -157,9 +157,9 @@ def represent(model_class, options)
 
         def extract_with(options)
           return unless options.key?(:with)
-          with_option = options[:with]
+          with_option = options.delete(:with)
```
dblock: You probably want to `dup` options first, since you're modifying the original hash.

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 
 * Your contribution here.
 
+#### Fixes
+
+* [#1365](https://github.com/ruby-grape/grape/pull/1365): Fix finding exception handler in error middleware - [@ktimothy](https://github.com/ktimothy)
```
dblock: Missing a period at the end, please ;)

dblock:
See minor-ish comments, otherwise good to merge?

ktimothy:
Yes, just made small fixes. Should I squash commits?

lib/grape/dsl/request_response.rb
```diff
@@ -104,7 +104,7 @@ def rescue_from(*args, &block)
             handler = block
           end
 
-          options = args.extract_options!
+          options = args.extract_options!.dup
```
dblock: That's not what I meant. This particular one actually returns a new hash every time according to the documentation, so a `dup` is just doing extra unnecessary work.

dblock: So what's wrong with just checking `options[:with]` below? Why does it need to `delete` that value?

spec/grape/dsl/request_response_spec.rb
```diff
@@ -173,7 +173,7 @@ def self.imbue(key, value)
           it 'sets a rescue handler declared through :with option for each key in hash' do
             with_block = -> { 'hello' }
             expect(subject).to receive(:namespace_stackable).with(:rescue_handlers, StandardError => an_instance_of(Proc))
-            expect(subject).to receive(:namespace_stackable).with(:rescue_options, with: with_block)
+            expect(subject).to receive(:namespace_stackable).with(:rescue_options, {})
```
dblock: What happened here?

dblock:
Sorry to be a pest, I have more questions ;)

lib/grape/dsl/request_response.rb
ktimothy: > This particular one actually returns a new hash every time

Oh, shame to me =)

Ok, we should have discussed it a little bit before I made my changes =). The option ```:with``` was used in middleware/error. I removed that usage, and stored its value in ```rescue_handlers``` hash. 

So I decided to delete ```:with``` from ```options``` hash just to clean it from unused key. And that's why I had to fix request_response_spec.rb.

spec/grape/dsl/request_response_spec.rb
ktimothy: ```rescue_options``` now do not contain with: option, see my comment above.

ktimothy:
No problem ;)

spec/grape/dsl/request_response_spec.rb
dblock: I don't get it, why is it a problem related to the original issue you're trying to fix?

dblock:
@ktimothy I am so sorry to be thick and slow here (typing this from an island in the Caribbean). Amend the code to what you want it to be, and re-explain the whole thing to me like a two year old. Really appreciate your patience.

ktimothy:
@dblock I hope you have a good time there =)

Ok, here is the whole thing.

At first I handled different exceptions using block syntax:
```ruby
rescue_from ArgumentError do |e|
  rack_response({error: 'argument_error'}.to_json)
end

rescue_from :all do |e|
  rack_response({error: 'internal_error'}.to_json)
end
```

Then I wanted to move exception handlers from api class to helpers module. So I tried to use options ```:with```:
```ruby
rescue_from ArgumentError, with: :rescue_argument_error
rescue_from :all, with: :all_errors_handler
```
which made my api code much better.

But, unfortunately, grape stopped handling ```ArgumentError```, calling ```:all_errors_handler``` instead. I started to dig into the problem, and found out, that ```:with``` is stored in ```:rescue_options``` hash, which is common for all exception types. Because of that ```:with``` option is overwritten every time you call ```rescue_from: ..., with: ...```. This behavior is definitely wrong, as we save last ```with:``` value in options.

As a solution I suggest to store ```with:``` in ```rescue_handlers``` and ```base_only_rescue_handlers```, where block handlers are now stored. So values in these two hashes can be either a block or a symbol. 

Error middleware now does not check rescue_options[:with], it tries to take handler from ```rescue_handlers`` and ```base_only_rescue_handlers```, then tries ```all_rescue_handler```. If handler is a symbol, we know that we should use helper method with corresponding name.

dblock:
Merged via https://github.com/ruby-grape/grape/commit/044dad76074f987c58de90bcdb061575d82fb784, thanks for the clear explanation!

ktimothy:
Awesome, thanks!

PR #1366, opened 2016-04-19
add back message key in order to have error_codes

mkou:
In my situation, in the exception validations, I need message keys as well as the translations in order to pass error codes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Your contribution here.
 
 #### Features
-* [#1366](https://github.com/ruby-grape/grape/pull/1366): Store back `message_key` on `Grape::Exceptions::Validation` - [@mkou](https://github.com/mkou).
+* [#1366](https://github.com/ruby-grape/grape/pull/1366): Store `message_key` on `Grape::Exceptions::Validation` - [@mkou](https://github.com/mkou).
 
 0.16.2 (4/12/2016)
 ==================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 
 * Your contribution here.
 
+#### Features
+* [#1366](https://github.com/ruby-grape/grape/pull/1366): Store back `message_key` on `Grape::Exceptions::Validation` - [@mkou](https://github.com/mkou).
+
 0.16.2 (4/12/2016)
 ==================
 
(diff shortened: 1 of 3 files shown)


dblock:
This needs tests and a CHANGELOG entry, please.

mkou:
Done. Thanks for the feedback

lib/grape/exceptions/validation.rb
```diff
@@ -9,6 +9,7 @@ class Validation < Grape::Exceptions::Base
       def initialize(args = {})
         fail 'Params are missing:' unless args.key? :params
         @params = args[:params]
+        @message_key = args[:message] if args.key?(:message) && args[:message].is_a?(Symbol)
         args[:message] = translate_message(args[:message]) if args.key? :message
         super
```
dblock: Can I nitpick? Is this better? Avoids checking `key?` twice.

```ruby
return unless args.key?(:message)
@message_key = args[:message] if args[:message].is_a?(Symbol)
args[:message] = translate_message(args[:message])
```

mkou: Good idea, I will use an ``if`` in order not to bypass the ``super`` method

mkou: Would you prefer that version ?
````
super && return unless args.key?(:message)
@message_key = args[:message] if args[:message].is_a?(Symbol)
args[:message] = translate_message(args[:message])
super
```

mkou: Would you prefer that version ? 
```
super && return unless args.key?(:message)
@message_key = args[:message] if args[:message].is_a?(Symbol)
args[:message] = translate_message(args[:message])
super
```

ridiculous: definitely the latter 👍

dblock: likewise I'm with @ridiculous

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 
 * Your contribution here.
 
+#### Features
+* [#1366](https://github.com/ruby-grape/grape/pull/1366): Store back `message_key` on `Grape::Exceptions::Validation` - [@mkou](https://github.com/mkou).
```
dblock: I'd remove the "back" word ;)

mkou:
Thanks! 😄  What do you think?

dblock:
Merged squashed, thanks.

PR #1378, opened 2016-04-27
Fix double negation violations

ochagata:
3 style cops remain, otherwise all metrics cops


lib/grape/path.rb
```diff
@@ -22,11 +22,11 @@ def root_prefix
     end
 
     def uses_specific_format?
-      !!(settings[:format] && Array(settings[:content_types]).size == 1)
+      (settings[:format] && Array(settings[:content_types]).size == 1)
```
ochagata: This and the other violation in this file are going to return a `boolean` value anyways, no need for double bang

lib/grape/router.rb
```diff
@@ -11,6 +11,14 @@ def initialize(pattern, attributes = {})
       end
     end
 
+    def self.normalize_path(path)
```
ochagata: This method had been implemented as private before. Ruby doesn't actually allow for private class methods, so moving it above the private block and taking care of the violation doesn't change anything

lib/grape/middleware/versioner/header.rb
```diff
@@ -159,7 +159,7 @@ def version_options
         # header, one can set the `:cascade` option to `false`.
         def cascade?
           if version_options && version_options.key?(:cascade)
-            !!version_options[:cascade]
+            version_options[:cascade]
```
ochagata: Wanted to make sure on these. `:cascade` is going to be set as either `TrueClass` or `FalseClass` as I understood it, which makes the double bang not needed

dblock:
Note a legit spec failure.

lib/grape/path.rb
```diff
@@ -22,11 +22,11 @@ def root_prefix
...
     end
 
     def uses_path_versioning?
-      !!(settings[:version] && settings[:version_options][:using] == :path)
+      (settings[:version] && settings[:version_options][:using] == :path)
```
ochagata: @dblock 
The error in the specs is related to this line change.
My problem is with the way ruby style-guide/rubocop addresses this violation.

```
# bad
!!(settings[:format] && Array(settings[:content_types]).size == 1)

#good
!(settings[:format] && Array(settings[:content_types]).size == 1).nil?
```

This raises the problem of if this operation returns `false`, it will evaluate to true, which makes it return the incorrect value.
I personally am not opposed to `!!` and think it is in fact better than the alternative `!foo.nil?`
Potentially, I could rework this PR to simply disable this check and move it to `.rubocop.yml` as disabled

Thoughts?

dblock: I think keeping a double-negation is fine. I also don't agree with Rubocop here :)

ochagata:
Alright, should be good once travis check finishes

.rubocop.yml
```diff
@@ -3,4 +3,7 @@ AllCops:
     - vendor/**/*
     - bin/**/*
 
+Style/DoubleNegation:
```
dblock: We often just run `rubocop --auto-gen-config`, can we just do that instead?

ochagata: Sure, but I'll have to keep this one in here so it ignores this

dblock: Na, just remove it from here and let the auto-gen-config flag it as a violation. Maybe one day we'll change our mind on whether we like `!!` or not.

ochagata: Got it, I'll have this be a PR to fix the lint violation then

dblock: Note that the other double-negation fixes *are* valuable, they just waste CPU cycles, so I am happy you did them!

lib/grape/path.rb
```diff
@@ -22,11 +22,19 @@ def root_prefix
     end
 
     def uses_specific_format?
-      !!(settings[:format] && Array(settings[:content_types]).size == 1)
+      if settings.key?(:format) && settings.key?(:content_types)
```
ochagata: Figured out why it was failing here and how to implement it properly.
Added a check for the presence of the keys, if either is `nil`, it will simply move on to the `else` block

dblock:
Excellent.

PR #1380, opened 2016-04-28
allow_blank for Time types 

ipkes:
Based on [this](https://github.com/ruby-grape/grape/issues/723#issuecomment-214208198)
Valid value for `Time` attribute causes an error when attribute marked `allow_blank: false`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 #### Fixes
 
 * [#1365](https://github.com/ruby-grape/grape/pull/1365): Fix finding exception handler in error middleware - [@ktimothy](https://github.com/ktimothy).
+* [#1380](https://github.com/ruby-grape/grape/pull/1380): `allow_blank: false` for `Time` attribute with valid value causes an error [@ipkes](https://github.com/ipkes).
 
 0.16.2 (4/12/2016)
 ==================
--- a/lib/grape/validations/validators/allow_blank.rb
+++ b/lib/grape/validations/validators/allow_blank.rb
@@ -20,7 +20,7 @@ module Grape
 
         return unless should_validate
 
-        return if value == false || value.present?
+        return if false == value || value.present?
 
         raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: message(:blank)
       end
--- a/spec/grape/validations/validators/allow_blank_spec.rb
+++ b/spec/grape/validations/validators/allow_blank_spec.rb
@@ -488,6 +488,11 @@ describe Grape::Validations::AllowBlankValidator do
       get '/disallow_boolean_blank', val: false
       expect(last_response.status).to eq(200)
     end
+
+    it 'accepts value when time allow_blank' do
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 #### Fixes
 
 * [#1365](https://github.com/ruby-grape/grape/pull/1365): Fix finding exception handler in error middleware - [@ktimothy](https://github.com/ktimothy).
+* #1367: `allow_blank: false` for `Time` attribute with valid value causes an error [@ipkes](https://github.com/ipkes).
```
dblock: Please stick to the format of other CHANGELOG entries. Pay attention to what's a link, then name, the period at the end :)

lib/grape/validations/validators/allow_blank.rb
```diff
@@ -20,7 +20,7 @@ def validate_param!(attr_name, params)
 
         return unless should_validate
 
-        return if value == false || value.present?
+        return if false == value || value.present?
```
dblock: You're such a C programmer ;)

spec/grape/validations/validators/allow_blank_spec.rb
```diff
@@ -488,6 +488,11 @@ def app
       get '/disallow_boolean_blank', val: false
       expect(last_response.status).to eq(200)
     end
+
+    it 'accepts value when time allow_blank' do
+      get '/disallow_datetime_blank', val: Time.now
```
dblock: How did this fail before, just for completeness?

ipkes: Here a start of the error stack:
<pre>
Failure/Error: return if value == false || value.present?
     
     NoMethodError:
       undefined method `to_datetime' for false:FalseClass
     # .../.rvm/gems/ruby-2.3.0/gems/activesupport-3.2.19/lib/active_support/core_ext/date_time/calculations.rb:141:in `<=>'
     # ./lib/grape/validations/validators/allow_blank.rb:24:in `=='
     # ./lib/grape/validations/validators/allow_blank.rb:24:in `validate_param!'
</pre>

Here 'active_support/core_ext/date_time/calculations.rb:141':
<pre>
  # Layers additional behavior on DateTime#<=> so that Time and
  # ActiveSupport::TimeWithZone instances can be compared with a DateTime.
  def <=>(other)
    super other.to_datetime
  end
</pre>
I'm not sure why ActiveSupport is calling (if you have any idea, please share, I'll be very grateful). I think the problem is much deeper, but I'm to lazy to digging it up. So, maybe solution what offered is not a solution, it's looks like hook, but it works fine and looks pretty good =)
This works fine too:
<pre>return value.present? || false == value</pre>

ipkes:
I'm trying to do all according by contributing instruction. Changelog was updated, as you asked. Please correct me if I wrong.

dblock:
Merged via https://github.com/ruby-grape/grape/commit/a711cab410367831359b129a41672a423b0bacb6. Thank you.

PR #911, opened 2015-01-27
Yet more failing tests for optional parameter behavior around Arrays

al2o3cr:
This is another group of failing tests for parameter parsing. See also related issues / PRs:

* #615 encounters similar behavior with defaults inside an optional Array

* #895 tests some of the same behavior, but with optional params inside an optional Array

* #723 encounters identical behavior (nested optional Array containing required params forces all elements of the array to have an optional value)

Creating this PR for discussion of the cases + any resolution I come up with.

PR #1384, opened 2016-05-03
Empty optional nested array

ipkes:
Based on [issue](https://github.com/ruby-grape/grape/issues/723) and [tests](https://github.com/ruby-grape/grape/pull/911)
1. Some test modification
2. Possible solution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * [#1365](https://github.com/ruby-grape/grape/pull/1365): Fix finding exception handler in error middleware - [@ktimothy](https://github.com/ktimothy).
 * [#1380](https://github.com/ruby-grape/grape/pull/1380): Fix `allow_blank: false` for `Time` attributes with valid values causes `NoMethodError` - [@ipkes](https://github.com/ipkes).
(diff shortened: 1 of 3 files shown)


ipkes:
I'm afraid solution is insufficient and needs some work here. A major problem here with digging deeper in array. Need more time to think

dblock:
Start by rebasing and squashing this, there're a bunch of unrelated changes, hard to grok.

ipkes:
1. correctly done rebasing
2. another spec modify, more complex example (to digging deeper)
3. and sufficient solution
----------
problem was with checking on blank, first we should check not for all elements, just for any and second sometimes we have already blank parameter and we should check it first.

ipkes:
And for information. When I tried to find solution, I analyzed elements and parameters, which came, without complete understanding why they are coming. I can't be sure that anything else parameters will be work as well, but to already described parameters, it's ok.

dblock:
This makes total sense to me. We want to do an `allow_blank` validation if anything is blank here. Merging.

dblock:
Merged via https://github.com/ruby-grape/grape/commit/9a083581204a492ac7153cbd4d05e385f4a94793, thank you.

PR #1386, opened 2016-05-05
Replace merge by merge!

glaucocustodio:
As described [here](https://github.com/JuanitoFatas/fast-ruby#hashmerge-vs-hashmerge-code), `merge!` is 24x faster than `merge`.

### DIFF
--- a/lib/grape/cookies.rb
+++ b/lib/grape/cookies.rb
@@ -32,7 +32,7 @@ module Grape
     end
 
     def delete(name, opts = {})
-      options = opts.merge(value: 'deleted', expires: Time.at(0))
+      options = opts.merge!(value: 'deleted', expires: Time.at(0))
       self.[]=(name, options)
     end
   end
--- a/lib/grape/dsl/desc.rb
+++ b/lib/grape/dsl/desc.rb
@@ -51,7 +51,7 @@ module Grape
           end
           options = config_class.settings
         else
-          options = options.merge(description: description)
+          options = options.merge!(description: description)
         end
 
         namespace_setting :description, options
--- a/lib/grape/dsl/desc.rb
+++ b/lib/grape/dsl/desc.rb
@@ -51,7 +51,7 @@ module Grape
           end
           options = config_class.settings
         else
-          options = options.merge!(description: description)
+          options[:description] = description
         end
 
         namespace_setting :description, options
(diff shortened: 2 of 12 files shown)


dblock:
See build and rubocop errors, there's notably one you can do a lot faster than `merge!`. 

Generally you cannot be modifying the hash being passed in, so not all your changes of `merge` for `merge!` are ok. Finally if you're going to be doing `merge!`, you don't need to assign the result, it's modifying the hash in-place.

glaucocustodio:
Take a look now please @dblock.

lib/grape/cookies.rb
```diff
@@ -32,7 +32,7 @@ def each(&block)
     end
 
     def delete(name, opts = {})
-      options = opts.merge(value: 'deleted', expires: Time.at(0))
+      options = opts.merge!(value: 'deleted', expires: Time.at(0))
```
dblock: This modifies opts being passed in, you cannot do that.

lib/grape/dsl/desc.rb
```diff
@@ -51,7 +51,7 @@ def desc(description, options = {}, &config_block)
           end
           options = config_class.settings
         else
-          options = options.merge(description: description)
+          options[:description] = description
```
dblock: This modifies options passed in, you cannot do that.

dblock:
Most of these continue to have the same problem: modifying the `Hash` argument being passed in. This *will* have unpredictable side effects - It might be faster, but it only works because we get lucky or we have to think hard about where that hash came from.

lib/grape/cookies.rb
glaucocustodio: Why? Both `options` and `opts` are local variables, they aren't used somewhere else except there.

lib/grape/dsl/desc.rb
glaucocustodio: Same here, no?!

lib/grape/cookies.rb
dblock: When you say local, local to the class? Clearly here `opts` is not a local variable. 

```ruby
2.2.4 :001 > def doit(opts)
2.2.4 :002?>     opts[:foobar] = 1
2.2.4 :003?>   end
 => :doit 
2.2.4 :004 > 
2.2.4 :005 >   opts = {}
 => {} 
2.2.4 :006 > doit(opts)
 => 1 
2.2.4 :007 > p opts
{:foobar=>1}
 => {:foobar=>1} 
```

Modifying a hash within the function causes a side effect of modifying the caller's hash. So it's bad practice, finding bugs caused by such side effect is really hard.

glaucocustodio: 😯 I thought the arguments were copied, but they are the same to those which are passed to the method. Very nice explanation 👏.

dblock:
Lets close this, thanks for hanging in here @glaucocustodio.

Issue #1388, opened 2016-05-06
Nested params mixing


mkou:
I have a strange error, whenever I POST the parameters 
````
{ a: [ {  b: 'foo', c: 'bar' }, { d: 'foo2', c: 'bar' }]
```
I get the wrong params value in my api
```
{ "a"=>[{"b"=>"foo", "c"=>"bar", "d"=>"foo2"}, {"c"=>"bar"}]}
```

dblock:
Are you POSTing content-type JSON or doing this in a spec? This is not something you can post as a query string with curl for example.

I'll close this. Please use [the Grape mailing list](https://groups.google.com/forum/#!forum/ruby-grape) for questions.

mkou:
I created a test https://github.com/ruby-grape/grape/pull/1389 to show what my issue is

PR #1389, opened 2016-05-06
Illustrating issue #1388

mkou:
https://github.com/ruby-grape/grape/issues/1388


dblock:
I still think this can't work and Rack::Test is just making it more confusing. Lets take a step back: write a dummy API that takes an array of hashes and try to make it work outside of a spec. I think you can't without changing the content-type of what you're POSTing to something like JSON.

mkou:
Got it ;)

Issue #794, opened 2014-10-29
Insert middleware before/after


camertron:
At the moment, it doesn't appear as if Grape allows inserting middleware at arbitrary points in the middleware stack. Rails, for example, provides `insert_before` and `insert_after` methods for this purpose.

In our particular case, we'd like to add custom error handling middleware to our API. Specifically, we'd like to be able to submit all errors to Rollbar (an error aggregator) via the [rollbar](https://github.com/rollbar/rollbar-gem) gem. As far as I can tell, we can't use Grape's built-in error handling for this (i.e. via `rescue_from`) because that may override functionality in the API class. We could also monkey-patch the [existing error handling middleware](https://github.com/intridea/grape/blob/v0.9.0/lib/grape/middleware/error.rb) or the [`build_middleware`](https://github.com/intridea/grape/blob/v0.9.0/lib/grape/endpoint.rb#L227) method in `Endpoint`, but monkey-patching is brittle and means we'd have to copy and paste a ton of code. It would be great if we could simply insert our middleware at the top of the middleware stack. That way, it could temporarily bypass the existing error middleware, report the error via rollbar, then either re-raise the exception or return the appropriate rack response and continue up the middleware chain.

Since there's currently no way to arbitrarily insert middleware, I've come up with a rather [hack-ish solution](https://gist.github.com/camertron/0422ce551111cd23afdc). It works by wrapping `build_middleware` in the equivalent of an alias method chain. When the new method is invoked, it calls the original `build_middleware` method, removes the last item in the chain, and adds it back again at index 0. Clearly not optimal, but it works.

Any thoughts, suggestions, improvements? Thanks!

douglascamata:
For noise purposes, this issue could be closed as it was fixed in #1390.

dblock:
Closing, thanks @douglascamata.

dblock:
Closing, thanks @douglascamata.

PR #1390, opened 2016-05-08
Allow inserting middleware at arbitrary points in the middleware stack

Rosa:
This pull request is inspired by #794. I was using Grape and found myself needing this feature, so I have built a possible way of supporting `insert`, `insert_after` and `insert_before` in a similar way as Rails does.

Adding a new middleware can now be done through `use`, `insert`, `insert_after` and `insert_before`, and the operation chosen to do so is stacked together with the middleware class, args, and blocks. The stack is built using the operations in `Grape::Endpoint#build_stack` using a new class `Grape::Middleware::Stack`, based on `ActionDispatch::MiddlewareStack` with some simplifications (for example, only classes are supported for middleware, no symbols or strings, which is being deprecated in `ActionDispatch` in any case).

The commit message offers more details on the implementation itself. If anyone has a better idea of how this feature could be implemented I'd be glad to modify my changes to follow a different approach.

This is my first PR in the Grape project and I have tried to follow the same style when adding new tests for the feature and updating the `README`, but I'd be more than happy to adapt those according to feedback 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -2807,7 +2807,27 @@ Your middleware can overwrite application response as follows, except error case
 ```ruby
 class Overwriter < Grape::Middleware::Base
   def after
-    [200, { 'Content-Type' => 'text/plain' }, ['Overwrited.']]
+    [200, { 'Content-Type' => 'text/plain' }, ['Overwritten.']]
+  end
+end
+```
+
+You can add your custom middleware with `use`, that push the middleware onto the stack, and you can also control where the middleware is inserted using `insert`, `insert_before` and `insert_after`.
+
+```ruby
+class CustomOverwriter < Grape::Middleware::Base
+  def after
+    [200, { 'Content-Type' => 'text/plain' }, [@options[:message]]]
+  end
+end
+
+
+class API < Grape::API
+  use Overwriter
+  insert_before Overwriter, CustomOverwriter, message: 'Overwritten again.'
+  insert 0, CustomOverwriter, message: 'Overwrites all other middleware.'
+
+  get '/' do
   end
 end
 ```
--- a/lib/grape/util/stackable_values.rb
+++ b/lib/grape/util/stackable_values.rb
@@ -3,16 +3,16 @@ module Grape
     class StackableValues
       attr_accessor :inherited_values
       attr_reader :new_values
-      attr_reader :froozen_values
+      attr_reader :frozen_values
 
       def initialize(inherited_values = {})
         @inherited_values = inherited_values
         @new_values = {}
-        @froozen_values = {}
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -2812,6 +2812,26 @@ class Overwriter < Grape::Middleware::Base
...
+class CustomOverwriter < Grape::Middleware::Base
+  def after
+    [200, { 'Content-Type' => 'text/plain' }, [@options[:message]]]
+  end
+end
+
+
+class API < Grape::API
+  use Overwriter
+  insert_before Overwriter, CustomOverwriter, message: 'Overwrited again.'
```
dblock: Overwrited => Overwritten :)

README.md
```diff
@@ -2812,6 +2812,26 @@ class Overwriter < Grape::Middleware::Base
...
+  insert 0, CustomOverwriter, message: 'Overwrited definitive.'
```
dblock: Overwritten definitive doesn't sound great in English, how about "Overwrites all other middleware."?

spec/grape/dsl/middleware_spec.rb
```diff
@@ -6,24 +6,44 @@ module MiddlewareSpec
       class Dummy
         include Grape::DSL::Middleware
       end
+
+      class MyMiddleware; end
+      class MyOtherMiddleware; end
```
dblock: I know `Dummy` is declared above the same way, but these are global scope so we should use `let(:dummy) { Class.new(...) }` instead, if you don't mind changing those.

dblock:
This is excellent, code is spot on. I made minor text or other comments if you could please amend and I'll merge.

README.md
Rosa: Sounds good! I'll change also the original `Overwriter` response with a similar sentence.

spec/grape/dsl/middleware_spec.rb
Rosa: You are totally right, I'll change them 😄

Rosa:
Thanks so much for the quick review and the feedback! I think I have addressed all the comments now, but feel free to add more comments if you find anything else that can be improved :)

lib/grape/util/stackable_values.rb
```diff
@@ -3,16 +3,16 @@ module Util
     class StackableValues
       attr_accessor :inherited_values
       attr_reader :new_values
-      attr_reader :froozen_values
+      attr_reader :frozen_values
```
dblock: Those froozen values were just extra frozen. Thanks for spotting this ;)

dblock:
Merged.

PR #1393, opened 2016-05-12
Middleware can be inserted before or after defaults

ridiculous:
Builds onto the new middleware stack introduced in https://github.com/ruby-grape/grape/pull/1390. This is useful because it allows positioning of middleware around default Grape middleware.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 #### Features
 
+* [#1393](https://github.com/ruby-grape/grape/pull/1393): Middleware can be inserted before or after default Grape middleware - [@ridiculous](https://github.com/ridiculous).
 * [#1390](https://github.com/ruby-grape/grape/pull/1390): Allow inserting middleware at arbitrary points in the middleware stack - [@Rosa](https://github.com/Rosa).
 * [#1366](https://github.com/ruby-grape/grape/pull/1366): Store `message_key` on `Grape::Exceptions::Validation` - [@mkou](https://github.com/mkou).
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 #### Features
 
+* [#1393](https://github.com/ruby-grape/grape/pull/1393): Middleware can be inserted before or after default Grape middleware - [@ridiculous](https://github.com/ridiculous).
```
dblock: I think the CHANGELOG entry below covers this functionality, I would just add `, [#1393]` to it. No big deal.

dblock:
Very nice, merged.

Rosa:
Nice :) Although with only #1390 you could position your middleware around the default middleware, for example with:
```
insert_before Grape::Middleware::Error, MyMiddleware
```

Just out of curiosity, what was the use case not considered?

dblock:
Sorry @Rosa I think I don't understand the question :)

Rosa:
@dblock sorry for not explaining myself better! :sweat_smile: I meant to ask about which new feature this PR introduces that wasn't already covered by #1390. I believed you could already position middleware around defaults, so I was wondering if I had missed some use case that is now covered.

ridiculous:
@Rosa It let's you reference middleware added after https://github.com/ruby-grape/grape/pull/1393/files#diff-8e15cddb80f42ca7d43872846d80ecbeR263, which is pretty much just the formatter.

Rosa:
Ah, got it! I was using my changes just to insert a middleware _before_ `Grape::Middleware::Error`, which was fine but as you noticed, it didn't cover all cases. Thanks for the clarification! 😄

PR #1394, opened 2016-05-12
Add #middleware_spec hook method to Middleware.use

ridiculous:
@dblock @namusyaka What do you think about this? The benefit is it provides a way for middleware to position themselves when added with `use`

### DIFF
--- a/lib/grape/dsl/middleware.rb
+++ b/lib/grape/dsl/middleware.rb
@@ -12,10 +12,13 @@ module Grape
         # to the current namespace and any children, but
         # not parents.
         #
-        # @param middleware_class [Class] The class of the middleware you'd like
-        #   to inject.
-        def use(middleware_class, *args, &block)
-          arr = [:use, middleware_class, *args]
+        # @param [Class] klass The class of the middleware you'd like to inject.
+        def use(klass, *args, &block)
+          arr = if klass.respond_to?(:middleware_spec)
+                  klass.middleware_spec.push(*args)
+                else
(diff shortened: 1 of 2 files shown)


dblock:
Sounds dangerous. What's a real world scenario?

ridiculous:
The real world example I'm thinking of is for the [Grape::Middleware::Logger](https://github.com/ridiculous/grape-middleware-logger), which should always be positioned after `Grape::Middleware::Formatter` in the stack. This is because it uses an `env` key set by it for retrieving POST params. Thanks to the work @Rosa did, we can now specify the position with:
```ruby
insert_after Grape::Middleware::Formatter, Grape::Middleware::Logger
```
Which works great, and it's acceptable! But it seems cleaner to allow users to simply call:
```ruby
use Grape::Middleware::Logger
``` 
and let the middleware worry about it's position dependency. Ideally, there'd be some sort of initialization process for gems to hook into, enabling them to configure these things.. But in lieu of that, hook methods!  👻

dblock:
It hides middleware order though, making something that was explicit much less explicit and harder to debug. I'll leave this open for now, maybe other people have strong opinions?

ridiculous:
Any update on this? Yay or nay?

dblock:
I vote _nay_. I think the solution we have now is acceptable enough. Going to close this.

dblock:
Thanks much for your suggestion @ridiculous and the PR, no hard feelings.

ridiculous:
Ok, no prob, thank you

PR #1398, opened 2016-05-13
adding option to allow grape to handle Grape::Exception when rescue all is set

mmclead:
This will allow Grape to use the built in Grape Exceptions defined in /lib/grape/exceptions when rescue :all is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 * Your contribution here.
 
+* [#1398](https://github.com/ruby-grape/grape/pull/1398): Added rescue_from :grape_exceptions option to allow Grape to use the built in Grape::Exception handing and use rescue :all behavior for everything else - [@mmclead](https://github.com/mmclead).
+
 #### Features
 
 * [#1393](https://github.com/ruby-grape/grape/pull/1393): Middleware can be inserted before or after default Grape middleware - [@ridiculous](https://github.com/ridiculous).
--- a/README.md
+++ b/README.md
@@ -1853,6 +1853,17 @@ class Twitter::API < Grape::API
 end
 ```
 
+Grape can also rescue from all exceptions and still use the built-in exception handing.
+This will give the same behavior as `rescue_from :all` with the addition that Grape will use the exception handling defined by all Exception classes that inherit Grape::Exceptions::Base.
+
+The intent of this setting is to provide a simple way to cover the most common exceptions and return any unexpected exceptions in the API format.
+
+```ruby
+class Twitter::API < Grape::API
+  rescue_from :grape_exceptions
+end
+```
+
 You can also rescue specific exceptions.
 
 ```ruby
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -12,8 +12,9 @@ module Grape
           formatters: {},
           error_formatters: {},
           rescue_all: false, # true to rescue all exceptions
+          rescue_grape_exceptions: false,
           rescue_subclasses: true, # rescue subclasses of exceptions listed
-          rescue_options: { backtrace: false }, # true to display backtrace
+          rescue_options: { backtrace: false }, # true to display backtrace, true to let Grape handle Grape::Exceptions
           rescue_handlers: {}, # rescue handler blocks
           base_only_rescue_handlers: {}, # rescue handler blocks rescuing only the base class
           all_rescue_handler: nil # rescue handler block to rescue from all exceptions
@@ -34,7 +35,7 @@ module Grape
           end)
         rescue StandardError => e
           is_rescuable = rescuable?(e.class)
-          if e.is_a?(Grape::Exceptions::Base) && !is_rescuable
+          if e.is_a?(Grape::Exceptions::Base) && (!is_rescuable || rescuable_by_grape?(e.class))
             handler = ->(arg) { error_response(arg) }
           else
             raise unless is_rescuable
@@ -61,6 +62,11 @@ module Grape
       def rescuable?(klass)
         return false if klass == Grape::Exceptions::InvalidVersionHeader
         rescue_all? || rescue_class_or_its_ancestor?(klass) || rescue_with_base_only_handler?(klass)
+      end
+
+      def rescuable_by_grape?(klass)
+        return false if klass == Grape::Exceptions::InvalidVersionHeader
+        options[:rescue_grape_exceptions]
       end
 
       def exec_handler(e, &handler)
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 0.16.3 (Next)
 =============
 
-* Your contribution here.
```
dblock: Put this back please.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
...
+* [#1398](https://github.com/ruby-grape/grape/pull/1398): Option to allow Grape to handle Grape::Exception when rescue :all is set [@mmclead](https://github.com/mmclead)
```
dblock: Missing a period. Should describe what was added, such as "added rescue_grape_errors to ...".

dblock:
Would you please explain a bit more in detail what this achieves? I understand what it does, but I cannot think of a scenario where this is really useful.

mmclead:
I added this so that I could easily respond with JSON errors and not HTML that leaked too much data and still have Grape handle 404's, 406's, and all the other exceptions that are handled out of the box.  

So, if you were to add ```rescue_from :all, rescue_grape_errors: true``` to your API, you could cover all exception handling in one go, and then add specific ```error!``` calls as needed throughout the code.

Does that make sense?   I can try to explain better if needed.

dblock:
There's still something odd about this feature. It skips a specific exception, handles the rest, so that seems very "special" to begin with.

I want to talk it over more, I am sure we can come up with something better. 

Can't the same thing be accomplished with `rescue_from Grape::Exceptions::Base`, minus the special handling of `Grape::Exceptions::InvalidVersionHeader`? If that's the case maybe we could alias that long exception name with `rescue_from :grape_exceptions`?

mmclead:
The  exception ```Grape::Exceptions::InvalidVersionHeader```  is skipped for the same reason that it is skipped in ```rescuable?(klass)```  which is explained in the README:
 
>  Unrescuable Exceptions
> 
> `Grape::Exceptions::InvalidVersionHeader`, which is raised when the version in the request header doesn't match the currently evaluated version for the endpoint, will _never_ be rescued from a `rescue_from` block (even a `rescue_from :all`) This is because Grape relies on Rack to catch that error and try the next versioned-route for cases where there exist identical Grape endpoints with different versions.

I just copied the same logic.   I wanted to put this repeated logic into one place but couldn't think of a clean way to do it.


I don't think it'll work the same if we do ```rescue_from Grape::Exceptions::Base``` because an ```ActiveRecord::RecordInvalid``` wouldn't be caught.   The purpose of this PR is to allow users to be lazy about their error catching if they don't have specific use cases for certain errors.  As soon as the user has need of specific error handling then ```error!``` is available.   
Until then, this PR gives a 1 line way to catch all errors and use the previously built in error handling that Grape currently bypasses when ```rescue :all``` is set.

dblock:
Ok, but don't you think a DSL like `rescue_from :grape_exceptions` would be cleaner? Same implementation roughly.

mmclead:
yeah.  I like that name better.   it doesn't explicitly say that its also rescuing :all but a decently descriptive README might make up for it. 

I'll make that change and we can take a look at it.

mmclead:
@dblock change  completed. let me know what you think.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 
 * Your contribution here.
 
+* [#1398](https://github.com/ruby-grape/grape/pull/1398): Added rescue_from :grape_exceptions option to allow Grape to use the built in Grape::Exception handing and use rescue :all behavior for everything else. [@mmclead](https://github.com/mmclead)
```
dblock: Make this look like every other line please, the period goes to the end and there's a dash before your name ;)

README.md
```diff
@@ -1853,6 +1853,17 @@ class Twitter::API < Grape::API
 end
 ```
 
+Grape can also be told to rescue from all exceptions and still use the built in exception handing.
```
dblock: Grape can also rescue ...

built-in I think is spelled with a dash.

README.md
```diff
@@ -1853,6 +1853,17 @@ class Twitter::API < Grape::API
...
+This will give the same behavior as ```rescue_from :all``` with the addition that Grape will use the exception handling defined by all Exeption classes that inherit Grape::Exceptions::Base
```
dblock: You don't need three back ticks on `rescue_from :all`. 

This is a full sentence so needs a period.

dblock: Exeption is mispelled, missing c.

lib/grape/middleware/error.rb
```diff
@@ -12,8 +12,9 @@ def default_options
           formatters: {},
           error_formatters: {},
           rescue_all: false, # true to rescue all exceptions
+          rescue_grape_exceptions: false,
           rescue_subclasses: true, # rescue subclasses of exceptions listed
-          rescue_options: { backtrace: false }, # true to display backtrace
+          rescue_options: { backtrace: false, rescue_grape_errors: false }, # true to display backtrace, true to let Grape handle Grape::Exceptions
```
dblock: This is still referencing `rescue_grape_errors`?

spec/grape/exceptions/body_parse_errors_spec.rb
```diff
@@ -52,6 +52,43 @@ def app
...
+
+    def app
+      subject
+    end
+
+    context 'with content_type json' do
+      it 'returns body parsing error message' do
+        post '/beer', 'test', 'CONTENT_TYPE' => 'application/json'
+        expect(last_response.status).to eq 400
+        expect(last_response.body).to include 'message body does not match declared format'
```
dblock: What's the full body here? Maybe we can make an exact match?

dblock:
Made some comments, it's mostly ready. Try to shorten the README, get to the point :)

spec/grape/exceptions/body_parse_errors_spec.rb
mmclead: The full body is 
```
Problem:
  message body does not match declared format
Resolution:
  when specifying application/json as content-type, you must pass valid application/json in the request's 'body'
```

Kind of a pain to match the full thing.   Totally could do.   Was following the precedent set on the tests below:  
https://github.com/mmclead/grape/blob/68717c623ef5c3c0ad4f9180c5553c8ce758fa92/spec/grape/exceptions/body_parse_errors_spec.rb#L111

mmclead:
@dblock updated again.  Sorry my grammar is terrible.   And, I answered comment on test.  Your call on whether you want the full match or continue with the way other examples are.

spec/grape/exceptions/body_parse_errors_spec.rb
dblock: That's fine, thanks.

dblock:
Merged, thank you.

Issue #1403, opened 2016-05-19
`format :json` does not seem to work as documented


ciaoben:
I want my API to accept only `application/json` as a Content-Type. Reading the docs it seems that this should do the work:

```
class BaseApi < Grape::API
  format :json
  prefix :api
end
``` 

but my API still accept other Content-Types

dblock:
It's slightly more complicated, the doc says:

<img width="916" alt="screen shot 2016-05-20 at 7 12 40 am" src="https://cloud.githubusercontent.com/assets/542335/15426098/5077217a-1e5a-11e6-8833-3f156bfbe629.png">

What content-type are you sending that is not producing a 406? If it's none of the above, try turning this into a spec?

ciaoben:
I will do it as soon as possibile

ciaoben:
@dblock 
I am having problem pushing my specs to the repo, I followed the steps of `CONTRIBUTING.md` but it gives me:

```
remote: Permission to ruby-grape/grape.git denied to ciaoben.
fatal: unable to access 'https://github.com/ruby-grape/grape.git/': The requested URL returned error: 403
```

dblock:
You should make changes and push to your fork, not to this repo, [this](https://help.github.com/articles/fork-a-repo/) should be helpful.

ciaoben:
Thanks, I have created the [PR](https://github.com/ruby-grape/grape/pull/1407), sorry but I had never done it.

dblock:
This works by design, the documentation is correct. A JSON API does parse application/x-www-form-urlencoded date. I would be curious to see what a PR "fixing" this looks like, and would like to open it for discussion.

dblock:
https://github.com/rack/rack/blob/b0818268fd9f4026a1b0fb205cc1060ee38a23cc/lib/rack/request.rb#L334

PR #1407, opened 2016-05-26
Create a test that shows bug #1403

ciaoben:
I have created an example that shows that even if it is setted `format :json`, if I send a post request with content_type: application/x-www-form-urlencoded it does return `201`

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -652,6 +652,18 @@ describe Grape::Endpoint do
       expect(last_response.body).to eq('{"error":"The requested content-type \'application/xml\' is not supported."}')
     end
 
+    it 'responds with a 406 for an unsupported content-type for a post request' do
+      subject.format :json
+      # subject.content_type :json, "application/json"
+      subject.post '/request_body' do
+        params[:user]
+      end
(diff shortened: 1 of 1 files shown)


dblock:
Isn't this what the doc says? "it will not parse any other input than ... application/x-www-form-urlencoded"?

```
The formats apply to parsing, too. The following API will only respond to the JSON content-type and will not parse any other input than application/json, application/x-www-form-urlencoded, multipart/form-data, multipart/related and multipart/mixed. All other requests will fail with an HTTP 406 error code.
```

ciaoben:
Ok, I definitely need a vacation. I have red this sentence hundred times, and all the time I misunderstand it. 

My bad, sorry this time waste!

dblock:
It's not obvious that it's supposed to behave this way, I think the biggest reason is simply that Rack does some of the parsing and we can't force it not to.

dblock:
I would be curious to see what a PR "fixing" this looks like, and would like to open it for discussion. I'm going to reopen this.

ciaoben:
Do you mean introducing the case that if using `format :json`, it actually accept ONLY json?

I am not familiar with your codebase, but if you said that if it is due to some `Rack` level parsing, I can only imaging creating some sort of "extension" in the form of a Rack middleware.

dblock:
Right, I mean treating this as a bug and fixing it.

PR #1408, opened 2016-05-27
The `include_parent_namespaces` option is broken

faucct:
Fixed the tests to point that out.

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -464,51 +464,66 @@ describe Grape::Endpoint do
   end
 
   describe '#declared; call from child namespace' do
-    before do
-      subject.format :json
-      subject.namespace :something do
-        params do
-          requires :id, type: Integer
-        end
-        resource ':id' do
+    context 'include_parent_namespaces: true' do
+      before do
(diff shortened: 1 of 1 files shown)


dblock:
Just to save us time, what's the simplest example of where it's broken? Want to attempt a fix and it looks like we don't document it, would love a README entry.

faucct:
The tests have the example:

```ruby
        namespace :something do
          params do
            requires :id, type: Integer
          end
          resource ':id' do
            params do
              requires :happy
              optional :days
            end
            get '/test' do
              {
                params: params,
                declared_params: declared(params, include_parent_namespaces: false)
              }
            end
          end
        end
```

dblock:
What does the above output and what do you expect?

faucct:
Can you check the updated tests, please? It clearly states what behavior I do expect and in which case.

dblock:
Makes sense. If someone is trying to fix this, lets please add specs for nested namespaces as well (parent of parent) and document the option. Looking forward to a PR.

dblock:
Closed via https://github.com/ruby-grape/grape/pull/1415.

PR #1409, opened 2016-06-01
add a method set_common_block

liukgg:
Sometimes people need to add some common block (such as headers) for a 'desc' method in a given class which inherits from class Grape::API. So this method may be useful.

By this method:      
You can set a common block for a specified class.
      It works only when you do not pass a block for the `desc` method.
      
      This can reduce many repeated code when you need to add a same header for many requests.
      You can just put them in the same class which inherits from class Grape::API.
      
      @example
      By the following code, you can add a common header 'Token' for the first two requests.
      But the third one is not added the header 'Token'.
      
          init_common_block do
            headers Token: { description: 'Token', required: true }
          end
      
          desc 'create a user'
          post '/users' do
            # ...
          end
      
          desc 'find a user'
          get '/user/:id' do
            # ...
          end
      
          desc 'update a user' do
          end
          put '/user/:id' do
            # ...
          end

### DIFF
--- a/lib/grape/dsl/desc.rb
+++ b/lib/grape/dsl/desc.rb
@@ -38,7 +38,9 @@ module Grape
       #     end
       #
       def desc(description, options = {}, &config_block)
-        if block_given?
+        config_block = @_common_block if !block_given? && @_common_block
+
+        if block_given? || config_block
           config_class = desc_container
 
           config_class.configure do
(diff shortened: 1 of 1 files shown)


dblock:
I am not a big fan of automatically reused blocks. We already have reusable `params` via `use`. Maybe we want a similar approach for the rest of the things?

liukgg:
Thanks for your reply!

I'm sorry but I've not known exactly what you mean by saying  "We already have reusable params via use.".

However, there are two reasons that make me want to add this method:
1,"reusable params via use "  will add something in the url rather than in the HTTP Header. Am I right?
In my project, I'd like to add something in the Header, such as "Token" which is for security. And this is something I refer to:
https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#General_Recommendation:_Synchronizer_Token_Pattern

2,I want to only add common headers for specific class which inherits form Grape::API, not all classes under a version. So in my opinion, this new method may make things flexible.

Thanks again!

dblock:
I see the need for shared headers (or shared anything). What I am saying is that I'd rather see the DSL be the same for `headers` as for `params`, aka:

```ruby
helpers do
   headers :owasp do
       header ...
   end
end

headers do
  use :owasp
end
```

dblock:
Cool. Lets close this for now.

liukgg:
ok.

PR #1412, opened 2016-06-06
helper "per_page" mistyped (per)

emmanueldd:


dblock:
Actually this is taken from the syntax that Kaminari uses, https://github.com/amatsuda/kaminari. So it's "by design". Thanks for bringing this up, lmk if I am wrong of course :)

Issue #1400, opened 2016-05-16
API version can't take an array of versions on 0.16.2


dangnm:
This code works
```
 version 'v1', using: :path, vendor: 'api'
    resource :test do
      desc "Returns pong."
      get "ping" do
        { ping: "pong" }
      end
    end
```

This code doesn't work but it can work on 0.15.0
  
```
 version %w(v2 v1), using: :path, vendor: 'api'
    resource :test do
      desc "Returns pong."
      get "ping" do
        { ping: "pong" }
      end
    end
```

> Started GET "/api/v1/test/ping" for ::1 at 2016-05-16 18:50:34 +0700
> 
> ActionController::RoutingError (No route matches [GET] "/api/v1/test/ping"):
>   actionpack (4.2.0) lib/action_dispatch/middleware/debug_exceptions.rb:21:in `call'
>   web-console (2.2.1) lib/web_console/middleware.rb:39:in `call'
>   actionpack (4.2.0) lib/action_dispatch/middleware/show_exceptions.rb:30:in `call'
>   railties (4.2.0) lib/rails/rack/logger.rb:38:in `call_app'
>   railties (4.2.0) lib/rails/rack/logger.rb:20:in `block in call'
>   activesupport (4.2.0) lib/active_support/tagged_logging.rb:68:in `block in tagged'
>   activesupport (4.2.0) lib/active_support/tagged_logging.rb:26:in `tagged'
>   activesupport (4.2.0) lib/active_support/tagged_logging.rb:68:in `tagged'
>   railties (4.2.0) lib/rails/rack/logger.rb:20:in `call'
>   quiet_assets (1.1.0) lib/quiet_assets.rb:27:in `call_with_quiet_assets'
>   request_store (1.2.0) lib/request_store/middleware.rb:8:in `call'
>   actionpack (4.2.0) lib/action_dispatch/middleware/request_id.rb:21:in `call'
>   rack (1.6.4) lib/rack/methodoverride.rb:22:in `call'
>   rack (1.6.4) lib/rack/runtime.rb:18:in `call'
>   activesupport (4.2.0) lib/active_support/cache/strategy/local_cache_middleware.rb:28:in `call'
>   rack (1.6.4) lib/rack/lock.rb:17:in `call'
>   actionpack (4.2.0) lib/action_dispatch/middleware/static.rb:113:in `call'
>   rack (1.6.4) lib/rack/sendfile.rb:113:in `call'
[...]

PR #1414, opened 2016-06-06
Fix multiple version definitions for path versioning

304:
It is a fix for issue #1400

I created a new spec for this case and found out why it was not working.

### DIFF
--- a/lib/grape/dsl/routing.rb
+++ b/lib/grape/dsl/routing.rb
@@ -28,23 +28,23 @@ module Grape
         #
         def version(*args, &block)
           if args.any?
-            args.flatten!
             options = args.extract_options!
             options = options.reverse_merge(using: :path)
+            requested_versions = args.flatten
 
             raise Grape::Exceptions::MissingVendorOption.new if options[:using] == :header && !options.key?(:vendor)
 
-            @versions = versions | args
+            @versions = versions | requested_versions
 
             if block_given?
               within_namespace do
-                namespace_inheritable(:version, args)
+                namespace_inheritable(:version, requested_versions)
                 namespace_inheritable(:version_options, options)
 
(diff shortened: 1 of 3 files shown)


spec/grape/api_spec.rb
```diff
@@ -313,6 +315,13 @@ def subject.enable_root_route!
         it 'with a format' do
           get '/v1/.json'
         end
+
+        context 'when array of versions provided' do
```
dblock: The above should be in its own contect context since for example `let(:version)` isn't used here.

lib/grape/dsl/routing.rb
```diff
@@ -28,6 +28,7 @@ module ClassMethods
         #
         def version(*args, &block)
           if args.any?
+            args.flatten!
```
dblock: You should not be modifying the parameters passed in, so something like `args = args.flatten`. However, do investigate why you're getting an array of arrays here, it's possible that there's a better fix upstream.

spec/grape/api_spec.rb
304: Done => b0b9586

lib/grape/dsl/routing.rb
304: Good point. I've extracted version parameter into local variable to describe the intention in a more explicit way => 82e1f21

In this case the upstream is defined by user. I've checked Grape documentation, but I've not found any examples how to properly define several versions for the endpoint. Looks like it is an undocumented feature that was supporting arrays before routing refactoring.

Should it be like that `version ['v1', 'v2']` or like that `version 'v1', 'v2'` or it is not a correct way to define a version at all?

dblock:
This is good, merging.

PR #1415, opened 2016-06-06
Fix logic for `include_parent_namespaces` option

304:
Based on PR #1408.

* Add specs to test different combinations of namespaces
* Add documentation for the `include_parent_namespaces` option

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [#1380](https://github.com/ruby-grape/grape/pull/1380): Fix `allow_blank: false` for `Time` attributes with valid values causes `NoMethodError` - [@ipkes](https://github.com/ipkes).
 * [#1384](https://github.com/ruby-grape/grape/pull/1384): Fix parameter validation with an empty optional nested `Array` - [@ipkes](https://github.com/ipkes).
 * [#1414](https://github.com/ruby-grape/grape/pull/1414): Fix multiple version definitions for path versioning - [@304](https://github.com/304).
+* [#1415](https://github.com/ruby-grape/grape/pull/1415): Fix `declared(params, include_parent_namespaces: false)` - [@304](https://github.com/304).
 
 0.16.2 (4/12/2016)
 ==================
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -29,7 +29,10 @@ module Grape
         def declared(params, options = {}, declared_params = nil)
           options = options.reverse_merge(include_missing: true, include_parent_namespaces: true)
 
-          declared_params ||= (!options[:include_parent_namespaces] ? route_setting(:declared_params) : (route_setting(:saved_declared_params) || [])).flatten(1) || []
+          all_declared_params = route_setting(:declared_params)
+          current_namespace_declared_params = route_setting(:saved_declared_params).last
+
+          declared_params ||= options[:include_parent_namespaces] ? all_declared_params : current_namespace_declared_params
 
           raise ArgumentError, 'Tried to filter for declared parameters but none exist.' unless declared_params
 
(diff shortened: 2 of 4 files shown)


lib/grape/dsl/inside_route.rb
```diff
@@ -29,7 +29,10 @@ module PostBeforeFilter
         def declared(params, options = {}, declared_params = nil)
           options = options.reverse_merge(include_missing: true, include_parent_namespaces: true)
 
-          declared_params ||= (!options[:include_parent_namespaces] ? route_setting(:declared_params) : (route_setting(:saved_declared_params) || [])).flatten(1) || []
+          all_declared_params = route_setting(:declared_params)
+          current_namespace_declared_params = route_setting(:saved_declared_params).last
+
+          declared_params ||= options[:include_parent_namespaces] ? all_declared_params : current_namespace_declared_params
```
304: I think it won't harm to provide a short explanation what is going on here.

`route_setting(:declared_params)` keep every parameter in a flat array, for example:
```
route_setting(:declared_params) # [:parent_name, :child_name, :child_age, :grandchild_name]
```
However, `route_setting(:saved_declared_params)` stores parameters in groups by namespace.
```
route_setting(:saved_declared_params) # [[:parent_name], [:child_name, :child_age], [:grandchild_name]]
```
The last element here contains only parameters that were defined in current namespace.

lib/grape/dsl/inside_route.rb
```diff
@@ -29,7 +29,10 @@ module PostBeforeFilter
         def declared(params, options = {}, declared_params = nil)
           options = options.reverse_merge(include_missing: true, include_parent_namespaces: true)
 
-          declared_params ||= (!options[:include_parent_namespaces] ? route_setting(:declared_params) : (route_setting(:saved_declared_params) || [])).flatten(1) || []
+          all_declared_params = route_setting(:declared_params)
+          current_namespace_declared_params = route_setting(:saved_declared_params).last
```
faucct: Something can be broken, because you have removed that `route_setting(:saved_declared_params) || []` piece.

faucct: I think that just replacing `flatten(1)` with `last` is enough.

dblock:
I like this. This just needs a CHANGELOG entry, please.

lib/grape/dsl/inside_route.rb
304: I tried to investigate the reason why these additional checks were introduced, but it's still unclear for me (PR #739 and specifically [here](https://github.com/ruby-grape/grape/pull/739/files#diff-e9e8424a5238d48301e313d8fe285697R22)).

I also played around endpoint structure to detect the case when `flatten` was really needed. But specs were green no matter what I did. I would personally prefer to clean up the code that does not break tests. What do you think?

304:
@dblock, CHANGELOG updated => 8a7a24c

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 * [#1365](https://github.com/ruby-grape/grape/pull/1365): Fix finding exception handler in error middleware - [@ktimothy](https://github.com/ktimothy).
 * [#1380](https://github.com/ruby-grape/grape/pull/1380): Fix `allow_blank: false` for `Time` attributes with valid values causes `NoMethodError` - [@ipkes](https://github.com/ipkes).
 * [#1384](https://github.com/ruby-grape/grape/pull/1384): Fix parameter validation with an empty optional nested `Array` - [@ipkes](https://github.com/ipkes).
+* [#1415](https://github.com/ruby-grape/grape/pull/1415): Fix logic for `include_parent_namespaces` option - [@304](https://github.com/304).
```
dblock: It doesn't say where, it's in `declared`, maybe say "Fix `declared(params, include_parent_namespaces: true)`"?

dblock:
Merged via https://github.com/ruby-grape/grape/commit/7d1b3e334f5d3d5b045e9a48b08a8f1ed03a9f5e.

Issue #1413, opened 2016-06-06
API versioning strict mode not returning a 406 as expected


nicholaspufal:
I created a very simple endpoint to make it easier to spot this issue. Say I have the following API:

```ruby
module MyApp
  class API < Grape::API
    format :json
    version 'v1', using: :accept_version_header, strict: true

    get 'healthcheck' do
      status 200
    end
  end
end
```

And the following tests:

```ruby
describe 'MyApp::API' do
  def app
    MyApp::API
  end

  describe 'versioning' do
    it 'fails if the client does not specify a version' do
      get '/healthcheck'

      expect(last_response.status).to eq(406)
    end

    it 'returns successfully if a version if specified' do
      header 'Accept-Version', 'v1'

      get '/healthcheck'

      expect(last_response.status).to eq(200)
    end
  end
end
```

What's happening is that for the first scenario it's returning a `404`. 

After investigating a bit deeper the issue it seems like that `versioner/accept_version_header.rb` is doing the right job - I can see a 406 from there, however later in the process `Grape::Middleware::Error#error_response` appends a `X-Cascade: pass` to the headers and that seems to be used later by `Grape::Router#transaction` to search for another match of that route (lines 86 to 90 in `lib/grape/router.rb`). As the router can't find it ends up returning a `404`.

I'm not exactly sure what's suppose to be going on as part of that last stage, so maybe someone with more context on it can help me here.

The workaround for now is commented out as part of https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/versioner/accept_version_header.rb#L48

So in my example, it would mean that this line:

```ruby
version 'v1', using: :accept_version_header, strict: true
```

...would become this:

```ruby
version 'v1', using: :accept_version_header, strict: true, cascade: false
```

That work wonders but it's still unclear to me why that behavior is not default.

Thoughts?

Thanks!
Nicholas

nicholaspufal:
After taking another look at the docs, this behavior is actually being mentioned:

>When an invalid Accept header is supplied, a 406 Not Acceptable error is returned if the :cascade option is set to false. Otherwise a 404 Not Found error is returned by Rack if no other route matches.

Wouldn't the other way around make more sense? Make the default `cascade: false`?

dblock:
It could make more sense, but I bet the reason is historical and I would vote not to change this behavior so that everybody doesn't have to suffer. If you agree I think we should close this?

nicholaspufal:
[...]

PR #1417, opened 2016-06-08
Improve docs related to the `Accept-Version Header`

nicholaspufal:
This is related to issue #1413 

It's not clear by reading the docs that when using the `Accept-Version header` a `406` will only be returned IF `:cascade` is set to `false`.This PR aims to make that behavior more visible in the docs.


dblock:
I'll take it.

Issue #1420, opened 2016-06-09
helpers injection pollutes Grape::Middleware::Error


etehtsea:
[This](https://github.com/ruby-grape/grape/commit/951ae51cc0eccec6c5225c2aaf16f091829062dd) commit introduces very strange issues in our projects:

1. helpers are including to the middleware which is global to all code that use Grape. So if you have two helpers with the same name you are in trouble.

  ```
  >> auth = session(Auth)
  => #<Rack::Test::Session:0x007fb6b582abe0 @headers={}, @env={}, @rack_mock_session=#  <Rack::MockSession:0x007fb6b582ac30 @app=Auth, @after_request=[], @default_host="example.org", @last_request=nil, @last_response=nil>, @default_host="example.org">
  >> auth.get('/').body
  => "auth"
  >> $sessions.each { |s| s.get('/').body }; nil
  => nil
  >> auth.get('/').body
  => "API"
  ```
  Upsy daisy!

2. Historically in order to not copy pasting `format`/`helpers` and other to each endpoint we have `Base` endpoint like in example from which we subclass all other. After this commit `helpers` are injecting over and over and ancestors list grows from `~11` to `N * endpoints + 11`.

  ```
  >> Grape::Middleware::Error.ancestors.count
  => 10
  >> session(Auth).get('/')
  >> Grape::Middleware::Error.ancestors.count
  => 10
  >> $sessions.each { |s| s.get('/') }; nil
  => nil
  >> Grape::Middleware::Error.ancestors.count
  => 210
  ```
  Possibly this is not an issue by itself and could be fixed by moving helpers to the root class where all other endpoints are mounted, but I could not get it to work (#1418).

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'grape', '>= 0.15.0'
  gem 'rack-test'
  gem 'test-unit'
  gem 'pry'
end

require 'grape'
require 'rack/test'
require 'test/unit'

class Auth < Grape::API
  helpers do
    extend Grape::API::Helpers

    params :foo do
      optional :ololo
    end

    def respond_error!
      error!(format_err_msg, 401)
    end

    def format_err_msg
      'auth'
    end
  end

  rescue_from RuntimeError do |_|
    respond_error!
  end

  get do
    raise 'auth error'
  end
end

class Base < Grape::API
  def self.inherited(subclass)
    super

    subclass.instance_eval do
      rescue_from RuntimeError do |_|
        respond_error!
      end

[...]

PR #1421, opened 2016-06-09
Avoid polluting Grape::Middleware::Error

namusyaka:
ref #1420 
/cc @etehtsea @dblock

### DIFF
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -1456,6 +1456,36 @@ XML
       expect(last_response.body).to eq 'hello bob'
     end
 
+    context 'with multiple apis' do
+      let(:a) { Class.new(Grape::API) }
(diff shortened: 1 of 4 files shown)


spec/grape/api_spec.rb
```diff
@@ -1456,6 +1456,34 @@ def custom_error!(name)
       expect(last_response.body).to eq 'hello bob'
     end
 
+    it 'avoids polluting global namespace' do
+      class A < Grape::API
```
dblock: These should be `let(:A) { Class.new(...) }` to avoid polluting (see what I did there :) the global namespace.

namusyaka: OMG you're right ;) Thanks!

dblock:
Merged.

PR #1422, opened 2016-06-11
Concatenate parent declared params with current one

namusyaka:
ref #1085 
I cherry-picked changes from @plukevdh and @rnubel commits, and fixed a few points.
The issue should be fixed in this pull request.

@dblock Could you review this?
/cc @plukevdh @rnubel


dblock:
Looks great, merging.

PR #1425, opened 2016-06-14
Run tests with warnings enabled

bf4:
Should help clean up warnings in gems or apps that use Grape ;)

refs:
- http://mislav.net/2011/06/ruby-verbose-mode/


dblock:
I looked at the output in https://travis-ci.org/ruby-grape/grape/jobs/137408689 and there're some juicy ones that seem like they should be fixed. Would you like to address them as part of this PR?

bf4:
Depends on how soon you'd like this merged... :) we run AMS tests w warnings on and grape is main offender, but otherwise I'm juggling a bunch of stuff right now (as are we all)

B mobile phone

dblock:
I'll wait or will pick this up. I agree that this is important.

Issue #1424, opened 2016-06-13
What is expected before/after workflow for 405/Method not found?


jnardone:
When in the workflow should a 405 be returned for the wrong method?  At what point with regards to the various handlers should the code exit and return?

Based on searches of various issues (including #1402) I would expect that the "before" handler would run, and that's it.  

However, that does not seem to be the case.  It appears that "before", "before_validation", and "after_validation" are run before I get the 405.  This leads to weird issues where common code in these handlers gets called (and perhaps returns a different error, e.g. a 404 for invalid route data) even though the more fundamental issue is the 405.

(It is also interesting to note that in 0.15.0 headers set in before/before_validation/after_validation are returned with the 405 error, but in 0.16.2 only the headers set in 'before' are returned, even though all 3 blocks are still called.)

    module Foo
      class API < Grape::API
        prefix 'api'
        version 'v1', using: :path
        format :json
        do_not_route_head!

        before do
          puts 'in before'
          header 'x-b', 'before'
        end

        before_validation do
          puts 'in before_validation'
          header 'x-bv', 'before_validation'
        end

        after_validation do
          puts 'in after_validation'
          header 'x-av', 'after_validation'
        end

        after do
          puts 'in after'
          header 'x-a', 'after'
        end

        get do
          puts "in do"
        end
      end
    end

If I do a PUT to /api/v1 I will only get the "x-b" header (expected) but "before", "before_validation", "after_validation" will all run (unexpected).  However, there is no documented explanation of what should happen.  So... what should happen?

jlfaber:
I added some debug trace to the grape specs, and it looks like the `before` filter is actually getting called twice: once by the `method_not_allowed` method in `router.rb`, and then again as part of the endpoint evaluation in `endpoint.rb`.  For some reason, the side-effects (like header setting) of the latter call are not "sticking".  This also means the other two filters (`before_validation` and `after_validation`) are also getting called stealthily.

Here's a spec that demonstrates the issue.

````ruby
it 'skips before_validation filter on bad method' do
  subject.namespace :test do
    before { header 'X-Custom-Header1', 'foo' }
    before_validation { header 'X-Custom-Header2', 'foo' }
    # before_validation { raise 'THIS SHOULD NOT HAPPEN' }
    get
  end

  post 'test'
  expect(last_response.status).to eql 405
  expect(last_response.headers['X-Custom-Header1']).to eql 'foo'
  expect(last_response.headers['X-Custom-Header2']).to be_nil
end
[...]

PR #1426, opened 2016-06-14
New tests to demonstrate #1424

jlfaber:
This presumes that only the before filter should run when an
unsupported method is called on an endpoint, and that other filters
(before_validation, after_validation, and after) should be skipped.

It also demonstrates that the before filter is called more than
once in that case.


dblock:
Closed via https://github.com/ruby-grape/grape/pull/1465.

PR #1430, opened 2016-06-16
specs for declared(params) inside route_param bug.

Arkanain:
I check some specs on my project and I find out that this functionality work little bit wrong with declared when we have several route_param in one namespace.
For example we have:

```ruby
class Artists
  namespace :artists do
    route_param :id, type: Integer do
      get do
        declared(params)
      end

      params do
        requires :filter, type: String
      end
      get :some_route do
        declared(params)
      end
    end
    route_param :artist_id, type: Integer do
      mount Compositions
    end
  end
end
```
`declared(params)` should return us next hash `{id: 1}`, but it return hash `{id: 1, artist_id: nil}`.
Also in first `route_param` in second endpoint(`get :some_route`) when we call `declared(params)` we have only `filter` params in returned hash `{filter: 'some_filter'}`

P.S.: I find out where we have problem! Problem is in options for route_param. where I use route_param :id without type: Integer everything is ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 #### Fixes
 
+* [#1430](https://github.com/ruby-grape/grape/pull/1430): Fix for `declared(params)` inside `route_param` - [@Arkanain](https://github.com/Arkanain).
 * [#1405](https://github.com/ruby-grape/grape/pull/1405): Fix priority of `rescue_from` clauses applying - [@hedgesky](https://github.com/hedgesky).
 * [#1365](https://github.com/ruby-grape/grape/pull/1365): Fix finding exception handler in error middleware - [@ktimothy](https://github.com/ktimothy).
 * [#1380](https://github.com/ruby-grape/grape/pull/1380): Fix `allow_blank: false` for `Time` attributes with valid values causes `NoMethodError` - [@ipkes](https://github.com/ipkes).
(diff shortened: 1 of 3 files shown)


Arkanain:
@dblock @namusyaka please review this.

dblock:
Well these specs pass... weren't they supposed to fail?

Arkanain:
@dblock It's a regression specs. They represent a problem. Or you want to see a regular specs instead of regression specs?

Arkanain:
@dblock change specs. not it failed and show problem correct.

Arkanain:
Hi guys. Any news about this issue?

Arkanain:
@dblock @namusyaka  Please review my code. I find out where was a problem and fixed it. I really need this fix in my project so if everything ok can you please merge it ASAP.

namusyaka:
Looks good to me.
@dblock Any thoughts on this?

dblock:
Looks good, needs a CHANGELOG entry please.

Arkanain:
@dblock Done.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 #### Fixes
 
+* [#1430](https://github.com/ruby-grape/grape/pull/1430): Fix for declared(params) inside route_param. - [@Arkanain](https://github.com/Arkanain).
```
dblock: Sorry to be a pain. There shouldn't be a period after `route_param`, also maybe quote the code as below?

Arkanain:
@dblock update.

dblock:
Excellent, merging.

Issue #1438, opened 2016-07-12
Modifying params in the endpoint will change the default


ashkan18:
When using `default` of a param, modifying the param within the request will change the default. Here is a spec for the issue:
```ruby
require 'spec_helper'

module FreezeDefaultForParamsSpec
  
  class Users < Grape::API
    resource :users do
      params do
        optional :sort, default: '-created_at'
      end
      get do
        params[:sort].slice!(0)
        params[:sort]
      end
    end
  end
  
  class Main < Grape::API
    mount Users
  end
end

describe Grape::API do
  subject { FreezeDefaultForParamsSpec::Main }

  def app
    subject
  end

  it 'should not change the default for the param' do
    get '/users'
    expect(last_response.status).to eql 200
    expect(last_response.body).to eql 'created_at'

    get '/users'
    expect(last_response.status).to eql 200
    expect(last_response.body).to eql 'created_at'
  end
end
```
For the first `get` in the spec, default is correct which is `-created_at` but on the second get, default sort param has changed to `created_at` because of `params[:sort].slice!(0)` in `get`. 

Ideally we'd need to pass a `clone` of params, as an alternative we may need to `freeze` default.

dblock:
I think we probably don't want to clone the default every time, which could be unnecessarily expensive for every request, but definitely freeze it.

jlfaber:
@dblock If we freeze default, that might cause existing client code to suddenly start throwing RuntimeErrors (if they are mutating param values).  Admittedly, they're probably not getting the behavior they expect on subsequent calls after the default has been modified.  So there's already an error present.  But it seems to me that the best course **is** to clone the default when it is assigned to a param.

This is also appealing because it means that params are uniformly modifiable whether they have a default value or not.  Freezing the default would cause the two cases to behave differently, which I think is a much more confusing semantic.

Also, I haven't actually verified it, but I think the test spec above will only pass if default is cloned.  If we freeze default, the slice! call is going to raise a RuntimeError.

dblock:
I worry about unintentional performance overheads, but maybe I am worried about nothing. Either way I'll take a PR.

dblock:
This was fixed in https://github.com/ruby-grape/grape/commit/8b721447fa6a92aa42e4c99b1e32a38b12b96ea7.

ashkan18:
awesome! Thanks! 👍

PR #1439, opened 2016-07-14
Clone non-frozen default params with each use (#1438)

jlfaber:


jlfaber:
This appears to be failing because of #1440.  Once that fix merges, I'll rebase to pick it up and re-run the CI build.

jlfaber:
Rebased.  All tests now passing, so this is ready for review.  @dblock

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
  
 #### Fixes
 
+* [#1438](https://github.com/ruby-grape/grape/pull/1439): Fix for default param modification bug - [@jlfaber](https://github.com/jlfaber).
```
dblock: Lets write something more useful, maybe "Default values in `params` are now cloned on assignment"?

lib/grape/validations/validators/default.rb
```diff
@@ -7,7 +7,24 @@ def initialize(attrs, options, required, scope)
       end
 
       def validate_param!(attr_name, params)
-        params[attr_name] = @default.is_a?(Proc) ? @default.call : @default unless params.key?(attr_name)
+        return if params.key? attr_name
+        params[attr_name] = if @default.is_a? Proc
```
dblock: Lots of nested `if`s and `rescue`, refactor this into a method?

lib/grape/validations/validators/default.rb
```diff
@@ -7,7 +7,24 @@ def initialize(attrs, options, required, scope)
...
+                                  @default == false ||
+                                  @default.is_a?(Symbol) ||
+                                  @default.is_a?(Numeric)
+                              # We know clone won't work in these cases, so just use the object.
+                              @default
+                            else
+                              # for the rest, try to clone but use the object if that fails.
+                              begin
+                                @default.clone
+                              rescue TypeError
```
dblock: When does this happen? Why `TypeError` and not anything else?

dblock: Also, `clone` retains `frozen`, I just want to ensure that this is what we want here. Maybe it should be `dup`?

jlfaber: If the default value is `frozen` to begin with, we could safely continue to use it as-is since there's no danger of side-effects via modification (which is the bug we're fixing). Doing that would not be a change from how things work today, so it won't surprise anyone. 

And, if the user really wants the default to be a `dup` of a `frozen` object, they have the ability to do that by using a `Proc`.  In fact, I kind of like the idea that the user can use the `frozen` state of the default object to control whether or not Grape should expend resources cloning it each time it's used.

When the default value is _not_ frozen, I agree that `dup` is probably preferable to `clone`.

jlfaber: Regarding `TypeError`, I think that's what normally gets raised when an object is not allowed to be cloned or duped.  I hesitate to expand the rescue for fear of masking other errors.

jlfaber:
@dblock, I've updated the PR:
- simplify `validate!` by refactoring
- reword CHANGELOG message
- use `dup` instead of `clone`
- use `frozen` default values as-is, without `dup`ing, since they're not subject to the bug.  (See line notes for more on this).

This should be ready to go if you concur with the changes.

lib/grape/validations/validators/default.rb
```diff
@@ -6,8 +6,30 @@ def initialize(attrs, options, required, scope)
         super
       end
 
+      # return true if we know we cannot dup this object
+      private def cannot_dup(obj)
```
dblock: Should be `....?`, but I would write this as the other way around, so maybe `duplicatable?`, which is how ActiveRecord calls these things I believe.

lib/grape/validations/validators/default.rb
```diff
@@ -6,8 +6,30 @@ def initialize(attrs, options, required, scope)
         super
       end
```
dblock: In Ruby `private def` does nothing AFAIK, it treats it as two separate things. Write:

```ruby
private

def duplicatable?(...)
...
end

def duplicate(...)
end
```

dblock:
This is good. I have nitpicks on naming, otherwise it's ready to go.

lib/grape/validations/validators/default.rb
jlfaber: `private def` works fine for me.  It causes just that method to be marked private without altering the visibility of subsequent methods.

````ruby
$ irb
irb(main):001:0> class Foo
irb(main):002:1>   private def bar
irb(main):003:2>     'bar'
irb(main):004:2>   end
irb(main):005:1>   def baz
irb(main):006:2>     'baz'
irb(main):007:2>   end
irb(main):008:1> end
=> :baz
irb(main):009:0> foo = Foo.new
=> #<Foo:0x007fcae281d800>
irb(main):010:0> foo.baz
=> "baz"
irb(main):011:0> foo.bar
NoMethodError: private method `bar' called for #<Foo:0x007fcae281d800>
````
But I'm happy to break it up if that's what you'd prefer.

jlfaber:
Renamed and reformatted as requested by @dblock .

dblock:
Merged via https://github.com/ruby-grape/grape/commit/8b721447fa6a92aa42e4c99b1e32a38b12b96ea7, cc: @ashkan18 check it out.

Issue #1440, opened 2016-07-14
activesupport 5.0.0 breaks CI tests


jlfaber:
Seeing these errors on ruby versions below 2.2.2:

> Gem::InstallError: activesupport requires Ruby version >= 2.2.2.
> An error occurred while installing activesupport (5.0.0), and Bundler cannot continue.
> Make sure that `gem install activesupport -v '5.0.0'` succeeds before bundling.

Looks like grape was picking up 4.2.6 prior to June 30 when 5.0.0 came out.

There's no version currently listed for activesupport in grape.gemspec.  Perhaps it should be changed to this?

`  s.add_runtime_dependency 'activesupport', '~> 4.2'`

jlfaber:
Using ~> 4.2 didn't work since rails actually uses older versions of activesupport.  Trying again with < 5.0.0.

jlfaber:
That seems to have fixed it.  I'll force-push a revision to add a CHANGELOG line and correct the commit message, then I think this is good to go.  (Assuming everyone is ok with holding off on taking activesupport 5.0.0 for now.)

dblock:
Just bringing my comment from #1441, lets not restrict grape to AS 4.x, but fix the build by removing incompatible rubies.

jlfaber:
Updated #1441 to remove pre-2.2.2 rubies and add explicit checks for 2.2.2 - 2.3.1

PR #1441, opened 2016-07-14
Remove pre-2.2.2 rubies since they no longer work with AS 5 (#1440)

jlfaber:


dblock:
Grape itself is compatible with AS 5. When we add it to .gemspec we're saying that it's not, which isn't true. I would take a PR that fixes the build and removes old(er) rubies instead?

jlfaber:
Changed commit to remove pre-2.2.2 versions of ruby (2.2, 2.1, 2.0.0, jruby-19mode) from .travis.yml, and replace them with 2.2.2, 2.2.3, 2.2.4, 2.2.5 and 2.3.1

.travis.yml
```diff
@@ -3,12 +3,13 @@ language: ruby
 sudo: false
 
 rvm:
+  - 2.3.1
   - 2.3.0
-  - 2.2
-  - 2.1
-  - 2.0.0
+  - 2.2.5
```
dblock: I think we don't need to build all those, how about just the latest 2.2 (does just 2.2 work)?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
  
 #### Fixes
 
+* [#1440](https://github.com/ruby-grape/grape/pull/1441): Remove pre-2.2.2 rubies from .travis.yml - [@jlfaber](https://github.com/jlfaber).
```
dblock: While this is technically correct, what we're doing is officially dropping support for Ruby < 2.2, so this should be in the features section and say something like "Supporting Ruby 2.2 or newer". I also think we could use a paragraph in UPGRADING about it.

.travis.yml
jlfaber: I think '2.2' will pick up 2.2.0, not 2.2.5 (which is the latest 2.2.x release).  How about 2.2.5 and 2.3.1?  (omitting 2.2.2, 2.2.3, 2.2.4, and 2.3.0)

dblock: Sounds good. I would leave 2.3.0 just cause it's recent enough.

CHANGELOG.md
jlfaber: That's a good point.  Although what we're actually doing is dropping support for 2.2.0 and 2.2.1 as well (i.e. Ruby < 2.2**.2**).  I'll adjust the PR.

jlfaber: Also, I wonder whether this change is worthy of a bump to 0.17.0 since it is technically a breaking change for people on older rubies.

jlfaber:
Updated PR to build only 2.2.5, 2.3.0, 2.3.1.  Moved CHANGELOG entry to features section. Added text in UPGRADING.

dblock:
Merged.

Issue #1395, opened 2016-05-12
Extend `given` to check against the value


danieldraper:
It would be nice to allow a value to be specified on a `given` so that different parameters could be included for example:

``` ruby
params do
  requires :grant_type, type: String, values: [:password, :refresh_token]
  requires :client_id, type: String

  given :grant_type, value: ->(*) { |attr| attr == :password } do
    requires :username, type: String
    requires :password, type: String
  end

  given :grant_type, value: ->(*) { |attr| attr == :refresh_token } do
    requires :refresh_token, type: String
  end
end
```

dblock:
I like it.

danieldraper:
I'm happy to look into implementation, and pointers on accessing the value from https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/parameters.rb or should I be looking elsewhere?

rnubel:
@danieldraper that section of code is executed right when your API loads, and is what defines the new parameter scope with a `dependent_on` value. You'll need to update that to save the `value` lambda for later reference, and update [ParamsScope#should_validate?](https://github.com/ruby-grape/grape/blob/7e6daeb0d86a476dbf9c15668aa53a1109e45532/lib/grape/validations/params_scope.rb#L41-42) to verify that, if the scope is dependent on an attribute, it also satisfies the `value` lambda.

Zooming out a bit, right now, `given` can take more than one parameter. You could either enforce the value check for *all* dependent parameters, or use a different format (maybe something like,
```given foo: ->(val) { val == :bar }, baz: ->(val) { val == :qaz }```
... though it's a bit limiting and not as clear).

glaucocustodio:
@danieldraper can you check https://github.com/ruby-grape/grape/pull/1443 and close this issue please? Thanks.

towanda:
@dblock Could you close this, please?

dblock:
Closing via #1443.

PR #1442, opened 2016-07-18
Allow given to receive a proc

glaucocustodio:
Implements https://github.com/ruby-grape/grape/issues/1395.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@
 * [#1390](https://github.com/ruby-grape/grape/pull/1390): Allow inserting middleware at arbitrary points in the middleware stack - [@Rosa](https://github.com/Rosa).
 * [#1366](https://github.com/ruby-grape/grape/pull/1366): Store `message_key` on `Grape::Exceptions::Validation` - [@mkou](https://github.com/mkou).
 * [#1398](https://github.com/ruby-grape/grape/pull/1398): Added `rescue_from :grape_exceptions` - allow Grape to use the built-in `Grape::Exception` handing and use `rescue :all` behavior for everything else - [@mmclead](https://github.com/mmclead).
+* [#1442](https://github.com/ruby-grape/grape/pull/1442): Extended `given` to receive a `Proc` - [@glaucocustodio](https://github.com/glaucocustodio).
 * Your contribution here.
- 
+
 #### Fixes
 
 * [#1430](https://github.com/ruby-grape/grape/pull/1430): Fix for `declared(params)` inside `route_param` - [@Arkanain](https://github.com/Arkanain).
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,8 +8,9 @@
 * [#1390](https://github.com/ruby-grape/grape/pull/1390): Allow inserting middleware at arbitrary points in the middleware stack - [@Rosa](https://github.com/Rosa).
 * [#1366](https://github.com/ruby-grape/grape/pull/1366): Store `message_key` on `Grape::Exceptions::Validation` - [@mkou](https://github.com/mkou).
 * [#1398](https://github.com/ruby-grape/grape/pull/1398): Added `rescue_from :grape_exceptions` - allow Grape to use the built-in `Grape::Exception` handing and use `rescue :all` behavior for everything else - [@mmclead](https://github.com/mmclead).
+* [#1442](https://github.com/ruby-grape/grape/pull/1442): Allow `given` to receive a `Proc` to check more conditions other than presence only - [@glaucocustodio](https://github.com/glaucocustodio)
```
dblock: This is missing a period at the end. Also maybe just shorten it to "Extended `given` to receive a `Proc`".

README.md
```diff
@@ -987,6 +987,21 @@ params do
 end
 ```
 
+In the example above Grape will use `blank?` to check if the `shelf_id` param is present.
```
dblock: if -> whether?

README.md
```diff
@@ -987,6 +987,21 @@ params do
...
+
+But you can use a `Proc` with custom code as well. If it returns true, the block will be evaluated.
```
dblock: This is documentation, so keep things succinct. Just continue the previous phrase with something like "given also takes a proc". You don't need to say "if it returns true", because that's not ... true, anything truthy will work. Anyway, just try to say the most with the least amount of words :)

README.md
```diff
@@ -987,6 +987,21 @@ params do
...
+```ruby
+params do
+  optional :category
+  given category: ->(val) { val == 'foo' } do
+    requires :description
+  end
+end
+```
+
+The param `description` is required only if the value of `category` is equal `foo`.
```
dblock: Try to move this up to the text before the example.

dblock:
Code is good. My comments are all about the text. Amend and I'll merge. Appreciate if you could squash the commits to to keep the git history clean.

glaucocustodio:
I failed to squash, git created two more commits. Do you know how to solve it? I can open a new PR with all the changes just in one commit as well.

glaucocustodio:
@dblock please take a look at https://github.com/ruby-grape/grape/pull/1443, you can delete this closed PR.

PR #1446, opened 2016-07-19
Correctly initialize env and headers

leifg:


lib/grape/router.rb
```diff
@@ -136,6 +136,13 @@ def method_not_allowed(env, methods, endpoint)
       env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED] = true
       current = endpoint.dup
       current.instance_eval do
+        @env = env
+        @header = {}
+
+        @request = Grape::Request.new(env)
+        @params = @request.params
+        @headers = @request.headers
```
dblock: I am merging this. But I wonder whether we can introduce a module/class/code that can be injected here and in other places where these properties would be memoized the first time we request them. Especially the construction of `Grape::Request` is not free as it parses the environment.

dblock:
Thanks much @leifg. Merged. See my comment in case we can improve this further.

PR #1447, opened 2016-07-19
Failing UT spec to demonstrate missing params with mounts (#1085)

jlfaber:
This spec demonstrates a variant of #1085 that appears not to have been fixed by #1422 (@namusyaka).

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -655,6 +655,38 @@ describe Grape::Endpoint do
     end
   end
 
+  describe '#declared; from a nested mounted endpoint - case 2' do
+    before do
+      mounted = Class.new(Grape::API) do
+        params { requires :inner_val, type: Integer }
+        get { declared(params) }
+      end
+
+      subject.format :json
+      subject.namespace :outer do
+        params { requires :outer_val, type: Integer }
+        get { declared(params) }
+        namespace :inner do
+          mount mounted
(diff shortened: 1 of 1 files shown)


dblock:
This looks legit @jlfaber.

spec/grape/endpoint_spec.rb
```diff
@@ -655,6 +655,38 @@ def app
...
+  describe '#declared; from a nested mounted endpoint - case 2' do
+    before do
+      mounted = Class.new(Grape::API) do
+        params { requires :inner_val, type: Integer }
+        get { declared(params) }
+      end
+
+      subject.format :json
+      subject.namespace :outer do
+        params { requires :outer_val, type: Integer }
```
namusyaka: Well, this should be executed before `namespace :outer`, I guess it should be something like:
```ruby
      subject.format :json
      subject.params { requires :outer_val, type: Integer }
      subject.namespace :outer do
        get { declared(params) }
        namespace :inner do
          mount mounted
        end
```

And then your specs are passed locally. What do you think? /cc @dblock

jlfaber: Yes.  You are correct.  I had mistakenly thought that once a params block was declared, it would remain in effect until superceded by another params block.  That turns out not to be the case; the block is consumed by the get in my example leaving nothing for the mounted endpoint to inherit.

Closing.

PR #1448, opened 2016-07-20
Remove hashie dependency

etehtsea:
Because less dependencies is much better than more.
This is breaking change and I'm open for discussion.

Additional point for doing this is that in our app we are still locked to `hashie ~> 2.1.1` (we aren't using it in the project) just because later versions require `https://github.com/Maxim-Filimonov/hashie-forbidden_attributes` gem. This is kinda weird from my point of view to require "workaround" gem for internal grape dependency.

P.S. Spec suite is green, but I'm not sure that there is no unexpected behavior changes.


etehtsea:
Didn't see that it was already made in https://github.com/ruby-grape/grape/pull/1279
Leaving it open because this PR is based on `HashWithIndifferentAccess`.

dblock:
I think we want this to be a plain `Hash`, *and* we want to provide a way to wrap that with either a `Hashie::Mash` class, `HashWithIndifferentAccess` or whatever else for backward compatibility. I'd take a PR that accomplishes that.

nbulaj:
I agree that Grape needs to delete `Hashie` dependency and work with any any class that maybe respond to `to_h`. @dblock what do you think about that?

dblock:
I'm with you @nbulaj, see above for more detailed comments about what I suggest the plan to move forward could be.

etehtsea:
@dblock I couldn't make it, so please reopen it as issue if needed.

dblock:
Lets leave this open, it's a valid feature request.

etehtsea:
@dblock I meant that this PR won't be ever merged so it will be better to open issue that someone could assign to.

dblock:
Lets leave this as is. Issues, PRs, all the same ;)

dblock:
Closing via #1610.

PR #1449, opened 2016-07-20
Replace multi_json with json

etehtsea:
JSON is good option for default.
It was [made in Rails in 2013](https://github.com/rails/rails/pull/10576).


etehtsea:
Dup of https://github.com/ruby-grape/grape/pull/1292

P.S. Despite the fact that I don't think that it should be configurable at all.

dblock:
@etehtsea On being configurable it's more about backward compatibility than anything else.

dblock:
https://github.com/ruby-grape/grape/pull/1623

Issue #1451, opened 2016-07-20
Helpers injected in rescue_from may conflict with middleware base methods


dblock:
```ruby
require 'spec_helper'

describe Grape::API do
  subject do
    Class.new(Grape::API) do
      format :json
      rescue_from :all do |e|
        rack_response({ message: e.message }, 400)
      end
      helpers do
        def error!(message, status, headers)
          raise 'this should never be called'
        end
      end
      rescue_from Grape::Exceptions::MethodNotAllowed do |e|
        error! e.message, e.status, e.headers
      end
      post do
        { ok: true }
      end
    end
  end

  def app
    subject
  end

  it '201 on post' do
    post '/'
    expect(last_response.status).to eq 201
  end

  it '405s on get' do
    get '/'
    expect(last_response.status).to eq 405
  end
end
```

In Grape 0.16.2 this passes.

In Grape 0.16.3 this fails with 

```
Grape::API
  201 on post
  405s on get (FAILED - 1)

Failures:

  1) Grape::API 405s on get
     Failure/Error: raise 'this should never be called'
     
     RuntimeError:
       this should never be called
     # ./spec/grape/integration/error_in_middleware_spec.rb:12:in `error!'
     # ./spec/grape/integration/error_in_middleware_spec.rb:16:in `block (4 levels) in <top (required)>'
     # ./lib/grape/middleware/error.rb:76:in `instance_exec'
     # ./lib/grape/middleware/error.rb:76:in `exec_handler'
     # ./lib/grape/middleware/error.rb:45:in `rescue in call!'
     # ./lib/grape/middleware/error.rb:32:in `call!'
     # ./lib/grape/middleware/base.rb:23:in `call'
     # /Users/dblock/.rvm/gems/ruby-2.3.1/gems/rack-1.6.4/lib/rack/head.rb:13:in `call'
     # ./lib/grape/endpoint.rb:222:in `call!'
     # ./lib/grape/endpoint.rb:216:in `call'
     # ./lib/grape/router.rb:153:in `method_not_allowed'
     # ./lib/grape/router.rb:98:in `transaction'
     # ./lib/grape/router.rb:67:in `identity'
     # ./lib/grape/router.rb:52:in `block in call'
     # ./lib/grape/router.rb:115:in `with_optimization'
     # ./lib/grape/router.rb:51:in `call'
     # ./lib/grape/api.rb:119:in `call'
     # ./lib/grape/api.rb:45:in `call!'
[...]

PR #1452, opened 2016-07-20
Documented caveat from #1451, incremented version to 0.17.0.

dblock:
@namusyaka Take a look. This is related to #1451.


namusyaka:
👍 Great. Thanks!

PR #1456, opened 2016-07-26
Add Rails 5 to Appraisals

glaucocustodio:

PR #1457, opened 2016-07-26
Remove the is trivial reference in the Dangerfile correctly

orta:
Fixes the build in #1456

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,7 +1,7 @@
 # Has any changes happened inside the lib dir?
 has_app_changes = !git.modified_files.grep(/lib/).empty?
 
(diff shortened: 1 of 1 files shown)


Dangerfile
```diff
@@ -1,7 +1,7 @@
 # Has any changes happened inside the lib dir?
 has_app_changes = !git.modified_files.grep(/lib/).empty?
 
-if git.modified_files.include?('CHANGELOG.md') == false && !declared_trivial && has_app_changes
+if git.modified_files.include?('CHANGELOG.md') == false && has_app_changes
```
dblock: Just some DSL feedback for @orta - I think this should have `has_app_changes?`

PR #1458, opened 2016-07-26
Reduce build matrix, run Danger once.

dblock:


dblock:
Also restored older Ruby builds by running them with Rails 3 and 4 only, cc: @jlfaber.

dblock:
@orta see this PR for how Danger can be run once.

PR #1459, opened 2016-07-26
Presence validator ensures value is not nil

arempe93:
I was surprised today by a `NoMethodError` for `nil:NilClass` on a endpoint for a param declared using `requires`. I discovered it by submitting an empty form on a javascript app (still in prototyping, ie. no form validations) that requests my grape api. The app posts json to the api, and javascript replaced the unfilled fields with `null` and caused the above error.

I tracked this behavior to the presence validator, which currently only checks that the key is present in the request. I would say that it should reject `nil` values since using `requires`  (especially in conjunction with a `type` validator) seems like a promise to the programmer that the param will have a value in the endpoint.

I had to change some other specs that dealt with nil params. I tried to keep them focused on the behavior in question by changing the params to `optional`, but in one case I had to remove a spec. I feel it's redundant - but any feedback on that would be great.

### DIFF
--- a/lib/grape/validations/validators/presence.rb
+++ b/lib/grape/validations/validators/presence.rb
@@ -7,7 +7,7 @@ module Grape
       end
 
       def validate_param!(attr_name, params)
-        return if params.respond_to?(:key?) && params.key?(attr_name)
+        return if params.respond_to?(:key?) && !params[attr_name].nil?
         raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: message(:presence)
       end
     end
--- a/spec/grape/validations/validators/allow_blank_spec.rb
+++ b/spec/grape/validations/validators/allow_blank_spec.rb
@@ -123,7 +123,7 @@ describe Grape::Validations::AllowBlankValidator do
 
         resources :custom_message do
           params do
-            requires :name, allow_blank: { value: false, message: 'has no value' }
+            optional :name, allow_blank: { value: false, message: 'has no value' }
           end
           get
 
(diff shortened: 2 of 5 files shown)


lib/grape/validations/validators/presence.rb
```diff
@@ -7,7 +7,7 @@ def validate!(params)
       end
 
       def validate_param!(attr_name, params)
-        return if params.respond_to?(:key?) && params.key?(attr_name)
+        return if params.respond_to?(:key?) && !params[attr_name].nil?
```
dblock: This feels major. Before it allowed `nil` to be a valid value. Thoughts?

dblock:
This clearly changes the behavior of the presence validator. It wants to allow `nil` as a differentiated value. Thoughts?

arempe93:
My opinion is when I see - 

```ruby
  requires :email, type: String
```

then I think "ok, it has to be there and has to be a String" and its surprising to me when nil passes through both validations.

dblock:
Isn't that what `allow_blank` is for? See https://github.com/ruby-grape/grape#allow_blank. I think it's `true` by default for historical and backwards compatibility reasons.

arempe93:
Hmm. I would think it should be `false` by default for `requires`...but I know it doesn't always work out that nice when backwards compatibility is involved 😄 

I guess I will drop this in favor of using `allow_blank`. Thanks for the time anyway

dblock:
I would be OK with changing this default over time. It's obviously a major change, but I'd be open to PRs in that direction with a nice UPGRADING and other such things.

PR #1464, opened 2016-07-27
Add Rails 5 appraisal.

dblock:
Replacing https://github.com/ruby-grape/grape/pull/1456.


dblock:
cc: @glaucocustodio, with build fixes for Rails 5, all issues are caused by rack-contrib not being compatible with rack 2.0 (see https://github.com/rack/rack-contrib/issues/127).

glaucocustodio:
I saw it, checkout https://github.com/rack/rack/pull/1103.

dblock:
That PR into Rack is great @glaucocustodio. I think I should still move forward with workarounds here (assuming I can fix the remaining build issues). Please do let me know if you feel otherwise.

glaucocustodio:
As you prefer @dblock. Depending on `rack-jsonp` instead `rack-contrib` seems better if possible.

dblock:
Yep, we have a winner with `rack-jsonp`. Merging.

PR #1465, opened 2016-07-27
Fix filters being called multiple times, consistent intialization for…

jsteinberg:
… method_not_found errors

I also think this is a more consistent fix to this: https://github.com/ruby-grape/grape/pull/1446

### DIFF
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -248,9 +248,12 @@ module Grape
 
         run_filters befores, :before
 
+        allowed_methods = env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED]
+        raise Grape::Exceptions::MethodNotAllowed, header.merge('Allow' => allowed_methods) if allowed_methods
+
         run_filters before_validations, :before_validation
 
-        run_validators validations, request unless env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED]
+        run_validators validations, request
 
         run_filters after_validations, :after_validation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 #### Fixes
 
+* [#1465](https://github.com/ruby-grape/grape/pull/1465): Fix 'before' being called twice when using not allowed method - [@jsteinberg](https://github.com/jsteinberg).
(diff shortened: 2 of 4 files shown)


jsteinberg:
This resolves https://github.com/ruby-grape/grape/pull/1426 / https://github.com/ruby-grape/grape/issues/1424

lib/grape/endpoint.rb
```diff
@@ -248,6 +248,8 @@ def run
 
         run_filters befores, :before
 
+        raise Grape::Exceptions::MethodNotAllowed, header.merge('Allow' => env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED]) if env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED]
```
namusyaka: Could you get rid of the duplication of `env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED]` ?

namusyaka:
@jsteinberg Awesome. Your patch seems to work well. 
Thank you so much!

lib/grape/endpoint.rb
jsteinberg: Sure.  It only is duplicated if it is present so I thought it was okay.  Do you have a preference between these two?

1:
```ruby
if (allowed_methods = env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED])
  raise Grape::Exceptions::MethodNotAllowed, header.merge('Allow' => allowed_methods)
end
```

2:
```ruby
allowed_methods = env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED]
raise Grape::Exceptions::MethodNotAllowed, header.merge('Allow' => allowed_methods) if allowed_methods
```

namusyaka: Maybe 1 will be aborted by rubocop, so I'd like to vote to 2 approach.
Btw, there are two reasons why I have pointed out.

1. By assiging variable, we can reduce number of characters of the line.
2. The assigned variable can be used in three places.

jsteinberg: rubocop was fine with both(the parens around it quieted it).  But it caused the class to have too many lines.  option 2 seems to take care of that restraint, but either that limit will need to be raised soon or....

and the 3rd instance of `env[Grape::Env::GRAPE_METHOD_NOT_ALLOWED]` is actually completely redundant since we are raising an exception.  So I removed it with my latest change.

dblock:
Excellent, merging.

PR #1478, opened 2016-08-26
Fix typo in UPGRADING.md

BastianHofmann:
Extra `}` causes `SyntaxError`.


namusyaka:
Thanks! Btw, you can avoid running test by adding `[ci skip]` to your commit message.

PR #1481, opened 2016-09-03
Add MIT license badge to the README

kmewhort:
I ran a license compatibility check against Grape's gems:  all 20 dependencies are under permissive licenses (MIT or Apache), so definitely check-out as compatible with the MIT license!

The badge here gives a quick indicator that the license is MIT, and that the libraries used are compliant with that.


dblock:
That's cool! I didn't know about git.legal.

Issue #1485, opened 2016-09-07
Values NOT allowed Validator?


jonmchan:
the ```values``` validator is great for defining values for the parameter that you are allowed to have. I would like to define values that are NOT allowed for the parameter and to return an error if the value passed matches a value on the list. Is there any way to this currently? If not, I don't mind submitting a PR for this.

```values_not_allowed:```, ```exclusions```, or ```disallowed_values``` or something of that sort?

dblock:
Maybe `except: [ ... ]` as an extension to the current values would be cleaner? Please do PR!

dblock:
Related, I would want to support regular expressions in `values` as opposed to `regexp: ` that we do today.

PR #1486, opened 2016-09-08
implemented except in values validator

jonmchan:
This implements the feature requested in #1485 .

Some notes:

* This PR allows for except to be used exclusively allowing all other values except what is defined in except as well as with values, defining a range and then marking a subset of the range as invalid.
* This PR allows for custom except message outside of the regular validation fail message, but falls back to the regular validation fail message if no ```except_message``` is defined.
* There was slight difficulty maintaining compatibility with allowing values to be passed in to ```values:``` without the ```value:``` key. Please ensure that line 5-8 of values.rb https://github.com/ruby-grape/grape/compare/master...jonmchan:feature/except?expand=1#diff-45b62b4ed0aed7ceff346c347f76570eR8 do not cause any incompatibility issues.
* Side effect of this implementation is that ```values: []``` accepts all passed parameters. Previous implementation, ```values: []``` would reject every value. I figure nobody would be trying to create a parameter that would allow no value so I did not bother to maintain this functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ==================
 
 * [#1480](https://github.com/ruby-grape/grape/pull/1480): Use the ruby-grape-danger gem for PR linting - [@dblock](https://github.com/dblock).
+* [#1486](https://github.com/ruby-grape/grape/pull/1486): implemented except in values validator - [@jonmchan](https://github.com/jonmchan).
 * Your contribution here.
 
 #### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 ==================
 
 * [#1480](https://github.com/ruby-grape/grape/pull/1480): Use the ruby-grape-danger gem for PR linting - [@dblock](https://github.com/dblock).
-* [#1486](https://github.com/ruby-grape/grape/pull/1486): implemented except in values validator - [@jonmchan](https://github.com/jonmchan).
+* [#1486](https://github.com/ruby-grape/grape/pull/1486): Implemented except in values validator - [@jonmchan](https://github.com/jonmchan).
 * Your contribution here.
 
 #### Fixes
--- a/README.md
+++ b/README.md
@@ -1065,6 +1065,26 @@ params do
 end
 ```
 
+The values validator can also validate that the value is explicitly not within a specific
+set of values by passing ```except```. ```except``` accepts the same types of parameters as
+values (Procs, ranges, etc.).
+
+```ruby
+params do
+  requires :browsers, values: { except: [ 'ie6', 'ie7', 'ie8' ] }
+end
+```
+
+Values and except can be combined to define a range of accepted values while not allowing
+certain values within the set. Custom error messages can be defined for both when the parameter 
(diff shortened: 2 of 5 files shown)


jonmchan:
I don't know how to fix the danger error:

```
[!] Invalid `temporary_danger.rb` file: undefined local variable or method `changelog' for #<Danger::Dangerfile:0x0000000378e850>. Updating the Danger gem might fix the issue.
```

I can't even run danger on my local system:

```
$ bundle exec danger
bundler: failed to load command: danger (/home/jchan/.rbenv/versions/2.3.1/bin/danger)
NameError: undefined local variable or method `ci_klass' for #<Danger::Executor:0x0056436b27bbe0>
Did you mean?  class
  /home/jchan/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/danger-2.1.6/lib/danger/danger_core/executor.rb:16:in `run'
  /home/jchan/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/danger-2.1.6/lib/danger/commands/runner.rb:56:in `run'
  /home/jchan/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/claide-1.0.0/lib/claide/command.rb:334:in `run'
  /home/jchan/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/danger-2.1.6/bin/danger:5:in `<top (required)>'
  /home/jchan/.rbenv/versions/2.3.1/bin/danger:23:in `load'
  /home/jchan/.rbenv/versions/2.3.1/bin/danger:23:in `<top (required)>'
```

jonmchan:
@dblock any feedback?

dblock:
The danger problem fixed in https://github.com/ruby-grape/grape/pull/1480. Rebase.

dblock:
Overall this looks great. My biggest question is whether it should be

```
requires :number, type: Integer, values: { value: 1..20 except: [4,13], except_message: 'includes unsafe numbers', message: 'is outside the range of numbers allowed' }
```

vs.

```
requires :number, type: Integer, values: 1..20, except: [4,13], except_message: 'includes unsafe numbers', message: 'is outside the range of numbers allowed'
```

The latter would be consistent with how we do all options today.

jonmchan:
If I am understanding the difference in what you are suggesting, you are saying that:

#1.) ```value``` is omitted and #2.) the parameters are not passed into a hash to ```values```. 

However, I don't find this consistent with the currently implementation in terms of tests:

eg,

https://github.com/ruby-grape/grape/blob/master/spec/grape/validations/validators/values_spec.rb#L26
```ruby
requires :type, values: { value: ValuesModel.values, message: 'value does not include in values' }
```

https://github.com/ruby-grape/grape/blob/master/spec/grape/validations/validators/values_spec.rb#L33
```ruby
            optional :type, values: { value: -> { ValuesModel.values }, message: 'value does not include in values' }, default: 'valid-type2'
```

If you are interested in seeing values refactored, I can make that change, but it is currently very different (and potentially breaking) to how it currently behaves.

jonmchan:
@dblock fixed all errors. This should be good to go unless you want me to refactor values.

dblock:
Yes you're totally right @jonmchan. Merging this, looks great.

PR #1490, opened 2016-09-13
Switch to Ruby-2.x+ syntax

namusyaka:


dblock:
How fancy! I'll play.

Issue #1433, opened 2016-06-22
When using `given` it does not actually check that it's required


andrew13nguyen:
I have code similar to this:

```
params do
  optional :button_link, type: String
  given :button_link do
    requires :button_text, type: String
  end
end

post :something do
  puts params.inspect
end
```

When I post to "something" with no params, everything is valid, but I see that params has:

```
params.inspect
=> {"button_text"=>nil}
```

When I change the params block to this:

```
params do
  optional :button_link, type: String
  given :button_link do#, :authenticated_button_link do
    requires :button_text, type: String, allow_blank: false
  end
end
```

I get a validation error (button_text is empty). Am I using `given` incorrectly or is this a bug?

dblock:
This doesn't seem right, you should write a spec next.

andrew13nguyen:
Okay, so what I posted before was lacking and the tests passed. After tinkering around I got this to fail and maybe I'm using grape incorrectly, but I feel like this is a valid use case:

(Borrowing from https://github.com/ruby-grape/grape/blob/v0.16.2/spec/grape/validations/params_scope_spec.rb#L319-L343)
```
it 'applies the validations of multiple parameters' do
  subject.params do
    optional :a
    given :a do
      requires :c, type: String
    end
    optional :b
    given :b do
      requires :c, type: String
    end
    mutually_exclusive :a, :b
  end
  subject.get('/multiple') { declared(params).to_json }

  get '/multiple'
  expect(last_response.status).to eq(200)

  get '/multiple', a: true, c: 'test'
  expect(last_response.status).to eq(200)

  get '/multiple', b: true, c: 'test'
  expect(last_response.status).to eq(200)

  # this passes
  get '/multiple', a: true
  expect(last_response.status).to eq(400)
  expect(last_response.body).to eq('c is missing')

  # this fails
  get '/multiple', b: true
  expect(last_response.status).to eq(400)
  expect(last_response.body).to eq('c is missing')

  get '/multiple', a: true, b: true, c: 'test'
  expect(last_response.status).to eq(400)
  expect(last_response.body).to eq('a, b are mutually exclusive')
end
```

If I change `requires :c, type: String` to `requires :c` then it passes, but I would think enforcing a check on type is desired.

dblock:
Right, turn it into a failing PR?

giedriusr:
I've discovered the same thing as @andrew13nguyen. If you remove type: String then everything works. Is this a bug?

[...]

PR #1498, opened 2016-09-21
Skip running param validators inside inactive given blocks (#1433)

jlfaber:
This should address both the primary issue identified in #1433, which was related to the ```type``` coercer, and related issues with ```allow_blank```.  In both cases, these validators were being run even when their scopes indicated they should not (since the associated ```given``` criterion had not been met).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 #### Fixes
 
+* [#1498](https://github.com/ruby-grape/grape/pull/1498): Skip validations in inactive given blocks - [@jlfaber](https://github.com/jlfaber).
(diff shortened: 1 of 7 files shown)


jlfaber:
Hat tip to @andrew13nguyen for the UT code that demonstrates multiple type validators for the same parameter.  I tweaked it to make the types different just to add another layer of chaos.

dblock:
This is some excellent work @jlfaber & @andrew13nguyen. Merging.

jlfaber:
Thanks @dblock for the merge!  Any chance we can get this tagged for release?

dblock:
We try not to release too often, but I'll queue this up for myself for this/next week. Usually that involves running grape next against a very large API I am looking at over here.

dblock:
Of course if @namusyaka @dm1try @marshall-lee @adamgotterer want to cut a release before that go for it!

namusyaka:
👍 Great job!

Issue #1423, opened 2016-06-13
Ability to fail fast on params validation


akozhemiakin:
Correct me if I'm wrong, but currently all validators will be executed even if some of them fails during the validation process. I believe this is the corresponding code chunk, showing the nature of this behavior:
```ruby
def run_validators(validators, request)
  validation_errors = []

  validators.each do |validator|
    begin
      validator.validate(request)
    rescue Grape::Exceptions::Validation => e
      validation_errors << e
    end
  end

  validation_errors.any? && raise(Grape::Exceptions::ValidationErrors, errors: validation_errors, headers: header)
end
```

I believe that it should be possible to instruct Grape to fail fast during the validation process.

One possible case: We need to validate that some string matches some pattern (for example it is an email) and than we need to validate if this email is in the white list. For now even if this email is malformed, it will be checked against the white list (which has no sense cause it is malformed anyway).

The DSL may look like this:
```ruby
params do
  requires :email, allow_blank: false, regexp: /@/, fail_fast: true
end
```

dblock:
I like this idea as long as we ensure the validation order.

connorshea:
Is this resolved now?

dgasper:
Yes, should probably be closed.

dblock:
Closing via #1499.

PR #1499, opened 2016-09-28
Support fail_fast param validation option

dgasper:
[#1423](https://github.com/ruby-grape/grape/issues/1423)
This change will enable
```ruby
params do
  required :beer, fail_fast: true
  required :wine
end
```
which will result in skipping of subsequent validation checks if a fail_fast required or optional fail validation.


lib/grape/validations/params_scope.rb
```diff
@@ -140,9 +140,13 @@ def require_optional_fields(context, opts)
       end
 
       def validate_attributes(attrs, opts, &block)
+        # check fail_fast flag
+        fail_fast = opts.delete(:fail_fast) || false
```
dblock: You should not be modifying `opts` being passed in here, combine the `clone` below.

README.md
```diff
@@ -1355,6 +1355,23 @@ subject.rescue_from Grape::Exceptions::ValidationErrors do |e|
 end
 ```
 
+Grape returns all validation and coercion errors found by default. To skip all subsequent validation checks when a specific param is found invalid, use `fail_fast: true`.
+For example:
```
dblock: Maybe space things up a bit. Minor.

README.md
```diff
@@ -1355,6 +1355,23 @@ subject.rescue_from Grape::Exceptions::ValidationErrors do |e|
...
+```ruby
+params do
+  required :beer, fail_fast: true
+  required :wine
+end
+```
+will not check if `:wine` is present unless it finds `:beer`. The result of empty params would be a single `Grape::Exceptions::ValidationErrors` error.
```
dblock: I think a lot of grape examples try to make full sentences, so prefer "The following example will not ..." to having "For example: [code] will not".

lib/grape/validations/params_scope.rb
```diff
@@ -201,7 +205,7 @@ def configure_declared_params
         end
       end
 
-      def validates(attrs, validations)
+      def validates(attrs, validations, fail_fast = false)
```
dblock: I think the argument should be an options hash here, so we can make it extensible for the future. So `def validates(attrs, validations, opts = {})`.

lib/grape/validations/validators/base.rb
```diff
@@ -10,10 +10,11 @@ class Base
       # @param options [Object] implementation-dependent Validator options
       # @param required [Boolean] attribute(s) are required or optional
       # @param scope [ParamsScope] parent scope for this Validator
-      def initialize(attrs, options, required, scope)
+      def initialize(attrs, options, required, fail_fast, scope)
```
dblock: Should `fail_fast` be rolled into one of these params?

dblock:
This is excellent. Most of my comments are small implementation-related things, I'd like `fail_fast` to roll in as yet another option on the validator rather than its own big kind of thing. I am sure we'll have future other options.

lib/grape/validations/validators/base.rb
dgasper: @dblock I've thought about it. It could be a part of the scope, but I am not sure if it is the best place for validation options. Thoughts?

I've changed the constructor to the following to keep existing custom validators working.
`def initialize(attrs, options, required, fail_fast, scope, opts = {})`

lib/grape/validations/params_scope.rb
dgasper: :+1:

lib/grape/validations/params_scope.rb
```diff
@@ -140,9 +140,14 @@ def require_optional_fields(context, opts)
       end
 
       def validate_attributes(attrs, opts, &block)
```
dblock: Have you tried to do this?

```
def validate_attributes(attrs, validations, opts = {}, &block)
```

This way there wouldn't be any reason to do the selection below.

dblock: Either way this seems to be a bit too much future-proof ;)

```ruby
validations = opts.clone
validations[:type] ||= Array if block

options = {}
options[:fail_fast] = validations.delete(:fail_fast)
```

dblock: ... which begs the question, why is `fail_fast` so special? Why does it need to be split at all?

dgasper: Then I would have to split the options [here](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/parameters.rb#L97), no?

Hmm, I guess I got a bit carried away with the future other options idea :)

In my opinion, the difference is that the rest of the `validations` refer to a specific validation type with options whereas `fail_fast` (and possible future validation options) influences the validation process across multiple validations.

But I could be wrong. I just dove into grape code for the first time :)

dblock: Try to make it simpler for now, if not splitting is less code, we should go with that.

dblock:
This is better. Merging.

PR #1503, opened 2016-10-06
Allow to use regexp validator with arrays

akoltun:
Ex. this will work

```
params do
  requires :names, type: Array[String], regexp: /^[a-z]+$/
end
```

### DIFF
--- a/lib/grape/validations/validators/regexp.rb
+++ b/lib/grape/validations/validators/regexp.rb
@@ -2,7 +2,7 @@ module Grape
   module Validations
     class RegexpValidator < Base
       def validate_param!(attr_name, params)
-        return unless params.respond_to?(:has_key?) && params.has_key?(attr_name)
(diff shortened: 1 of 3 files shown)


lib/grape/validations/validators/regexp.rb
```diff
@@ -2,7 +2,8 @@ module Grape
   module Validations
     class RegexpValidator < Base
       def validate_param!(attr_name, params)
-        return unless params.key?(attr_name) && !params[attr_name].nil? && !(params[attr_name].to_s =~ (options_key?(:value) ? @option[:value] : @option))
+        return unless params.respond_to?(:has_key?) && params.has_key?(attr_name)
```
namusyaka: Rubocop said, please use `key?` instead of `has_key`.

namusyaka:
@akoltun Thanks for your pull request. This change looks like reasonable. 
Please check my comment out.

dblock:
This looks good, merging.

Issue #1482, opened 2016-09-05
Make allow_blank false when the type is defined


azbshiri:
Hey, we're working on a project and we have a lot of required parameters with their type defined but allow_blank is true by default so we should make allow_blank false for each one and that's bad (DRY). is there any option we can set to make it false when at least the type is defined or just for the requires method?

intuxicated:
+1

dblock:
I would take a pull request that configures global defaults for various validators.

namusyaka:
@azbshiri @intuxicated Could you confirm [this pull request](https://github.com/ruby-grape/grape/pull/1507)?

azbshiri:
I really appreciate your great work. I think it does the job 💯

namusyaka:
Thanks for confirming the issue. Closed.

PR #1507, opened 2016-10-07
Feature: add group attributes for parameter definitions

304:
It is a proposal with a working prototype for a group attributes feature. If it makes sense then I will continue working on this PR. Your feedback is welcome. 

I was looking into this issue (https://github.com/ruby-grape/grape/issues/1482) recently. I think the presented use case is common for complex APIs schemas. Currently, if I want to apply the same validator (or the same type) for a several endpoint attributes, then I have to define it for every parameter.

What if we introduce a way of defining common parameter settings? Something like [Rails conditional validations](http://guides.rubyonrails.org/active_record_validations.html#grouping-conditional-validations).

For example, this parameters definition:

```ruby
params do
  requires :first_name, type: String, allow_blank: false
  requires :last_name, type: String, allow_blank: false
end
```

can be replaced with
```ruby
params(with: { type: String, allow_blank: false }) do
  requires :first_name
  requires :last_name
end
```

It is also possible to extract group attributes into a class method and use it instead of default params in your code. It should provide the solution for the problem from #1482.
```ruby
class MyAPI < Grape::API do
  def self.not_blank_params(&block)
    params(with: { allow_blank: false }, &block)
  end

  not_blank_params do
    requires :first_name # This param won't accept blank lines 
    requires :last_name, allow_blank: true # This param *will* accept blank lines. 
  end
end
```

/cc @dblock

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 Next Release
 ============
 
-* Your contribution here.
 * [#1503](https://github.com/ruby-grape/grape/pull/1503): Allow to use regexp validator with arrays - [@akoltun](https://github.com/akoltun).
+* [#1507](https://github.com/ruby-grape/grape/pull/1507): Add group attributes for parameter definitions - [@304](https://github.com/304).
+* Your contribution here.
 
 0.18.0 (10/7/2016)
 ==================
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@
   - [Multiple Allowed Types](#multiple-allowed-types)
   - [Validation of Nested Parameters](#validation-of-nested-parameters)
   - [Dependent Parameters](#dependent-parameters)
+  - [Group Options](#group-options)
   - [Built-in Validators](#built-in-validators)
   - [Namespace Validation and Coercion](#namespace-validation-and-coercion)
   - [Custom Validators](#custom-validators)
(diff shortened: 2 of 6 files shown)


rnubel:
I think the use case where *all* parameters for an endpoint share the same options is pretty rare; even if your initial endpoint has only homogenous params, eventually you'll want to add an extra param of a different type and be forced to undo all the shared options. However, like you point out, having a group of parameters with shared options is very common. What about a solution which introduces a new method within the params DSL? Something like this:

```ruby
params do
  with(type: String, allow_blank: false) do
    requires :first_name
    requires :last_name
  end

  requires :age, type: Integer, allow_blank: true
end
```

dblock:
Thanks @304, this is great. I think @rnubel's suggestion is spot on and is even better, since this gives us a proper DSL to deal with within `params` and something much more extensible for the future. I'd definitely take that PR.

dgasper:
@rnubel I guess this should already be [possible](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/parameters.rb#L88) like this:

```ruby
params do
  requires :first_name, :last_name, type: String, allow_blank: false
  requires :age, type: Integer, allow_blank: true
end
```

jlfaber:
You can use an array of attribute names rather than a single one to define a bunch of identical attributes, right?
```
params do
  requires %i(first_name last_name), type: String, allow_blank: false
end
```
One could also use a params helper to dynamically specify the name(s) of attrs that need a specific validation.
```
helpers do
  params :name do |options|
    optional options[:attr], type: String, allow_blank: false
  end
end

params do
  use :name, attr: :first_name
  use :name, attr: :last_name
end
```

304:
Thank you for the feedback! I like @rnubel's suggestion too. I think it is better to have it this way than as a part of params method. I will try to implement it and I will update this PR when I have something to show.

304:
It is ready. I've added a new method `with` to Grape DSL. It allows to extract and group common options for a list of parameters. The documentation was also updated. Can you please take a look?

/cc @dblock

dblock:
I'm going to merge this. Seems like a sensible add-on and doesn't hurt at the very least. If anyone hates it, speak up so we can iterate before the next release.

Issue #1508, opened 2016-10-11
Inconsistent validation messages


dgasper:
When defining parameters in separate `requires`, the validation errors include errors for all defined params. For example the following returns `one is missing, two is missing`

``` ruby
params do
  requires :one, type: Integer
  requires :two, type: Integer
end
```

In contrast, the following usage of a single `requires` for multiple params only returns `one is missing`.

``` ruby
params do
  requires :one, :two, type: Integer
end
```

I've tracked down where the validation is stopped to [this line](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/base.rb#L48). Removing it results in both above examples returning `one is missing, two is missing`.

Is this something we would want or is the above behavior expected?

dblock:
This looks like a bug. The two should be equivalent IMO.

towanda:
@dblock Could you please close this? It's fixed by https://github.com/ruby-grape/grape/pull/1510

dblock:
Closed via #1510.

PR #1510, opened 2016-10-12
Fix inconsistent validation

dgasper:
Fixes [#1508](https://github.com/ruby-grape/grape/issues/1508)


dblock:
So looks like that code was there intentionally, but without specs. I think it's worth mentioning the behavior change in UPGRADING, please?

UPGRADING.md
```diff
@@ -1,6 +1,21 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 0.18.0
```
dblock: This is incorrect, we have already shipped 0.18.0, so this should be 0.19.0.

dblock:
If you really want to be nice, squash your commits after you've fixed the version in UPGRADING. I'll merge.

UPGRADING.md
dblock: Maybe 0.18.1, I am not sure whether this is major/minor release worthy.

dgasper:
Oh, sorry. Force push doesn't result in a notification. Glad I could contribute. Thanks!

dblock:
Thanks for your excellent work @dgasper!

PR #1512, opened 2016-10-19
Test for nested params problem and also fix

krbs:


dblock:
@krbs The bot is telling you to update CHANGELOG :) ^^^

lib/grape/dsl/parameters.rb
```diff
@@ -193,21 +193,25 @@ def declared_param?(param)
 
       alias group requires
 
+      def map_params(params, element)
+        result = if params.is_a?(Array)
```
dblock: you don't need to assign result here, just leave the `if` block it returns all the right things.

dblock:
Merged. Thank you.

PR #1514, opened 2016-10-28
Replace Hashie::Mash with ActiveSupport::HashWithIndifferentAccess

msroot:
This will replace `Hashie::Mash` with `ActiveSupport::HashWithIndifferentAccess`


spec/grape/validations/validators/coerce_spec.rb
```diff
@@ -285,7 +285,7 @@ class User
           requires :file, type: Rack::Multipart::UploadedFile
         end
         subject.post '/upload' do
-          params[:file].filename
+          params[:file][:filename]
```
dblock: So this is a change in API as only the top keys are with indifferent access?

dblock:
This is similar to #1448, but better.

I think we also need to provide a backward compatible way for existing users to return a `Hashie::Mash`. Here's what I would like the DSL to be:

```ruby
class API < Grape::API
  # params is a regular Hash
end
```

```ruby
class API < Grape::API
   include Grape::Extensions::...::Hashie::Mash
   # params is Hashie::Mash
end
```

```ruby
class API < Grape::API
   include Grape::Extensions::...::ActiveSupport::HashWithIndifferentAccess
   # params is HashWithIndifferentAccess
end
```

You get the idea ...

spec/grape/validations/validators/coerce_spec.rb
msroot: nope! to all sub keys

msroot:
@dblock sure but the default should be `ActiveSupport::HashWithIndifferentAccess` yes?
For the  `Hashie::Mash` does the user has to explicitly require the Hashie::Mash gem then?

About the DSL, is this something like this?

```
if defined? Hashie::Mash
params = Hashie::Mash.new(rack_params)
```

dblock:
Yes, I think the default should be `Hash`.

The user should explicitly include `hashie` in their Gemfile in order to use the extension, and we should not have it in `.gemspec` anymore. For tests, it should be a development dependency.

jdurand:
@dblock any timeline on merging this?

dblock:
This PR won't be merged without the changes I'm asking for above.

dblock:
Replaced with https://github.com/ruby-grape/grape/pull/1594.

dblock:
Please see https://github.com/ruby-grape/grape/pull/1610 as a working end-to-end PR.

Issue #1392, opened 2016-05-11
Ability to stream entities


raulpopadineti:
Hi,

We'd like to build an API that is able to paginate entities and send them in chunks. 

A really raw way of sending data in chunks would be:

```
require 'grape'

module Stream
  class API < Grape::API
    prefix 'api'
    format :txt
    default_format :txt

    class Streamer
      def initialize(pages:, page_size:, latency:)
        @pages = pages
        @page_size = page_size
        @latency = latency
      end

      def each
        (1..@pages).lazy.each do |p|
          yield "#{p}\n" * @page_size
          sleep(@latency)
        end
      end
    end

    resources :users do
      params do
        optional :pages, type: Integer, default: 10
        optional :page_size, type: Integer, default: 10
        optional :latency, type: Float, default: 1.0
      end

      get do
        status 200
        p = declared(params)
        stream Streamer.new(pages: p.pages, page_size: p.page_size, latency: p.latency)
      end
    end
  end
end
```

With this approach I get the deprecation warning:

`[DEPRECATION] Argument as FileStreamer-like object is deprecated. Use path to file instead.`

because Grape assumes the response is a file.

Will you support streaming of a `Grape::Entity` or any other non-file objects in the future?

dblock:
I think the deprecation is just for files. It's still supporting a pure streaming interface. You should look at the code a bit more in depth and see what's missing? Generally I don't see why we wouldn't want this, so I'll leave this open.

lfidnl:
Hello. I guess, now `stream` method can stream only files. Maybe we should to do:
1. Add option `stream: true` to `file` method. This option will replace current `stream` method.
2. Rewrite `stream` method in such a way that it can receive object which respond to `each` method and streams it.

raulpopadineti:
I've managed to stream JSON by polishing a bit on the `Streamer` class I added:

```
```

lfidnl:
@raulpopadineti one question: is it works now with deprecation warning, or not?

raulpopadineti:
@lfidnl it works with deprecation warning.

lfidnl:
@raulpopadineti good, I think, @dblock should approve my suggestion and someone can create PR for fix this issue :)

dblock:
I think what you suggest is sensible, would like to see a PR.

raulpopadineti:
@dblock to which solution are you referring?

dblock:
[...]

PR #1520, opened 2016-11-14
un-deprecate stream-like objects

urkle:
- this splits the "stream" and "file" inside_route methods 
- file is purely for specifying a file (and utilizing sendfile when it can)
- stream is purely for stream-like objects.. This allows you to have streaming responses of generated data

This PR is in reference to issue #1392 

Pending Questions
- why are the headers only set for the "stream" method and not the "file" method in inside_routes ?
- if the behavior is different because of those headers then I should update this to NOT deprecate passing a filename to "stream"
- what additional tests are needed for this change?


dblock:
So the new code is handling in `stream`? I think there needs to be tests for any new paths. Maybe explicit deprecation tests that `warn` is called?

Needs an UPGRADING too. Thanks.

milgner:
This looks very promising! @urkle any chance you can adapt this to the current `master`? Otherwise if anyone is interested, I'd be willing to take this on, too.

urkle:
@milgner it maybe a week or so before I get back around to this. It's been on my TODO list to rebase, add tests and re-submit this

linchus:
@urkle Any news in this PR?
Do you need any help?

urkle:
@linchus  this needs to be rebased and tests added (as requested). Right now it has been a matter of finding the time to work on it.

crondaemon:
Can we help in some way this PR to be merged? The deprecation warning is really noisy when used in regression tests and clutters the output. Since we're using a non-deprecated object, it would be really useful to remove the warning.

dblock:
> Can we help in some way this PR to be merged? The deprecation warning is really noisy when used in regression tests and clutters the output. Since we're using a non-deprecated object, it would be really useful to remove the warning.

Reopen the PR, fix per comments above, make it green?

urkle:
Right now I'm finishing a project in the next week or so upgrading our apps to using Grape 1.3.x.  Once that is done I'll be able to revisit this against grape master (as we are still using this functionality)

crondaemon:
@urkle thanks for your effort!

dblock:
This looks ready code-wise, thanks for following up! We need to document the new behavior in README? Anything to add to UPGRADING?

urkle:
> This looks ready code-wise, thanks for following up! We need to document the new behavior in README? Anything to add to UPGRADING?

I just rebased it right now.   Going through my changes and refreshing my memory and going to be adding more tests and updating README still.  I'll let you know when it's all ready.

@dblock also did you have any thoughts on my first question in the PR? Why the "stream' method sets headers but not the 'file' method?

urkle:
@dblock ok.. I pushed up new tests and updates to UPGRADING and README.  Let me know what else you want me to adjust/tweak.

dblock:
> * why are the headers only set for the "stream" method and not the "file" method in inside_routes ?


I am guessing because for a file you want to let Rack do its thing for content-length and transfer-encoding and generate an e-tag, but for a stream none of these make sense.

> * if the behavior is different because of those headers then I should update this to NOT deprecate passing a filename to "stream"

I think what you did was right. A file is a file, a stream is a stream. Until someone points out something broken I say leave it.

> * what additional tests are needed for this change?

I would ensure tests for those headers, if we made a mistake at least we will have tests to say that it was thought through behavior.

README.md
```diff
@@ -3168,7 +3168,7 @@ end
 
 Use `body false` to return `204 No Content` without any data or content-type.
 
-You can also set the response to a file with `file`.
+You can also set the response to a file with `file` and it will be streamed using Rack::Chunked.
```
dblock, line 3171: Quote `Rack::Chunked`

README.md
```diff
@@ -3178,12 +3178,22 @@ class API < Grape::API
 end
 ```
 
-If you want a file to be streamed using Rack::Chunked, use `stream`.
+If you want to stream non-file data you use the stream method and a Stream object.
```
dblock, line 3181: Quote classes

README.md
```diff
@@ -3178,12 +3178,22 @@ class API < Grape::API
...
+This is simply an object that responds to each and yields for each chunk to send to the client.
```
dblock, line 3182: Remove "simply", weasel word ;)

Quote `each`

UPGRADING.md
```diff
@@ -1,6 +1,42 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.3.4
+
+#### Reworking stream and file and un-deprecating stream like-objects
+
+Previously in 0.16 stream-like objects were deprecated. This release restores their functionality for use-cases other than file streaming.
+
+For streaming files, simply use file always.
```
dblock, line 10: Remove simply, quote `file`

UPGRADING.md
```diff
@@ -1,6 +1,42 @@
...
+
+```ruby
+class API < Grape::API
+  get '/' do
+    file '/path/to/file'
+  end
+end
+```
+
+If you want to stream other kinds of content from a streamer object you may.
```
dblock, line 20: Or shorter ...

Use `stream` to stream other kinds of content. In the following example a streamer class ...

UPGRADING.md
```diff
@@ -1,6 +1,42 @@
...
+An example would be a streamer class that fetches several pages of data from a database and streams the formatted responses back.
+
+```ruby
+class MyObject
+  def each
+    yield '['
+    # maybe do some paginated DB fetches and return each page
```
dblock, line 27: Remove maybe, another weasel word

urkle:
> > file is purely for specifying a file (and utilizing sendfile when it can)
> 
> from the user's point of view, it still sounds ambiguous. at least it should be documented "when it can".
> 
> though I would use `file` only for utilizing `sendfile`(or even rename it to `sendfile`) and use `stream` to produce a chunked response(this could be a file too): those different delivery mechanisms require different changes to the stack, so I would use different interfaces for them to force a user to make an explicit choice.
> moreover `sendfile` is not about streaming, it's about an optimization of static file delivering using web server facilities.

@dm1try @dblock 
OK.. so how about this.

1) deprecate fully "file"
2) add sendfile that does what "file" does today in master
3) allow "stream" to take a filename (string) or a stream-like object

dblock:
I think what is in this PR is good (enough), before we go into further renaming. What do you think about merging this as here @dm1try and then we can try the suggested change?

urkle:
@dblock I pushed a fixup that does that rework.. but I can extract to a separate PR if you wish.. Up to you.. just let me know.

urkle:
@dblock I think the rework is definitely cleaner and more obvious.

dm1try:
> I think what is in this PR is good (enough) as it does what it advertises. Before we go into further renaming - @dm1try what do you think about merging this on 🍏 as here and then we can try the suggested change?

I’m fine with it)

UPGRADING.md
```diff
@@ -1,6 +1,70 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.3.4
+
+#### Reworking stream and file and un-deprecating stream like-objects
+
+Previously in 0.16 stream-like objects were deprecated. This release restores their functionality for use-cases other than file streaming.
+
+This release also renamed `file` to `sendfile` to better document it's purpose.
```
dblock, line 10: it's -> its

We're deprecating `file`, so I would say "this release deprecated `file` in favor of `sendfile`"

UPGRADING.md
```diff
@@ -1,6 +1,70 @@
...
+
+To deliver a file via `sendfile` if available do this.
```
dblock, line 12: "if available" doesn't make sense to me, I would just remove

to deliver "in chunks" or something similar?

UPGRADING.md
```diff
@@ -1,6 +1,70 @@
...
+```ruby
+class API < Grape::API
+  get '/' do
+    sendfile '/path/to/file'
+  end
+end
+```
+
+Use `stream` to stream files
```
dblock, line 21: "stream file conntent"?

UPGRADING.md
urkle: @dblock look at how I re-worded it.. as it is not really "in chunks" as if the web server you are running through supports sendfile then it hands it off to that functionality.

spec/grape/integration/rack_sendfile_spec.rb
```diff
@@ -4,12 +4,16 @@
 
 describe Rack::Sendfile do
   subject do
-    send_file = file_streamer
+    contet_object = file_object
```
dblock, line 7: content_object, typo, missing `n`

UPGRADING.md
```diff
@@ -1,15 +1,15 @@
...
 
 #### Reworking stream and file and un-deprecating stream like-objects
 
 Previously in 0.16 stream-like objects were deprecated. This release restores their functionality for use-cases other than file streaming.
 
-This release also renamed `file` to `sendfile` to better document it's purpose.
+This release deprecated `file` in favor of `sendfile` to better document its purpose.
 
-To deliver a file via `sendfile` if available do this.
+To deliver a file via the Sendfile support in your web server and have the Rack::Sendfile middleware enabled. See [`Rack::Sendfile`](https://www.rubydoc.info/gems/rack/Rack/Sendfile)
```
dblock, line 12: Add a period in the end.

UPGRADING.md
```diff
@@ -18,7 +18,7 @@ class API < Grape::API
 end
 ```
 
-Use `stream` to stream files
+Use `stream` to stream file content in chunks
```
dblock, line 21: Add a period.

dblock:
Since we're bumping to 1.4, change version.rb and the version in CHANGELOG too please.

dblock:
Merged, thank you!

dblock:
I suggest sending an email to [the Grape mailing list](https://groups.google.com/forum/#!forum/ruby-grape) describing the change (mostly copy from upgrading) and asking people to comment/test/etc. early.

PR #1526, opened 2016-11-25
Avoid noise when code runs with Ruby warnings

cpetschnig:
New versions of RSpec have warnings turned on by default:

    # This setting enables warnings. It's recommended, but in some cases may
    # be too noisy due to issues in dependencies.
    config.warnings = true

I believe these warnings are helpful in general to write better code. However, Grape and Grape Entity are very noisy. I get about 20 lines of warnings for one single test where Grape is involved.

I hope you agree. Many of my changes could also be solved differently for sure. I am happy to discuss and change those cases. I also changed the specs for less warnings and turned them on in the test suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Next Release
 * [#1517](https://github.com/ruby-grape/grape/pull/1517), [#1089](https://github.com/ruby-grape/grape/pull/1089): Fix: priority of ANY routes - [@namusyaka](https://github.com/namusyaka), [@wagenet](https://github.com/wagenet).
 * [#1512](https://github.com/ruby-grape/grape/pull/1512): Fix: deeply nested parameters are included within `#declared(params)` - [@krbs](https://github.com/krbs).
 * [#1510](https://github.com/ruby-grape/grape/pull/1510): Fix: inconsistent validation for multiple parameters - [@dgasper](https://github.com/dgasper).
+* [#1526](https://github.com/ruby-grape/grape/pull/1526): Reduce warnings caused by instance variables not initialized - [@cpetschnig](https://github.com/cpetschnig).
 
 0.18.0 (10/7/2016)
 ==================
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -95,6 +95,8 @@ module Grape
       @options[:route_options] ||= {}
 
       @lazy_initialize_lock = Mutex.new
+      @lazy_initialized = nil
+      @block = nil
 
       return unless block_given?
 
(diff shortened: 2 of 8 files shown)


dblock:
I think initializers such as `@status ||= nil` should be `@status = nil` in `def initialize`. Otherwise this looks sensible.

spec/spec_helper.rb
```diff
@@ -18,5 +18,7 @@
   config.include Rack::Test::Methods
   config.raise_errors_for_deprecations!
 
+  config.warnings = true
```
LeFnord: please set it to `false`, it adds many additional warnings, and deprecation warnings could be overseen
->  run with ruby 2.4 to see what I mean

cpetschnig:
I have done so in all the classes. The cases where I didn't are all modules in the `DSL` namespace.

dblock:
Because those are included? I am not a huge fan of side-effects of assigning something on return to `nil`, warning or convention or not. Anyone else has an opinion, maybe @namusyaka or @lefnord?

cpetschnig:
I agree that `@body ||= nil` is not a very good pattern, rather hard to read/understand. I removed the two occurrences from the PR.

dm1try:
Sorry to bother you guys, I'm personally :-1: for this. 
In prefer to move  "instance variable @var not initialized" warning to the other verbosity level than add more code to deal with it :)
`bundle exec ruby -Ispec -W2 spec/grape/validations/validators/default_spec.rb` - I'll happy to see it only here.

There are some good points in [the `sequel` repo](https://github.com/jeremyevans/sequel/issues/1184) that can help with the decision.

cpetschnig:
The discussion is very interesting, also the references to the Ruby development. The `warning` gem in Ruby 2.4 will definitely help. The sequel gem should be a good reference.

I see that some of my changes may affect the performance of every API call, while others are only executed when the app is started. What do you think of changing only the occurrences in start-up code?

dblock:
I would be down with what you're proposing @cpetschnig. Initialization in `def initialize` is ok with me, but side effects in properties aren't.

cpetschnig:
I updated the PR: 
Instance variables are now only initialized in `def initialize`. Also removed the `config.warning = true` from `spec_helper.rb`.

lib/grape/middleware/formatter.rb
```diff
@@ -5,6 +5,11 @@ module Middleware
     class Formatter < Base
       CHUNKED = 'chunked'.freeze
 
+      def initialize(*_)
+        super
+        @app_response = nil
```
dm1try: maybe `Base` class is the better place for such initialization?

cpetschnig:
Very true, I've changed it.

dm1try:
It's very good, thanks @cpetschnig!

LeFnord:
🎉

dblock:
Perfect, thanks!

PR #1532, opened 2016-12-04
sets 204 as default status for delete

LeFnord:


LeFnord:
@namusyaka, @dblock … any objections, meanings?

namusyaka:
This change looks reasonable.
In fact, the reasonable status code that can occur in the normal route of the delete route would be 204, 202, and 200.
I think that 204 is the most frequently used among them.

UPGRADING.md
```diff
@@ -41,6 +41,19 @@ Prior to this version the response would be `one is missing`.
 
 See [#1510](https://github.com/ruby-grape/grape/pull/1510) for more information.
 
+#### Change default status code of DELETE to 204
+
+**Breaking change**: Sets the default response status code for a delete request to 204.
+Doing it makes the response more explicit, cause a delete request returns in most cases an empty body – the resource was deleted/voided. By setting it to 204, other as 200, makes the handling on consumer side easier.
+
+To achieve the old behaviour, one has to set it explicitly:
```
dblock: Lets make this American, so _behavior_ :)

UPGRADING.md
```diff
@@ -41,6 +41,19 @@ Prior to this version the response would be `one is missing`.
 
 See [#1510](https://github.com/ruby-grape/grape/pull/1510) for more information.
 
+#### Change default status code of DELETE to 204
+
+**Breaking change**: Sets the default response status code for a delete request to 204.
+Doing it makes the response more explicit, cause a delete request returns in most cases an empty body – the resource was deleted/voided. By setting it to 204, other as 200, makes the handling on consumer side easier.
```
dblock: How about _A status of 204 makes the response more distinguishable and therefore easier to handle on the client side, particularly because a DELETE request typically returns an empty body as the resource was deleted or voided._

UPGRADING.md
```diff
@@ -41,6 +41,19 @@ Prior to this version the response would be `one is missing`.
...
+
+To achieve the old behaviour, one has to set it explicitly:
+```ruby
+delete :id do
+  status 200
+  'foo successfully deleted'
+end
```
dblock: Add a note like _For more information see ..._ referencing this pull request.

UPGRADING.md
```diff
@@ -41,6 +41,19 @@ Prior to this version the response would be `one is missing`.
 
 See [#1510](https://github.com/ruby-grape/grape/pull/1510) for more information.
 
+#### Change default status code of DELETE to 204
```
dblock: Should the change, so _The default status code for DELETE is now 204 instead of 200._

dblock:
I support this. Made some minor UPGRADING comments. 

I think we should document this in README as well, along with the default status code being 200 for the other verbs.

LeFnord:
@dblock … please can you provide a sample, how the changelog should looks like, sp that danger accept it?

dblock:
Sounds like you got it, thanks!

dblock:
I'm merging this.

urkle:
This seems an odd change to make..  or, IMHO if the request did return a response it should switch back to 200.. (e.g. 204 should only be returned IF there really is no content).. As it is incorrect (according to the HTTP spec) to return ANY body when a 204 response is returned.

PR #1540, opened 2016-12-25
API.logger is confusing statement

greenfieldtech-davidb:
It is confusing because it may look like `API` means `Grape::API`


dblock:
We mostly use TwitterAPI in the documentation, maybe we want to standardize on that everywhere? Care to review the rest of the readme? No need to edit CHANGELOG for this one if you don't feel like it, danger is being annoying :)

PR #1544, opened 2016-12-27
Declaring an endpoint in the namespace affects its state

faucct:
Please, see the 2 specs I have added. First one has no other endpoints in the namespace, while the second one has another endpoint defined before the one we are testing. The first spec succeeds in building `declared(params)`, while the second one tells that there are none declared.
I would be glad to locate the issue in the source code, but it was too hard, so I gave up.

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -1303,4 +1303,50 @@ describe Grape::Endpoint do
       )
     end
   end
+
+  context 'bug' do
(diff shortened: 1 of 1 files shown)


dblock:
Looks like a legit problem.

dblock:
Closing via https://github.com/ruby-grape/grape/pull/1564, @faucct you might want to check that out and potentially add some of these tests (which should now pass).

faucct:
Yes, for sure it does and have similar specs.

PR #1543, opened 2016-12-27
Support ruby 2.4

namusyaka:
[Fixnum is deprecated now](https://bugs.ruby-lang.org/issues/12005), so I guess we should drop Fixnum and use Integer instead.
@dblock @dm1try Thoughts?

PR #1545, opened 2016-12-27
adds ruby 2.4.0 support

LeFnord:


lib/grape/dsl/inside_route.rb
```diff
@@ -122,7 +122,7 @@ def status(status = nil)
         when Symbol
           raise ArgumentError, "Status code :#{status} is invalid." unless Rack::Utils::SYMBOL_TO_STATUS_CODE.keys.include?(status)
           @status = Rack::Utils.status_code(status)
-        when Fixnum
+        when integer_type
```
dblock: Was `Fixnum` removed? Can this just be `Fixnum, Integer`?

namusyaka: Nope, Fixnum is not removed but deprecated.
This should be `Integer` because `1 === Integer` will be true. Also `Fixnum` is `Integer` on ruby-2.4.

```ruby
p Fixnum #=> Integer
```

namusyaka:
Thanks for your pull request!
However, this duplicates https://github.com/ruby-grape/grape/pull/1543

LeFnord:
@namusyaka … please 😄 … didn't see it, but this one didn't drop support for ruby < 2.4.0 😉

namusyaka:
@LeFnord My description is bad. In fact, Fixnum cannot be dropped, it may be used in Ruby < 2.4.0.
Please take a look.

LeFnord:
@namusyaka … yeap I know

lib/grape/dsl/inside_route.rb
```diff
@@ -139,6 +139,15 @@ def status(status = nil)
...
       end
 
+      # TODO: maybe it should be moved out for better separation
+      def integer_type
+        if RUBY_VERSION < '2.4.0'
+          Fixnum
+        else
+          Integer
+        end
+      end
```
namusyaka: I think this separation is not needed.

LeFnord:
closed in favour of https://github.com/ruby-grape/grape/pull/1543

PR #1547, opened 2016-12-29
Remove changelog placeholder for released version

Razer6:


namusyaka:
Good catch, thank you!

PR #1548, opened 2016-12-30
Avoid failing even if given path does not match with prefix

namusyaka:
I partially reverted changes and disabled rubocop in the line. Also, I added a spec for this case.

Rubocop seems to have a bug about NumericPredicate.
@dblock @thomas-peyric Thoughts?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,3 +18,6 @@ Style/MultilineIfModifier:
 
 Style/RaiseArgs:
   Enabled: false
+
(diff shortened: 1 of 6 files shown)


dblock:
Let's merge this and deal with rubocop separately (open a bug for them @namusyaka?).

dblock:
@thomas-peyric can you please confirm this fixes your problem?

thomas-peyric:
It's OK, problem solved
thanks

Issue #1549, opened 2017-01-04
Revert #1532 set default 204 for deletes


JanStevens:
This MR: https://github.com/ruby-grape/grape/pull/1532 is plain wrong.

204 is only when there is NO CONTENT meaning I explicitly as a developer set `body false`. I now have to go through all my delete calls and explicitly set status 200, which makes no sense.

For example returning the deleted object is quite common, now we don't follow the spec because by default (even when there is body!) we return 204.

**mind blown**

dblock:
@LeFnord @namusyaka ^^^ thoughts?

dblock:
Also @urkle

JanStevens:
Or at least we should evaluate if a body is returned and then return 200 else 204.

Less magic, more explicitness, happy devs 👍

dblock:
@JanStevens you could help by PRing your suggestion on empty body, I would support that

jthornec:
This just bit me too. We're using Grape with `grape-roar` something like this:

```
delete do
  model.destroy!
  present model, with: ModelPresenter
end
```

After updating Grape, our API `DELETE` endpoints started returning `#<Model:0x00000000>` instead of the expected `{ "model": "json" }` because the new 204 status code is causing the [formatter to be skipped](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/formatter.rb#L25-L26). 

This behaviour is a bit surprising, as I didn't immediately connect the new `204 No Content` header to Grape skipping the formatter but still returning content.

@JanStevens if you don't have time to put a PR together in the next few days, I might be able to do it.

JanStevens:
@jthornec That would be awesome pretty swamped with elixir atm

dblock:
The formatter being skipped is a bigger deal than I thought, thanks for bringing it up @jthornec :(

LeFnord:
sorry for the trouble :( … but @JanStevens, this change isn't wrong, it is only a modern interpretation of possibilities, means responding the deleted entity of a resource makes not much sense IMHO, cause the API consumer always has it … and we should reduce traffic 😉 

I'm totally support your suggestion 
> Or at least we should evaluate if a body is returned and then return 200 else 204.
> Less magic, more explicitness, happy devs 👍

and think this would be the best solution for all

btw, we should think about a re-ordering of the specs, and smaller, better specs, so such failures as the formatting stuff could be found much earlier :(

JanStevens:
[...]

PR #1550, opened 2017-01-04
Use 200 as default status for deletes that reply with content

jthornec:
As per #1549, having deletes use 204 as the default status when a body is present causes some issues. Sending a 204 No Content with content will likely surprise some clients, and it causes the [formatter to be skipped](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/formatter.rb#L25-L26).

This PR adds a check for `@body.present?` so that we only default to 204 No Content when there's actually no content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 #### Fixes
 
 * [#1548](https://github.com/ruby-grape/grape/pull/1548): Avoid failing even if given path does not match with prefix - [@thomas-peyric](https://github.com/thomas-peyric), [@namusyaka](https://github.com/namusyaka).
+* [#1550](https://github.com/ruby-grape/grape/pull/1550): Use 200 as default status for deletes that reply with content - [@jthornec](https://github.com/jthornec).
 * Your contribution here.
 
 ### 0.19.0 (12/18/2016)
--- a/README.md
+++ b/README.md
@@ -1764,7 +1764,7 @@ cookies.delete :status_count, path: '/'
 
 ## HTTP Status Code
 
-By default Grape returns a 201 for `POST`-Requests, 204 for `DELETE`-Requests and 200 status code for all other Requests.
+By default Grape returns a 201 for `POST`-Requests, 204 for `DELETE`-Requests that don't return any content, and 200 status code for all other Requests.
 You can use `status` to query and set the actual HTTP Status Code
 
 ```ruby
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,36 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 0.19.1 (next)
+
+#### DELETE now defaults to status code 200 for responses with a body, or 204 otherwise
+
+DELETE responses now use a 200 status code by default when a body is present, or 204 otherwise.
+
+- In versions < 0.19.0, all DELETE requests defaulted to a 200 OK status code.
+- In version 0.19.0, all DELETE requests defaulted to a 204 No Content status code, even when content was included in the response.
+- As of version 0.19.1, DELETE requests default to a 204 No Content status code, unless content is supplied, in which case they default to a 200 OK status code.
+
+To achieve the old behavior, one can specify the status code explicitly:
+
+```ruby
+delete :id do
+  status 204 # or 200, for < 0.19.0 behavior
+  'foo successfully deleted'
+end
+```
+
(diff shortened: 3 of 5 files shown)


JanStevens:
👍 On the MR!

I'm really thinking we should just use a helper that combines status + body false, like I suggested before in the discussion:

```ruby
def return_no_content
  status 204
  body false
end
```

Why is this much better?

1. Explicit
2. No magic, maybe I want to return a 200 without any body, this will be converted to a 204 but I got a legacy client from 1989 that doesn't understand 204 (we all have our dragons :D)
3. We should have logical defaults and less magic, this really helps making Grape solid and consistent: This means we always return status 200/201(POST) unless set explicitly.

Also now we only set 204 on a delete when there is no body, this is just not consistent with other HTTP methods (why is delete so special?)

dblock:
@jthornec Add a paragraph about this to UPGRADING please?

Consider the `return_no_content` improvement too, although we can take that as a subsequent PR from @JanStevens.

@LeFnord @namusyaka can you please chip in here to make sure I don't merge something we don't want?

dblock:
I like this change because it looks like it corrects 1 specific bug: if there's a body on DELETE, don't return 204, but 200. I can live with that.

Now, I do believe there're a lot of sensible arguments in #1549 around what to do "in general", notably the following does make sense to me too:

```ruby
if request.request_method == Grape::Http::Headers::POST
  201
elsif @body.empty?
  204
else
  200
end
```

JanStevens:
@dblock Again why the magic behaviour for a specific method + no body returned? Helper would be more than sufficient. I keep bitching about this since I strongly feel this is magical and quite unexpected behaviour (what if i want to delete return nothing but want status 200, legacy clients for example?)

I vote for helper only :D

jthornec:
I've updated this PR to include the `return_no_content` helper and a section in `UPGRADING.md` explaining the changes.

dblock:
@JanStevens I hear you. If we hadn't shipped this change in a previous version I would probably just say "you're right" and leave things as is. @LeFnord @namusyaka I need some help deciding.

namusyaka:
yes, l'll confirm this within a few hours.

namusyaka:
First of all, we have to apologize that [that changes](https://github.com/ruby-grape/grape/pull/1532) in the case were insufficient.
Actually, the behavior that the formatter skipped, for example, wasn't what we meant.

Well, regarding this pull request, never this change is wrong.
Let's check rfc.

> A successful response SHOULD be 200 (OK) if the response includes an entity describing the status, 202 (Accepted) if the action has not yet been enacted, or 204 (No Content) if the action has been enacted but the response does not include an entity. 
>
>https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7

As written here, this changes is reasonable (but it's not complete..).

>Also now we only set 204 on a delete when there is no body, this is just not consistent with other HTTP methods (why is delete so special?)

It's true that there is no consistency. However, we should commit considering changes at 0.19.0.
I can't say that this change certainly keeps consistent behavior in grape.
From developers who respected the last change, I think this change is the least damaging approach.
Also, The delete route can be said to be special because we made a mistake.

Therefore, I'd like to vote to both helper and this change.

dblock:
Thanks @namusyaka, @LeFnord?

LeFnord:
as I said here https://github.com/ruby-grape/grape/issues/1549#issuecomment-270461521, I see this as the best solution
 > Also now we only set 204 on a delete when there is no body, this is just not consistent with other HTTP methods (why is delete so special?)

good point, for that @dblock's suggestion for a general handling 
```ruby
when Grape::Http::Headers::POST
  201
when !@body.present?
  204
else
  200
end
```
should be implemented

but this can also be done in a different PR IMHO

dblock:
I am going to merge this. Lets let it sit for a bit on master. I'd like @JanStevens and @jthornec to at least confirm you're happy (enough) with this change.

jthornec:
I'm happy with this change.

My issues were that the formatter wasn't running, and the 204 with content might confuse my client. This fixes both of those things, and all my app's tests pass now, so I'm good to move forward with the upgrade once this gets released.

JanStevens:
Thanks @jthornec for the effort

dblock:
I am going to cut a release today.

LeFnord:
@dblock … 👍 cool, so I can run tests and cut new ones for grape-entity and grape-swagger

dblock:
And done.

PR #1555, opened 2017-01-12
Added code coverage w/ Coveralls.

dblock:


LeFnord:
@dblock would be nice to see it in the other gems, I'm not sure if I can do it 😉 

How about using of [hound ci](https://houndci.com), with that the commiter see there rubocop offences and can adapt theyr PRs …
… it would be really nice to use it

dblock:
Yes you can, enabling it in coveralls (login as yourself via github) and the rest is just code as in this PR.

dblock:
I've tried Hound, I personally am not in love with it because it makes rubocop an external thing. But obviously feel free to do it for anything you're maintaining.

PR #1556, opened 2017-01-15
Add information about #mount inheritance

### LINKED ISSUE #646: rescue_from must precede mount
Documentation does not say this, but it appears that rescue_from statements must precede mount statements in an api class, otherwise rescue_from is ignored.

dm1try:
closes #646 [ci skip]


dblock:
Thanks @dm1try

PR #1559, opened 2017-01-18
should be able to pass `nil` to an optional array values param

ghiculescu:
this bug was introduced in https://github.com/ruby-grape/grape/commit/3540ea25a6cc82a700a572b2c8f483b66302d7e6#diff-45b62b4ed0aed7ceff346c347f76570e

previously, you were able to pass `nil` to an endpoint for an attribute that was marked as optional and had a set array of values. after the above change, this was no longer possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 #### Fixes
 
-* Your contribution here.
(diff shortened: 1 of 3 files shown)


dblock:
Makes sense, thanks, merged.

PR #1564, opened 2017-01-26
Failing test to expose a declared params bug

bmarini:
I'm working on enforcing declared params usage in my codebase, and in doing so I started running into issues where I thought a declared param should be available via a parent scope, but for some reason wasn't.

This is one of those cases. It works as expected if I remove the route defined just above the namespace 'foo'.


bmarini:
I'll be diving in to try and figure out whats wrong, but if this is expected behavior or someone has some hints about a fix, please let me know.

bmarini:
Well, I got the tests to pass, BUT I doubt the way I fixed it is the right approach. The problem is I don't really understand the difference between `route_setting(:delcared_params)` and `route_setting(:saved_declared_params)`, or how and why they get set under different nesting scenarios.

The fix was:

``` ruby
          # Declared params including parent namespaces
          all_declared_params = route_setting(:saved_declared_params).flatten | Array(route_setting(:declared_params))

          # Declared params at current namespace
          current_namespace_declared_params = route_setting(:saved_declared_params).last & Array(route_setting(:declared_params))
```

bmarini:
Same issue as https://github.com/ruby-grape/grape/pull/1544 looks like.

dblock:
I think this is fine, I'm going to merge it.

The saved declared params are just whatever was there at endpoint construction, see [this](https://github.com/ruby-grape/grape/blob/dc5e4b099f09c0214706cf7cc636d8e963d52c07/lib/grape/endpoint.rb#L83).

PR #1565, opened 2017-01-26
Expose bug with response codes

bmarini:
This is related to the 204 no content changes made #1532 and #1549 #1550 

Seems like the behavior should be as follows:

```
1. If status was not set: use that.
2. If status was not set:
    a) If body is formatted to an empty string, use 204 No Content
    b) If body is formatted to a non empty string, use a default status for method, (GET: 200, POST: 201, PATCH: 200, DELETE: 200). 
```

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -115,6 +115,26 @@ describe Grape::Endpoint do
       expect(memoized_status).to eq(201)
       expect(last_response.body).to eq('Hello')
     end
+
+    it 'is set as default to 204 for delete with no content' do
+      subject.delete('/home') do
+      end
+
+      delete '/home'
+      expect(last_response.status).to eq(204)
+      expect(last_response.body).to eq('')
+    end
+
+    it 'is set as default to 200 for delete with content' do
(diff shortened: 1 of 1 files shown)


bmarini:
I ran into some issues trying to fix at the status method in `inside_route.rb`:

``` ruby
      def status(status = nil)
        case status
        when Symbol
          raise ArgumentError, "Status code :#{status} is invalid." unless Rack::Utils::SYMBOL_TO_STATUS_CODE.keys.include?(status)
          @status = Rack::Utils.status_code(status)
        when Integer
          @status = status
        when nil
          return @status if @status
          case request.request_method.to_s.upcase
          when Grape::Http::Headers::POST
            201
          when Grape::Http::Headers::DELETE
            if @body.present?
              200
            else
              204
            end
          else
            200
          end
        else
          raise ArgumentError, 'Status code must be Integer or Symbol.'
        end
```

1. @body isn't as the return value of the endpoint block
2. If @body is `{}` or `[]` it'll still evaluate as not present, even though it'll serialize to json as a non empty string.

So I think the logic has to be moved til after the formatters run to know if the status should be set to 204 or not.

Issue #1535, opened 2016-12-16
Parameter maximum value


CharlesP:
When adding a parameter (i.e. per_page), is there a way to set the maximum value?

The cleanest (only) way I've found so far is to pass the entire array of possible options:
`optional :per_page, values: [*1..100], default: 10, type: Integer`

Is there a better way to achieve this?

dblock:
I'll label this as a feature request to support `min` and `max`, obviously there be dragons for non enumerable types ;)

jlfaber:
#1568 added a new feature to the ValuesValidator that allows a Proc to be supplied for anonymous custom validations.  Using this, your example above might look like this.
```ruby
optional :per_page, values: { proc: ->(v) { v.positive? && v <= 100 } }, default: 10, type: Integer
```
I should note that the ValuesValidator already accepted Ranges, so you also could have done this.
```ruby
optional :per_page, values: 1..100, default: 10, type: Integer
```
For both your original example and my Range alternative, the validator will call `include?` on the passed value.  And, in both cases, that could result in up to 100 individual tests as it enumerates the Array or Range to determine whether the supplied value is a member.  This is probably not catastrophic with a relatively small Range, but doing something like 0..1000000 could cause some performance problems.

A way to avoid this is (again) to use `proc` to test using `cover?` instead of `include?`, which will instead check that the value is between the endpoints of the range.
```ruby
optional :per_page, values: { proc: ->(v) { (1..100).cover? v } }, default: 10, type: Integer
```

dblock:
Should we close this?

CharlesP:
@jlfaber Thanks, that's a big help.

@dblock Yes, this looks like a valid workflow for this.

PR #1568, opened 2017-01-31
Add test_with option to values validator

jlfaber:
This PR adds a new option `test_with` to the `values` validator.  If present, it must be a Proc that accepts a single argument (the param value) and returns a truthy value if the argument is valid.

I don't think this has been explicitly asked for, and one could certainly do the same thing using a custom validator.  But this seemed a little more convenient for simple validations that aren't already handled by the existing built-ins.

This could be an acceptable workaround/solution for #1535 which requests the ability to specify a maximum parameter value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * [#1555](https://github.com/ruby-grape/grape/pull/1555): Added code coverage w/Coveralls - [@dblock](https://github.com/dblock).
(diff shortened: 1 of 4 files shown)


dblock:
Quick thought. Maybe `test_with` could be `compare_with` or just `with`?

jlfaber:
Fair point.  Not in love with `compare_with` as it's not necessarily a compare.  I kind of like `proc` though.  I'll tweak that and repush.

jlfaber:
Renamed `test_with` to `proc` and cleaned up some unused test code to (slightly) improve coverage.

dblock:
Looks great, merged.

Issue #1511, opened 2016-10-16
Upgrade mustermann dependency ?


jarthod:
I'm currently having trouble updating an app to Rails 5 because of this dependency, here's why:
- `rails 5` depends on `rack 2`
- `rack 2` requires `sinatra 2+`
- `sinatra 2` requires `mustermann 1`
- `grape 0.18` depends on `mustermann-grape 0.4` which depends on `mustermann 0.4`

So this is a deadlock and I can't bundle this. I tried to find `mustermann-grape` to see if we could upgrade the dependency but I can't find it, It seems with a github search that the Grape extension is now in [mustermann-contrib](https://github.com/sinatra/mustermann/blob/master/mustermann-contrib/lib/mustermann/grape.rb).

So I guess grape is now supposed to depend on `mustermann-contrib` ?

namusyaka:
@jarthod Thanks for yourt report! I'm going to investigate the problem in a few days.

dblock:
I heard from @zzak that the conttrib stuff was a result of merging many gems? Maybe he can help shed some light on what's going on?

dblock:
I think we should use whatever is maintained.

namusyaka:
Yes,  both `mustermann-contrib` and `mustermann-grape` can be maintained.
Btw I think `mustermann-grape` is hosted by me, so we can feel free to cut a release of the gem.

dblock:
Sounds good. Maybe you want to move it into the Grape organization @namusyaka?

namusyaka:
Exactly. Mustermann will be continued to be used by us.
In other words, the gem is considered to be the most desirable that we can handle.

dblock:
If you don't have org permissions move it to me and I'll move it to the org @namusyaka.

namusyaka:
I extracted [mustermann-grape gem](https://github.com/ruby-grape/mustermann-grape), and [dropped the code base from Mustermann repository](https://github.com/sinatra/mustermann/pull/54).

zzak:
Sounds good to me, let me know if you need anything from me.

Thanks

shaneog:
Also stuck due to the same upgrade. Would be great to use the new gem and release. Is there anything I can do to help?

namusyaka:
Sorry fory the late reply. I've just released [v1.0.0.beta2](https://rubygems.org/gems/mustermann-grape/versions/1.0.0.beta2).

I'm going to close this issue, but if you still have the problem, please let me know.
[...]

PR #1569, opened 2017-02-01
upgrade mustermann-grape to 1.0.0

namusyaka:
ref #1511

### DIFF
--- a/grape.gemspec
+++ b/grape.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.license     = 'MIT'
 
   s.add_runtime_dependency 'rack', '>= 1.3.0'
-  s.add_runtime_dependency 'mustermann-grape', '~> 0.4.0'
+  s.add_runtime_dependency 'mustermann-grape', '~> 1.0.0'
   s.add_runtime_dependency 'rack-accept'
   s.add_runtime_dependency 'activesupport'
   s.add_runtime_dependency 'multi_json', '>= 1.3.2'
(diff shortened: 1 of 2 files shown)


dblock:
That's a very rigid dependency and prevents us from releasing Grape v. next. Would `~> 1.0.0` include the beta?

dblock:
I would also add a CHANGELOG line that says that #1511 is fixed by this PR.

namusyaka:
>Would ~> 1.0.0 include the beta?

Nope, `~> 1.0.0` does not include beta release.
But maybe we don't have to unify versioning with official mustermann. What do you think?

>I would also add a CHANGELOG line that says that #1511 is fixed by this PR.

Fixed.

dblock:
I don't think unifying versioning with mustermann is needed, but otherwise I think you should release 1.0 (not a beta) and call it a day. I don't think we should merge this PR until Grape is forward compatible.

namusyaka:
I got your point, it's reasonable.
Okay, I'm going to confirm the plan for releasing mustermann-1.0.0.

namusyaka:
@dblock mustermann-1.0.0 has been released, so I cut a new release of mustermann-grape-1.0.0.

grape.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.license     = 'MIT'
 
   s.add_runtime_dependency 'rack', '>= 1.3.0'
-  s.add_runtime_dependency 'mustermann-grape', '~> 0.4.0'
+  s.add_runtime_dependency 'mustermann-grape', '1.0.0'
```
dblock: Should be `~> 1.0.0` or it will lock at 1.0 no?

namusyaka: Right. Fixed.

PR #1570, opened 2017-02-01
Make versioner consider the mount destination path

### LINKED ISSUE #636: API version not found when api nested
I have two api(APIOne and APITwo) mounted in API, which is mounted in routes, but when I access `http://lvh.me:3000/one/v1/hello`, I got `404 API Version Not Found`.

``` ruby
class API < Grape::API
  mount APIOne => '/one'
  mount APITwo => '/two'
end
```

``` ruby
class APIOne < Grape::API
  version 'v1', :using => :path

  get '/hello' do
    'hello'
  end
end
```

``` ruby
class APITwo < Grape::API
  version 'v1', :using => :path

  get '/world' do
    'world'
  end
end
```

``` ruby
Testgrape::Application.routes.draw do
  mount API, :at => "/"
end
```

I have traced the code about parsing path, I found problem is here https://github.com/intridea/grape/blob/master/lib/grape/middleware/versioner/path.rb#L34 The `path` here is `/one/v1/hello`, then `potential_version = pieces[1]`, `potential_version` got `one`, instead of `v1`.

namusyaka:
Fixes #636 

If my perception is correct, it is enough that only this changes is enough.
@dblock Thoughts?


lib/grape/endpoint.rb
```diff
@@ -295,7 +295,8 @@ def build_stack(helpers)
         stack.use Grape::Middleware::Versioner.using(namespace_inheritable(:version_options)[:using]),
                   versions: namespace_inheritable(:version) ? namespace_inheritable(:version).flatten : nil,
                   version_options: namespace_inheritable(:version_options),
-                  prefix: namespace_inheritable(:root_prefix)
+                  prefix: namespace_inheritable(:root_prefix),
+                  mount_path: namespace_stackable(:mount_path)[0]
```
dblock: Maybe `.first`?

namusyaka: Fixed.

lib/grape/middleware/versioner/path.rb
```diff
@@ -25,6 +25,7 @@ def default_options
 
         def before
           path = env[Grape::Http::Headers::PATH_INFO].dup
+          path.sub!(mount_path, '') if mount_path && path.start_with?(mount_path)
```
namusyaka: Indeed, I have just force-pushed changes

namusyaka:
@dblock Could you confirmed current changes?

dblock:
Looks good.

Issue #1574, opened 2017-02-14
should i see nil values for nested declared params?


thogg4:
I have an endpoint that needs either a nested address or an address line:

```
params do
  optional :address_line

  optional :address do
    optional :street
    optional :city
    optional :state
  end

end
```

If no params are passed in, i get this in `declared`:

```
{
  address_line: null,
  address: []
}
```

I would expect this (with `include_missing` set to true):

```
{
  address_line: null,
  address: {
    street: null,
    city: null,
    state: null
  }
}
```

Am I doing something wrong, or is this how it should work?

dblock:
I think the default here is to treat `address` as an array (for historical and backward compatibility reasons), does adding `optional :address, type: Hash do ...` change this?

thogg4:
@dblock im sorry i forgot to add the `type`
I have been using `type: Hash`, but getting the same result.

thogg4:
@dblock it would break backward compatibility, but would you have any interest in another option (similar to `include_missing`), that would add this?

dblock:
It seems to me that with `type: Hash` this is a bug. But I didn't investigate. Maybe start by writing a test for this - and see if there's already a test in this vein.

thogg4:
It looks like this may be intended based on this commit:
https://github.com/ruby-grape/grape/commit/f60e58eec0d9cf56ddab5bd078f7a0436a71650d

thogg4:
From what I can tell `declared` doesn't take the `type` that is set in the `params` block into account at all.

dblock:
Closed via https://github.com/ruby-grape/grape/pull/1575.

PR #1575, opened 2017-02-15
refactor declared and include nil values for missing nested params

thogg4:
addresses #1574

### DIFF
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -37,8 +37,6 @@ module Grape
           end
         end
 
-        private
-
         def declared_array(passed_params, options, declared_params)
           passed_params.map do |passed_param|
             declared(passed_param || {}, options, declared_params)
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -37,6 +37,8 @@ module Grape
           end
         end
 
+        private
+
         def declared_array(passed_params, options, declared_params)
           passed_params.map do |passed_param|
             declared(passed_param || {}, options, declared_params)
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -61,10 +61,10 @@ module Grape
         def get_correct_declared_params(options)
           if options[:include_parent_namespaces]
             # Declared params including parent namespaces
-            declared_params = all_declared_params = route_setting(:saved_declared_params).flatten | Array(route_setting(:declared_params))
+            declared_params = route_setting(:saved_declared_params).flatten | Array(route_setting(:declared_params))
           else
             # Declared params at current namespace
-            declared_params = current_namespace_declared_params = route_setting(:saved_declared_params).last & Array(route_setting(:declared_params))
+            declared_params = route_setting(:saved_declared_params).last & Array(route_setting(:declared_params))
           end
 
           raise ArgumentError, 'Tried to filter for declared parameters but none exist.' unless declared_params
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -47,8 +47,8 @@ module Grape
 
         def declared_hash(passed_params, options, declared_params)
           declared_params.each_with_object(Hashie::Mash.new) do |declared_param, memo|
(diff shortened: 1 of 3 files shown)


thogg4:
@dblock Tests are passing now. Can I get some feedback?

spec/grape/endpoint_spec.rb
```diff
@@ -257,10 +257,29 @@ def app
...
+            optional :sixth
+            optional :nested_three, type: Hash do
+              optional :seventh
+            end
+          end
         end
       end
     end
 
+    it 'should show nil for nested params if include_missing and include_missing_nested are true' do
```
dblock: There's no more `include_missing_nested`.

lib/grape/dsl/inside_route.rb
```diff
@@ -26,42 +26,56 @@ def self.post_filter_methods(type)
       # Methods which should not be available in filters until the before filter
       # has completed
       module PostBeforeFilter
-        def declared(params, options = {}, declared_params = nil)
+        def declared(passed_params, options = {}, declared_params = nil)
```
dblock: Maybe unrename `passed_params` to just `params`, will be easier to grok.

lib/grape/dsl/inside_route.rb
```diff
@@ -26,42 +26,56 @@ def self.post_filter_methods(type)
...
           else
-            declared_params.each_with_object(Hashie::Mash.new) do |key, hash|
-              key = { key => nil } unless key.is_a? Hash
+            declared_hash(passed_params, options, declared_params)
+          end
+        end
 
-              key.each_pair do |parent, children|
-                output_key = options[:stringify] ? parent.to_s : parent.to_sym
+        def declared_array(passed_params, options, declared_params)
```
dblock: These should probably be `private`.

lib/grape/dsl/inside_route.rb
```diff
@@ -26,42 +26,56 @@ def self.post_filter_methods(type)
...
+          passed_params.map do |passed_param|
+            declared(passed_param || {}, options, declared_params)
+          end
+        end
 
-                next unless options[:include_missing] || params.key?(parent)
+        def declared_hash(passed_params, options, declared_params)
+          declared_params.each_with_object(Hashie::Mash.new) do |declared_param, memo|
+            # if it is not a Hash then it does not have children.
```
dblock: Nitpick, this comment should not have a period if it's not a full sentence starting with a Capital.

spec/grape/endpoint_spec.rb
```diff
@@ -257,10 +257,29 @@ def app
...
+      inner_params = nil
+      subject.get '/declared' do
+        inner_params = declared(params, include_missing: true)
```
dblock: I realize we do this elsewhere, but it's bad practice - maybe fix this and other instances?

This works, but is hacky and abuses the fact that the inner function still has access to outer variables. Below you check that the GET call succeeded with a 200, but then reach in for the value of `inner_params`. Be consistent, return `declared(...)` and examine the JSON response instead of relying on this.

dblock:
I have fairly cosmetic change requests, otherwise this looks OK. Mostly trusting the fact that no existing specs broke :)

lib/grape/dsl/inside_route.rb
thogg4: I actually added this in because I thought it made it a little easier to grok.
When this is called and the declared child params params are passed as an argument, its a bit easier to remember that `passed_params` are always the params passed into the method and `declared_params` are the params that were declared in the `params` block

thogg4:
@dblock Thanks for your comments, I have done the majority of those things. Please see my comment about naming `params` and let me know what you think.

lib/grape/dsl/inside_route.rb
dblock: It's totally cool.

dblock:
Great work @thogg4, merged.

tjwp:
This change breaks one of our existing applications. We are using `declared(params, include_missing: false)` with params containing a nested Hash. There are params within the Hash that are optional without a default. With this change, the unspecified params are included in the Hash with nil values. Prior to this change, keys for the unspecified params are not present, and this is the behavior I would expect with `include_missing: false`.

Issue #1572, opened 2017-02-09
Invalid rack response.


dopykuh:
Please see here https://github.com/puma/puma/issues/1208

namusyaka:
@dopykuh Could you provide a failing project as github repository?

dopykuh:
Its not as easy to reproduce, it could also possibly depend on nginx proxy params.
```
proxy_set_header Host $http_host;
proxy_set_header X-Real-IP $remote_addr;
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-Proto $scheme;
proxy_redirect off;
proxy_read_timeout 86400;
proxy_set_header Connection '';
proxy_http_version 1.1;
chunked_transfer_encoding off;
proxy_buffering off;
proxy_cache off;
```

dopykuh:
This 2 lines will produce my exception https://github.com/ruby-grape/grape/commit/dab3c095a7470fdf8b9dff4475fe451a04d7b7a0#diff-e9e8424a5238d48301e313d8fe285697

dblock:
I wouldn't be so sure that this is a grape bug. It looks like the response is a hash that is not being formatted into a string. Which content type is here? What is the response from the `DELETE` (implemented by the user)?

dopykuh:
**Request on v.18.0**
```
curl localhost:9000/api/v1/ldap?mac=00%3A50%3A56%3A00%3A00%3A00 -X DELETE -i; echo
HTTP/1.1 201 Created
Content-Type: application/json
Content-Length: 96

{"ip":"10.10.10.1","mac":"00:50:56:00:00:00","gateway":"10.10.10.10","netmask":"255.255.255.224"}
```
**Request on v.19.0**
```
curl localhost:9000/api/v1/ldap?mac=00%3A50%3A56%3A00%3A00%3A00 -X DELETE -i; echo
HTTP/1.1 500 Internal Server Error


```

dopykuh:
The issue could also been appeared earlier, but it looks like an grape bug. You only setup a default right now, but response_body shouldn't be served as hash or not?

dblock:
This is a little more complicated. In 0.19.0 we made the server return 204 on a DELETE, then in 0.19.1 we fixed https://github.com/ruby-grape/grape/pull/1550 to return 200 when there's no content. That's ultimately what's causing this, but it's not necessarily a bug. 

[...]

PR #1579, opened 2017-02-19
PR for issue #1578 

### LINKED ISSUE #1578: DELETE endpoint returned value not formatted and results in 204
Hi,

Grape `0.19.1`

The return value of a DELETE endpoint is not considered has `@body.present?`. Therefore, the response isn't formatted has expected and it returns a 204. Is it normal ?

The following spec fails :

```ruby
describe 'delete 200, with return value has body' do
      it 'responds to /example' do
        subject.send(:delete, '/example') do
          'deleted'
        end
        send(:delete, '/example')
        expect(last_response.status).to eql 200
        expect(last_response.body).not_to be_empty
      end
    end
```
Thanks

eproulx-petalmd:
Closes #1578 and #1572


dblock:
I think this is the same as https://github.com/ruby-grape/grape/issues/1572, and we've been wanting a fix, right?

lib/grape/endpoint.rb
```diff
@@ -265,7 +265,10 @@ def run
         run_filters afters, :after
         cookies.write(header)
 
-        # The Body commonly is an Array of Strings, the application instance itself, or a File-like object.
+        #status verifies body presence when DELETE
+        @body ||= response_object if response_object.present?
```
dblock: I think the `if` is redundant.

dblock:
Run `rubocop -a` then `rubocop --auto-gen-config` to cleanup whatever rubocop is unhappy about.

lib/grape/endpoint.rb
eproulx-petalmd: Indeed, i'll remove it

lib/grape/endpoint.rb
```diff
@@ -265,7 +265,10 @@ def run
         run_filters afters, :after
         cookies.write(header)
 
-        # The Body commonly is an Array of Strings, the application instance itself, or a File-like object.
+        # status verifies body presence when DELETE
+        @body ||= response_object
+
+        # The Body commonly is an Array of Strings, the application instance itself, or a File-like object
         response_object = file || [body || response_object]
```
dblock: I believe `body` here is the same as `body`, so this should be `file || body`?

spec/grape/endpoint_spec.rb
```diff
@@ -1149,6 +1149,39 @@ def memoized
       end
     end
 
+    describe 'delete 200, with a return value (no explicit body)' do
+      it 'responds to /example delete method' do
+        subject.send(:delete, '/example', anchor: false) do
```
dblock: I don't think you need to `send` these, `delete` is public, so `subject.delete ...`

spec/grape/endpoint_spec.rb
```diff
@@ -1149,6 +1149,39 @@ def memoized
...
+          'deleted'
+        end
+        send(:delete, '/example/')
```
dblock: Same, just `delete '/example'`, these are rack helpers.

lib/grape/endpoint.rb
eproulx-petalmd: Yes, it's the same. I forgot to remove it. On it 👍 
Thanks

dblock:
You need a CHANGELOG and it's good to go. Thx.

dblock:
Merged, thanks.

PR #1583, opened 2017-02-27
Fix spec warnings

axfcampos:
Hi all,

When running specs with warnings on, grape outputs a lot of warnings.

This PR attempts to fix all of them.

With regards to the warnings: `warning: instance variable @disable_warnings not initialized`, by the `hashie` gem, I have opened a PR there to fix those: https://github.com/intridea/hashie/pull/416


dblock:
Interesting. This feels like a workaround rather than a fix. Why wouldn't something like `@status` for example _not_ be defined?

axfcampos:
Hi @dblock the warnings occur because `@status` is in fact not defined when it is accessed. This may only happen in the specs...

I did not look into implementation details, my objective was to clear three warnings that were showing in my projects specs, but since I'm here might as well fix them all 😄 

Having warnings not suppressed has saved me from some bugs in the past, but it comes with the caveat of having to fix them if we want a sane output when running tests 😛 

Would you like me to change something on the PR?

dblock:
Yes, lets have this discussion in https://github.com/intridea/hashie/pull/416 first? Same problem but fewer changes.

namusyaka:
Thanks for your contribution.
I'd like to postpone this until after 1.0.0 release is made.

dblock:
I want to revive this. I think a fix that says `if defined(:@foobar)` is not what we want. Lets initialize variables where needed instead? What do you think @axfcampos?

PR #1589, opened 2017-03-09
Fix for issue #726

inclooder:
This changes should fix issue described in #726. 
Default format is used when there is no Content-type.
When Content-type is provided but it's not on the list then the error is thrown.
Normal behavior otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [#1564](https://github.com/ruby-grape/grape/pull/1564): Fix declared params bug with nested namespaces - [@bmarini](https://github.com/bmarini).
 * [#1567](https://github.com/ruby-grape/grape/pull/1567): Fix values validator when value is empty array and apply except to input array - [@jlfaber](https://github.com/jlfaber).
 * [#1569](https://github.com/ruby-grape/grape/pull/1569), [#1511](https://github.com/ruby-grape/grape/issues/1511): Upgrade mustermann-grape to 1.0.0 - [@namusyaka](https://github.com/namusyaka).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [#1564](https://github.com/ruby-grape/grape/pull/1564): Fix declared params bug with nested namespaces - [@bmarini](https://github.com/bmarini).
 * [#1567](https://github.com/ruby-grape/grape/pull/1567): Fix values validator when value is empty array and apply except to input array - [@jlfaber](https://github.com/jlfaber).
 * [#1569](https://github.com/ruby-grape/grape/pull/1569), [#1511](https://github.com/ruby-grape/grape/issues/1511): Upgrade mustermann-grape to 1.0.0 - [@namusyaka](https://github.com/namusyaka).
+* [#1589](https://github.com/ruby-grape/grape/pull/1589): Fix for issue #726 - [@inclooder](https://github.com/inclooder).
```
dblock: Describe what the problem is, otherwise users have click on issues in changelog to figure out what the deal is.

inclooder:
@dblock Should be OK

AndrewRayCode:
How do you get logging around this? Lost a few days trying to track down why bumping grape blocked a pubsub system, and it's because it's coming in as `text/plain`. Makes sense it errors, however I can't find anything in the logs except a `406` returned from the API, I had to hack into the guts of the gem to figure out it was coming from here. I also installed the grape logging middleware, and tried adding my own, but was unable to get useful logs to show up.

dblock:
The detail is in the HTTP response, but logging the body seems like a bad idea generally, no? Feel free to open a new issue or suggest what we can do here.

AndrewRayCode:
Am I correct in my analysis that this changes discards anything set by `?format=` in the URL?

https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/formatter.rb#L120-L127

dblock:
@AndrewRayCode Why do you think that? If anything was dropped it wasn't a spec-ed behavior.

PR #1590, opened 2017-03-09
allow undefined route options

thogg4:
I would like to be allowed to add arbitrary options into a route description like this:

```
desc 'get an object' do
  option awesome_value
end
```

and then have that option be available in `route.options`

after a little looking into it, using `method_missing` like this works, but i wonder if there is a solution that would be preferred over this.

Any thoughts?

### DIFF
--- a/lib/grape/util/strict_hash_configuration.rb
+++ b/lib/grape/util/strict_hash_configuration.rb
@@ -26,6 +26,10 @@ module Grape
         def to_hash
           @settings.to_hash
         end
+
+        def method_missing(setting_name, value)
+          @settings[setting_name] = value
+        end
       end
 
       def self.config_class(*args)


dblock:
What's your scenario for this?

Generally we want strong DSLs, ie. no arbitrary settings.

thogg4:
@dblock Yeah I see what you mean.
I am trying to use it in the grape-swagger gem, so the settings I would add would be specific to that gem (such as `hidden` for hiding a route), and that would make them inappropriate to add to grape. How else would you suggest accomplishing something like that?

Maybe using `route_settings`?

LeFnord:
@dblock … think the background of it is this [deprecation warning](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/desc.rb#L50), which pops-up in the setup of @thogg4 – maybe you can provide some details on it – but not in the travis build of grape-swagger, where the option hash would be used very often.

I see two options:
1. remove it, and allow option hash again, so it would be possible by other gems to add gem specific keys, without defining it in grape itself
2. use method missing, as it would be suggested here

If I interpret it correctly, the options of the `desc` wouldn't be used very often from grape itself, 
and sometimes they would be handled different → see: `params` key, vs. `params` block.

Issue #1604, opened 2017-04-03
allow_blank for values


brlo:
Hello.
Option `allow_blank` ignores when using option `values` like this:

```ruby
optional :test_param_name, type: String, values: ['A', 'B'], allow_blank: true
```

This request:
`GET /test_method_name?test_param_name=`

returning error:
`{"error":"invalid_params","message":"test_param_name does not have a valid value"}, status: 400`

grape (0.17.0)

workaround: `values: ['A', 'B', '']`

dblock:
First, you should try with the latest Grape, there have been a lot of fixes around `allow_blank`. Then write a spec that reproduces this here, please? Maybe even a fix?

brlo:
ok. i'll do.

PR #1607, opened 2017-04-05
Allow blank with values

kimquy:
Fix this issue https://github.com/ruby-grape/grape/issues/1604

### DIFF
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -343,7 +343,7 @@ module Grape
 
       validators.each do |validator|
         begin
-          validate.validators = validators
+          validator.validators = validators
           validator.validate(request)
         rescue Grape::Exceptions::Validation => e
           validation_errors << e
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -343,6 +343,7 @@ module Grape
 
       validators.each do |validator|
         begin
+          validate.validators = validators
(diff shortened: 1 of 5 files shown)


lib/grape/endpoint.rb
```diff
@@ -339,10 +338,9 @@ def lazy_initialize!
     end
 
     def run_validators(validators, request)
-      validation_errors = []
-
-      validators.each do |validator|
+      validators.each_with_object(validation_errors = []) do |validator|
         begin
+          validator.validators = validators
```
dblock: This introduces state during validation into the validator proper, I don't think that's OK since these are singleton objects.

Or do we do this elsewhere?

dblock:
I think it's odd that a validator has to look at another validator, but then again I don't know whether we do this elsewhere. Maybe we can introduce a more reliable interface giving a validator access to all other validators at runtime?

kimquy:
Hi @dblock, thanks for the feedback. I updated the PR. I look at couple of other places to find the best place for this `validators`. I feel that the `Base` class seem ok. I also feel that the `Base` class got access to all current validators might come in handy for many cases do you think ?

dblock:
The idea is right, but the way you implemented this makes it _global_, so one API is going to affect another! I didn't try to write a repro, but I suspect it will break everything :)

The "other validators" have to come from a route setting.

PR #1610, opened 2017-04-09
Replace Hashie::Mash with ActiveSupport::HashWithIndifferentAccess.

dblock:
Replaces #1594, includes a way to set the class name globally.


dblock:
@namusyaka @marshall-lee @james2m Would appreciate comments/feedback will iterate to green and everyone being happy. I am a little concerned that we're losing method access, since it probably will affect quite a few people but I think it's the right thing to do.

dblock:
💚

james2m:
@dblock thanks for finishing that up, looks similar to what I have locally, I just didn't have time to polish up and complete as I'm traveling until 17th. The cleaned up tests is a big relief too! 😬

dblock:
You've done an excellent job up this far @james2m, thanks for your big effort!

namusyaka:
💚 👍 
I also agree that method access of parameters isn't good design.

dblock:
How are we feeling about merging this?

james2m:
I'm happy with it. Using it in a new app with Rails 5.1 release candidate. If I had more tinkering time I'd probably reorganise the files under /extensions as it feels like a dumping ground right now, but nothing functional.

dblock:
I merged this, lets see what happens ;)

PR #1615, opened 2017-04-13
Bugfix: values without value breaks type and default

jlfaber:
The checking of `type` and `default` validators in `params_scope.rb` is broken in cases where `values` is a Hash containing only the `except` attribute (and no `value` attribute).  For example both of the following are broken:
```
params do
  requires :foo, type: String, values: { except: ['bar', 'baz'] }
  requires :foo2, values: { except: ['bar2', 'baz2'] }, default: 'bing'
end
```

### DIFF
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -232,17 +232,20 @@ module Grape
         default = validations[:default]
         doc_attrs[:default] = default if validations.key?(:default)
 
-        values = validations[:values]
-        if values.is_a? Hash
-          excepts = values[:except]
-          values = values[:value]
-          values = nil if values.is_a? Proc
+        if (values_hash = validations[:values]).is_a? Hash
+          values = values_hash[:value]
+          excepts = values_hash[:except]
+        else
+          values = validations[:values]
         end
-
         doc_attrs[:values] = values if values
 
+        # NB. values and excepts should be nil, Proc, Array, or Range.
+        # Specifically, values should NOT be a Hash
+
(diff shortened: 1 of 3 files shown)


lib/grape/validations/params_scope.rb
```diff
@@ -232,16 +232,25 @@ def validates(attrs, validations)
         default = validations[:default]
         doc_attrs[:default] = default if validations.key?(:default)
 
-        values = options_key?(:values, :value, validations) ? validations[:values][:value] : validations[:values]
+        values = validations[:values]
+        if values.is_a? Hash
+          excepts = values[:except]
+          values = values[:value]
+          values = nil if values.is_a? Proc
```
dblock: This is a little hard to read. Maybe flatten this thing, if I understand correctly you can do something like this:

```
values = options_key?(:values, :value, validations) ? validations[:values][:value] : validations[:values]
excepts = ...
```

Give it a try?

jlfaber: I'm looking for a little different behavior that's tricky with a one-liner.  I really don't ever want `values` to be a Hash.  But that's what the one-liner above will do if the Hash doesn't have a `value` attribute.  Fair point about reassigning variables though.  I'll push another commit that clarifies it a bit (I hope).

jlfaber: I should add that, tbh, I'm not a fan of the current syntax for except.  After playing around with it, I think we'd be better served by a separate top level validator to handle value exclusion. But in the mean time, the current syntax should at least work properly.  ;)

dblock:
Better, thx.

PR #1623, opened 2017-05-02
Don't require multi_json and multi_xml.

dblock:
I think we need to add explicit configuration to this and tests that use different json and xml libraries.


dblock:
@namusyaka How do we feel about this?

lib/grape/util/json.rb
```diff
@@ -0,0 +1,13 @@
...
+end
+
+module Grape
+  if Object.const_defined? :MultiJson
+    Json = ::MultiJson
+  else
+    Json = ::JSON
+    Json::ParseError = Json::ParserError
+  end
+end
```
namusyaka: I'd like to declare `require 'json'` explicitly.

lib/grape/util/xml.rb
```diff
@@ -0,0 +1,13 @@
...
+rescue LoadError
+end
+
+module Grape
+  if Object.const_defined? :MultiXml
+    Xml = ::MultiXml
+  else
+    Xml = ::ActiveSupport::XmlMini
+    Xml::ParseError = StandardError
+  end
```
namusyaka: Likewise, I'd like to declare `require 'active_support/xml_mini'` explicitly. 
At the moment we don't declare `require 'active_support'` itself.

namusyaka:
Sorry for the delay. I strongly agree with the purpose of this pull request.
Thank you!

lib/grape/util/json.rb
dblock: What do you mean exactly? That the user has to `require 'json'`?

lib/grape/util/xml.rb
dblock: Or are you saying that I should add `require 'active_support/xml_mini'` here after `else`?

dblock:
@namusyaka I made the change that requires you to require multi_* manually. If you care to take a look? I'm going to merge soon.

lib/grape/util/xml.rb
namusyaka: Yes. Do I understand things correctly?

dblock: Ok, take a look at the PR as now, any issues? Thank you.

namusyaka:
👍 👍 👍

namusyaka:
Thank you for your great work! 🎉

Issue #1602, opened 2017-03-23
Parameter dependecy using `given` does not work with nested arrays and custom block 


nicooga:
## What's happening
Given I declared some rules to be dependent on other param using `given parameter: -> v { v }` and I did it inside a nested array param, it raises error.

## Expected behavior
It should apply the rules.

## Extended Explanation
Given this params:

~~~ruby
params do
  optional :comments, type: Array do
    optional :_destroy, coerce: Boolean
    given _destroy: -> v { !v } do
      requires :name
    end
  end
end

put('/asdf') do
  declared(params)
end
~~~
When I call the controller with this params:

~~~json
{"comments": [{ "_destroy": true}]}
~~~
It fails with:

~~~ruby
TypeError (no implicit conversion of Symbol into Integer):
  
activesupport (5.0.1) lib/active_support/core_ext/object/try.rb:17:in `[]'
activesupport (5.0.1) lib/active_support/core_ext/object/try.rb:17:in `public_send'
activesupport (5.0.1) lib/active_support/core_ext/object/try.rb:17:in `try!'
activesupport (5.0.1) lib/active_support/core_ext/object/try.rb:6:in `try'
grape (0.19.1) lib/grape/validations/params_scope.rb:49:in `block in should_validate?'
~~~

[This is the source from grape where it fails](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/params_scope.rb#L47-L49):

~~~ruby
          if dependency.is_a?(hash)
            dependency_key = dependency.keys[0]
            proc = dependency.values[0]
            require 'pry'
            binding.pry
            return false unless proc.call(params(parameters).try(:[], dependency_key))
~~~

I have put a `binding.pry` here on the source to see what's happening.

~~~ruby
[1] pry(#<Grape::Validations::ParamsScope>)> params(parameters)
=> [{"_destroy"=>true}]
[3] pry(#<Grape::Validations::ParamsScope>)> dependency_key
=> :_destroy
~~~
The validations are being run over the entire array instead of over each element of it. It basically runs `[{"_destroy"=>true}][:_destroy]` which raises the `TypeError (no implicit conversion of Symbol into Integer)` error.

dblock:
Closing via https://github.com/ruby-grape/grape/pull/1625

Issue #1600, opened 2017-03-21
Dependent parameters with specific values don't support Arrays


squarebracket:
With params like this:

```
        optional :captions, type: Array do
          requires :output, type: String, value_set: ['cea608', 'cea708', 'dvbsub', 'teletext']
          given output: ->(val) { val == 'cea608' } do
            requires :language, type: String
            requires :input_pid, type: String
          end
          given output: ->(val) { val == 'dvbsub' } do
            requires :input_pid, type: String
          end
        end
```

grape has an exception. It's because of this code in `lib/grape/validations/params_scope.rb`:

```
        @dependent_on.each do |dependency|
          if dependency.is_a?(Hash)
            dependency_key = dependency.keys[0]
            proc = dependency.values[0]
            return false unless proc.call(params(parameters).try(:[], dependency_key))
          elsif params(parameters).try(:[], dependency).blank?
            return false
          end
        end if @dependent_on
```

Since parameters is an array here, and not a hash, it can't do `[]` since it wants an integer. This code needs to be amended to support arrays. From what I've been able to figure out on the code, the `try(:[])` method should be applied to every item in the array `params(parameters)`.

I'm looking at the spec tests in `spec/grape/validations/params_scope_spec.rb` to try to figure out what to copy/paste so that I can work on implementing the behaviour. I guess lines 476 -> 586, but without the error conditions?

rnubel:
I think you're right; the use case of `type: Array` seems to just be an oversight. This test reproduces your error:

```ruby
  context 'when validations are dependent on a parameter within an array param' do
    before do
      subject.params do
        requires :foos, type: Array  do
          optional :foo_type, :baz_type
          given :foo_type do
            requires :bar
          end
        end
      end
      subject.get('/test') { declared(params).to_json }
    end

    it 'applies the constraint within each value' do
      get '/test', foos: [ { foo_type: 'a', bar: 'b' }, { baz_type: 'c' } ]

      expect(last_response.status).to eq(200)
    end
  end
```

I'm taking a quick look into a fix while on the train; I'll update if I find anything.

rnubel:
This technically fixes it, but is a bit rough: https://github.com/ruby-grape/grape/compare/ruby-grape:master...rnubel:fix/given-within-array-param I will revisit tomorrow and make a PR, or you can run with it if you see a better path.

squarebracket:
It appears that I am running into two problems with the current master branch:
[...]

PR #1625, opened 2017-05-05
Bugfix: Correctly handle `given` in Array params

rnubel:
Addresses #1600 and #1602.

Array parameters are handled as a parameter that opens a scope with `type: Array`; `given` opens up a new scope, but was always setting the type to Hash. This patch fixes that, as well as correctly labeling the error messages associated with array fields.

### DIFF
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -47,17 +47,17 @@ module Grape
       end
 
       def meets_dependency?(params)
-        if @dependent_on
-          params = params.with_indifferent_access
+        return true unless @dependent_on
 
-          @dependent_on.each do |dependency|
-            if dependency.is_a?(Hash)
-              dependency_key = dependency.keys[0]
-              proc = dependency.values[0]
-              return false unless proc.call(params.try(:[], dependency_key))
-            elsif params.respond_to?(:key?) && params.try(:[], dependency).blank?
-              return false
-            end
+        params = params.with_indifferent_access
+
+        @dependent_on.each do |dependency|
+          if dependency.is_a?(Hash)
+            dependency_key = dependency.keys[0]
+            proc = dependency.values[0]
+            return false unless proc.call(params.try(:[], dependency_key))
+          elsif params.respond_to?(:key?) && params.try(:[], dependency).blank?
+            return false
           end
         end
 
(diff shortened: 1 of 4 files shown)


dblock:
Build needs to be fixed please.

timothysu:
@rnubel Including an extra test to check for each element being independently validated fails.  Given your test, removing `bar` from the second element exemplifies the issue, unless this is intended behavior.

```
it 'applies the constraint within each element' do
      post '/test',
           { foos: [{ foo_type: 'a' }, { baz_type: 'c' }] }.to_json,
           'CONTENT_TYPE' => 'application/json'

      expect(last_response.body).to eq('foos[0][bar] is missing')
    end
```
results in

```
       expected: "foos[0][bar] is missing"
            got: "foos[0][bar] is missing, foos[1][bar] is missing"
```

`foos[1]` should not have `bar` required because it does not contain a `foo_type`

dblock:
I'll wait for @rnubel to comment on the above before merging. Code looks OK to me otherwise.

rnubel:
Good catch, @timothysu. The problem was that an Array block defines a single scope, for which `should_validate?` is only called once; the AttributesIterator within applies the validations to each attribute. So, I've updated the PR to perform the dependency check within that iteration, but it needs some polish and further testing.

avellable:
Any update on this?

dblock:
Please contribute @avellable.

dblock:
Still needs CHANGELOG and stuff.

lib/grape/validations/params_scope.rb
```diff
@@ -42,37 +42,45 @@ def should_validate?(parameters)
...
-            return false
-          end
-        end if @dependent_on
-
         return true if parent.nil?
         parent.should_validate?(parameters)
       end
 
+      def meets_dependency?(params)
+        if @dependent_on
```
dblock: Make this a guard, `return unless ...`

avellable:
@dblock Done: https://github.com/ruby-grape/grape/pull/1645

rnubel:
@avellable appreciate the help. `dependency` is the correct spelling of the word, though, and your changelog format is a bit off, so I've pulled the changes into this PR instead and put your name on the changelog entry as well.

@timothysu would you be able to test this patch to verify it works in real-life usage? I can test it against contrived scenarios, but there's no substitute for a real app.

avellable:
@rnubel Thanks! Sorry I forgot to add your name in the CHANGELOG of my changes.

timothysu:
@rnubel It doesn't seem to solve real-life usage issues, during a quick test.  I'll double check and follow up with more actionable info when I get a chance.

dblock:
I am going to merge this since it fixes at least one problem (per spec), @timothysu add anything on top that's failing in a new issue/spec please.

Issue #1582, opened 2017-02-24
warnings when running tests


thogg4:
I imagine I am missing a config option or something, but I can't figure out why I am seeing these errors when running tests using `Rack::Test` against a grape app:

```
/ruby/gems/2.3.0/bundler/gems/grape-4004faf0bf46/lib/grape/validations/validators/values.rb:21: warning: instance variable @excepts not initialized
/ruby/gems/2.3.0/bundler/gems/grape-4004faf0bf46/lib/grape/validations/validators/values.rb:31: warning: instance variable @proc not initialized
/ruby/gems/2.3.0/bundler/gems/grape-4004faf0bf46/lib/grape/validations/validators/values.rb:21: warning: instance variable @excepts not initialized
/ruby/gems/2.3.0/bundler/gems/grape-4004faf0bf46/lib/grape/validations/validators/values.rb:21: warning: instance variable @excepts not initialized
/ruby/gems/2.3.0/bundler/gems/grape-4004faf0bf46/lib/grape/validations/validators/values.rb:31: warning: instance variable @proc not initialized
/ruby/gems/2.3.0/bundler/gems/grape-4004faf0bf46/lib/grape/dsl/inside_route.rb:237: warning: instance variable @file not initialized
/ruby/gems/2.3.0/bundler/gems/grape-4004faf0bf46/lib/grape/dsl/inside_route.rb:147: warning: instance variable @status not initialized
```

Anyone have any idea about how to get rid of these?

dblock:
These are actual things we should fix in Grape.

terry90:
Someone working on it ? What is the cause of all this ?

dblock:
Crickets, so no, nobody is working on these. And they are uninitialized scoped variables, we assume they are `nil`.

PR #1629, opened 2017-05-10
Silence warnings "instance var not initialized"

terry90:
Add `defined?` checks for instance variables.

I will need feedback to get rid of more warnings, I am not aware of every warning triggered yet.
If someone has warnings too, let me know and I'll update the PR.

Reference to #1582

### DIFF
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -145,7 +145,7 @@ module Grape
         when Integer
           @status = status
         when nil
-          return @status if @status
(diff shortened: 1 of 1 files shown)


dblock:
This looks like a hack to prevent warnings and introduces a dynamic lookup. The whole point of the warning is to actually initialize these in an initializer, I would take a PR that accomplishes that.

terry90:
Fair enough

thogg4:
@dblock maybe like this? #1632

dblock:
I'll close this one.

PR #1631, opened 2017-05-10
Dynamic declared class

thogg4:
#1610 causes the params to be passed to `declared` in the correct class, but once they get there that method only returns a `Hash`


dblock:
Makes sense, but needs a test/changelog/etc., please.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 #### Fixes
 
+* [#1631](https://github.com/ruby-grape/grape/pull/1631): Dynamic declared class - [@thogg4](https://github.com/thogg4).
```
dblock: Maybe something more useful?

spec/grape/endpoint_spec.rb
```diff
@@ -298,6 +298,65 @@ def app
...
+
+          requires :first
+          optional :second, default: 'second-default'
+        end
+        subject.get '/declared' do
+          d = declared(params, include_missing: true)
+          ''
+        end
+
+        puts d
```
nbulaj: I think there is no need in that piece of code, and the same for above

spec/grape/endpoint_spec.rb
```diff
@@ -298,6 +298,65 @@ def app
...
+        d = nil
+
+        subject.params do
+          build_with Grape::Extensions::Hash::ParamBuilder
+
+          requires :first
+          optional :second, default: 'second-default'
+        end
+        subject.get '/declared' do
+          d = declared(params, include_missing: true)
```
dblock: You should return the declared output and compare that, don't use a variable out of scope.

spec/grape/endpoint_spec.rb
```diff
@@ -298,6 +298,65 @@ def app
       end
     end
 
+    context 'when params are not built with default class' do
+      it 'should return an object that corresponds with the params class - hash with indifferent access' do
```
dblock: Instead of "should return", just say "returns"

spec/grape/endpoint_spec.rb
thogg4: @dblock what is returned by the api call is always a json string. What we are testing here is what is returned by `declared`. Is there a better place to put this to test just that?

dblock: Have the API return `d.class` as JSON or a string and compare the result.

thogg4: @dblock duh, im not sure why i didn't think of that. thanks 😄

spec/grape/endpoint_spec.rb
```diff
@@ -298,6 +298,56 @@ def app
       end
     end
 
+    context 'when params are not built with default class' do
+      it 'returns an object that corresponds with the params class - hash with indifferent access' do
+        subject.params do
+          build_with Grape::Extensions::ActiveSupport::HashWithIndifferentAccess::ParamBuilder
+
+          requires :first
```
dblock: The actual parameters are not material for the tests, remove them. Make this as short as possible.

dblock:
Looks like there's a legit failure with one of the builds, check it out until this build is green please.

dblock:
This looks good, just needs a rebase/squash please.

PR #1635, opened 2017-05-19
Instrument validators with ActiveSupport::Notifications

ktimothy:
It can be useful to measure time spent in validators, especially when you have a lot of them.
This allows to instrument execution of validators by adding `endpoint_run_validators.grape` event.


dblock:
Perfect, thanks.

PR #1645, opened 2017-06-09
Bugfix: Handle given nested in Array params correctly

avellable:
Reference: https://github.com/ruby-grape/grape/pull/1625


avellable:
Original PR https://github.com/ruby-grape/grape/issues/1625 was updated. Let me close this one.

PR #1646, opened 2017-06-12
Add ability to include an array of modules as helpers

pablonahuelgomez:
Changing helpers DSL to allow the inclusion of many modules.
This attempts to bring a better readability, since it seems to be
more intuitive to send a list of modules when the message in
question is called `helpers`.

Want to move from a syntax like:

```ruby
helpers MyHelper
helpers MyOtherHelper
```
To a syntax like:
```ruby
helpers MyHelper, MyOtherHelper
```


dblock:
Beautiful. I merged a bit fast it needed a README update. Can you please PR on top @pablonahuelgomez?

pablonahuelgomez:
Sure @dblock, will do! Thanks for merging this PR 👍

PR #1647, opened 2017-06-12
Update README according to new `helpers` macro interface

pablonahuelgomez:
@dblock here goes the update to the README file.


dblock:
Thanks

Issue #1592, opened 2017-03-10
validators and instance variables: weird behaviour


loicginoux:
Hi, I am posting this here in case someone already had this case. I think there is a bug
I am using:
- grape (0.16.2)
- ruby 1.9.3
- rails 3.2.22.1

With the following code

```
class AlertIdValidation < Grape::Validations::Base
  include NewApi::Alerti::V3::Helpers::Authentication

  def validate_param!(attr_name, params)
    if !current_user.alerts.where(id: params[attr_name]).exists?
      raise Api::Exceptions::AlertNotExist, "alert #{params[attr_name]} does not exist"
    end
  end
end

class NewApi::Alerti::V3::Endpoints::Alerts < Grape::API
  helpers do
    def authenticate!
      raise NewApi::Alerti::V3::Exceptions::InvalidOrExpiredToken unless current_user
    end

    def current_user(req_params = nil)
      return @current_user unless @current_user.nil?
      @current_user = get_user(req_params) #not detailed, it will get user from the request access token 
    end
  end

  before do
    authenticate!
  end

  params do
    requires :alert_id, type: Integer, alert_id_validation: true
  end
  route_param :alert_id do
    # GET /alerts/:alert_id/
    get "/", jbuilder: 'alerts/show' do
      @alert = current_user.alerts.where(id: params[:alert_id]).first
    end
  end
end
```

I have a request `GET /alerts/:alert_id`

I call it once for a user A having an alert with id 1
`curl -X GET "http://api.my_site.dev/v3/alerts/1?token=d1071245ed8dc94dfa336f629e1e23cf2671b21f82d5e41714c514401027"`

It brings back the alert 1 correctly

I call the same endpoind for a user B having an alert with id 2
`curl -X GET "http://api.my_site.dev/v3/alerts/2?token=6aa880cbc0413aaac411d260e158ec598102f6b3d0da6ce511570bcdd9e7"`

the AlertIdValidation raises an error wrongly, the token is fine and the user B owns the alert 2.

After investigating, I have noticed that inside the validator, when calling the method current_user:
`if !current_user.alerts.where(id: params[attr_name]).exists?`

the current_user method will have a reference `@current_user` which is pointing to user A. everywhere else where I call current_user in the second call, I have the correct `@current_user` pointing to user B.

also implementing without using instance variable fix the issue when doing the second call, so I guess there is something that does not play well between instance variable and validators

```
def current_user(req_params = nil)
[...]

PR #1649, opened 2017-06-17
Create a new validator instance for every validation done

anakinj:
Had some problems related to shared validators between requests. Probably related to #1592.

Also the way the validators are registered is preventing the creation of dynamic validators, so made some modification to that also.


anakinj:
Any idea what this ```Grape::Validations::CoerceValidator coerce converter does not build Virtus::Attribute multiple times``` test is verifying? Is it safe/ok to just increase the expected call count by one?

anakinj:
Investigated this a little and if we want per request instances of the validators and don't wan't to call this ```Virtus::Attribute.build``` on every request we would need to have some kind of cache for the thing that method is building for the types.

anakinj:
Did a quick benchmark comparison between current master and this branch. There was a really significant performance hit calling this ```Virtus::Attribute.build``` on every request.

Added this quick POC of the cache for the ```build_coercer``` method to get the tests passing. Maybe need to add some Mutex to ensure the thread safety of this?

dblock:
I don't think I am clear with what this actually fixes, but I understand that it ensures validators aren't shared. So, what were the problems in sharing validators? These shouldn't have state and therefore that should be OK, no?

What is the improvement in performance? Could you please post some of these benchmark numbers?

anakinj:
In my case i have a feature toggle to control the behaviour of the validator. I used this kind of lazy-loading instance variable to get the data available to all the methods in the validator.

```
def feature_toggle_state
  @feature_toggle_state ||= fetch_toggle_state
end
```

When sharing instances ```@feature_toggle_state``` was never loaded again after the first request. When a validator is a class I was expecting the instance to be unique, of course now when I know how the validators behave I would know not to do it like this. 

I think @loicginoux is having the same problem in #1592 

Related to performance, this change of having a unique instance per request was slowing down the coercion related validators that were using the ```Virtus::Attributes.build``` method. So this caching of the results from that method does not make current versions any faster, only restores the performance to the state it's currently in.

lib/grape/validations/types/build_coercer.rb
```diff
@@ -54,8 +58,22 @@ def self.build_coercer(type, method = nil)
 
         # Virtus will infer coercion and validation rules
         # for many common ruby types.
-        Virtus::Attribute.build(conversion_type, converter_options)
+        to_cache(type, method, Virtus::Attribute.build(conversion_type, converter_options))
       end
+
+      def self.from_cache(type, method)
```
dblock: Typically caching is written with a block:

```ruby
def cache(type, method, &block)
  if block_given?
    # write result to cache and yield
  else
    # return result from cache
  end
end
```

lib/grape/validations/types/build_coercer.rb
```diff
@@ -54,8 +58,22 @@ def self.build_coercer(type, method = nil)
...
+
+      def self.to_cache(type, method, value)
+        @virtus_cache[cache_key(type, method)] = value
+      end
+
+      def self.cache_key(type, method)
+        [type.to_s, method].flatten.join('_')
+      end
+
+      instance_variable_set(:@virtus_cache, {})
```
dblock: For consistency I would either call this `@cache` or rename the method to read/write cache. Also what's the scope of this variable? Maybe it needs to be something like `__cache` or some other more internal variation?

lib/grape/validations/types/build_coercer.rb
anakinj: Like this block approach, did something similar to the latest version. Also not sure if we need to lock the write access to the hash for thread-safety.

The non locking version of this method would be something like:
```
def self.cache_instance_no_lock(type, method, &_block)
  @__cache.fetch(cache_key(type, method)) do |key|
    @__cache[key] = yield
  end
end
```

benchmark/simple_with_type_coercer.rb
```diff
@@ -0,0 +1,22 @@
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', 'lib'))
```
anakinj: Don't know what's the policy with adding these benchmarks, this can be removed if seen obsolete.

The results between not cached and cached version were about these:
```
No cache                   1.457k (± 8.8%) i/s -      7.250k in   5.017311s
Caching enabled            3.440k (± 5.6%) i/s -     17.435k in   5.085603s
```

anakinj:
What do you think should be added to the README? Did not notice any notes about the old behaviour. Removed some threadsafety notes from the ```Grape::Validations::Base``` documentation.

Is this worth mentioning in the UPGRADING notes? if someone is relying on that the validators are only instantiated once per process.

dblock:
In README maybe document under custom validators that they can contain state?

I do think UPGRADING deserves a note.

benchmark/simple_with_type_coercer.rb
dblock: Seems worth it!

spec/grape/validations/instance_behaivour_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        optional :param_to_validate, instance_validator: true
+      end
+      get do
+        'noop'
+      end
+    end
+  end
+
+  it 'passes validation every time' do
+    get '/', param_to_validate: 'value'
```
dblock: Nitpick you can do `2.times do ...`

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 * [#1632](https://github.com/ruby-grape/grape/pull/1632): Silence warnings - [@thogg4](https://github.com/thogg4).
 * [#1615](https://github.com/ruby-grape/grape/pull/1615): Fix default and type validator when values is a Hash with no value attribute - [@jlfaber](https://github.com/jlfaber).
 * [#1625](https://github.com/ruby-grape/grape/pull/1625): Handle `given` correctly when nested in Array params - [@rnubel](https://github.com/rnubel), [@avellable](https://github.com/avellable).
+* [#1649](https://github.com/ruby-grape/grape/pull/1649): Stop sharing validator instances between requests - [@anakinj](https://github.com/anakinj).
```
dblock: Maybe "Don't share ..." instead of "stop sharing", sounds more English.

spec/grape/validations/instance_behaivour_spec.rb
anakinj: This was a good one. I added some more expectations to the test at the same time.

anakinj:
Changed the CHANGELOG and added something to the README about the validations being standalone instances.

dblock:
Merged, thanks for your excellent work @anakinj!

Issue #1551, opened 2017-01-05
Boolean type not working as expected


Crystark:
Hi,

I'm having trouble using the Boolean type with proper boolean values.

What I initially wanted to do:
```ruby
optional :active, type: Boolean, default: true
```

Always returns `active is invalid` should I send `{  "active": true }` or `{  "active": false }`

What i had to do is actually add `coerce_with` using the identity lambda...

```ruby
optional :active, type: Boolean, default: true, coerce_with: ->(v) { v }
```

Did I miss something here ?
I'm using version `0.18`

dblock:
`Boolean` is actually `Virtus::Boolean` if I am not mistaken. Either way try to write a spec for this, you obviously shouldn't need to coerce a boolean.

tuttiq:
I just experienced the same problem when using `type: Boolean`, any value is considered invalid.
My workaround was to use the default coercion method, changing `type` for `coerce` in the param declaration.

```ruby
optional :flag, coerce: Boolean
```

Then it works with any value that can be coerced into a `Boolean`, like `{flag: true}`, `{flag: 'true'}`, `{flag: 1}`, `{flag: '1'}`.

LeFnord:
try to use `Virtus::Attribute::Boolean` instead, see: https://github.com/ruby-grape/grape/issues/1115

dblock:
We alias `Boolean` to that in https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/coerce.rb#L3.

dblock:
Closing via more tests in https://github.com/ruby-grape/grape/pull/1650.

PR #1650, opened 2017-06-30
Add extra specs for Boolean type field

tiarly:
This is commit adds additional specs to make sure Boolean type
is indeed working as expected.

https://github.com/ruby-grape/grape/issues/1551


dblock:
Merging this, thanks.

PR #1658, opened 2017-07-13
#1656: Update specs after upgrading rack-test to 0.7.0

### LINKED ISSUE #1656: Upgrade to rack-test 0.7.0
Newer versions of rack-test pass empty arrays. Needs some spec work. See #1655.

ashkan18:
Rack test was [ignoring](https://github.com/rack-test/rack-test/issues/122) empty arrays in its output which was fixed [here](https://github.com/rack-test/rack-test/pull/125). With these changes specs now show proper validation which doesn't match old ones anymore.

Updated `rack-test` to `0.7.0` and updated specs to match error messages we now receive with more details about missing params.

Fixes #1656

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#1658](https://github.com/ruby-grape/grape/pull/1658): Update `rack-test` to `0.7.0` and fix failing specs - [@ashkan18](https://github.com/ashkan18).
 * [#1652](https://github.com/ruby-grape/grape/pull/1652): Add the original exception to the error_formatter the original exception - [@dcsg](https://github.com/dcsg).
 
 #### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 #### Features
 
 * Your contribution here.
-* [#1658](https://github.com/ruby-grape/grape/pull/1658): Update `rack-test` to `0.7.0` and fix failing specs - [@ashkan18](https://github.com/ashkan18).
+* [#1658](https://github.com/ruby-grape/grape/pull/1658): Update `rack-test` to `0.7.0` - [@ashkan18](https://github.com/ashkan18).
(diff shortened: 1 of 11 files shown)


dblock:
You have to re-generate appraisals, with `appraisal generate --travis ; rubocop -a gemfiles/*`.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#1658](https://github.com/ruby-grape/grape/pull/1658): Update `rack-test` to `0.7.0` and fix failing specs - [@ashkan18](https://github.com/ashkan18).
```
dblock: I would remove "and fix failing specs" cause that's not very interesting or useful.

ashkan18:
Looks like Rails 5 doesn't like this upgrade:
```
In Gemfile:
    rack-test (~> 0.7.0)
    rails (= 5.0.0) was resolved to 5.0.0, which depends on
      actionpack (= 5.0.0) was resolved to 5.0.0, which depends on
        rack-test (~> 0.6.3)
```
Will look into it later today.

dblock:
I would imagine there's a newer Rails 5 that is ok with a later rack-test.

dblock:
Bump @ashkan18

ashkan18:
Sorry for delay on this, I actually looked at different Rails V5 versions, `rack-test` is [used](https://github.com/rails/rails/blob/master/actionpack/actionpack.gemspec#L30) by `actionpack` and it seems all of the versions are set to use 
```
s.add_dependency "rack-test", "~> 0.6.3"
```
Not sure what the best way to proceed on this one now. I'm happy to close this PR till Rails upgrade.

dblock:
I see. I think you should either open an issue in Rails or PR the upgrade in Rails and link it from here.

laertispappas:
Shouldn't Rails_3.gemfile have `rack-test v0.6.3` instead? Why should all Rails version have an updated version of rack test?

dblock:
@laertispappas If you can make that work, great. I think because the behavior of the spec changes it's not that easy.

dnesteryuk:
This one might be closed, Grape already uses Rack-test 1.1.0

Issue #1659, opened 2017-07-13
regression: parameters bound by `given` always applied/checked when nested


jnardone:
This is a regression from 0.19.2 -> 1.0.0

Consider a declaration like:

```
     subject.params do
        requires :a, type: String, allow_blank: false, values: %w(x y)
        given a: ->(val) { val == 'x' } do
          requires :inner1, type: Hash, allow_blank: false do
            requires :foo, type: Integer, allow_blank: false
          end
        end
        given a: ->(val) { val == 'y' } do
          requires :inner2, type: Hash, allow_blank: false do
            requires :bar, type: Integer, allow_blank: false
          end
        end
      end
```

The parameters for "inner1" / "inner2" should only be applied when the `given` block matches, but in the nested case (like above) the parameter validations are always applied even if they do not match the `given` criteria.  

For example, doing a request where `a` is `x`, you will get a 400 with `inner2[bar] is missing, inner2[bar] is empty`

Seems like this must have broken with #1625 @rnubel  @avellable ?

rnubel:
Thanks for the great write-up! #1661 should resolve this, and has some explanation as to why it regressed.

dblock:
Closed via https://github.com/ruby-grape/grape/pull/1661

PR #1660, opened 2017-07-13
Repro issue #1659

jnardone:


dblock:
Looks legit.

jnardone:
Closing due to fix in #1661, though @dblock how soon are you thinking about a 1.0.1 tag?  We'd like to move to 1.0.x but can't until next tag...

dblock:
Opened https://github.com/ruby-grape/grape/issues/1664 for next release, we don't make any promises, but any maintainer can cut one any time.

PR #1661, opened 2017-07-13
[Bugfix] Handle deeply-nested dependent params

rnubel:
Addresses https://github.com/ruby-grape/grape/issues/1659. The core issue was that param validations are a flat list, and the tree structure was only enforced by the recursion in `should_validate?`. When I moved the dependency check into the actual validator, the structure was no longer respected past one level deep. To resolve, I've added a similar recursive check into `ParamsScope#meets_dependency?`. I also added a few more test cases onto @jnardone's repro to test out the behavior works with arrays and nested arrays.


dblock:
Thanks for this!

PR #1665, opened 2017-07-23
Make helpers available in subclasses

### LINKED ISSUE #1617: Helpers are not being inherited from parent classes
Suppose I have a `Base` class with some basic stuff I want to include in every endpoint:
```
module Example
  module V1
    class Base < Grape::API
      helpers do
        params :user_params do
          requires :foo, type: Integer
        end
      end
    end
  end
end
```

I have a `Root` class which inherits from `Base`, where I mount my `Users` endpoint:
```
module Example
  module V1
    class Root < Base
      format :json
      mount Users::Root
    end
  end
end
```

The `Users` endpoint also inherits from `Base`:
```
module Example
  module V1
    module Users
      class Root < Base
        resource :users do
          mount Index
          mount Show
        end
      end
    end
  end
end
```

And here's the `Show`:
```
module Example
  module V1
    module Users
      class Show < Base
        desc 'Shows one user'
        params do
          requires :id, type: Integer
          use :user_params # <= This is the line that blows up
        end
        get ':id' do
          User.find(params[:id])
        end
      end
    end
  end
end
```

I would expect that the `use` statement in the `Show` endpoint would work because I expect that my `:user_params` helper is available, having been declared on `Base`. However, it does not. In fact the whole API fails to load because of that statement.

I am not sure if this is a bug, a feature, or user error, but I don't understand what's going on here. My only working solution is to put those helpers in a module that extends `Grape::API::Helpers` and include it on every single endpoint which needs that helper. In small projects this is trivial (and I probably wouldn't even bother with the helper module), but it would be very helpful to have this behavior in large projects.

Edit: I made a test project illustrating this here: https://github.com/flanger001/grape_example

pablonahuelgomez:
Updating settings code to define `:top_level_setting` with parent's
`:inheritable_setting` when available. This will emulate inheritance,
with its expected capabilities (sharing and overriding).

This fixes #1617

### DIFF
--- a/lib/grape/dsl/settings.rb
+++ b/lib/grape/dsl/settings.rb
@@ -2,7 +2,7 @@ require 'active_support/concern'
 
 module Grape
   module DSL
-    # Keeps track of settings (impemented as key-value pairs, grouped by
+    # Keeps track of settings (implemented as key-value pairs, grouped by
     # types), in two contexts: top-level settings which apply globally no
     # matter where they're defined, and inheritable settings which apply only
     # in the current scope and scopes nested under it.
@@ -13,7 +13,7 @@ module Grape
 
       # Fetch our top-level settings, which apply to all endpoints in the API.
       def top_level_setting
-        @top_level_setting ||= Grape::Util::InheritableSetting.new
+        @top_level_setting ||= obtain_top_level_setting
       end
 
       # Fetch our current inheritable settings, which are inherited by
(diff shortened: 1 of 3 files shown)


lib/grape/dsl/settings.rb
```diff
@@ -162,6 +162,18 @@ def within_namespace(&_block)
 
         result
       end
+
+      private
+
+      # Obtains the current class :inheritable_setting. If available, it returns the superclass's :inheritable_setting.
+      # Otherwise, a clean :inheritable_setting is returned.
+      def obtain_top_level_setting
```
dblock: I don't love the `obtain` name, maybe just `get`, another more common variation is `build`?

pablonahuelgomez: Agreed! Updated in https://github.com/ruby-grape/grape/pull/1665/commits/1a3ba3f0cb47a2a3f20bffa682c03718ab6c2694

dblock:
Merged, thank you.

Issue #1673, opened 2017-08-02
Param alias


glaucocustodio:
Hi folks, given the following api:

```ruby
resource :users do
  params do
    requires :email_address, allow_blank: false
    requires :password, allow_blank: false
  end
  post do
    User.create!(email: params[:email_address], password: params[:password])
  end
end
```

Do you think the following api would be nice? (alias `as:`)

```ruby
resource :users do
  params do
    requires :email, allow_blank: false, as: :email_address
    requires :password, allow_blank: false
  end
  post do
    User.create!(params)
  end
end
```

dblock:
I think that would be valuable.

glaucocustodio:
Do you think the first suggestion is better than the below one @dblock?

```ruby
resource :users do
  params do
    requires :email_address, allow_blank: false, as: :email
    requires :password, allow_blank: false
  end
  post do
    User.create!(params)
  end
end
```
I just implemented this one `requires :email_address, allow_blank: false, as: :email` but I'm not sure which is better.

dblock:
I think that if the model takes `:email` and the API takes `:email_address` you're trying to rename the API's `:email_address` into `:email`, so I think `as:` means "rename what's required into the value of `:as`, aka the second suggestion. Otherwise it feels backwards.

glaucocustodio:
Do you think `requires :email_address, as: :email` would lead to mistakes given that in Grape Entity we would do `expose :email, as: email_address` to expose the key as email_address.

dblock:
I mean it's what renaming is about ;)

tlabeeuw:
PS: this currently doesn't seem to work for nested params:
```ruby
group :foo, alias: :bar do
  requires :a
end
requires :baz, alias: :cat do
  requires :b
end
```
These above both fail.

dblock:
@tlabeeuw Can you please see if this is still broken on HEAD? Maybe PR a spec?

PR #1674, opened 2017-08-07
Add parameter alias

glaucocustodio:
Closes https://github.com/ruby-grape/grape/issues/1673.

Please, check if it's ok.

The methods `requires` and `optional` have a `if opts[:using]` (`grape/dsl/parameters.rb`) but I don't know how this condition can be true, I also didn't find any mention to `using` on readme.


glaucocustodio:
Does anybody have any clue why the tests (only on Rails 3) are failing with the message ` undefined method 'slice' for Hash`??

README.md
```diff
@@ -1094,6 +1095,23 @@ params do
 end
 ```
 
+### Alias
+
+It's possible to set alias for parameters using `as` (it can be useful when refactoring an existing API):
```
dblock: You can set an alias for parameters using `as`, which can be useful when refactoring existing APIs.

README.md
```diff
@@ -1094,6 +1095,23 @@ params do
...
+    requires :email_address, as: :email
+    requires :password
+  end
+  post do
+    User.create!(declared(params)) # User takes email and password
+  end
+end
+```
+
+The value passed to `as` will be the key when calling `declared(params)` (calling only `params` will not set the alias).
```
dblock: Maybe ", but `params` will not use the alias."

lib/grape/validations/params_scope.rb
```diff
@@ -115,10 +115,15 @@ def required?
 
       # Adds a parameter declaration to our list of validations.
       # @param attrs [Array] (see Grape::DSL::Parameters#requires)
-      def push_declared_params(attrs)
+      def push_declared_params(attrs, param_alias = nil)
```
dblock: Instead of `param_alias` I would make it `opts` and a `Hash` of different options, starting with `as`, for future proofing.

glaucocustodio:
I am not sure if would be cool to use alias with `params`, because today it seems that `params` is set directly from rack (`@params = @request.params`) and we would have to create another method (or an option for `params`) to get the raw parameters.

glaucocustodio:
>  I feel like when I use as, params should just be the target structure and maybe there should be some other way to access the "raw" parameters. That seems more predictable, no?

When we use `coerce_with:` the value is coerced when calling `params` as well, so I took a look in the implementation and achieved the same for alias, now if you call `params` the value of `as` will override the original key. You can check the tests @dblock.

dblock:
So much better! Merging.

dblock:
Nice work @glaucocustodio.

Issue #1662, opened 2017-07-18
Sending a single element in an array using XML format fails with the request being invalid


ramkumar-kr:
Hi,

I have an API which accepts requests using XML. It accepts an array of products. However for a case where there is only one product in the array, validation errors occur.
Sample request -
`curl -X POST \
  http://localhost:3000/v1/products \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/xml' \
  -d '<?xml version="1.0" encoding="UTF-8" ?>
<admin>
	<product>
		<id>1</id>
		<price>100</price>
	</product>
</admin>'`

This is because when there is a single element in an array, the product will be parsed to a hash instead of an array of hash objects.
Expected - `{ admin: { products: [{ id: "1", price: "100" }] } }`
Actual - `{ admin: { products: { id: "1", price: "100" } } }`

the parameters are defined as 
```
params do
  requires :admin, type: Hash do
    requires :products, type: Array do
      requires :id, type: String
      requires :price, type: String
    end
  end
end
```
Currently to overcome this, I run the following `before_validation` block
```
before_validation do
  params[:admin][:products] = [params[:admin][:products]] if params[:admin][:products].is_a?(Hash)
end
```
I was wondering since grape knows the request type (Array or Hash), can it convert the hash to an array before validating (similar to the before_validation block).

Notes:
Using Grape version : 1.0.0
Sample code - [https://github.com/ramkumar-kr/grape-xml-error](https://github.com/ramkumar-kr/grape-xml-error)

dblock:
In your example you send a single product, it cannot guess that it's an array, but maybe you're right and it should infer and coerce it it because of what's in the `params` block. I say write a spec for this next and lets see if we can implement/fix this.

ramkumar-kr:
Hi,
Sorry for the late reply.
I have created a Spec and have sent a [pull request](https://github.com/ruby-grape/grape/pull/1675) for it.

Smeedy:
We also bumped into this one the other day. This is due to the XML parsing of the arrays using the default ActiveSupport::XmlMini. We noticed that Nokogiri handles the XML arrays correctly using some test XML on the irb.

So the less intrusive solution is using MultiXML and Nokogiri as the parser in the project and not rewrite the params in the _before_ hook.

So the use of an initializer like 
`require 'multi_xml'

# Force all the XML parsing to be handled by Nokogiri
MultiXml.parser = MultiXml::Parsers::Nokogiri
`
[...]

PR #1675, opened 2017-08-10
#1662 - Add a spec for sending single element to an array

ramkumar-kr:


dblock:
I tried to fix this at no avail. Also tried `requires :products, type: Array, coerce_with: ->(val) { Array(val) }` but it gets overwritten somehow. I would focus on that one first, because it sounds that it could be an acceptable work-around.

dblock:
You can work-around/hack it with a `before` block. This passes:

```ruby
  context 'array params with XML content type' do
    let(:single_element_request) do
      '<?xml version="1.0" encoding="UTF-8" ?>
       <admin>
         <products>
           <price>100</price>
         </products>
       </admin>'
    end

    let(:multiple_element_request) do
      '<?xml version="1.0" encoding="UTF-8" ?>
       <admin>
         <products>
           <price>100</price>
         </products>
         <products>
           <price>200</price>
         </products>
       </admin>'
    end

    before do
      subject.format :xml
      subject.content_type :xml, 'application/xml; charset=utf-8'
      subject.before do
        params[:admin][:products] = [params[:admin][:products]] unless params[:admin][:products].is_a?(Array)
      end
      subject.params do
        requires :admin, type: Hash do
          requires :products, type: Array do
            requires :price, type: String
          end
        end
      end
      subject.post do
        params[:admin][:products].map do |product|
          { price: product[:price] }
        end
      end
    end

    context 'with one element' do
      it 'returns a successful response' do
        post '/', single_element_request, 'CONTENT_TYPE' => 'application/xml'
        expect(MultiXml.parse(last_response.body)).to eq('objects' => [{ 'price' => '100' }])
      end
    end

    context 'with multiple elements' do
      it 'returns a successful response' do
        post '/', multiple_element_request, 'CONTENT_TYPE' => 'application/xml'
        expect(MultiXml.parse(last_response.body)).to eq('objects' => [{ 'price' => '100' }, { 'price' => '200' }])
      end
    end
  end
```

Issue #1642, opened 2017-06-02
declared transforms empty array to hash


karhatsu:
I am using 0.19.2 and I have the following endpoint:

```
        resource :foos do
          params do
            requires :foo, type: Hash do
              requires :name, type: String
              optional :arr, type: Array do
                requires :param1, type: String
                requires :param2, type: String
              end
            end
          end

          post do
            puts "original: #{params.to_hash['foo']}"
            puts "declared: #{declared(params).to_hash['foo']}"
            {}
          end
        end
```

When I call it with

```
    {
        foo: {
            name: 'test',
            arr: []
        }
    }
```

I get this output:

```
original: {"name"=>"test"}
declared: {"name"=>"test", "arr"=>{"param1"=>nil, "param2"=>nil}}
```

In other words, for some reason the empty `arr` array has become a hash containing `arr`'s required fields.

dblock:
How do you call it? Are you posting JSON with the correct content-type?

karhatsu:
Yes, it's application/json.

dblock:
Try to build a spec for this next.

karhatsu:
Unfortunately I'm not able to write a failing test for this. I tried both with master and v0.19.2.

hwaterke:
I am having the same issue after updating Grape to the latest version.

```ruby
params :instance_params do
  requires :name, type: String
  optional :locations, type: Array do
    requires :latitude, type: Float
    requires :longitude, type: Float
  end
end
```

When sending something without the `locations` key (as it is optional), the array is transformed into a Hash.

Sending
```json
{
  "name": "FooBar",
  "locations": []
}
```
Works fine

But sending:
```json
{
  "name": "FooBar"
}
```

Results in the following `declared(params)`:
```ruby
{
  "name" => "FooBar",
  "locations" => {
    "latitude" => nil,
    "longitude" => nil
  }
}
```

jeradphelps:
I experienced this issue when attempting to upgrade to 19.2 as well.  It looks like there's some spec detailing this functionality for hashes but not for arrays.

This looks to have changed here: https://github.com/ruby-grape/grape/pull/1575.  There's some brief discussion [in this ticket](https://github.com/ruby-grape/grape/issues/1574).  From the sounds of it, the expectation is that we'd keep the functionality of nested arrays being populated as an array with `declared`.

Here's a diff with an added spec that demos this bug. I can work on a PR if we confirm the expectation:

[...]

PR #1678, opened 2017-08-24
#1642 Added spec for optional array param

hwaterke:
This is related to issue #1642.

When a param of type array is declared as optional, if it is not provided it is converted by `declared()` to a hash instead of an empty array.


dblock:
Closing via https://github.com/ruby-grape/grape/pull/1710

Issue #1676, opened 2017-08-22
Params of incorrect type are still processed by values: condition


nicholasserra:
Apologies if this is a dupe or if it turns out i'm doing something wrong :)

Reporting an odd behavior with types and values conditions.

Example: optional param that has to be a positive integer
```
optional :page, type: Integer, values: ->(v) { v.positive? }, default: 1
```

If I pass a string to this param I would expect to get back a `page is invalid` error, as the type is Integer. If you omit the `values` condition, this works.

But in my example, it seems the `values` condition is processed before the type is checked, resulting in ```#<NoMethodError: undefined method `positive?' for \"testing\":String>``` exception.

Does the values condition have priority by design? Should any values conditions be robust enough to handle various types?

Thanks :)

dblock:
There's clearly some order, in this case you're always getting the "raw" value, not a coerced one, so I think the "right" way here would be to rewrite `-> (v) { v.is_a?(Numeric) && v.positive? }`, no?

I'll close this, but feel free to reopen if you feel strongly otherwise and describe what you think should happen here.

Issue #1670, opened 2017-07-31
Can't use a Proc with values


gtournie:
On my project, this code:
```
requires :number, type: Integer, values: ->(v) { v > 0 }
```
throws an `ArgumentError (wrong number of arguments (0 for 1))`
If I remove the `values` option, everything is ok.
Seems like the proc doesn't like the arg.

I use grape 1.0.0

dblock:
This should work, can you turn it into a spec in the grape repo, please?

dblock:
Looks related to @jlfaber's https://github.com/ruby-grape/grape/issues/1616

jlfaber:
This should work. It was both implemented and spec'd. I tried adding your exact param spec in the UT code under both 1.0.0 and HEAD (Ruby 2.4.1) and it works fine for me.  Can you pass on which ruby version you're using, and at least a portion of the stack trace you're seeing?

I do notice that the validator does not rescue RuntimeErrors raised within the lambda literal.  So, in your example, sending in a request that's missing the number param will raise NoMethodError: undefined method `>' for nil:NilClass.  We could probably put a rescue around the call in the validator code.  But you can also avoid that by carefully defining the lambda.

nicholasserra:
Had the same issue. Bumped from grape 0.19.2 to 1.0.0 and now it works. I was using the example in the README:

```
requires :page, type: Integer, values: ->(v) { v.even? && v < 25 }
```

dblock:
I'm going to close this.

nicholasserra:
This resurfaced for me somehow. Tried bumping to ruby 2.4.1 and still seeing it.

PR #1679, opened 2017-08-24
Rescue StandardError from explicit values validator procs

jlfaber:
Example:
```ruby
params do
  requires :number, type: Integer, values: ->(v) { v.even? && v < 25 }
end
```
Right now, sending a non-numeric string or a nil value for `number` will result in a stack trace because neither String nor NilClass have an `even?` method. One could address this with an additional type check in the proc, but that would result in a lot of extra code that doesn't really contribute to the overall clarity of the declaration.

Instead, we should rescue any StandardError raised from an arity-one values validator proc and treat it as a false result.

### DIFF
--- a/lib/grape/validations/validators/values.rb
+++ b/lib/grape/validations/validators/values.rb
@@ -32,7 +32,7 @@ module Grape
           unless check_excepts(param_array)
 
         raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: message(:values) \
-          unless check_values(param_array)
+          unless check_values(param_array, attr_name)
 
         raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: message(:values) \
           if @proc && !param_array.all? { |param| @proc.call(param) }
@@ -40,12 +40,13 @@ module Grape
 
       private
 
-      def check_values(param_array)
+      def check_values(param_array, attr_name)
         values = @values.is_a?(Proc) && @values.arity.zero? ? @values.call : @values
         return true if values.nil?
         begin
           return param_array.all? { |param| values.call(param) } if values.is_a? Proc
-        rescue StandardError => _e
+        rescue StandardError => e
+          warn "Error '#{e}' raised while validating attribute '#{attr_name}'"
           return false
         end
         param_array.all? { |param| values.include?(param) }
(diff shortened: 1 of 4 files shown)


jlfaber:
This issue came up in both #1670 and #1676 so it's likely to keep biting people.

lib/grape/validations/validators/values.rb
```diff
@@ -43,7 +43,11 @@ def validate_param!(attr_name, params)
       def check_values(param_array)
         values = @values.is_a?(Proc) && @values.arity.zero? ? @values.call : @values
         return true if values.nil?
-        return param_array.all? { |param| values.call(param) } if values.is_a? Proc
+        begin
+          return param_array.all? { |param| values.call(param) } if values.is_a? Proc
+        rescue => _e
```
dblock: Maybe `rescue StandardError => e`, even if it's implicit to make it clear?

jlfaber:
@dblock - Fair point. Yeah, it definitely hides the details of the exception. How would you feel about doing this inside the rescue?
```ruby
warn "Error '#{e}' raised while validating attribute '#{attr_name}'"
```

lib/grape/validations/validators/values.rb
```diff
@@ -43,7 +43,11 @@ def validate_param!(attr_name, params)
       def check_values(param_array)
         values = @values.is_a?(Proc) && @values.arity.zero? ? @values.call : @values
```
dblock: Why aren't we also rescuing this `@values.call` here in that case? (and would need a spec)

jlfaber: I think that an error raised from an arity-zero proc is a different and more serious condition.  The output there is expected to be an array of valid values (or invalid values in the case of except_values) and a raised error means that the validator is completely broken. Yes, we _could_ return an empty array and warn in that case, but that doesn't feel like the right thing to do. In the values case it wouldn't ever accept anything; and, worse, in the except_values case it would always accept everything.

The primary reason I want to rescue in the arity-one case is because of the likelihood that unexpected input values (non-numeric string or nil in the example above) will raise errors in simple validation code that doesn't do thorough type-safety checks.  But in the arity-zero case, there is no input value so that's not a consideration.

jlfaber:
Added a warning for the arity-one case only.  Let me know if you feel strongly about rescuing in the arity-zero case.  Otherwise, this should be ready to merge.

dblock:
Merged. Thank you.

Issue #1685, opened 2017-09-28
JSON support is still not first class


timothysu:
First class JSON support is confusing, as some valid JSON is not supported due to the current implementation expecting the parsed result to either be a `Hash` or `Array[Hash]`, while the output of `JSON.parse` could be `String`, `Integer`, or mixed type arrays.

The initial thought would be to have JSON targeting all valid JSON, while being able to take modifiers to limit the possible valid outputs, like:
```
requires :foo, type: JSON # this could result in  `Hash`, `String`, `Integer`, or `Array`, accepting any of the inputs detailed in the two lines below
requires :bar, type: JSON[Hash] # this limits the output to `Hash`, so  "{\"a\":\"hash\"} is valid while "1",  "\"foo\"", and "[1, 2]" would be invalid inputs
requires :baz, type: JSON[Hash, String] # this would accept "\"foo\"" and "{\"a\":\"hash\"} as valid, while "1" and "[1, 2]" are invalid
```

When the API is given a type that is already what the JSON is parsed to (currently that would be `Hash` or `Array[Hash]`, it still attempts to coerce it despite being already valid.  For example, `"{\"a\":\"hash\"}` is currently valid while `{ a: 'hash' }` is not despite the desired parsed representation being the same.  A PR modifying that logic will be linked soon.

PR #1686, opened 2017-09-28
Avoid coercion of a value if it is valid

timothysu:
This PR aims to simplify the logic (and expensiveness) of coercion by not coercing the param if it is already a valid value unless a custom coercer is provided, which resolves issues we're run into around JSON until further work can be done to improve the experience.  Relates to #1685

### DIFF
--- a/lib/grape/validations/types/custom_type_coercer.rb
+++ b/lib/grape/validations/types/custom_type_coercer.rb
@@ -137,7 +137,7 @@ module Grape
             # passed, or if the type also implements a parse() method.
             type
           elsif type.is_a?(Enumerable)
-            ->(value) { value.all? { |item| item.is_a? type[0] } }
+            ->(value) { value.respond_to?(:all?) && value.all? { |item| item.is_a? type[0] } }
           else
             # By default, do a simple type check
             ->(value) { value.is_a? type }
--- a/lib/grape/validations/validators/coerce.rb
+++ b/lib/grape/validations/validators/coerce.rb
@@ -16,7 +16,7 @@ module Grape
 
       def validate_param!(attr_name, params)
         raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: message(:coerce) unless params.is_a? Hash
-        return if valid_type?(params[attr_name]) && !custom_type_coercer?
+        return if valid_type?(params[attr_name]) && !coerce_with?
         new_value = coerce_value(params[attr_name])
         raise Grape::Exceptions::Validation, params: [@scope.full_name(attr_name)], message: message(:coerce) unless valid_type?(new_value)
         params[attr_name] = new_value
@@ -65,8 +65,9 @@ module Grape
         @option[:type].is_a?(Hash) ? @option[:type][:value] : @option[:type]
       end
 
-      def custom_type_coercer?
-        converter.coercer.class.name.demodulize == 'CustomTypeCoercer'
+      def coerce_with?
+        # JSON types cannot have coerce_with
+        converter.coercer.respond_to?(:method) && !converter.is_a?(Grape::Validations::Types::Json)
       end
     end
   end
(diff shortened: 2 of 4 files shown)


timothysu:
Let me take a look at failing specs - didn't encounter them locally...

timothysu:
Ready for review

lib/grape/validations/types/custom_type_coercer.rb
```diff
@@ -137,7 +137,7 @@ def infer_type_check(type)
             # passed, or if the type also implements a parse() method.
             type
           elsif type.is_a?(Enumerable)
-            ->(value) { value.all? { |item| item.is_a? type[0] } }
+            ->(value) { value.respond_to?(:all?) && value.all? { |item| item.is_a? type[0] } }
```
dblock: `Enumerable` by definition has `all?`, how are we hitting a path where `respond_to?(:all?)` is needed?

lib/grape/validations/validators/coerce.rb
```diff
@@ -63,6 +64,11 @@ def coerce_value(val)
       def type
         @option[:type].is_a?(Hash) ? @option[:type][:value] : @option[:type]
       end
+
+      def requires_coercion?(value)
+        # JSON types do not require coercion if value is valid
+        !valid_type?(value) || converter.coercer.respond_to?(:method) && !converter.is_a?(Grape::Validations::Types::Json)
```
dblock: Can you explain `respond_to?(:method)` here?

Then, these seen unrelated, so to make it readable I would break it up.

```ruby
return true if valid_type?(value)
return true if ...
false
```

lib/grape/validations/types/custom_type_coercer.rb
timothysu: As I'm now checking for `valid_type?` before `value` is coerced, it means that `value` now could be anything.  The `Proc` that's defined here is used in `def success?` on line 96, which has the annotation of `# This method is called from somewhere within +Virtus::Attribute::value_coerced?+ in order to assert that the value has been coerced successfully.`  `valid_type?` directly calls `value_coerced?`, bringing this full circle.  I think the confusion is around `elsif type.is_a?(Enumerable)`, where `type` refers to the target/desired post-coercion type (eg. `Array`) rather than the type of `value`.

Without it, specs that go from `'1 2 3'` to `[1, 2, 3]` would fail as the type check on `'1 2 3'` would attempt to directly call `'1 2 3'.all? ...` which can't be done on a `String`.

dblock: Makes sense.

lib/grape/validations/validators/coerce.rb
timothysu: `.method` is defined if a custom `coerce_with` had been specified by the user, or the destination `type` has a defined coercion that everything needs to go through despite validity.  I added a test that shows why this is needed, if that's the behavior we want (which is the current behavior).

The conditionals are all directly related because of the following statements:
```
If the value is not valid, we require coercion.

Otherwise (if the value is valid), we require coercion it if a `method`  has been defined and it is not JSON.
```
which is written explicitly as:
```
def requires_coercion?(value)
  if !valid_type?(value)
    return true
  else
    if converter.coercer.respond_to?(:method) && !converter.is_a?(Grape::Validations::Types::Json)
      return true
    else
      return false
    end
  end
end
```
which could go to:
```
def requires_coercion?(value)
  return true if converter.coercer.respond_to?(:method) && !converter.is_a?(Grape::Validations::Types::Json)
  return true if !valid_type?(value)
  false
end
```
which condenses to the logic given.  Is the middle one still what is preferred?

dblock: I don't really care much, this is fine as is ;)

dblock:
Merged, thank you.

PR #1689, opened 2017-10-05
Fix incredibly minimal typo on namespace.rb documentation

yurighensev:
Fix incredibly minimal typo on `namespace.rb` documentation.


dblock:
Thanks!

Issue #1681, opened 2017-09-15
`default` value gets used when `given` block evaluates to `false`


astorije:
So, given this hypothetical scenario (using `grape` v1.0.1):

```rb
params do
  requires :foo, type: Symbol, values: [:a, :b]
  given foo: ->(val) { val == :a } do
    optional :bar, type: Symbol, default: :my_default
  end
end
put do
  binding.pry # Setting a breakpoint to highlight the following examples
end
```

I get the following when `foo` is `:a`, as expected:

```rb
> request.url
=> "http://example.com/api?foo=a"
> params
=> {"foo"=>:a, "bar"=>:my_default}
```

However, when `foo` is `:b`, this seems unexpected:

```rb
> request.url
=> "http://example.com/api?foo=b"
> params
=> {"foo"=>:b, "bar"=>:my_default}
```

Reason why I was surprised by this is that the parameters contract says that the default value of `:my_default` is only relevant when `foo` is `a`. Instead, I should be getting the following:

```rb
> params
=> {"foo"=>:b}
```

Is this a known case, or a bug?

rnubel:
Looks like a bug to me. Default-value population is probably not looking at the `given` constraints.

dblock:
@darren987469 another one? :)

darren987469:
No problem, I will try it. 😄

darren987469:
@dblock close this?

PR #1691, opened 2017-10-06
[Bug] Fix validations nested under a `given`

rnubel:
Addresses #1681.

Multi-param and default validators, all of which implemented their own `validate!` method, were not respecting the `given` dependency as expected. This change copies the check into all those validators.

The bug was likely introduced by #1625, which required that the dependency check be moved out of the scope itself and into the validators. Rolling back that change would require fundamentally re-architecting how array parameters are validated (possibly with a new kind of scope?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#1691](https://github.com/ruby-grape/grape/pull/1691): Fix validations nested under `given` - [@rnubel](https://github.com/rnubel).
 * Your contribution here.
 
 ### 1.0.1 (9/8/2017)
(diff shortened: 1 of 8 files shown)


dblock:
This needs a `rubocop -a ; rubocop --auto-gen-config` and ^ as well.

dblock:
Bump @rnubel ?

rnubel:
@dblock https://github.com/ruby-grape/grape/pull/1755 actually resolved this one; I just verified that locally by running the specs I added for this PR against master's code, and they pass.

dblock:
Great. Double check please that there's a spec for this exact scenario and PR that if not. Appreciate it.

PR #1700, opened 2017-11-01
Add failing specs for coercion of Arrays of custom types

javmorin:
Per discussion on the google group, here are two broken specs that show coercion failure for an Array of custom type (ie. not a built-in type).

First spec tests directly, second is via a Virtus::Attribute.

Google Groups discussion: https://groups.google.com/d/msg/ruby-grape/hYWBmgRgpps/itWH3UGoBgAJ

### DIFF
--- a/spec/grape/validations/validators/coerce_spec.rb
+++ b/spec/grape/validations/validators/coerce_spec.rb
@@ -225,6 +225,44 @@ describe Grape::Validations::CoerceValidator do
           expect(last_response.status).to eq(200)
           expect(last_response.body).to eq('1')
         end
+
+        it 'Array of URI::HTTP' do
+          module URI
+            class HTTP
+              def self.parse(value)
+                URI.parse(value)
+              end
+            end
+          end
+          subject.params do
+            requires :uri, coerce: Array[URI::HTTP]
+          end
+          subject.get '/uri_array' do
+            params[:uri][0].class
+          end
+          get 'uri_array', uri: ['http://www.google.com']
+          expect(last_response.status).to eq(200)
+          expect(last_response.body).to eq('URI::HTTP')
+        end
+
+        it 'Array of URI::HTTP (via Virtus::Attribute)' do
+          class URIAttribute < Virtus::Attribute
+            primitive URI::HTTP
+            def coerce(input)
+              URI.parse(input.to_s)
+            end
+          end
+
+          subject.params do
+            requires :uri, coerce: Array[URIAttribute]
+          end
+          subject.get '/uri_array' do
+            params[:uri][0].class
+          end
+          get 'uri_array', uri: ['http://www.google.com']
+          expect(last_response.status).to eq(200)
+          expect(last_response.body).to eq('URI::HTTP')
+        end
       end
 
       context 'Set' do


dblock:
There's actually a spec in here already that does this:

```ruby
requires :arry, coerce: Array[CoerceValidatorSpec::User]
```

The difference is that `User` includes `Virtus.model`. You can fix your specs by adding `include Virtus.model`, so:

```ruby
      class CustomParamType
        include Virtus.model
        attr_reader :value
        def initialize(value)
          @value = value
        end

        def self.parse(value)
          new(value)
        end
      end

      class CustomVirtusParamType < Virtus::Attribute
        primitive CustomParamType
        def coerce(input)
          CustomParamType.parse(input.to_s)
        end
      end
```

I am not sure what methods that brings in and whether there's a smaller include that can be used, but it works.

I'll close this, reopen/comment if that doesn't help.

javmorin:
The object in question would be a BSON::ObjectId, not a model in the normal sense. I didn't want to add a dependency on BSON for the spec so I just added a contrived class to demonstrate the issue.

As mentioned in the group conversation, the singular use case works, it's just the array use case that fails. Adding all of virtus.model to BSON::ObjectId seems like a bad idea just to support the Array use case.

dblock:
I agree. Try just inheriting from BSON::ObjectId and extending that?

javmorin:
For a better test, I have swapped URI::HTTP for my contrived type class and the same methods of coercion specification (direct and via Virtus::Attribute definition) fail.  I also tried adding the Virtus.model to a custom URI::HTTP inheriting class and while it fails differently (with a "<field name> is invalid" message), it still fails. Should I push these as a extension of the PR, or a replacement?

dblock:
Yes, push that up please, lets reopen the issue.

javmorin:
@dblock just another commit, or shall I reset the PR with just the URI example?

dblock:
Not important :)

javmorin:
I reset the PR because the other contrived examples were less clear about the issue as the URI::HTTP example.

dblock:
I think the easiest is to think of an Array of URI::HTTP as a thing (a type), so the following works:

```ruby
        it 'Array of URI::HTTP (via Virtus::Attribute)' do
          class URIArrayAttribute < Virtus::Attribute
            primitive Array[URI::HTTP]
            def coerce(input)
              input.map { |el| URI.parse(el.to_s) }
            end
          end

          subject.params do
            requires :uri, coerce: URIArrayAttribute
          end
          subject.get '/uri_array' do
            params[:uri][0].class
          end
          get 'uri_array', uri: ['http://www.google.com', 'http://www.artsy.net']
          expect(last_response.body).to eq('URI::HTTP')
        end
```

I do think that anything of type `Array[type that responds .parse]` should work. I tried to implement this in `build_coercer.rb`, but didn't succeed, maybe @dslh could give us a hand here?

javmorin:
Good to know there is that workaround, but I agree with you that supporting the same `Array[type that responds to .parse or is a Virtus::Attribute]` is the "right" thing. I wonder if the later case should be brought up with the virtus people.

dblock:
I think this is very much a Grape bug. I am just not smart enough to fix it.

dblock:
Closing via #1711.

PR #1710, opened 2017-11-19
Fix wrong transformation of empty Array in declared params

pablonahuelgomez:
Updating `PostBeforeFilter` module to handle empty given arrays as expected:
returning an empty array instead of an empty hash.

Specs have been taken from #1678 and from [this commit](https://github.com/ruby-grape/grape/commit/696db42c1e6ab6627b18e525a7011900d4740c5d).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 #### Fixes
 
+* [#1710](https://github.com/ruby-grape/grape/pull/1710): Fix wrong transformation of empty Array in declared params - [@pablonahuelgomez](https://github.com/pablonahuelgomez).
 * Your contribution here.
 
 ### 1.0.1 (9/8/2017)
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -47,9 +47,20 @@ module Grape
 
         def declared_hash(passed_params, options, declared_params)
           declared_params.each_with_object(passed_params.class.new) do |declared_param, memo|
-            # If it is not a Hash then it does not have children.
-            # Find its value or set it to nil.
-            if !declared_param.is_a?(Hash)
+            if declared_param.is_a?(Hash)
+              declared_param.each_pair do |declared_parent_param, declared_children_params|
+                next unless options[:include_missing] || passed_params.key?(declared_parent_param)
+
+                passed_children_params = passed_params[declared_parent_param] || passed_params.class.new
+                memo_key = optioned_param_key(declared_parent_param, options)
+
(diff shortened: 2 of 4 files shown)


lib/grape/dsl/inside_route.rb
```diff
@@ -26,30 +26,38 @@ def self.post_filter_methods(type)
...
+        def declared(passed_params, options = {}, declared_params_ = nil)
           options = options.reverse_merge(include_missing: true, include_parent_namespaces: true)
-          declared_params ||= optioned_declared_params(options)
-
-          if passed_params.is_a?(Array)
-            declared_array(passed_params, options, declared_params)
-          else
-            declared_hash(passed_params, options, declared_params)
-          end
+          @declared_params = declared_params_ || optioned_declared_params(options)
```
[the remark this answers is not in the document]
pablonahuelgomez: Agreed! Didn't see this from that perspective, will revert this approach 👍

pablonahuelgomez:
@dblock changes applied. Travis is red due to a worker issue:
```
oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 #### Fixes
 
+* [#1710](https://github.com/ruby-grape/grape/pull/1710): Make `declared` transform passed params to what is expected - [@pablonahuelgomez](https://github.com/pablonahuelgomez).
```
dblock: This is a bit too vague. The bug being fixed is about nested objects or something like that, care to make it a bit more specific?

lib/grape/dsl/inside_route.rb
```diff
@@ -60,17 +71,26 @@ def declared_hash(passed_params, options, declared_params)
...
-                next unless options[:include_missing] || passed_params.key?(declared_parent_param)
-
-                passed_children_params = passed_params[declared_parent_param] || passed_params.class.new
-                memo[optioned_param_key(declared_parent_param, options)] = declared(passed_children_params, options, declared_children_params)
-              end
             end
           end
         end
 
+        def handle_passed_param(declared_param, passed_children_params)
```
dblock: Probably want a `&_block` in the parameters since this yields.

pablonahuelgomez: Sure, updated.

CHANGELOG.md
pablonahuelgomez: Of course, tried to make it more specific.

dblock:
Merged

PR #1711, opened 2017-11-20
Coerce collections of parseable types

dslh:
Hi @dblock and @javmorin!

This implements support for "`type: Array[type that responds to .parse]`", as discussed in #1700. `CustomTypeCoercer` is leveraged, so the type may also implement `coerced?` or `parsed?` to implement custom type-checking (or also `call`, which I admit seems a bit weird in this instance).

Supporting `type: Array[Virtus::Attribute subclass]` should also be possible but I think it needs some different plumbing to make it work. From the discussion it wasn't clear to me that there's anyone who is specifically looking for it, or if this PR would make it redundant. Let me know if there's any interest and I can take a look at it.

Supports `Set` declarations as well as `Array`. `Hash` didn't make much sense to me in this context but if anyone has a usage example in mind I could take a look.

A couple more tests could be appropriate, I'll also get the readme and changelog filled in when I can, but now seemed like a good time to get get some feedback. Is there anything that could be changed or added?

### DIFF
--- a/lib/grape/validations/types/custom_type_collection_coercer.rb
+++ b/lib/grape/validations/types/custom_type_collection_coercer.rb
@@ -1,0 +1,71 @@
+module Grape
+  module Validations
+    module Types
+      # Instances of this class may be passed to
+      # +Virtus::Attribute.build+ as the +:coercer+
+      # option, to handle collections of types that
+      # provide their own parsing (and optionally,
+      # type-checking) functionality.
+      #
+      # See {CustomTypeCoercer} for details on types
+      # that will be supported by this by this coercer.
+      # This coercer works in the same way as +CustomTypeCoercer+
+      # except that it expects to receive an array of strings to
+      # coerce and will return an array (or optionally, a set)
+      # of coerced values.
+      #
+      # +CustomTypeCoercer+ is already capable of providing type
+      # checking for arrays where an independent coercion method
+      # is supplied. As such, +CustomTypeCollectionCoercer+ does
+      # not allow for such a method to be supplied independently
+      # of the type.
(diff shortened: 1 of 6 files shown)


lib/grape/validations/types/custom_type_collection_coercer.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        # This method is called from somewhere within
+        # +Virtus::Attribute::coerce+ in order to coerce
+        # the given value.
+        #
+        # @param value [Array<String>] an array of values to be coerced
+        # @return [Array,Set] the coerced result. May be an +Array+ or a
+        #   +Set+ depending on the setting given to the constructor
+        def call(value)
+          coerced = value.map { |item| super(item) }
```
dslh: I'm aware this will fail if `value` is a string but I wasn't sure of the best thing to do about it. Maybe `Array.wrap`?

javmorin: To allow for a single item to be passed in as such (without the array wrapping/declaration) is a convenience, but could be improper. I suppose it depends on how strict Grape wants to be.

dblock: I would deal with that scenario later with a proper spec.

dslh:
Readme and changelog are updated. It didn't seem like there was much to be said about the functionality added in this PR; once you have custom types, wrapping them in an array or a set is something that should "just work". But I did notice that the existence of the `parsed?` method was undocumented so I added a note about that and also a quick test.

dblock:
Perfect.

PR #1714, opened 2017-11-23
Removing references to `hashie-forbidden_attributes` from readme

alagos:
Hashie was [removed as dependency](https://github.com/ruby-grape/grape/commit/1bd1a21525e58768c2188a291488d0b3ba2832c5), so `hashie-forbidden_attributes` should not be installed anymore.


dblock:
Thanks!

PR #1716, opened 2017-12-09
upgrade rubocop version

namusyaka:
Fixes potential security vulnerability.
ref https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-8418


dblock:
You should upgrade to the latest rubocop, fix appraisals and what not, too. Then `rubocop -a ; rubocop --auto-gen-config`.

namusyaka:
Fixed.

dblock:
Thanks. The encoding header was removed everywhere, which is right, but a newline was added instead. Care to fix that? Sorry to be a pest :)

lib/grape/exceptions/incompatible_option_values.rb
```diff
@@ -1,4 +1,4 @@
-# encoding: utf-8
+
```
dblock: This newline was added.

namusyaka:
Oh.. sorry. Thanks for pointing this out.

Issue #1717, opened 2017-12-15
exactly_one_of is not functioning as expected and says missing


gencer:
**Definition:**
```
params do
   optional :folder, type: String, default: nil
   optional :tag, type: Array, default: nil
   exactly_one_of :folder, :tag
end
```
**header**:
```
Content-Type: application/json;
```
**request body**:
```
{
	"tag": [1,2,3]
}
```
or
```
{
	"folder": "/hello"
}
```
**got**:
```
{
	"error": "folder, tag are mutually exclusive"
}
```

rnubel:
Does removing `default: nil` for both params change the behavior?

gencer:
Definitely, did! Also, error message now shows the correct message. (not for mutually)

Instead of default value, I set `allow_blank` and/or `regexp` for better data integrity.

dblock:
Still a bug, right?

gencer:
Yes. Default value can be useful. Exactly one of means one of them will be empty **or have a default value I gave.**

dblock:
Would appreciate a pull request even if it's just failing specs.

gencer:
Before I send a PR, lets make sure my steps are correct.

After I review the code, I found that, In fact `:default` should be dropped. In `exactly_one_of` we should **not use default**. Because, if we do that then we cannot know which parameter given by user or by default. Due to this grape simply tell us that they are mutually exclusive, in other mean it is **limited**.

So, I no longer thinks this is a bug. This is an expected behavior.

I think clearly stating this on documentation will be enough.

dblock:
Aha. If both values have a default it already violates `exactly_one_of`, but maybe if one of them has a default and you don't pass anything, then it should work?

gencer:
Yes. To be clear;

Definition:
```
params do
   optional :folder, type: String, default: nil
   optional :tag, type: Array, default: nil
   exactly_one_of :folder, :tag
end
```

Request:
```
{
	"folder": "/hello"
}
```

It will fail. Why? Because I only give `:folder` but I also give `default: nil` for `:tag` in definition too. `:tag` should **NOT exists** in params as a key. It should not have a value like nil or empty string. It should not exists in the first place.

I believe current behavior is the more efficient and correct behavior. We just need to tell that "do not assign even `nil` values" to any `exactly_one_of` definitions.

At first I thought giving a default value should be okay. But we should not. Exactly one parameter should be exists in params bag. So we can check which key is exists not by nil or emptyness.

[...]

PR #1720, opened 2017-12-18
added note for avoid using :default on exactly_one_of

gencer:


dblock:
This is related to #1717. I think this note makes it even more confusing, what should happen is that if you specify a default for more than one parameter (or any parameter) in a mutually exclusive relationship you should get an error on mounting the API.

gencer:
But, I am sure people will fall into this if they simply forget :default in their optional params when they are using exactly_one_of. To prevent this, and let them know that those params should not exists even as a nil, we should add a note/warning.

Still, It is your decision of course :)

gencer:
And you think the right thing. I agree on that.

I amend and squash changes (removed changelog but change the readme)

dblock:
Thanks, merged.

Issue #1721, opened 2017-12-20
Defining a catch-all after multiple versions of an endpoint hide endpoints after the first definition


zherr:
Given the following test:
```ruby
shared_examples_for 'versioning' do
  context 'with different versions for the same endpoint' do
    context 'catch-all' do
      before do
        subject.version 'v1', macro_options
        subject.get 'version' do
          'v1'
        end

        subject.format :txt
        subject.version 'v2', macro_options
        subject.get 'version' do
          'v2'
        end

        subject.route :any, '*path' do
          error!('Not Found', 404)
        end
      end

      it 'finds v1 of the endpoint' do
        versioned_get '/version', 'v1', macro_options
        expect(last_response.status).to eq(200)
        expect(last_response.body).to eq('v1')
      end

      it 'finds v2 of the endpoint' do 
        versioned_get '/version', 'v2', macro_options
        expect(last_response.status).to eq(200) # >>>>> This fails! Results in 404 <<<<<
        expect(last_response.body).to eq('v2')
      end
    end
  end
end
```
For even more context, see: https://github.com/ruby-grape/grape/compare/master...zherr:multiple-endpoint-catchall

Before digging deeper into a fix I would like to gather feedback from contributors on whether this appears to be a bug, or if this behavior intentional.

Happy holidays!

dblock:
It certainly smells like doing `version v1` then `version v2` to the same api class overrides the previous definition, so split them up and `mount` them and that should work better.

[This](http://code.dblock.org/2013/07/19/evolving-apis-using-grape-api-versioning.html) might be helpful.

zherr:
Our current implementation splits up the versions by module like so:
```ruby
module MyApp
  class Base < Grape::API
      mount V1::Base
      mount V2::Base
      route :any, '*path' do
        error!('Not Found', 404)
      end
  end
end
```
Given this implementation, when a request specifies `Accept-Version=v2` in the header, for example, the `V2` endpoint is not found. When the version is left unspecified, the `V1` endpoint is found just fine.

Unfortunately, if we switch the mount order here as the blog post suggests, we still run into the same issue if the catch-all is defined; the `V1` is not found.

Removing the catch-all solves part of the issue, and allow requests to find the `V1` endpoint when not specifying the version, and allows requests that do specify `'v2'` to find the new version. But this leaves us unable to capture traffic that would typically 404 and let us render a reasonable message.

[...]

PR #1722, opened 2017-12-20
Spec and fix for #1721.

dblock:
#1721

With the catch all route fails to find v2.

@namusyaka I tried to fix this but failed, maybe you'll have a moment to help?


zherr:
@dblock Take a look at https://github.com/dblock/grape/pull/2 when you get a chance.

dblock:
Combine my specs and your code @zherr and PR the whole thing, thanks for the fix!

dblock:
I just merged that PR @zherr lets see if this passes and it's good to go if so.

zherr:
Cheers! 🍻

Issue #1719, opened 2017-12-17
headers set before error! is missing on response


gencer:
Whenever I execute `error!()`, response does not contains my headers. In fact none of the headers like `X-Request-Id`. I presume this is due to calling a new `Request` instance and return that instead.

How do I make sure that my important headers are in every request that grape gave? Even in `Rails::API` I am able to hold headers on any error.

Of course, I could add my own Middleware. Is there any better or easy solution before middleware option?

dblock:
What are "your" headers? Anything you set with `header ...` should be there.

gencer:
I have few headers that sent to user on every request in every action. I do not want to add them manually on each and every `error!` call. I already defined them on:

```
before do
  header ...
  header ...
end
```

based on request type, user and others. These headers can be different on each request by user or type. So, I would expect grape to respect my `before`.

Problem is, before gets called but all headers set in before are gone. If this is an expected behavior then I will stick with middlewares.

dblock:
Does adding a header inside the API work?

```ruby
get do
   header ...
   errror!
end
```

If it does, I think `before` and `after` should work the same way and it would be a bug.

Care to write some specs?

gencer:
**Note**: If you want I can create an another branch with test in `spec/` folder.

**Expected**: Both requests should pass.
**Actual**: Second test will fail.

**error_header_spec.rb**:
```
require 'spec_helper'

describe Grape::API do
  let(:options) do
	default_options.merge(request_headers)
  end
 
 let(:error_header) do
    Class.new(Grape::API) do
      version :v1, using: :accept_version_header, cascade: true
	  before do
	    header 'X-Grape-Before-Header', 'yeah'
	  end
      get '/success' do
		header 'X-Grape-Returns-Error', 'yeah'
	  end
      get '/error' do
		header 'X-Grape-Returns-Error', 'yeah'
		error!({ success: false })
	  end
    end
  end

  subject do
    ErrorHeader = error_header
    Class.new(Grape::API) do
      format :json
      mount ErrorHeader => '/'
    end
  end

  def app
    subject
  end

  it 'returns expected headers on success' do
    get '/success'
    expect(last_response.headers['X-Grape-Returns-Error']).to eq('yeah')
    expect(last_response.headers['X-Grape-Before-Header']).to eq('yeah')
  end
  
  it 'returns expected headers on error' do
    get '/error'
    expect(last_response.headers['X-Grape-Before-Header']).to eq('yeah')
[...]

PR #1723, opened 2017-12-21
added spec for expected headers when error raised

gencer:
When we set headers using `before`, `after` or just before `error!` raised, we expect them to transferred with error response together.

* First spec will be `success` due to there is no error.
* Second test will fail at the moment because when an error raised, all headers will be ignored. They shouldn't be.

See: #1719


gencer:
I've amended and squashed due to incorrect file format according to grape's rubocop rules. I've fixed it by this merge.

dblock:
You should break up the specs: are headers not added from before *and* after in the `error!` case? If so we should call this expected behavior and document it. Then we can discuss whether we want to change it.

gencer:
I will break up the specs asap.

For the question; `after`, `before` or above `error!`. On three cases, headers are stripped. The specs you see will fail on every one of them.

I will back with more proper specs as soon as possible.

dblock:
I think we shouldn't change this behavior to avoid regressions for all existing users. PR documentation that says that the headers will not be included on error and specs that expect the opposite behavior of the ones here.

Lets find a way to create the opposite behavior. The `error!` call takes headers today. Any other suggestions?

gencer:
Let's say we did not changed this behavior. As you mentioned we have `error!` with header support.

But fact is, for specific reasons, this could be enough. But I need to show some headers to clients even if they got `error!`.

One thing that i can suggest is making error-specific before/after statements like;

```
before do
   # usual before...
end

after do
   # usual after...
end

on_error do
    # execute and assign headers here on every error
end
# or...
error do
    # execute and assign headers here on every error (same as on_error just shorter name)
end

```

dblock:
I like `error do` that would be an alias and possibly supersede `rescue_from`.

gencer:
Definitely. This will allow us to specify headers, actions and anything else on every error, rather then define them on each `error!` call.

PR #1724, opened 2017-12-27
Optional nested array validation

### LINKED ISSUE #1636: Optional nested array not validated
Branch: `master`
Works under version : `0.13.0`

Hard to explain so here are 2 specs that should pass but the last one is failing :

```ruby
context 'nesting array validation' do
    before :each do
      subject.params do
        requires :my_array, type: Array do
          requires :a_integer, type: Integer, allow_blank: true
          optional :nested_array, type: Array, allow_blank: true, default: [] do
            requires :a_integer, type: Integer, allow_blank: true
          end
        end
      end
      subject.get('/nested_optional_array') { declared(params).to_json }
      get '/nested_optional_array', params
    end

    context 'when my_array contains only a hash with the optional nested_array' do
      let(:params){
        {
          my_array: [
            {
              an_integer: 1,
              nested_array: [
                {
                  an_integer: :a
                }
              ]
            }
          ]
        }
      }

      it 'should fail' do
        # success
        expect(last_response).to be_bad_request
      end
    end

    context 'when my_array contains hashes with the optional nested_array and one without' do
      let(:params){
        {
          my_array: [
            {
              an_integer: 1,
              nested_array: [
                {
                  an_integer: :a
                }
              ]
            },
            {
              an_integer: 2,
              nested_array: [
                {
                  an_integer: :b
                }
              ]
            },
            {
              an_integer: 1
            }
          ]
        }
      }

      it 'should fail' do
        # failure
        expect(last_response).to be_bad_request
      end
    end
  end
```

The `nested_array` is optional but when there's at least one element in the parent array that doesn't have it, all other elements with a `nested_array` aren't validated

I'll try to figure it out.

Thanks

ericproulx:
Fix #1636 Merry 🎄


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [#1688](https://github.com/ruby-grape/grape/pull/1688): Removes yard docs - [@ramkumar-kr](https://github.com/ramkumar-kr).
 * [#1702](https://github.com/ruby-grape/grape/pull/1702): Added danger-toc, verify correct TOC in README - [@dblock](https://github.com/dblock).
 * [#1711](https://github.com/ruby-grape/grape/pull/1711): Automatically coerce arrays and sets of types that implement a `parse` method - [@dslh](https://github.com/dslh).
+* [#1724](https://github.com/ruby-grape/grape/pull/1724): Optional nested array validation - [@ericproulx](https://github.com/ericproulx).
```
dblock: It's a fix, no?

spec/grape/validations/validators/default_spec.rb
```diff
@@ -99,6 +114,28 @@ def app
...
 
+  it 'does not allows faulty optional arrays' do
+    params = { some_things:
+                 [
+                   { foo: 'one', options: [{ name: 'wat', value: 'nope' }] },
+                   { foo: 'two', options: [{ name: 'wat' }] },
+                   { foo: 'three' }
+                 ] }
+    get '/nested_optional_array', root: params
+    expect(last_response.status).to eq(400)
```
dblock: Test that the actual response contains the correct value being wrong. Maybe break this one up more, too.

CHANGELOG.md
ericproulx: It is, my bad

spec/grape/validations/validators/default_spec.rb
ericproulx: I don't understand

dblock: In this case the status can be 400 for all kinds of reasons. Check `last_response.body` that it says something about the parameter being incorrect.

ericproulx: got it 👍

dblock:
Merged, thank you.

PR #1726, opened 2018-01-02
Refactor #generate_api_method to improve performance

### LINKED ISSUE #1672: Grape is slow to initialize
While profiling the initialisation phase of our app I stumbled upon a sizeable chunk due to Grape.

screen shot 2017-08-02 at 13 48 23

It seems most of the time is spent within [`Grape::Endpoint#generate_api_method`](https://github.com/ruby-grape/grape/blob/12af441979edb83c8e52fb79a084379b70cc9813/lib/grape/endpoint.rb#L47), because `Array.sort!` uses `<=>` on its elements, thus internally calling `Symbol#to_s` since symbols really are just integers.

This seems to negate what could be an attempt at optimisation in `#generate_api_method`: `if instance_methods.include?(method_name.to_sym) || instance_methods.include?(method_name.to_s)`, though what triggers the `sort!` dance I don't know, as I tried to isolate it with the following script run from the affected Rails application:

```ruby
# bundle exec ruby foo.rb
require File.expand_path('../config/environment.rb', __FILE__)
require 'ruby-prof'
result = RubyProf.profile do
  1000.times do
    Grape::Endpoint.instance_methods.include?(:foo)
  end
end
printer = RubyProf::FlatPrinter.new(result)
File.open('prof.out', 'wb') { |f| printer.print(f, {}) }
```

and its output:

```
Measure Mode: wall_time
Thread ID: 70191707060580
Fiber ID: 70191783619080
Total: 0.161279
Sort by: self_time

 %self      total      self      wait     child     calls  name
 84.55      0.136     0.136     0.000     0.000     1000   Module#instance_methods
  5.13      0.008     0.008     0.000     0.000     1000   Array#include?
  2.42      0.004     0.004     0.000     0.000     1000   Module#extend_object
  2.12      0.150     0.003     0.000     0.147     1000   Class#instance_methods
  1.69      0.161     0.003     0.000     0.159        1   Integer#times
  1.29      0.002     0.002     0.000     0.000     1000   Kernel#instance_variable_set
  0.88      0.006     0.001     0.000     0.004     1000   Kernel#extend
  0.86      0.001     0.001     0.000     0.000     1000   UnboundMethod#bind
  0.77      0.138     0.001     0.000     0.136     1000   Method#call
  0.26      0.000     0.000     0.000     0.000     1000   Module#extended
  0.02      0.161     0.000     0.000     0.161        1   Global#[No method]

* indicates recursively called methods
```

jkowens:
This is an attempt to resolve #1672.

Benchmark script:

```
require 'grape'
require 'benchmark/ips'

Benchmark.ips do |ips|
  ips.report('class_load') do
    class API < Grape::API
      prefix :api
      version 'v1', using: :path
      get '/' do
        'hello'
      end
    end
  end
end
```

Before:

```
          class_load   590.000  i/100ms
          class_load      5.589k (±10.3%) i/s -     27.730k in   5.029428s
```

After:

```
          class_load   860.000  i/100ms
          class_load      8.761k (±13.6%) i/s -     43.000k in   5.035511s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#1722](https://github.com/ruby-grape/grape/pull/1722): Fix catch-all hiding multiple versions of an endpoint after the first definition - [@zherr](https://github.com/zherr).
 * [#1724](https://github.com/ruby-grape/grape/pull/1724): Optional nested array validation - [@ericproulx](https://github.com/ericproulx).
 * [#1725](https://github.com/ruby-grape/grape/pull/1725): Fix `rescue_from :all` documentation - [@Jelkster](https://github.com/Jelkster).
+* [#1726](https://github.com/ruby-grape/grape/pull/1726): Improved startup performance during API method generation - [@jkowens](https://github.com/jkowens).
 * Your contribution here.
 
 ### 1.0.1 (9/8/2017)
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -44,7 +44,7 @@ module Grape
       # @return [Proc]
       # @raise [NameError] an instance method with the same name already exists
       def generate_api_method(method_name, &block)
-        if instance_methods.include?(method_name.to_sym) || instance_methods.include?(method_name.to_s)
+        if method_defined?(method_name)
           raise NameError.new("method #{method_name.inspect} already exists and cannot be used as an unbound method name")
         end
 


dblock:
I think this code is an MRI 1.8 artifact, since https://github.com/ruby-grape/grape/commit/0e0a94f6de1c664b6e7f90066982ea1a024f3f51 was fixing some issue there and we haven't touched it for 5 years, so LGTM.

dblock:
I think this deserves a CHANGELOG entry. Maybe something like "Improved performance on ..."?

jkowens:
I've added that changelog entry, thanks!

dblock:
Merged, thank you.

PR #1729, opened 2018-01-17
Add validator 'declared_only' to reject undeclared params

seanhandley:
See https://github.com/ruby-grape/grape/pull/1606. This PR continues the work started by @jlfaber.


dblock:
Thanks for continuing this!

I still feel like a naked `declared_only` is strange here and I am not a fan of the name. How do we feel about `strict`? And bringing this into an argument of `params`?

```ruby
params(strict: true) do
  optional :beer
  optional :wine
  optional :juice
end
```

What's the behavior of nested blocks? Do they inherit this? Feels like they should.

dblock:
cc: @jlfaber

seanhandley:
Maybe something like this @dblock @jlfaber 👍 

```ruby
params(undeclared_params: :ignore) do
  optional :beer
  optional :wine
  optional :juice
end
```

```ruby
params(undeclared_params: :raise) do
  optional :beer
  optional :wine
  optional :juice
end
```

And tie that in with a global default somewhere?

dblock:
I like what you're suggesting @seanhandley. I would drop the `params` part since this is inside a `params` block.

```ruby
params(undeclared: :ignore) do
  optional :beer
  optional :wine
  optional :juice
end
```

dblock:
@seanhandley Don't want to finish this?

seanhandley:
@dblock Sadly not.

I'm not actually using Grape in any projects right now, and I have a lot of other things I need to focus on.

Issue #1735, opened 2018-01-30
Error args passed into I18n.t


mlzhuyi:
The following is an example of customed validator in README.
``` ruby
class Admin < Grape::Validations::Base
  def validate(request)
    # return if the param we are checking was not in request
    # @attrs is a list containing the attribute we are currently validating
    # in our sample case this method once will get called with
    # @attrs being [:admin_field] and once with @attrs being [:admin_false_field]
    return unless request.params.key?(@attrs.first)
    # check if admin flag is set to true
    return unless @option
    # check if user is admin or not
    # as an example get a token from request and check if it's admin or not
    fail Grape::Exceptions::Validation, params: @attrs, message: 'Can not set admin-only field.' unless request.headers['X-Access-Token'] == 'admin'
  end
end
```
It send `@attrs`, an array of symbols, as its params value.<br>The params value would be passed into `Grape::Exceptions::Base#translate_attributes` or `Grape::Exceptions::Base#translate_attribute` method.<br> As a result, Every value of the :default option for `I18n.t` is a symbol. Is a `to_s` missing here?

dblock:
Possibly. Try to turn this into a spec in a PR?

dnesteryuk:
I think this one can be closed, the PR was merged.

dblock:
👍

PR #1737, opened 2018-01-31
Ensure the value of default option must be a string when calling I18n.t

mlzhuyi:
https://github.com/ruby-grape/grape/issues/1735

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1737](https://github.com/ruby-grape/grape/pull/1737): Fix translating error when passing symbols as params in custom validations - [@mlzhuyi](https://github.com/mlzhuyi).
 
 ### 1.0.2 (1/10/2018)
 
--- a/lib/grape/exceptions/base.rb
+++ b/lib/grape/exceptions/base.rb
@@ -71,6 +71,8 @@ module Grape
       end
 
       def translate(key, **options)
+        options = options.dup
+        options[:default] &&= options[:default].to_s
         message = ::I18n.translate(key, **options)
         message.present? ? message : ::I18n.translate(key, locale: FALLBACK_LOCALE, **options)
       end
--- a/spec/grape/exceptions/validation_errors_spec.rb
+++ b/spec/grape/exceptions/validation_errors_spec.rb
@@ -43,6 +43,15 @@ describe Grape::Exceptions::ValidationErrors do
         expect(subject).to contain_exactly('id is missing', 'name is missing')
       end
     end
+
+    context 'when attributes is an array of symbols' do
+      let(:validation_error) { Grape::Exceptions::Validation.new(params: [:admin_field], message: 'Can not set admin-only field') }
+      subject { described_class.new(errors: [validation_error]).full_messages }
+
+      it 'returns an array with an error full message' do
+        expect(subject.first).to eq('admin_field Can not set admin-only field')
+      end
+    end
   end
 
   context 'api' do


lib/grape/exceptions/base.rb
```diff
@@ -51,12 +51,12 @@ def resolution(key, attributes)
...
       def translate_attributes(keys, **options)
         keys.map do |key|
-          translate("#{BASE_ATTRIBUTES_KEY}.#{key}", default: key, **options)
+          translate("#{BASE_ATTRIBUTES_KEY}.#{key}", default: key.to_s, **options)
         end.join(', ')
       end
 
       def translate_attribute(key, **options)
-        translate("#{BASE_ATTRIBUTES_KEY}.#{key}", default: key, **options)
+        translate("#{BASE_ATTRIBUTES_KEY}.#{key}", default: key.to_s, **options)
```
dblock: Maybe I am nitpicking, but feels like we should do this inside `translate`, no?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1737](https://github.com/ruby-grape/grape/pull/1737): Ensure the value of default option must be a string when calling i18n.t - [@mlzhuyi](https://github.com/mlzhuyi).
```
dblock: This fixes a bug. What's the error that you get without this fix? That's what should be in the changelog, please.

lib/grape/exceptions/base.rb
mlzhuyi: It is generally better. But I'm not sure if there is any need to pass an alternative symbol key as value of :default in other cases, though, there isn't any example now.

mlzhuyi: It is generally better. But I'm not sure if there is any need to pass an alternative symbol key as value of :default in other cases, though, there isn't any example now.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1737](https://github.com/ruby-grape/grape/pull/1737): Fix translating error whe pass symbols as params in custom validations - [@mlzhuyi](https://github.com/mlzhuyi).
```
dblock: Typo: `when passing`

mlzhuyi:
It has been changed

lib/grape/exceptions/base.rb
```diff
@@ -71,6 +71,7 @@ def translate_message(key, **options)
       end
 
       def translate(key, **options)
+        options[:default] &&= options[:default].to_s
```
dblock: This modifies the caller's `options`, which can easily be a frozen hash or whatever.

mlzhuyi: `*` and `**` in arguments act a different way with common argument.

```ruby
def common_argument_test(args)
  p args.object_id
end

def keyword_argument_test(**args)
  p args.object_id
end

args = { a: 1 }
p args.object_id # => 70338004137000
common_argument_test args # => 70338004137000
keyword_argument_test args # => 70338004136780
```
In there code, `args` in `keyword_argument_test ` is a new hash, create by the method itself.I think we can treat it as a most basic hash.

dblock: I guess we always declared it the old fashioned way as `options = {}`. I worry that this is different version from version, especially reading [this SO](https://stackoverflow.com/questions/18289152/what-does-a-double-splat-operator-do). Can we just do this the old fashioned way and avoid surprises?

mlzhuyi: Do you mean refactoring out the code, for there here are many `**` now?

dblock: Sorry if I wasn't clear. The double splat is fine. I mean this:

```ruby
        options = options.dup
        options[:default] &&= options[:default].to_s
```

We dup arguments everywhere as far as I can see.

mlzhuyi: It's a really good habit in case of unknown changes. <br>I have added it :)

dblock:
Thanks, merged.

PR #1740, opened 2018-02-06
Fix exception being thrown when dependent parameter is a Hash

jvortmann:
When using `given` and the paramter type is a Hash,
a Grape::Exceptions::UnknownParameter is thrown. This is a wrong behavior
due to the fact that the method `declared_param?(param)` do not deal
properly with params that are hashes (even after flattening the
declared params).

This commit fixes this by checking if the param is a Hash and thus
returning its first key. This approach was already used in other part of
the code, so it was extracted and reused on `declared_params?(param)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 #### Fixes
 
+* [#1740](https://github.com/ruby-grape/grape/pull/1740): Fix dependent parameter validation using `given` when parameter is a `Hash` - [@jvortmann](https://github.com/jvortmann).
+
 * Your contribution here.
 
 ### 1.0.2 (1/10/2018)
--- a/lib/grape/dsl/parameters.rb
+++ b/lib/grape/dsl/parameters.rb
@@ -201,7 +201,7 @@ module Grape
       # @yield a parameter definition DSL
       def given(*attrs, &block)
         attrs.each do |attr|
-          proxy_attr = attr.is_a?(Hash) ? attr.keys[0] : attr
+          proxy_attr = first_hash_key_or_param(attr)
           raise Grape::Exceptions::UnknownParameter.new(proxy_attr) unless declared_param?(proxy_attr)
         end
         new_lateral_scope(dependent_on: attrs, &block)
@@ -213,7 +213,9 @@ module Grape
       def declared_param?(param)
         # @declared_params also includes hashes of options and such, but those
         # won't be flattened out.
-        @declared_params.flatten.include?(param)
+        @declared_params.flatten.any? do |declared_param|
+          first_hash_key_or_param(declared_param) == param
+        end
       end
 
       alias group requires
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 * Your contribution here.
 
+* [#1740](https://github.com/ruby-grape/grape/pull/1740): Fix dependent parameter validation using `given` when parameter is a `Hash` - [@jvortmann](https://github.com/jvortmann).
```
dblock: Nitpick, put "Your contribution here" below the line? Just more consistent with how we've been doing it.

spec/grape/validations/params_scope_spec.rb
```diff
@@ -466,6 +466,20 @@ def initialize(value)
       end.to raise_error(Grape::Exceptions::UnknownParameter)
     end
 
+    it 'DOES NOT raise an error if the dependent parameter is a Hash' do
```
dblock: Lets not shout. So "does not raise error if ..." Thanks :)

spec/grape/validations/params_scope_spec.rb
```diff
@@ -466,6 +466,20 @@ def initialize(value)
...
+      expect do
+        subject.params do
+          optional :d, type: Integer
```
dblock: I am nitpicking, but is this `optional :d` needed for the spec?

lib/grape/dsl/parameters.rb
```diff
@@ -238,6 +240,12 @@ def params(params)
         params = map_params(params, @element) if @element
         params
       end
+
+      private
+
+      def first_hash_key_or_param(parameter)
+        parameter.is_a?(Hash) ? parameter.keys[0] : parameter
```
dblock: Maybe `.keys.first` is more idiomatic?

jvortmann: Yes, it might.

The case is that I've just move the line that were in the `given` method `as is`.
Should I change the line then?

spec/grape/validations/params_scope_spec.rb
jvortmann: No problem.

The idea was not to shout, but to make it more visible when running using "documentation" format.

spec/grape/validations/params_scope_spec.rb
jvortmann: I did to populate declared_params with something more than a Hash.
Just removed as I believe there are other specs covering thath.

jvortmann:
@dblock just did the changes :)

lib/grape/dsl/parameters.rb
dblock: It's nbd

dblock:
Merged, thank you.

jvortmann:
@dblock nice :)

What are the plans to cut a release? Is there a particular time-box for that?

dblock:
No plans, maintainers cut releases fairly frequently so stay tuned.

douglaslise:
I have the same problem, so a gem release with this fix would be pleased

dblock:
#1748

Issue #1713, opened 2017-11-21
rescue_from :all not rescuing all exceptions


Jelkster:
According to the README, grape can rescue from all exceptions with the following:
```ruby
rescue_from :all
```

I recently stumbled across a case where the wrong method accidentally raised a `NotImplementedError` and caused the web request to hang as a result.  I was surprised to see this because I had the `rescue_from :all` included in the API.  I suspected that maybe `NotImplementedError` was not a `StandardError` and maybe that somehow had something to do with it.  Sure enough, I found that it was a `ScriptError`.  

I consulted the documentation again to see if I missed something, and actually saw an example of `rescue_from NotImplementedError` in the README.  So it appeared as though it was perfectly valid at some point in time.  That being said, I also saw the following message below an example:  _"In this case UserDefinedError must be inherited from StandardError"_.

After taking a look at [error.rb](https://github.com/ruby-grape/grape/blob/3c0be5eb0526c0824ea425350ffcb919a566d4bf/lib/grape/middleware/error.rb#L39) it appears as though only `StandardError` will get rescued.  I modified the source and replaced `StandardError` with `Exception` to see if that would resolve the issue and it did!

The last thing I did was inspect the history of the error.rb file to see if it was ever rescuing `Exception` before, and sure enough [it was](https://github.com/ruby-grape/grape/commit/f038ed5457cef803b0275e4ad7a99674e61e4394#diff-38c8cda4fd3b2789c88488c6e1e08d0eL28).  It appears Rubocop recommended the change.

dblock:
I think we definitely need to update the documentation first and possibly add a `rescue_from :exception` that will behave differently? Please contribute?

Jelkster:
I'm willing to try and help.  Regarding the documentation, would you like me to remove the `NotImplementedError` example only, or do we also need to address each mention of `"...all exceptions..."`?

dblock:
Whatever you think is appropriate, try something and we can look at it.

dblock:
Doc updated in https://github.com/ruby-grape/grape/pull/1725. What do we want to do about this issue?

Jelkster:
I think your `rescue_from :exception` proposal is good.  Want me to take a stab at that?

dblock:
Yes please @Jelkster

mtsmfm:
Is it really documentation problem?
[...]

PR #1745, opened 2018-03-01
Rescue truly all exceptions

mtsmfm:
Fix https://github.com/ruby-grape/grape/issues/1713

Currently, `rescue_from` can't rescue from exceptions which aren't inherited from `StandardError`.
Due to this constraint, we can't rescue exceptions which are inherited from `Exception`.
Some exceptions are inherited from `Exception` because they should be rescued explicitly. (ex. [`Rack::Timeout::RequestTimeoutException`](https://github.com/heroku/rack-timeout/#errors))

Before v0.6.0, it can rescue truly all exceptions but unintentionally it changed because of [rubocop](https://github.com/ruby-grape/grape/commit/f038ed5457cef803b0275e4ad7a99674e61e4394#diff-38c8cda4fd3b2789c88488c6e1e08d0eL28).

I think grape should be able to rescue from all exceptions [like Rails](https://github.com/rails/rails/blob/7a9b93805bd6f901de72c8f84c75e4f52476a872/actionpack/lib/action_controller/metal/rescue.rb#L23-L25).

Perhaps it'll affect a lot but I think current users assume `rescue_from :all` can rescue from truly all exceptions.


dblock:
So 0.6.0 is very old. A few things need to happen to make this change:

* A backward compatible way to rescue only StandardError errors. At that point I still wonder whether the current default is OK and we should just add support for `rescue_from :exception`. 
* UPGRADING.md
* Major version bump

mtsmfm:
Thank you for the reviewing!

> A backward compatible way to rescue only StandardError errors

I think `rescue_from StandardError` is the way

dm1try:
>  I think current users assume rescue_from :all can rescue from truly all exceptions.

Actually, I always thought that `rescue_from :all` is a counterpart for [standard ruby behavior](https://ruby-doc.org/core-2.2.0/StandardError.html) when `rescue` is called without an explicit exception type; so using `:all` is just a DSL "compromise" for this behavior(`.._from` suffix needs some param :grimacing: )
```ruby
# :(
rescue_from do
  ...
end
```
and read `rescue_from :all` like `default_rescue`)

> Currently, rescue_from can't rescue from exceptions which aren't inherited from StandardError.

So I suggest fixing this first and add to README that  `rescue_from :all` behaves like standard `rescue`.

Otherwise, the original intention of providing  special exceptions like `Rack::Timeout::RequestTimeoutException`  will be broken(the intention was to force a developer to handle such exceptions explicitly)

dblock:
Makes sense @dm1try. Thanks. I'm still looking for a mergeable PR, see my comments above.

dblock:
Rebase against https://github.com/ruby-grape/grape/pull/1749.

dblock:
What should we do with this @mtsmfm?

mtsmfm:
Sorry for being late

dblock:
Just seeing this, apologies for the delay. 

This is also a behavior change. We just merged #1765 which forced a major version bump with a backward incompatible change, so we should put this one in as well. @mtsmfm would you mind adding a paragraph to UPGRADING.md too please?

PR #1747, opened 2018-03-07
Updating activerecord without rails section on README.md

thiagoramos23:
The class `ActiveRecord::ConnectionAdapters::ConnectionManagement` no longer exists in ActiveRecord 5. 

We need to handle the connection management from now on.

Updated the documentation following a solution that worked at my company.

Solution is here: [Replacing ActiveRecord ConnectionManagement](https://stackoverflow.com/questions/41400202/replacing-activerecordconnectionadaptersconnectionmanagement-in-activerecord)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ### 1.0.3 (Next)
 
-#### Features
+#### Fixes
 
-* Your contribution here.
+* Changed ActiveRecord Without Rails section of the README.md to use another approach to ConnectionManagement. [@thiagoramos23](https://github.com/thiagoramos23)
 
 #### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Fixes
 
+* Your contribution here.
 * [#1747](https://github.com/ruby-grape/grape/pull/1747): Updating activerecord without rails section on README.md - [@thiagoramos23](https://github.com/thiagoramos23).
 
 #### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,14 +1,13 @@
 ### 1.0.3 (Next)
 
-#### Features
+#### Fixes
```
dblock: Uhm what happened here? Should be features. Generally we don't need a CHANGELOG for updating documentation so I think you should just revert this.

README.md
```diff
@@ -269,11 +269,36 @@ Grape will also automatically respond to HEAD and OPTIONS for all GET, and just
 If you want to use ActiveRecord within Grape, you will need to make sure that ActiveRecord's connection pool
 is handled correctly.
 
-The easiest way to achieve that is by using ActiveRecord's `ConnectionManagement` middleware in your
-`config.ru` before mounting Grape, e.g.:
+In the old days you could use ActiveRecord's `ConnectionManagement` middleware in your
```
dblock: Lets make two sections, one for ActiveRecord 4 and another for ActiveRecord 5 and not write "in the old days" and just provide instructions to what to do? 

What happened to the middleware? Is it really being replaced by custom code?

thiagoramos23: Ok. Actually active record does not implement that anymore. It's up to the client now. So, if you want the same functionality you will need to add a middleware. I will do the changes.

dblock: I found https://github.com/rails/rails/issues/26947 and https://stackoverflow.com/questions/41400202/replacing-activerecordconnectionadaptersconnectionmanagement-in-activerecord. Looks like someone already made a gem for this, lets use that instead of recommending rolling our own middleware?

thiagoramos23: I support that. Will do the changes. Although I didn't test this. I test the other solution, but seems like pretty solid.

PR #1749, opened 2018-03-11
Allow rescue from non-StandardError exceptions

dm1try:
ref #1745 #1713


dblock:
Perfect. This solves the actual bug of being able to rescue non-StandardError exceptions, so we can now talk about changing the behavior of `:all`.

PR #1754, opened 2018-03-27
Allow rescue from non-`StandardError` exceptions to use default error handling

Jelkster:
This continues to address issues discussed in #1713 and #1745.

 @dm1try recently [added the ability to explicitly rescue non-StandardError exceptions](https://github.com/ruby-grape/grape/pull/1749) but I found that his solution required a block to be provided in order to handle exceptions or else it would raise an error due to no handler being found. This PR corrects that by re-using the existing `find_handler` and `rescuable?` logic, with some minor modifications to account for `Exception` being rescued instead of `StandardError`. This meant that the additional `rescue Exception` block could go away completely.

**Edit:**  I forgot that @dm1try recommended we deprecate the use of `:all` so this additional option (`rescue_from :exception` - removed from PR after further discussion) is probably not necessary, as you can simply `rescue_from Exception`.  The same as you can `rescue_from StandardError` for `rescue_from :all`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [#1749](https://github.com/ruby-grape/grape/pull/1749): Allow rescue from non-`StandardError` exceptions - [@dm1try](https://github.com/dm1try).
 * [#1750](https://github.com/ruby-grape/grape/pull/1750): Fix a circular dependency warning due to router being loaded by API - [@salasrod](https://github.com/salasrod).
 * [#1752](https://github.com/ruby-grape/grape/pull/1752): Fix `include_missing` behavior for aliased parameters - [@jonasoberschweiber](https://github.com/jonasoberschweiber).
+* [#1754](https://github.com/ruby-grape/grape/pull/1754): Allow rescue from non-`StandardError` exceptions to use default error handling - [@jelkster](https://github.com/jelkster).
 * Your contribution here.
 
 ### 1.0.2 (1/10/2018)
--- a/README.md
+++ b/README.md
@@ -2289,7 +2289,7 @@ Here `'inner'` will be result of handling occured `ArgumentError`.
 
 Any exception that is not subclass of `StandardError` should be rescued explicitly.
 Usually it is not a case for an application logic as such errors point to problems in Ruby runtime.
-This is following [standard recomendations for exceptions handling](https://ruby-doc.org/core/Exception.html).
+This is following [standard recommendations for exceptions handling](https://ruby-doc.org/core/Exception.html).
 
 ### Rails 3.x
 
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -36,7 +36,7 @@ module Grape
           error_response(catch(:error) do
             return @app.call(@env)
           end)
-        rescue StandardError => e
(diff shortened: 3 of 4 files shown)


dm1try:
> we deprecate the use of :all so this additional option is probably not necessary, as you can simply rescue_from Exception. The same as you can rescue_from StandardError for rescue_from :all

yep, `all` adds ambiguity (as I mentioned in related issue) but it's so "sugar")
my point is that a user should not waste time to resolve this using README as he could just use his Ruby knowledge.

spec/integration/multi_xml/xml_spec.rb
```diff
@@ -1,7 +1,13 @@
 require 'spec_helper'
 
 describe Grape::Xml do
-  it 'uses multi_xml' do
-    expect(Grape::Xml).to eq(::MultiXml)
+  if Object.const_defined? :MultiXml
```
dblock: This change should not be required, the integration spec is specifically for mutli_xml, so it should never be in that `else`.

dblock:
Stepping back I think `:all` is bad all around. I am down with deprecating it completely and forcing users to do `rescue_from Exception` or `rescue_from StandardError` with a major version bump. That can be done in a separate PR.

Therefore, before we add `rescue_from :exception`, are you are you saying @Jelkster that we _cannot_ make a non-block version without introducing a symbol sugar? Otherwise I think we just want to fix that ability instead of adding `:exception` support.

Jelkster:
@dblock, no, this code can handle blocks no problem without the symbol sugar.  I was in the middle of implementing `rescue_from :exception` before @dm1try addressed the issue, so I just left it in there in case you wanted to take a look at the end result.  It seems like we are all realizing the lack of need (and clarity) for the sugar.

spec/integration/multi_xml/xml_spec.rb
Jelkster: That makes sense about not ever being in the `else` since this is for multi_xml so I'll definitely remove that, but if I don't check for the presence of the constant, this is the only spec that fails for me locally on my ubuntu VM.

dblock: Run specs with `rake`, I think we cleverly exclude those as needed and run them with their own bundle?

dblock:
So should we just close this and focus on deprecating `:all`? Maybe we still want README updates?

Jelkster:
This PR actually added the ability to handle non-blocks ([this test fails without it](https://github.com/ruby-grape/grape/pull/1754/files#diff-be9e2328dbd775eb503f282e5df842daR113)), so I think we'll still want that piece.  I can convert this into a fix instead of a feature.

spec/integration/multi_xml/xml_spec.rb
Jelkster: Ahh, ok, I was running `bundle exec rspec`  🤦‍♂️

dblock:
Makes sense, looking forward to an update @Jelkster. Thanks for hanging in there!

Jelkster:
@dblock updated!

dblock:
Great work @Jelkster. Want to tackle the deprecation of `:all` next?

Jelkster:
I'm actually on vacation this week, but I can definitely take a stab when it's over.

dblock:
What do we pay you for @Jelkster :)

Issue #1288, opened 2016-02-23
exactly_one_of seems not work for nested parameters


mmkhmk:
Hello,

I'm currently creating an API with the latest version(0.14.0) but exactly_one_of doesn't work as expected for nested parameters.

Here is the sample.

```
params do
  optional :additionalFacets, type: Hash do
    requires: label, type: String
    requires: facet, type: Hash do
      requires :field, type: String
      requires :type, type: String

      optional :limit, type: Integer
      optional :start, type: Integer

      exactly_one_of :limit, :start
    end
  end
end
```

In this case I meant the validation of exactly_one_of works only when additionalFacets are given.
But it works when it is not given as well.

Can you help check this?

dblock:
It sounds like a bug, try writing a spec that reproduces this?

mmkhmk:
Here is the spec that reproduces this.

```
context 'nested params in nested params' do
  before do
    subject.params do
      requires :nested, type: Hash do
        requires :nested_in_nested, type: Hash do
          optional :beer_nested
          optional :wine_nested
          optional :juice_nested
          exactly_one_of :beer_nested, :wine_nested, :juice_nested
        end
      end
      optional :nested2, type: Hash do
        requires :nested_in_nested2, type: Array do
          optional :beer_nested2
          optional :wine_nested2
          optional :juice_nested2
          exactly_one_of :beer_nested2, :wine_nested2, :juice_nested2
        end
      end
    end
    subject.get '/exactly_one_of_nested' do
      'exactly_one_of works!'
    end
  end

  it 'errors when none are present' do
    get '/exactly_one_of_nested'
    expect(last_response.status).to eq(400)
    expect(last_response.body).to eq 'nested is missing, nested[nested_in_nested] is missing, beer_nested, wine_nested, juice_nested are missing, exactly one parameter must be provided'
  end

  it 'succeeds when one is present' do
    get '/exactly_one_of_nested', nested: { nested_in_nested: { beer_nested: 'string' } }
    expect(last_response.status).to eq(200)
    expect(last_response.body).to eq 'exactly_one_of works!'
  end

  it 'errors when two or more are present' do
    get '/exactly_one_of_nested', nested: { nested_in_nested: { beer_nested: 'string' } }, nested2: { nested_in_nested2: [{ beer_nested2: 'string', wine_nested2: 'anotherstring' }] }
    expect(last_response.status).to eq(400)
[...]

PR #1755, opened 2018-03-28
Fix shared params with exactly_one_of

milgner:
Tentative fix for #1288 - not sure whether it is the right approach but the specs are green :wink:

I can't shake the feeling that there's some unnecessary complexity in combination with the changes from #1625 so I'm open to suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#1755](https://github.com/ruby-grape/grape/pull/1755): Fix shared params with exactly_one_of - [@milgner](https://github.com/milgner).
 * [#1740](https://github.com/ruby-grape/grape/pull/1740): Fix dependent parameter validation using `given` when parameter is a `Hash` - [@jvortmann](https://github.com/jvortmann).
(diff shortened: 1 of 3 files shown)


milgner:
Alrighty, waiting for Travis to do its thing and if everything is green there as well, I'll update the changelog...

rnubel:
It's been a while since I looked into this, but I like the simplicity of your fix. My own attempt at resolving the bug is https://github.com/ruby-grape/grape/pull/1691/files, but I don't like how it involved updating all of the validators individually (which is why I stalled on it). However, the specs that I added in that PR might be useful as additional test cases for this PR.

spec/integration/multi_xml/xml_spec.rb
```diff
@@ -1,7 +1,13 @@
 require 'spec_helper'
 
 describe Grape::Xml do
```
dblock: This shouldn't be necessary, integration specs run in isolation. Is this failing on HEAD?

dblock:
Fix looks good to me. The XML adapter spec should be undone (run `rake` for tests) otherwise it's all good.

milgner:
Sorry, I was using `rspec` to run the tests and it was failing, so I fixed it. But using `rake`, it works as expected so I dropped the commit from the branch again.

PR #1756, opened 2018-04-18
Grape exception handling

soylent:
This pull request fixes Grape exception handling for the following case:

```ruby
class App < Grape::API
  rescue_from :grape_exceptions

  rescue_from Grape::Exceptions::ValidationErrors do |error|
    # BUG: This block is never called
    error!('redefined', 403)
  end

  params { requires :foo, type: String }

  get('/') { 'Hello' }
end
```

I reworked `Grape::Middleware::Error` a bit:
* the code is easier to follow now
* it works faster because we iterate over `options[:rescue_handlers]` and `options[:base_only_rescue_handlers]` only once
* it allocates less garbage
* fixed another covert bug: `options[:base_only_rescue_handlers]` should have higher priority than `options[:rescue_handlers]` because it is more restrictive

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#1750](https://github.com/ruby-grape/grape/pull/1750): Fix a circular dependency warning due to router being loaded by API - [@salasrod](https://github.com/salasrod).
 * [#1752](https://github.com/ruby-grape/grape/pull/1752): Fix `include_missing` behavior for aliased parameters - [@jonasoberschweiber](https://github.com/jonasoberschweiber).
 * [#1754](https://github.com/ruby-grape/grape/pull/1754): Allow rescue from non-`StandardError` exceptions to use default error handling - [@jelkster](https://github.com/jelkster).
+* [#1756](https://github.com/ruby-grape/grape/pull/1756): Allow custom Grape exception handlers when the built-in exception handling is enabled - [@soylent](https://github.com/soylent).
 * Your contribution here.
 
 ### 1.0.2 (1/10/2018)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * [#1750](https://github.com/ruby-grape/grape/pull/1750): Fix a circular dependency warning due to router being loaded by API - [@salasrod](https://github.com/salasrod).
 * [#1752](https://github.com/ruby-grape/grape/pull/1752): Fix `include_missing` behavior for aliased parameters - [@jonasoberschweiber](https://github.com/jonasoberschweiber).
 * [#1754](https://github.com/ruby-grape/grape/pull/1754): Allow rescue from non-`StandardError` exceptions to use default error handling - [@jelkster](https://github.com/jelkster).
+* [#1756](https://github.com/ruby-grape/grape/pull/1756): Allow custom Grape exception handlers when the build-in exception handling is enabled - [@soylent](https://github.com/soylent).
```
dblock: typo: builT-in

lib/grape/middleware/error.rb
```diff
@@ -122,16 +85,51 @@ def format_message(message, backtrace, original_exception = nil)
 
       private
 
-      def rescue_all?
-        options[:rescue_all]
+      DEFAULT_RESCUE_HANDLER = :handle_error
```
dblock: Seems pretty but maybe unnecessary? I mean `:handle_error` is already a constant, since this isn't going to change I say copy-paste it everywhere.

dblock: ... or maybe even rename it to `:default_rescue_handler`?

dblock:
Perfect, thanks. See minor comments above.

soylent:
@dblock All done

dblock:
Merged, thanks.

PR #1758, opened 2018-05-04
Fix expanding LOAD_PATH in gemspec

2maz:
Otherwise Grape::VERSION might refer to another, e.g.,
 system wide installed gem version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * [#1759](https://github.com/ruby-grape/grape/pull/1759): Update appraisal for rails_edge - [@zvkemp](https://github.com/zvkemp).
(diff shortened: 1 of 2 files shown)


dblock:
Interesting. Care to look at the build failure? Probably unrelated, but lets exclude rails edge in that case.

2maz:
```
Bundler could not find compatible versions for gem `"activerecord":
  In rails_edge.gemfile:
    arel was resolved to 10.0.0.pre, which depends on
      activerecord (>= 6.a)
```
To me, it does not seem to be related. 
Issue lies probably in the dependency definition given in "the repository: 'rails/arel'".
Should I update the travis configuration with this PR?

dblock:
It was fixed in #1759, rebase please.

2maz:
updated

PR #1759, opened 2018-05-04
Instrument serialization and rendering as 'format_response.grape'

zvkemp:
A good portion of a grape API response is currently uninstrumented — when `endpoint_run.grape` closes, the response bodies still haven't been serialized or rendered into their appropriate templates. Adding `format_response.grape` instrumentation will allow us to measure time spent in this portion of the request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,6 @@ matrix:
     - rvm: 2.3.5
       gemfile: gemfiles/rack_1.5.2.gemfile
     - rvm: 2.3.5
-      gemfile: gemfiles/rails_edge.gemfile
-    - rvm: 2.3.5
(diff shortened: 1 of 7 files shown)


dblock:
LGTM, care to check out the build failure?

zvkemp:
@dblock Failing on edge rails. Looks like it's related to the recent arel/activerecord merge.

```
Bundler could not find compatible versions for gem "activerecord":
  In rails_edge.gemfile:
    arel was resolved to 10.0.0.pre, which depends on
      activerecord (>= 6.a)
```

Unrelated to this PR, but I'd be happy to take care of it (could just tack on another commit to this one, unless you prefer to keep the issues separate).

Issue #1762, opened 2018-05-22
Default formatter error can cause XSS rendering issue


ctennis:
If you issue a request into an API endpoint with a format specified that is not handled, you get back a 406 error with message 'The requested format 'format' is not supported.'

The name of the specified format is rendered in the error message, which defaults as HTML.  You can easily craft a format value that gets passed in and renders as an XSS.  An example:

http://example.com/api/endpoint?format=%3Cscript%3Ealert(document.cookie)%3C/script%3E

Renders as html:

The requested format '<script>alert(document.cookie)</script>' is not supported.

Which will cause a javascript popup if you visit this page in a browser.

dblock:
Looks legit. Would appreciate at least a spec, maybe a fix, please. We can just drop the actual value I think.

ctennis:
Happy to do so, but I'm not sure the best approach for fixing. The use of the format name in the script was done intentionally (#322 I believe).  Ideally we'd sanitize the output before displaying it as html, but we don't know it's going to be displayed as html. ActionSupport has a safebuffer that could be used for sanitizing the html text prior to rendering, but we don't want to do that if the text won't be rendered as html.

dblock:
Closed via #1763.

PR #1763, opened 2018-05-23
When returning an HTML error, make sure it's safe

ctennis:
When calling into an API specifying a crafted format that is HTML,
the returned error renders the HTML back to the user, causing a potential XSS
issue.  For example:

http://example.com/api/endpoint?format=%3Cscript%3Ealert(document.cookie)%3C/script%3E

Renders as html:

The requested format '<script>alert(document.cookie)</script>' is not supported.

When an error generates html back to the user, make sure it's properly escaped.

Fixes issue #1762

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 #### Fixes
 
+
+* [#1762](https://github.com/ruby-grape/grape/pull/1763): Fix unsafe HTML rendering on errors - [@ctennis](https://github.com/ctennis).
 * [#1759](https://github.com/ruby-grape/grape/pull/1759): Update appraisal for rails_edge - [@zvkemp](https://github.com/zvkemp).
 * [#1758](https://github.com/ruby-grape/grape/pull/1758): Fix expanding load_path in gemspec - [@2maz](https://github.com/2maz).
 * Your contribution here.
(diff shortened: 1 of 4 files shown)


ctennis:
I had to fix six specs to be "compliant", which really surprised me.  The ones that return json in particular seem like maybe they're not setting their content-type appropriately to json and as such are returning a json blob in html code.  Is this ok/correct?

dblock:
The ones that expect JSON should indeed set the correct content type, thanks for taking care of it.

Looks like Rails 3 may be a problem. If it gets complicated we can discuss dropping support.

dblock:
Better, but not 🍏  :(

ctennis:
@dblock ok specs passing now

rschultheis:
:wave:  I am a member of the GitHub team that sends security alerts.  This issue was assigned [CVE-2018-3769](https://nvd.nist.gov/vuln/detail/CVE-2018-3769) which is now disclosed publicly therefore therefore we would like to send security alerts for it.

However, in our curation process it was noticed that the latest version of `grape`, `1.0.3`, does not appear to include this fix.  Will a new version of grape be published soon with this fix?

Thank you for any information!  🙇

dblock:
I've released 1.1.0.

PR #1765, opened 2018-06-04
Use 415 status for request bodies of unsupported media type

jdmurphy:
Noticed that for requests where bodies are present (i.e. PUT, POST, PATCH, etc.), that if the content type header was something unsupported that a 406 would be returned. Reading over the original [RFC 2616](https://www.ietf.org/rfc/rfc2616.txt), a 415 should be used when a request entity is not in a supported format, but currently a 406 is returned. It seems like it should be as simple as changing [this logic](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/formatter.rb#L97-L99).

### DIFF
--- a/lib/grape/middleware/formatter.rb
+++ b/lib/grape/middleware/formatter.rb
@@ -95,7 +95,7 @@ module Grape
         fmt = request.media_type ? mime_types[request.media_type] : options[:default_format]
 
         unless content_type_for(fmt)
-          throw :error, status: 406, message: "The requested content-type '#{request.media_type}' is not supported."
+          throw :error, status: 415, message: "The provided content-type '#{request.media_type}' is not supported."
         end
         parser = Grape::Parser.parser_for fmt, options
         if parser
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -941,15 +941,15 @@ describe Grape::Endpoint do
       end
     end
 
-    it 'responds with a 406 for an unsupported content-type' do
+    it 'responds with a 415 for an unsupported content-type' do
       subject.format :json
       # subject.content_type :json, "application/json"
       subject.put '/request_body' do
         params[:user]
       end
       put '/request_body', '<user>Bobby T.</user>', 'CONTENT_TYPE' => 'application/xml'
-      expect(last_response.status).to eq(406)
-      expect(last_response.body).to eq('{"error":"The requested content-type \'application/xml\' is not supported."}')
+      expect(last_response.status).to eq(415)
+      expect(last_response.body).to eq('{"error":"The provided content-type \'application/xml\' is not supported."}')
     end
 
     it 'does not accept text/plain in JSON format if application/json is specified as content type' do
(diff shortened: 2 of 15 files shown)


dblock:
That is probably correct, but is a backwards incompatible change.

At the very least it needs to have a section in https://github.com/ruby-grape/grape/blob/master/UPGRADING.md, please.

spec/grape/endpoint_spec.rb
```diff
@@ -948,7 +948,7 @@ def app
         params[:user]
       end
       put '/request_body', '<user>Bobby T.</user>', 'CONTENT_TYPE' => 'application/xml'
-      expect(last_response.status).to eq(406)
+      expect(last_response.status).to eq(415)
```
dblock: The description of this test is now incorrect.

dblock:
Does this need more test scenarios? With or without body for example?

jdmurphy:
@dblock, I agree it could probably use more tests. Do you have a preference/idea of where you want them. The tests that failed are actually just in a `#params` testing block which doesn't seem like the best place, but the formatter_spec also doesn't really have anything like that.

Also, I can update the UPGRADING document, but where should I put it (in terms of heading)? Should I assume a 2.x release? Will this need to sit in a 2.x feature branch until any other planned or unplanned non-passive changes are in?

dblock:
No strong opinions on test locations, but definitely in specs :) Maybe something new.

It doesn't need to sit in a branch, just keep updating the PR. I think we should do 1.1, Grape hasn't been using the major version, we've spent years getting to 1.0. Appreciate any thoughts on that too.

jdmurphy:
@dblock, I ended up adding them to the formatter_spec in d2e2f7e (removed my accidental require pry addition in 3c7df26); I grouped it with some similar logic around the correct methods and reading of the body into the rack request form hash. Let me know what you think.

dblock:
This looks good. Maybe squash it if you don't mind?

I want to leave this open for a bit, maybe @namusyaka or @dm1try could pitch in that we want to do this?

jdmurphy:
@dblock, squashed all the things.

dm1try:
as for versioning

> I think we should do 1.1

I agree, up to 1.1.0 and add "incompatibility" note in UPGRADING.md

Grape does not claim that it uses SemVer, so for such incompatible fix is fine :) (moreover, this way is common in Ruby/Rails community itself, at least for now)

lib/grape/middleware/formatter.rb
```diff
@@ -95,7 +95,7 @@ def read_rack_input(body)
         fmt = request.media_type ? mime_types[request.media_type] : options[:default_format]
 
         unless content_type_for(fmt)
-          throw :error, status: 406, message: "The requested content-type '#{request.media_type}' is not supported."
+          throw :error, status: 415, message: "The requested content-type '#{request.media_type}' is not supported."
```
dm1try: nitpicking, but I would change the message (ex. from "requested" to  "provided") to clarify the difference between 406/415

dblock: @jdmurphy ^

dblock:
@jdmurphy can you please do the version bump and the upgrading part of this? thx

jdmurphy:
@dblock when you say do the version bump, what did you mean? Like update the version file to be 1.1.0 and update the changelog and upgrading doc to reflect the new version there as well?

lib/grape/middleware/formatter.rb
jdmurphy: Updated, rebased and pushed as a single commit

jdmurphy:
I updated the version file, README, UPGRADING and CHANGELOG, and then rebased and pushed.

dblock:
Merged. Great work @jdmurphy and thanks!

PR #1767, opened 2018-06-15
Update rails reloading instructions in README

yeonhoyoon:
Rails reloading insturctions no longer required as of Rails >= 5.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,14 +316,7 @@ run Rack::Cascade.new [API, Web]
 
 ### Rails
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -316,22 +316,15 @@ run Rack::Cascade.new [API, Web]
 
 ### Rails
 
-Place API files into `app/api`. Rails expects a subdirectory that matches the name of the Ruby module and a file name that matches the name of the class. In our example, the file name location and directory for `Twitter::API` should be `app/api/twitter/api.rb`.
-
-Modify `application.rb`:
-
-```ruby
-config.paths.add File.join('app', 'api'), glob: File.join('**', '*.rb')
```
dblock: Feels like we should leave this for older versions? Maybe just make sections 

```
#### Rails 5.2 or Newer
...

#### Rails 4.0

````

?

Issue #1769, opened 2018-07-17
Params 'as' option fails within 'given' block


jereynolds:
The parameter alias `as` option is ignored when a parameter is declared inside a `given` block

```
params do
  optional :label, as: :vertex_label, type: String
  given :label do
    optional :property, as: :vertex_property, type: String
  end
end
```

When a route with this params declaration is triggered, `declared(params)` evaluates to:
`{ "vertex_label" => "MyLabel", "property" => "MyProperty" }`

I expect the `property` parameter to be aliased to `vertex_property` in declared params, but this does not happen

dblock:
Looks legit, try writing a spec for it?

jereynolds:
Added fix to the PR

dblock:
Closing via https://github.com/ruby-grape/grape/pull/1771. Thanks @jereynolds!

jereynolds:
Do you know what the expected time until the fix is released is?

dblock:
Nope, but there're a bunch of changes so I'll do a release when I get to it or another maintainer beats me to it, https://github.com/ruby-grape/grape/issues/1773.

PR #1771, opened 2018-07-19
Fix param aliases within 'given' blocks

jereynolds:
Spec to display behavior of #1769 (as requested in that issue)


dblock:
Thanks. We don't need a changelog for the spec since it's not going to be merged without a fix. But now that you have a spec you can try a fix? :)

dblock:
Great job! Merging.

PR #1774, opened 2018-07-24
Change HTTP DELETE status code and body logic, use nil for HTTP 204

### LINKED ISSUE #1578: DELETE endpoint returned value not formatted and results in 204

basjanssen:
This PR has two goals:
- Change HTTP DELETE status code and body logic
presenting {}. [], false will be regarded as content and will use HTTP status code 200. The content body will be presented with the set content.

- Change using nil to set NO_CONTENT
Using nil will be more consistent with the way the content is regarded in the previous point. 

Fixes #1578 

- [ ] Receive and process first round of feedback
- [ ] Need to fix the spec at spec/grape/dsl/inside_route_spec.rb:144 (I do not understand the reason of failure).

### DIFF
--- a/spec/grape/dsl/inside_route_spec.rb
+++ b/spec/grape/dsl/inside_route_spec.rb
@@ -111,11 +111,42 @@ describe Grape::Endpoint do
       expect(subject.status).to eq 204
     end
 
-    it 'defaults to 200 on DELETE with a body present' do
-      request = Grape::Request.new(Rack::MockRequest.env_for('/', method: 'DELETE'))
-      subject.body 'content here'
-      expect(subject).to receive(:request).and_return(request)
-      expect(subject.status).to eq 200
+    describe 'defaulting to 200 on DELETE' do
+      it 'defaults to 200 on DELETE with a body present' do
+        request = Grape::Request.new(Rack::MockRequest.env_for('/', method: 'DELETE'))
+        subject.body 'content here'
+        expect(subject).to receive(:request).and_return(request)
+        expect(subject.status).to eq 200
+      end
+
+      it 'regards an empty string as content' do
+        request = Grape::Request.new(Rack::MockRequest.env_for('/', method: 'DELETE'))
+        subject.body ''
+        expect(subject).to receive(:request).and_return(request)
+        expect(subject.status).to eq 200
+      end
+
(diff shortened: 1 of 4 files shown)


dblock:
definitely would need a good UPGRADING.md change to understand all that's changing and will take a look at the details in code, too.

@namusyaka would appreciate another pair of eyes

spec/grape/dsl/inside_route_spec.rb
```diff
@@ -111,11 +111,42 @@ def initialize
...
+        subject.body []
+        expect(subject).to receive(:request).and_return(request)
+        expect(subject.status).to eq 200
+      end
+
+      # Still failing for some weird kind of reason.
+      it 'regards a nil as no content' do
+        request = Grape::Request.new(Rack::MockRequest.env_for('/', method: 'DELETE'))
+        subject.body nil
+        expect(subject).to receive(:request).and_return(request)
```
namusyaka: This test doesn't work because there is wrong code in using `expect ~ receive`.
Depending on the implementation you modified, in this case [the `status` method is called internally](https://github.com/basjanssen/grape/blob/44cd8fb8d3fd84552c9f1c9748b683ce2a5da7e2/lib/grape/dsl/inside_route.rb#L232).
As a result, [the `@status` variable is assigned](https://github.com/basjanssen/grape/blob/44cd8fb8d3fd84552c9f1c9748b683ce2a5da7e2/lib/grape/dsl/inside_route.rb#L174) and [the `request` member isn't called and it ends](https://github.com/basjanssen/grape/blob/44cd8fb8d3fd84552c9f1c9748b683ce2a5da7e2/lib/grape/dsl/inside_route.rb#L176).

You don't need test this expectation. Please remove this line.

PR #1776, opened 2018-08-05
Validates response processed by exception handler

### LINKED ISSUE #1757: Validation around error handler responses
Hi, I ran into an interesting issue the other day when debugging an ugly response from Grape when sending a request with missing parameters to one of our endpoints. 

Here's a stripped down version of the `V1::Endpoints::Base` definition:
```ruby
module V1::Endpoints::Base
  included do
    format :json

    rescue_from :all do |e|
      error! V1::Entities::Error.represent(e), 400
      Rails.logger.error "[API] -- #{e.message}\n #{e.backtrace.join("\n")}"
      Bugsnag.notify(e) if Rails.env.production?
    end
end
```
You'll notice that the `error!` call was mistakenly placed on the first line instead of the last line and since `return` was omitted, this method will return the result of `Bugsnag#notify`, which is `nil`. This is obviously a bug on our end, but it causes the request to fail with a 500 and returns this full stack trace to the requestor:

```
NoMethodError (undefined method `[]' for nil:NilClass):

grape (1.0.2) lib/grape/router.rb:163:in `cascade?'
grape (1.0.2) lib/grape/router.rb:95:in `transaction'
grape (1.0.2) lib/grape/router.rb:72:in `identity'
grape (1.0.2) lib/grape/router.rb:57:in `block in call'
grape (1.0.2) lib/grape/router.rb:137:in `with_optimization'
grape (1.0.2) lib/grape/router.rb:56:in `call'
grape (1.0.2) lib/grape/api.rb:119:in `call'
grape (1.0.2) lib/grape/api.rb:45:in `call!'
grape (1.0.2) lib/grape/api.rb:40:in `call'
actionpack (5.1.6) lib/action_dispatch/routing/mapper.rb:17:in `block in <class:Constraints>'
actionpack (5.1.6) lib/action_dispatch/routing/mapper.rb:46:in `serve'
actionpack (5.1.6) lib/action_dispatch/journey/router.rb:50:in `block in serve'
actionpack (5.1.6) lib/action_dispatch/journey/router.rb:33:in `each'
actionpack (5.1.6) lib/action_dispatch/journey/router.rb:33:in `serve'
actionpack (5.1.6) lib/action_dispatch/routing/route_set.rb:844:in `call'
rack (2.0.4) lib/rack/etag.rb:25:in `call'
rack (2.0.4) lib/rack/conditional_get.rb:38:in `call'
rack (2.0.4) lib/rack/head.rb:12:in `call'
activerecord (5.1.6) lib/active_record/migration.rb:556:in `call'
actionpack (5.1.6) lib/action_dispatch/middleware/callbacks.rb:26:in `block in call'
activesupport (5.1.6) lib/active_support/callbacks.rb:97:in `run_callbacks'
actionpack (5.1.6) lib/action_dispatch/middleware/callbacks.rb:24:in `call'
actionpack (5.1.6) lib/action_dispatch/middleware/executor.rb:12:in `call'
bugsnag (6.7.0) lib/bugsnag/integrations/rack.rb:46:in `call'
actionpack (5.1.6) lib/action_dispatch/middleware/debug_exceptions.rb:59:in `call'
actionpack (5.1.6) lib/action_dispatch/middleware/show_exceptions.rb:31:in `call'
railties (5.1.6) lib/rails/rack/logger.rb:36:in `call_app'
railties (5.1.6) lib/rails/rack/logger.rb:24:in `block in call'
activesupport (5.1.6) lib/active_support/tagged_logging.rb:69:in `block in tagged'
activesupport (5.1.6) lib/active_support/tagged_logging.rb:26:in `tagged'
activesupport (5.1.6) lib/active_support/tagged_logging.rb:69:in `tagged'
railties (5.1.6) lib/rails/rack/logger.rb:24:in `call'
actionpack (5.1.6) lib/action_dispatch/middleware/remote_ip.rb:79:in `call'
actionpack (5.1.6) lib/action_dispatch/middleware/request_id.rb:25:in `call'
rack (2.0.4) lib/rack/runtime.rb:22:in `call'
activesupport (5.1.6) lib/active_support/cache/strategy/local_cache_middleware.rb:27:in `call'
actionpack (5.1.6) lib/action_dispatch/middleware/executor.rb:12:in `call'
actionpack (5.1.6) lib/action_dispatch/middleware/static.rb:125:in `call'
rack (2.0.4) lib/rack/sendfile.rb:111:in `call'
railties (5.1.6) lib/rails/engine.rb:522:in `call'
```
IMO, there should be some validation of the return values of exception handlers when they're called by `Grape::Middleware::Error#exec_handler` to make sure the result is an error response since the status, headers, and body are passed directly back up to Rack. At best this is a fairly baffling message for a simple error, at worst it's a security vulnerability since this exposes which libraries a service is using along with their exact versions.

Please let me know if there's any other information I can provide.

darren987469:
Try to fix #1757.

### Problem  
```ruby
class Api < Grape::API
  rescue_from :all do
    error!('rain')
    nil
  end
end
```
The `rescue_from` block return nil, that cause response like
```
500,
NoMethodError (undefined method `[]' for nil:NilClass):

grape (1.0.2) lib/grape/router.rb:163:in `cascade?'
grape (1.0.2) lib/grape/router.rb:95:in `transaction'
grape (1.0.2) lib/grape/router.rb:72:in `identity'
grape (1.0.2) lib/grape/router.rb:57:in `block in call'
...
```
Response exposes the system information.

### After PR
Response would like
```
500, Internal Server Error(Invalid Response)
```
Not sure the message is proper or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1776](https://github.com/ruby-grape/grape/pull/1776): Validates response processed by exception handler - [@darren987469](https://github.com/darren987469).
 
 ### 1.1.0 (8/4/2018)
 
@@ -21,7 +22,6 @@
 * [#1758](https://github.com/ruby-grape/grape/pull/1758): Fix expanding load_path in gemspec - [@2maz](https://github.com/2maz).
 * [#1765](https://github.com/ruby-grape/grape/pull/1765): Use 415 when request body is of an unsupported media type - [@jdmurphy](https://github.com/jdmurphy).
 * [#1771](https://github.com/ruby-grape/grape/pull/1771): Fix param aliases with 'given' blocks - [@jereynolds](https://github.com/jereynolds).
-* [#1776](https://github.com/ruby-grape/grape/pull/1776): Validates response processed by exception handler - [@darren987469](https://github.com/darren987469).
 
 ### 1.0.3 (4/23/2018)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#1758](https://github.com/ruby-grape/grape/pull/1758): Fix expanding load_path in gemspec - [@2maz](https://github.com/2maz).
 * [#1765](https://github.com/ruby-grape/grape/pull/1765): Use 415 when request body is of an unsupported media type - [@jdmurphy](https://github.com/jdmurphy).
 * [#1771](https://github.com/ruby-grape/grape/pull/1771): Fix param aliases with 'given' blocks - [@jereynolds](https://github.com/jereynolds).
+* [#1776](https://github.com/ruby-grape/grape/pull/1776): Validates response processed by exception handler - [@darren987469](https://github.com/darren987469).
 
 ### 1.0.3 (4/23/2018)
 
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -73,7 +73,7 @@ module Grape
         if headers[Grape::Http::Headers::CONTENT_TYPE] == TEXT_HTML
           message = ERB::Util.html_escape(message)
         end
-        Rack::Response.new([message], status, headers).finish
+        Rack::Response.new([message], status, headers)
       end
 
       def format_message(message, backtrace, original_exception = nil)
@@ -127,7 +127,20 @@ module Grape
           handler = public_method(handler)
         end
 
-        handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        response = handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        return response if response.is_a?(Rack::Response)
+
+        # take object returned from handler as [message, status, headers]
+        # assign default value if they are nil
+        message, status, headers = response
+        message ||= default_options[:default_message]
+        status ||= default_options[:default_status]
+
+        if headers.present? && headers.is_a?(Hash)
+          error!(message, status, headers)
+        else
+          error!(message, status)
+        end
       end
     end
   end
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -128,18 +128,11 @@ module Grape
         end
 
         response = handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
-        return response if response.is_a?(Rack::Response)
 
-        # take object returned from handler as [message, status, headers]
-        # assign default value if they are nil
-        message, status, headers = response
-        message ||= default_options[:default_message]
-        status ||= default_options[:default_status]
-
-        if headers.present? && headers.is_a?(Hash)
-          error!(message, status, headers)
+        if response.is_a?(Rack::Response)
+          response
         else
-          error!(message, status)
+          run_rescue_handler(:default_rescue_handler, Grape::Exceptions::InvalidResponse.new)
         end
       end
     end
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -127,7 +127,13 @@ module Grape
           handler = public_method(handler)
         end
 
-        handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        response = handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        valid_response?(response) ? response : error!('Internal Server Error(Invalid Response)')
+      end
+
+      def valid_response?(response)
+        # Rack::Response.new(...).finish generates an array with size 3
+        response.is_a?(Array) && response.size == 3
       end
     end
   end
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -128,7 +128,7 @@ module Grape
         end
 
         response = handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
-        valid_response?(response) ? response : error!('Internal Server Error(Invalid Response)')
+        valid_response?(response) ? response : run_rescue_handler(:default_rescue_handler, error)
       end
 
       def valid_response?(response)
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -127,13 +127,7 @@ module Grape
           handler = public_method(handler)
         end
 
-        response = handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
-        valid_response?(response) ? response : run_rescue_handler(:default_rescue_handler, error)
-      end
-
-      def valid_response?(response)
-        # Rack::Response.new(...).finish generates an array with size 3
-        response.is_a?(Array) && response.size == 3
+        handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
       end
     end
   end
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -1724,30 +1724,16 @@ XML
       expect(last_response.body).to eq('Formatter Error')
     end
 
-    context 'validates response processed by exception handler' do
-      it 'calls default_rescue_handler when response is invalid' do
-        subject.rescue_from :all do
-          error!('Internal Server Error')
-          nil # invalid response caused by return nil
-        end
-        subject.get('/invalid_response') { raise 'rain!' }
-
-        expect_any_instance_of(Grape::Middleware::Error).to receive(:default_rescue_handler).and_call_original
-        get '/invalid_response'
-        expect(last_response.status).to eql 500
-        expect(last_response.body).to eq('rain!')
+    it 'can rescue exception no mater what returned by rescue_from' do
+      subject.rescue_from :all do
+        error!('Internal Server Error')
+        Grape::API.logger.error('Internal Server Error')
       end
+      subject.get('/') { raise 'rain!' }
 
-      it 'calls custom handler when response is valid' do
-        subject.rescue_from :all do
-          error!('Internal Server Error')
-        end
-        subject.get('/valid_response') { raise 'rain!' }
-
-        get '/valid_response'
-        expect(last_response.status).to eql 500
-        expect(last_response.body).to eq('Internal Server Error')
-      end
+      get '/'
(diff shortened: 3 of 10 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * [#1758](https://github.com/ruby-grape/grape/pull/1758): Fix expanding load_path in gemspec - [@2maz](https://github.com/2maz).
 * [#1765](https://github.com/ruby-grape/grape/pull/1765): Use 415 when request body is of an unsupported media type - [@jdmurphy](https://github.com/jdmurphy).
 * [#1771](https://github.com/ruby-grape/grape/pull/1771): Fix param aliases with 'given' blocks - [@jereynolds](https://github.com/jereynolds).
+* [#1776](https://github.com/ruby-grape/grape/pull/1776): Validates response processed by exception handler - [@darren987469](https://github.com/darren987469).
```
dblock: We released 1.1.0, so this belongs in the next release above.

lib/grape/middleware/error.rb
```diff
@@ -127,7 +127,13 @@ def run_rescue_handler(handler, error)
           handler = public_method(handler)
         end
 
-        handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        response = handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        valid_response?(response) ? response : error!('Internal Server Error(Invalid Response)')
+      end
+
+      def valid_response?(response)
```
dblock: Do we do anything similar elsewhere? Seems a bit fishy, possibly belongs in Rack (maybe there's a method somewhere to check that the response is valid)?

lib/grape/middleware/error.rb
```diff
@@ -127,7 +127,13 @@ def run_rescue_handler(handler, error)
           handler = public_method(handler)
         end
 
-        handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        response = handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        valid_response?(response) ? response : error!('Internal Server Error(Invalid Response)')
```
dblock: Should we re-raise the original error instead?

lib/grape/middleware/error.rb
darren987469: Sounds good. I will change the code to use default_rescue_handler to handle the original error. That means, if custom handlers failed to do their job (return valid response), default_rescue_handler will handle it.

lib/grape/middleware/error.rb
darren987469: Sounds good. I will change the code to use default_rescue_handler to handle the original error. That means, if custom handlers failed to do their job (return valid response), default_rescue_handler will handle it.

dblock:
Looking at this more carefully, do we really want this behavior?

The problem I see is that with this change we allow any kind of response from `rescue_from` and just swallow anything that doesn't look like something compatible with Rack. That seems too loose for an API framework. 

The current documentation doesn't document  what `rescue_from` should return. Maybe we want to instead document what that should be just like we document that anything returned from an API gets passed to a formatter and serialized?

dblock:
Would love some second opinion, maybe @namusyaka?

dblock:
I totally agree that this is an important problem. Lets fix it.

Does anyone actually expect to get a response from `rescue_from`? I bet most people think of it as a final exception handler. 

Are we breaking any existing specs if we start ignoring the response from `rescue_from` altogether? If the answer is no, then we should just do that and document it. There's always a backward compatible way to return data (call `error!` like most examples, call `body`, etc.).

spec/grape/api_spec.rb
```diff
@@ -1723,6 +1723,32 @@ class CustomError < Grape::Exceptions::Base; end
       expect(last_response.status).to eql 500
       expect(last_response.body).to eq('Formatter Error')
     end
+
+    context 'validates response processed by exception handler' do
+      it 'calls default_rescue_handler when response is invalid' do
+        subject.rescue_from :all do
+          error!('Internal Server Error')
+          nil # invalid response caused by return nil
```
darren987469: @dblock I’m not sure if I get what you mean.

> Are we breaking any existing specs if we start ignoring the response from rescue_from altogether? If the answer is no, then we should just do that and document it.

Do you mean if we start ignoring what returned from `rescue_from`? If so, there would be only one spec broken, which I added it here.

darren987469:
@dblock I’m not sure if I get what you mean.

> Are we breaking any existing specs if we start ignoring the response from rescue_from altogether? If the answer is no, then we should just do that and document it.

Do you mean if we start ignoring what returned from `rescue_from`? If so, I added a fail spec.

dblock:
I think "yes". I am reading the spec and it says `error!` which means I don't expect the next line to execute, maybe for spec purposes you should return `'whatever'` to demonstrate that it can be anything? What's the fix for the spec as written?

darren987469:
I add documentation to suggest developers call `error!` in the last line of `rescue_from` block. Developers call also return sting, array, hash, and Rack::Response or even nil in the `rescue_from` block. The returned object would be the error message in response. If developer return nil in the `rescue_from` block, default_message and default_status would be responded.

What do you think about this version?

dblock:
This feels too complicated to me, I feel like your previous version may have been better. 

Re-reading README we say _The `rescue_from` block must return a `Rack::Response` object, call `error!` or re-raise an exception._ So the original problem is when you don't do this you end up with a weird error, and my problem with that was the whole check for an array. What if enforced this exact spec by checking for `is_a?(Rack::Response)` instead for an array?

I totally understand that we have been going back and forth on this. I can try the above as an alternate PR if you're tired of me ;)

darren987469:
@dblock What is the behavior you image if the response from `rescue_from` is not a `Rack::Response`?

dblock:
I think as is the code works well.

If the response inside a rescue handler is garbage I expect unexpected things to happen :)

What happens when you throw an exception from a rescue handler? That can always happen and I think we should explain the behavior and make it predictable.

Try to write an UPGRADING to catch any changes users may beed to do. 

I'm  worried about those `.finish` being removed turning a Rack response into an array and what consequences that might have when grape is mounted in different places. 

@namusyaka I could use some help with another pair of eyes here

darren987469:
Changes:
1. Add behavior description when raise exception in `rescue_from`
1. Add UPGRADING

@dblock I use version 1.1.1 as next version since this is a fix in a security issue. But this PR also break grape usage. What do you think?

For your concern about the removal of `.finish` in Rack response in `Grape::Middleware::Error`, I found `Grape::Middleware::Base` response return `Rack::Response` without `.finish`[link](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/base.rb#L56-L57). Maybe it is the real object we expected in response?

dblock:
Looks good. I'll edit the language in UPGRADING a bit, but for now lets just wait for another maintainer to take a look, hopefully soon.

namusyaka:
Sorry for the late response.

@darren987469 
Is there a merit to remove `finish` with this change?
[Your referrenced block](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/base.rb#L56-L57) isn't actually used for returning HTTP response.

You removed `.finish` from `rescue_from` explicity.  If the developer look at the upgrade.md, what do developers expect for?

In many cases, the change won't be no matter. I'm not saying that this change is bad.
However, as we are concerned about that, we should explain the removal of reason or concept.

darren987469:
@namusyaka Thanks for your response. Your are right, [my referrenced block](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/base.rb#L56-L57) is not actually used for returning HTTP response. After digging more, I think it is ok to remove .finish method:

```ruby
response = Rack::Response.new('body', 200)
status, body, headers = response
```
The code above announces a `Rack::Response` object. When doing assignment, it will trigger `Rack::Response#finish` method.

Since grape uses `Rack::Head`, `Grape::Middleware::Error` and `Grape::Middleware::Formmatter` in [endpoint](https://github.com/ruby-grape/grape/blob/master/lib/grape/endpoint.rb#L280), the `Rack::Head` middleware will [do the assignment](https://github.com/rack/rack/blob/master/lib/rack/head.rb#L14) and call finish for the response of  `Grape::Middleware::Error` or `Grape::Middleware::Formmatter`. You can also find `Grape::Middleware` return a `Rack::Response` object without calling finish [here](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/formatter.rb#L47).

Current documentation says we should [return `Rack::Response` in `rescue_from` block](https://github.com/ruby-grape/grape/pull/1776/files#diff-04c6e90faac2675aa89e2176d2eec7d8L2257), but also says we can return `Rack::Response#finish` [here](https://github.com/ruby-grape/grape/pull/1776/files#diff-04c6e90faac2675aa89e2176d2eec7d8L2186). That confuses developers. This change not only fix the mismatch, but also add a validation to prevent problem stated in #1757.

dblock:
@darren987469 That makes a lot of sense, lets try to `.finish` this PR by removing all those `.finish` and seeing if specs still pass?

darren987469:
@dblock There is no more `.finish` call in grape. I wrote a spec to show `.finish` is called when doing assignment. [Rack::Response#finish source code](https://github.com/rack/rack/blob/master/lib/rack/response.rb#L62-L75)

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rack'
  gem 'rspec'
end

require 'rack'
require 'rspec/autorun'

RSpec.describe Rack::Response do
  it 'calls to_ary when doing assignment' do
    response = Rack::Response.new('body', 200)

    expect(response).to receive(:to_ary)

    status, headers, body = response
  end

  it 'call finish when doing assignment' do
    # mock finish method
    Rack::Response.class_eval do
      def finish(&block)
        [1, 2, 3]
      end
      alias to_ary finish
    end

    response = Rack::Response.new('body', 200)

    status, headers, body = response
    expect(status).to eq 1
    expect(headers).to eq 2
    expect(body).to eq 3
  end
end

# Rack::Response
#    calls to_ary when doing assignment
#    call finish when doing assignment

# Finished in 0.00745 seconds (files took 0.10826 seconds to load)
# 2 examples, 0 failures
```

dblock:
Closed via https://github.com/ruby-grape/grape/commit/c117bff7d22971675f4b34367d3a98bc31c8fc02. Thanks @darren987469!

darren987469:
@dblock @namusyaka Thank you, too!

PR #1780, opened 2018-08-14
Fix example

glaucocustodio:
The class method `decode` does not exist for `Base64`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -935,7 +935,7 @@ parameter, and the return value must match the given `type`.
 
 ```ruby
 params do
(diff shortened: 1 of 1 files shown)


dblock:
This is a good opportunity to ask the question whether we think calling `method` is OK here. It causes the service not to start because the method doesn't exist, maybe it should always be a lambda?

glaucocustodio:
I think it is good to support `method` because give devs more possibilities. It seems good to me as well the service not to start because the method does not exist, it is easier to point errors.

dblock:
Merged

PR #1783, opened 2018-08-23
Added spec to demonstrate bug with nested requirements.

darrellnash:
I believe there's a bug that was introduced sometime after Grape 0.13 (we were running a very old version). We have several nested namespaces that have requirements to allow for dots in the path parameters.

e.g. /:user_id/:attr_id, where both :user_id and :attr_id can have dots in them.

When I upgraded to the latest Grape, these stopped working for the first namespace (:user_id) - it appears it's because when you pass in a requirements hash on a route, it replaces any existing requirements on the wrapping namespace.

This spec demonstrates the expected (and previous) behavior, especially the last test case which is failing against master.

A simplified example:

With:
```ruby
namespace ':user_id', requirements: { user_id: %r{[^\/]+} } do
  get ':attr_id',    requirements: { attr_id: %r{[^\/]+} } do
    "#{params[:user_id]}/#{params[:attr_id]}"
  end
end
```

Expected:

GET `/user.with.dots/attribute.with.dots` should work.

Current behavior:

GET `/user.with.dots/attribute.with.dots` returns a 404 error.

### DIFF
--- a/spec/grape/api/routes_with_requirements_spec.rb
+++ b/spec/grape/api/routes_with_requirements_spec.rb
@@ -1,0 +1,59 @@
+require 'spec_helper'
+
+describe Grape::Endpoint do
+  subject { Class.new(Grape::API) }
+
+  def app
+    subject
+  end
+
+  context 'get' do
+    it 'routes to a namespace param with dots' do
+      subject.namespace ':ns_with_dots', requirements: { ns_with_dots: %r{[^\/]+} } do
+        get '/' do
+          params[:ns_with_dots]
+        end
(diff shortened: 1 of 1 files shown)


dblock:
Looks like a totally legit bug. Now that there's a spec, try a fix @darrellnash ! :)

The TOC thing is probably unrelated because danger-toc got automatically upgraded and we should take care of it soon (anyone feel free to contribute).

darrellnash:
Fixed by @darren987469 in #1788. Closing this PR.

PR #1785, opened 2018-09-02
Add note about using Doorkeeper for OAuth2 support

sethherr:


sethherr:
I don't believe this merits an entry in the changelog or requires adding to the readme TOC.

Please let me know if it does

dblock:
Thanks, fixed TOC in https://github.com/ruby-grape/grape/pull/1786

PR #1787, opened 2018-09-04
Add documented but not implemented feature so a user can insert a middleware at a specific point in the stack

michaellennox:
### What does this PR do?

Adds the ability for a user to insert a middleware at a specific point in the middleware stack using the `insert` method.

### Where should the reviewer start?

First commit demonstrates failing integration test case. Second fixes it. Recommended to step through.

### Any background context to be provided?

In the README of this project it is documented that a user can insert a middleware at a specific point in the stack using the `.insert` method. 

Notable passage:

```
You can add your custom middleware with use, that push the middleware onto the stack, and you can also control where the middleware is inserted using insert, insert_before and insert_after.
```

Notable example:

```ruby
class CustomOverwriter < Grape::Middleware::Base
  def after
    [200, { 'Content-Type' => 'text/plain' }, [@options[:message]]]
  end
end

class API < Grape::API
  use Overwriter
  insert_before Overwriter, CustomOverwriter, message: 'Overwritten again.'
  insert 0, CustomOverwriter, message: 'Overwrites all other middleware.'

  get '/' do
  end
end
```

This was added in https://github.com/ruby-grape/grape/commit/c2e41f72b344f478abdfdc780461963a994838d3 but it appears that the `insert` method was missed from the public API and only exposed on the internal middleware stack class.

### Any specific implementation details?

I added the CHANGELOG entry under `Fixes` as it is 'fixing' a discrepancy between the documentation and the code but unsure whether it belongs best under this or under `Features`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * Your contribution here.
 * [#1776](https://github.com/ruby-grape/grape/pull/1776): Validate response returned by the exception handler - [@darren987469](https://github.com/darren987469).
+* [#1787](https://github.com/ruby-grape/grape/pull/1787): Add documented but not implemented ability to `.insert` a middleware in the stack - [@michaellennox](https://github.com/michaellennox).
 
 ### 1.1.0 (8/4/2018)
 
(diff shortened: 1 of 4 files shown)


dblock:
Perfect, merged.

PR #1791, opened 2018-09-15
Support more options in desc block

### LINKED ISSUE #1789: is_array is not supported if we pass options as a block to `desc`
Old style desc works well for me
```ruby
desc 'Get list of user accounts',
  is_array: true,
  success: Entities::Account
```

But since passing it via hash is [deprecated](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/desc.rb#L50) I've tried to pass it via block and it didn't work for me

```ruby
desc 'Get list of user accounts' do
  success Entities::Account
  is_array true
end
```
same for this
```ruby
desc 'Get list of user accounts', is_array: true do
  success Entities::Account
end
```
How array response should be described in documentation using block ?

darren987469:
Fix #1789.

Support `summary`, `hidden`, `deprecated`, `is_array`, `nickname`, `produces`, `consumes`, `tags` options in desc block.

```ruby

desc 'Get users' do
  summary 'summary'
  hidden false
 ...
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 #### Features
 
 * Your contribution here.
-* [#1791](https://github.com/ruby-grape/grape/pull/1791): Support more options in desc block - [@darren987469](https://github.com/darren987469).
+* [#1791](https://github.com/ruby-grape/grape/pull/1791): Support `summary`, `hidden`, `deprecated`, `is_array`, `nickname`, `produces`, `consumes`, `tags` options in `desc` block - [@darren987469](https://github.com/darren987469).
 
 #### Fixes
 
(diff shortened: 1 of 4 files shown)


dblock:
Rubocop needs a `rubocop -a ; rubocop --auto-gen-config`. Lets turn this 🍏 .

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#1791](https://github.com/ruby-grape/grape/pull/1791): Support more options in desc block - [@darren987469](https://github.com/darren987469).
```
dblock: Maybe quote `desc`.

Are these just "more" options, all options, specific options? All kinds of specific options? I am nitpicking but "more" isn't telling me much.

darren987469:
Thanks for the comments. I fixed. Please take a look again, thanks!

dblock:
Merged thanks!

PR #1794, opened 2018-09-24
Add more JRuby version in the Travis matrix

guizmaii:


dblock:
This feels like a huge build matrix, is this really necessary?
 
In the past JRuby has been a huge burden on the maintainers and we have been adding/removing it often. Are you committed to working on forward compatibility with future versions of JRuby?

guizmaii:
Hi @dblock,

> In the past JRuby has been a huge burden on the maintainers and we have been adding/removing it often.

What kind of problem did you encounter ?

> Are you committed to working on forward compatibility with future versions of JRuby?

Yes, you can ping me on errors.

> This feels like a huge build matrix, is this really necessary?

The original matrix is already quite complex. 
I just copied it for 2 JRuby versions. 
Can I simplify my configs for the JRuby versions ?
Do you think that it's possible to simplify the original matrix ?


More generally, IMO, it could be a good thing that Grape, which seems to be widely used in the Ruby world,  ensure its compatibility with JRuby.

guizmaii:
> Do you think that it's possible to simplify the original matrix ?

Answering my own question: 

I think that we can because, for example, Rails 4 requires Ruby 2.4.
So, IMO, it's useless to test Rails >4 with Ruby version <2.4

WDYT ?

guizmaii:
I pushed a proposition for a new build matrix:

 - I droped Ruby 2.2
 - I added Ruby 2.5
 - Rails 3 isn't tested anymore for Ruby >=2.4
 - Rails <4 isn't tested anymore for Ruby <2.4
 - Rails <5 isn't tested anymore for Ruby <2.5

I'm not sure about the need of these:
```yaml
    - rvm: 2.5.1
      gemfile: gemfiles/multi_json.gemfile
      script:
      - bundle exec rake
    - rvm: 2.5.1
      script:
      - bundle exec rake
      - bundle exec rspec spec/integration/multi_xml
```

dblock:
Let me backtrack a bit. 

I think JRuby should be treated as another version of Ruby. We tried to have a build matrix that is not insanely large, but that has good coverage, so we have at least one build with some specific version of ruby, but not NxM. I think you should just add a single `- rvm: your favorite jruby version` to the list and we will merge it. No need to test all combinations of the [Rails x Grape] matrix again and again.

WDYT?

myxoh:
`I think that we can because, for example, Rails 4 requires Ruby 2.4.`

This is not accurate. https://guides.rubyonrails.org/4_0_release_notes.html: `Ruby 2.0 preferred; 1.9.3+ required` 
I'm sure there's plenty of organizations (ours included) that has `Rails 4` with pre-2.4. It could even be the majority as usually those who keep up-to-date with ruby keep up-to-date with rails.

Not sure how big of a deal that is, but I'm just pointing it out

PR #1795, opened 2018-09-27
Handle improper Accept headers

bschmeck:
We have observed situations where our API has received a request from a network scanner that has an improperly formatted Accept header.  In these cases, `Rack:Accept::Header.parse_media_type` returns an empty array (https://github.com/mjackson/rack-accept/blob/master/lib/rack/accept/header.rb#L42) and `subtype` is set to `nil`.

This PR checks that we actually received a subtype before checking against the vendor and version regexes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#1795](https://github.com/ruby-grape/grape/pull/1795): Fix Accept header parsing bug - [@bschmeck](https://github.com/bschmeck)
 * [#1791](https://github.com/ruby-grape/grape/pull/1791): Support `summary`, `hidden`, `deprecated`, `is_array`, `nickname`, `produces`, `consumes`, `tags` options in `desc` block - [@darren987469](https://github.com/darren987469).
 
 #### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 #### Features
 
 * Your contribution here.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#1795](https://github.com/ruby-grape/grape/pull/1795): Fix Accept header parsing bug - [@bschmeck](https://github.com/bschmeck)
```
dblock: Missing a period at the end. It could also be more specific: "Fix: vendor/subtype parsing in an invalid Accept header"

bschmeck:
Pushed a correction to the CHANGELOG, I had missed the comment from grape-bot.

`vendor?` and `version?` are private methods, which I generally don't write unit tests for, but am happy to do so if that's a common approach in grape.  Do you think they should live in `spec/grape/middleware/versioner/header_spec.rb`?  I ask because these will be the first tests for something other than `#call` in that spec.

dblock:
https://travis-ci.org/ruby-grape/grape/jobs/434730968 has a failure, I am not sure what's up with that, care to check it out?

bschmeck:
I will try to find some time to work on extracting the parsing logic and wrapping it in tests.

I'll dig into the failure, too.  It appears to be only on `rails_edge`, related to `HashWithIndifferentAccess#except` and there was just a PR merged to ActiveSupport around that.

dblock:
I am sure it's unrelated, but we do need a green build one way or another.

bschmeck:
The upstream problem in ActiveSupport was fixed and the build is now green for this PR.

dblock:
Thanks, merged.

PR #1796, opened 2018-09-29
 Fix crash when available locales are enforced but fallback locale unavailable

Morred:
- If I18n enforces available locales, in case of an exception return the translation key if the fallback locale is not supported
- Set I18n.enforce_available_locales to `true` in the spec helper because this reflects the default settings for a standard Rails app

### DIFF
--- a/spec/grape/exceptions/base_spec.rb
+++ b/spec/grape/exceptions/base_spec.rb
@@ -1,0 +1,61 @@
+require 'spec_helper'
+
+describe Grape::Exceptions::Base do
+  describe '#compose_message' do
+    subject { described_class.new.send(:compose_message, key, attributes) }
+
+    let(:key) { :invalid_formatter }
+    let(:attributes) { { klass: String, to_format: 'xml' } }
+
+    after do
+      I18n.available_locales = %i[en]
+      I18n.default_locale = :en
+    end
+
+    context 'when I18n enforces available locales' do
+      before { I18n.enforce_available_locales = true }
+
+      context 'when the fallback locale is available' do
+        before do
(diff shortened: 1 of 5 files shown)


spec/grape/exceptions/base_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    let(:key) { :invalid_formatter }
+    let(:attributes) { { klass: String, to_format: 'xml' } }
+
+    context 'when I18n enforces available locales' do
+      before { I18n.enforce_available_locales = true }
+      after do
+        I18n.available_locales = %i[de en]
+        I18n.enforce_available_locales = false
+      end
```
dblock: These probably mess up global state, so the entire spec here should have some kind of after block that restores I18n options to default rather than to `de/en`.

Morred: @dblock How about the enforce_available_locales setting? As far as I can see, the default setting for this is `true`, so should that be the default setting for all specs too?

Morred:
Not sure why the rails edge test is failing actually, seems to be an issue within ActiveSupport? 🤔

dblock:
Yeah, see https://github.com/ruby-grape/grape/pull/1795#issuecomment-425564968. First one to fix gets their PR merged ;)

spec/grape/exceptions/base_spec.rb
dblock: Sounds like it. If you want to be sure set it globally in spec_helper.rb.

dblock:
I merged this.

dblock:
Thanks @Morred !

Issue #570, opened 2014-02-04
Allow multiple mounts of the same API class


USvER:
So i have Assets and have many other models that has many Assets.
How do i expose this in API using grape?
For now i'm duplicating Assest class that is exposing Assets resource... But it's not DRY...
I have many relations... and my current project starts to look like copy-paste shit.

I want to have one class that exposes Assets resource no matter on wich level it is (as top level resource or nested). But when i mount that class twice, only last mounted resource works.

Thank you.

u2:
Intreasting!I want to make a PR,but it looks not easy for me.

elado:
Just encountered the same issue, made a spec to show the issue (can be used in a PR):

```ruby
require 'spec_helper'

describe Grape::API do
  subject do
    class Stats < Grape::API
      get :stats do
        "stats"
      end
    end

    class Main < Grape::API
      resource :one do
        mount Stats
      end

      resource :two do
        mount Stats
      end
    end

    Main
  end

  def app
    subject
  end

  it 'can mount api in multiple resources' do
    get '/one/stats'
    puts one: last_response.body # => Not Found
    get '/two/stats'
    puts two: last_response.body # => stats

    get '/one/stats'
    expect(last_response.body).to eq('stats')

    get '/two/stats'
    expect(last_response.body).to eq('stats')
  end
end
```

elado:
@lasseebert's hack works:

```ruby
require 'spec_helper'

class MountableAPI
  def self.anonymous_class
    Class.new(Grape::API).tap do |klass|
      klass.instance_eval(&@proc)
    end
  end

  def self.mounted(&block)
    @proc = block
  end
end


describe Grape::API do
  subject do
    class Stats < MountableAPI
      mounted do
        get :stats do
          "stats"
        end
      end
    end

    class Main < Grape::API
      resource :one do
        mount Stats.anonymous_class
      end

      resource :two do
        mount Stats.anonymous_class
      end
    end

    Main
  end

  def app
    subject
  end

  it 'can mount api in multiple resources' do
    get '/one/stats'
    puts one: last_response.body
    get '/two/stats'
    puts two: last_response.body

    get '/one/stats'
    expect(last_response.body).to eq('stats')

    get '/two/stats'
    expect(last_response.body).to eq('stats')
  end
end
```

dblock:
[...]

PR #1803, opened 2018-10-19
Remountable APIs, allows to re-mounting all APIs

myxoh:
Addresses this feature request: #570

A lot of times (specially when using versioning) we want to re-mount an endpoint in a different namespace, perhaps with some tiny bit of different configuration each time.

On our project, for example, we have a massive API using Grape, and while we want to introduce versioning we don't want to:
1- Have to copy & paste the whole project (nor re-write the whole project at once),
2- Have to re-write our project to use one of the proposed meta-programming approaches.

### DIFF
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -6,7 +6,8 @@ module Grape
   # should subclass this class in order to build an API.
   class API
     # Class methods that we want to call on the API rather than on the API object
-    NON_OVERRIDABLE = %I[define_singleton_method instance_variable_set inspect class is_a? ! kind_of? respond_to? const_defined? const_missing].freeze
+    NON_OVERRIDABLE = %I[define_singleton_method instance_variable_set inspect class is_a? ! kind_of?
+                         respond_to? const_defined? const_missing parent parent_name name equal? to_s parents].freeze
 
     class << self
       attr_accessor :base_instance, :instances
@@ -48,7 +49,9 @@ module Grape
       # Alleviates problems with autoloading by tring to search for the constant
       def const_missing(*args)
         if base_instance.const_defined?(*args)
-          base_instance.const_missing(*args)
+          base_instance.const_get(*args)
+        elsif parent && parent.const_defined?(*args)
+          parent.const_get(*args)
         else
           super
         end


README.md
```diff
@@ -366,6 +368,59 @@ class Twitter::API < Grape::API
 end
 ```
 
+## Remounting
+
+You can mount the same endpoints in two different locations
+
+```ruby
+class Voting::API < Grape::GrapeAPI
```
dblock: `Grape::API`

README.md
```diff
@@ -366,6 +368,59 @@ class Twitter::API < Grape::API
...
+  namespace 'votes' do
+    desc "Vote for your #{configuration[:votable]}"
+    get do
+      # Your logic
+    end
+  end
+end
+
+class Post::API < Grape::API
+  mount Voting::API, with: {votable: 'posts'}
```
dblock: Rubocop, add spaces between `{ votable: 'posts' }`.

spec/grape/remountable_api_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+describe Grape::API do
+  subject(:a_remountable_api) { Class.new(Grape::API) }
+  let(:root_api) { Class.new(Grape::API) }
+
+  def app
+    root_api
+  end
+
+  describe 'mounted RemountableAPI' do
```
dblock: We no longer have RemountableAPI.

dblock:
Other than our own fear, uncertainty and doubt, what are the downsides? 

Specs pass. We trust the specs.

I will do a thorough code review. In the meantime do consider if any parts of the remountable implementation can be simplified by moving them into the api instance class itself.

lib/grape/api_instance.rb
```diff
@@ -0,0 +1,233 @@
+require 'grape/router'
+
+module Grape
+  # The API Instance class, is the engine behind Grape::API. Each class that inherits
+  # from this will represent a different API
+  class APIInstance
```
dblock: Would this be better called `Grape::API::Instance`? Or `Grape::SomethingSomething::API` (to match how we have `Grape::DSL::API`)? Just thinking out loud, no strong feelings.

The naming could be important because if someone is having trouble with the new `Grape::API` they could just inherit from `Grape::API::Instance`, making the diff prettier ;)

lib/grape/api_instance.rb
```diff
@@ -0,0 +1,233 @@
+require 'grape/router'
+
+module Grape
+  # The API Instance class, is the engine behind Grape::API. Each class that inherits
```
dblock: .. represent a different API instance?

lib/grape/dsl/routing.rb
```diff
@@ -77,9 +77,12 @@ def do_not_route_options!
           namespace_inheritable(:do_not_route_options, true)
         end
 
-        def mount(mounts)
+        def mount(mounts, with: {})
```
dblock: Do we use named parameters elsewhere. I know we no longer support older Rubies, but if we don't maybe make this consistent as `opts = {}` and `opts[:with] || {}`?

lib/grape/dsl/routing.rb
```diff
@@ -77,9 +77,12 @@ def do_not_route_options!
...
           mounts = { mounts => '/' } unless mounts.respond_to?(:each_pair)
           mounts.each_pair do |app, path|
+            if app.respond_to?(:mount_instance)
+              return mount(app.mount_instance(configuration: with) => path)
```
dblock: This return seems odd, can you explain? What if we have more `mounts`? I think it should be a `if ... else` for whatever follows?

spec/grape/remountable_api_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'spec_helper'
```
dblock: More naming, this is just about remounting, so maybe `api_remount_spec.rb`? NBD

lib/grape/dsl/routing.rb
myxoh: So the point of this return is not to execute the rest of the code as part of a Grape::API, only as part of a Grape::API::Instance. However, I do agree that return is the wrong approach, I just need to call next.
I'm surprised no test complained. Perhaps we don't have test coverage for mounting multiple apps, path on a single call?
Can you help me write a failing spec to make sure I better understand what the use case of this `each_pair`?

Do we support something like:
```
mount(API1 => '/api_1', API2 => '/api_2'})
```
?

myxoh:
No good reason, it's mainly because this is my first contribution to Grape, and due to the potential impact I wanted to avoid modifying the core behaviour so drastically. However, if you are happy with giving it a large look, I'm happy with implementing and supporting this changes.
I'll address the naming in a little moment. Should we close the other PR?

lib/grape/api_instance.rb
myxoh: I'm happy with that naming. Regarding the solution if someone's having trouble: Should I add an entry that says something among those lines in `updating.md` ?

lib/grape/dsl/routing.rb
```diff
@@ -77,9 +77,12 @@ def do_not_route_options!
           namespace_inheritable(:do_not_route_options, true)
         end
 
-        def mount(mounts)
+        def mount(mounts, opts = {})
           mounts = { mounts => '/' } unless mounts.respond_to?(:each_pair)
           mounts.each_pair do |app, path|
+            if app.respond_to?(:mount_instance)
+              return mount(app.mount_instance(configuration: opts[:with] || {}) => path)
```
myxoh: The return issue is still not addressed, I still need a failing spec to see whether `next` is appropiate, or whether this is better handled by a refactor

myxoh:
Also waiting on what you think we should do on the updating (as technically no change is required, but might be helpful to have information explaining how to fallback)

lib/grape/dsl/routing.rb
dblock: We're definitely missing tests.

I didn't know that we support this kind of mounting, but it looks like! Thanks for looking into it.

dblock:
> No good reason, it's mainly because this is my first contribution to Grape, and due to the potential impact I wanted to avoid modifying the core behaviour so drastically. However, if you are happy with giving it a large look, I'm happy with implementing and supporting this changes.

What could go wrong :)

> I'll address the naming in a little moment. Should we close the other PR?

Yes, go for it.

dblock:
> Also waiting on what you think we should do on the updating (as technically no change is required, but might be helpful to have information explaining how to fallback)

You should write a paragraph in UPGRADING.

myxoh:
> > Also waiting on what you think we should do on the updating (as technically no change is required, but might be helpful to have information explaining how to fallback)
> 
> You should write a paragraph in UPGRADING.

Done. The only point is that I've added the upgrading to `upgrading to >= 1.1.1`, not sure whether this PR should go in a point release. Of course that's completely your decision 👍 .

Let me know anything else you need to get this over the line 😄

myxoh:
By trying it out on a big project, I've faced a bunch of breaking issues with no easy solution
- Grape swagger seemed completely broken
- Referencing a constant without the namespace is broken.

I'm back to the drawing board, and adding specs to represent these issues later today

myxoh:
So. After a lot of investigation, issues were rooted in 2 main causes.
1- Autoloading in rails is ffed up. I would probably put a warning in the `upgrading.md`, but there's no easy solution. There was not a problem with the namespaces as I feared.
2- Grape-Swagger is opening and closing `Grape::API` but it should be opening and closing `Grape::API::Instance` (And also modifying the reference in `GrapeSwagger#create_documentation_class` to Grape::API::Instance)

After fixing these 2 things, I've got my project back working.

I'm waiting on your thoughts and what's next then

myxoh:
Bumped up the version, and opened this PR on grape-swagger: https://github.com/ruby-grape/grape-swagger/pull/717 that had green specs locally (when changing the gemfile to point to the right branch)

dblock:
LGTM, I'll wait on someone else to comment (@namusyaka if you have a moment?) otherwise will merge in a few.

namusyaka:
Sorry, I overlooked this PR.
This feature looks really promising, thanks so much @myxoh !

I'll review the code & design this weekend.

myxoh:
Hi @dblock  I've wrote you an e-mail from myxoh10[at]gmail[dot]com

PR #1810, opened 2018-10-30
Fix alias on dependent param bug

### LINKED ISSUE #1808: Params 'as' option not accessible within 'given' block proc
It seems like we cannot access to the value inside the proc for an alias param

```ruby
requires :type,   as: :action
given type: ->(val) { ActionService::ACTIONS.include?(val) } do #val not present inside the proc
  requires :comment, type: String
end
```

darren987469:
Fixes #1808 

> It seems like we cannot access to the value inside the proc for an alias param
> 
> ```ruby
> requires :type, as: :action
> given type: ->(val) { ActionService::ACTIONS.include?(val) } do #val not present inside the proc
>   requires :comment, type: String
> end
> ```

The root cause is `as: :action` alias would change `params` from `{ type: 'some type' }` to `{ action: 'some type' }` in as validator.
https://github.com/ruby-grape/grape/blob/3bb835a747f8ef2a9542ef2e6b66c795f377dd87/lib/grape/validations/validators/as.rb#L9-L12
When checking `->(val) { ActionService::ACTIONS.include?(val) }` in the following snippet(line 61), dependency_key is still `:type`, so params[:type] is nil because the params is changed in previous validator. 
https://github.com/ruby-grape/grape/blob/3bb835a747f8ef2a9542ef2e6b66c795f377dd87/lib/grape/validations/params_scope.rb#L57-L65

The idea to fix is let attribute name in `given` the same as the aliased one. That means the usage now becomes:

```ruby
reuiqres :type, as: :action
given action: ->(val) { ... }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 * [#1776](https://github.com/ruby-grape/grape/pull/1776): Validate response returned by the exception handler - [@darren987469](https://github.com/darren987469).
 * [#1787](https://github.com/ruby-grape/grape/pull/1787): Add documented but not implemented ability to `.insert` a middleware in the stack - [@michaellennox](https://github.com/michaellennox).
 * [#1788](https://github.com/ruby-grape/grape/pull/1788): Fix route requirements bug - [@darren987469](https://github.com/darren987469), [@darrellnash](https://github.com/darrellnash).
-* [#1810](https://github.com/ruby-grape/grape/pull/1810): Fix alias on dependent param bug - [@darren987469](https://github.com/darren987469).
+* [#1810](https://github.com/ruby-grape/grape/pull/1810): Fix support in `given` for aliased params - [@darren987469](https://github.com/darren987469).
 
 ### 1.1.0 (8/4/2018)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * [#1776](https://github.com/ruby-grape/grape/pull/1776): Validate response returned by the exception handler - [@darren987469](https://github.com/darren987469).
 * [#1787](https://github.com/ruby-grape/grape/pull/1787): Add documented but not implemented ability to `.insert` a middleware in the stack - [@michaellennox](https://github.com/michaellennox).
 * [#1788](https://github.com/ruby-grape/grape/pull/1788): Fix route requirements bug - [@darren987469](https://github.com/darren987469), [@darrellnash](https://github.com/darrellnash).
+* [#1810](https://github.com/ruby-grape/grape/pull/1810): Fix alias on dependent param bug - [@darren987469](https://github.com/darren987469).
 
 ### 1.1.0 (8/4/2018)
 
--- a/spec/grape/validations/params_scope_spec.rb
+++ b/spec/grape/validations/params_scope_spec.rb
@@ -497,6 +497,23 @@ describe Grape::Validations::ParamsScope do
       expect(body.keys).to_not include('b')
     end
 
+    it 'allows aliasing of dependent on parameter' do
+      subject.params do
+        optional :a, as: :b
+        given b: ->(val) { val == 'x' } do
+          requires :c
+        end
+      end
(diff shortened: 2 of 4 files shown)


dblock:
Thanks! So what happens if I was using the unaliased param before, that is `given a: ` in the spec below instead of `given b:`? 

```
subject.params do
        optional :a, as: :b
        given b: ->(val) { val == 'x' } do
          requires :c
        end
      end
```

I am going to guess that without this fix `given a:` works, but `given b:` fails, and vice-versa. Could you please add a spec for this?

If that's the case we should either support both or clearly state what happens here in UPGRADING.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * [#1776](https://github.com/ruby-grape/grape/pull/1776): Validate response returned by the exception handler - [@darren987469](https://github.com/darren987469).
 * [#1787](https://github.com/ruby-grape/grape/pull/1787): Add documented but not implemented ability to `.insert` a middleware in the stack - [@michaellennox](https://github.com/michaellennox).
 * [#1788](https://github.com/ruby-grape/grape/pull/1788): Fix route requirements bug - [@darren987469](https://github.com/darren987469), [@darrellnash](https://github.com/darrellnash).
+* [#1810](https://github.com/ruby-grape/grape/pull/1810): Fix alias on dependent param bug - [@darren987469](https://github.com/darren987469).
```
dblock: Maybe say something more explicit? Eg. "Fix support in `given` for aliased params"?

darren987469: Yeah, it is more clear.

darren987469:
Let me describe `given a:`, `given b:` cases with or without this fix.

Without this fix case `given a:`
```ruby
subject.params do
  optional :a, as: :b                # 'as: :b' change params form { a: 'somthing' } to { b: 'something' }
  given a: ->(val) { val == 'x' } do # val = params[:a] = nil, condition doesn't meet
    requires :c
  end
end
subject.get('/') { declared(params) }

get '/', a: 'x'
expect(last_response.status).to eq 400              # fail, since given condition doesn't meet 
expect(last_response.body).to eq 'c is missing' 
```
Without this fix case `given b:`
```ruby
subject.params do
  optional :a, as: :b                           
  given b: ->(val) { val == 'x' } do # raise Grape::Exceptions::UnknownParameter. @declared_params in grape/lib/grape/dsl/parameters.rb is [:a], which doesn't contain :b. That why this fix try to solve.
    requires :c
  end
end
```

With this fix case `given a:`
```ruby
subject.params do
  optional :a, as: :b                # 'as: :b' change params form { a: 'somthing' } to { b: 'something' }
  given a: ->(val) { val == 'x' } do # raise Grape::Exceptions::UnknownParameter.  @declared_params in grape/lib/grape/dsl/parameters.rb is now [:b], which doesn't contain :a. 
    requires :c
  end
end
```

With this fix case `given b:`
```ruby
subject.params do
  optional :a, as: :b            # 'as: :b' change params form { a: 'somthing' } to { b: 'something' }               
  given b: ->(val) { val == 'x' } do # val = params[:b] 
    requires :c
  end
end

get '/', a: 'x'
expect(last_response.status).to eq 400              # success, since given condition meets
expect(last_response.body).to eq 'c is missing' 
```

I think it make sense to force to use `given b:`. Because we change params from `{ a: 'somthing' }` to `{ b: 'something' }`.

spec/grape/validations/params_scope_spec.rb
```diff
@@ -497,6 +497,34 @@ def initialize(value)
...
+
+      get '/', a: 'y'
+      expect(last_response.status).to eq 200
+    end
+
+    it 'raises an error if the dependent parameter is not the aliased one' do
+      expect do
+        subject.params do
+          optional :a, as: :b
+          given :a do
```
darren987469: I add a spec to show with this fix, `given :a`(not the aliased param `:b`) case.

dblock:
Perfect, thanks for the specs.

PR #1804, opened 2018-10-22
Added test for failing nested given validation

andreacfm:
Using nested given validations looks unsupported. This is just a simple test case to show the issue.

PR #1811, opened 2018-11-01
Support nested dependent parameters

### LINKED ISSUE #1800: Unsupported nested given blocks
Example:

```ruby 
optional :filters, type: Array, default: [] do
  requires :name, type: String, values: [.....]
  requires :operator, values: [......]
  given name: ->(val) {val == 'name'} do
    requires :operator, values: ['a', 'b']
    given operator: -> (val) {val == 'b'} do
      requires :values, type: Array[String]
    end
    given operator: -> (val) { val == 'b' } do
      requires :value, type: Numeric
    end
  end
end
```

This fails because the params :operator seems not defined when passed to the nested given block. Is this use case supported?

darren987469:
Fixes #1800. Continue of #1804.

`given` will check whether attribute is declared in its scope.
If not, it [raises an error](https://github.com/ruby-grape/grape/blob/b8b85af12ad4860b6ee4ae2074f864a849400209/lib/grape/dsl/parameters.rb#L202-L208).

```ruby
optional :a
given a: ->(val) { val == 'a' } do
  optional :b
  given b: ->(val) { val == 'b' } do
    requires :c
  end
end
```

In the example, [`optional :a` creates a scope](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/parameters.rb#L160). [`given a: ...` creates
another scope](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/parameters.rb#L207), which parent scope is the one created by `optional :a`,
and [attributes `:b` declared in `optional :b` is pushed to parent scope](https://github.com/ruby-grape/grape/blob/b8b85af12ad4860b6ee4ae2074f864a849400209/lib/grape/validations/params_scope.rb#L117-L129).

The bug here is `given b: ...` cannot find `:b` in its scope. Since the
attribute `:b` is pushed to its parent scope. So, [fix the code](https://github.com/ruby-grape/grape/pull/1811/files#diff-720e18632121176ab27887e38663f186R216) to
check attribute in its parent scope if it has parent scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * [#1787](https://github.com/ruby-grape/grape/pull/1787): Add documented but not implemented ability to `.insert` a middleware in the stack - [@michaellennox](https://github.com/michaellennox).
 * [#1788](https://github.com/ruby-grape/grape/pull/1788): Fix route requirements bug - [@darren987469](https://github.com/darren987469), [@darrellnash](https://github.com/darrellnash).
 * [#1810](https://github.com/ruby-grape/grape/pull/1810): Fix support in `given` for aliased params - [@darren987469](https://github.com/darren987469).
+* [#1811](https://github.com/ruby-grape/grape/pull/1811): Support nested dependent parameters - [@darren987469](https://github.com/darren987469), [@andreacfm](https://github.com/andreacfm).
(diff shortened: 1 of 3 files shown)


dblock:
🚀 thanks!

PR #1812, opened 2018-11-02
Add spec for default value in given block

darren987469:
Add spec for #1681. The issue seems already fixed.

### DIFF
--- a/spec/grape/validations/params_scope_spec.rb
+++ b/spec/grape/validations/params_scope_spec.rb
@@ -663,6 +663,32 @@ describe Grape::Validations::ParamsScope do
     end
   end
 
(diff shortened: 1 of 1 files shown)


dblock:
There's some dependency/ruby version changes that failed the build here. Appreciate if you could please try to fix that? Upgrading those rubies seems to be the right way to go. Thanks.

darren987469:
I open another PR to add ruby 2.5 support and drop 2.2 support (#1813 ). After 1813 merged, I will rebase it and CI should pass.

dblock:
Merged it, thanks @darren987469

dblock:
Merged, thank you.

Issue #1821, opened 2018-11-19
Validation Error is not raised when optional Hash type parameter is received String type value and exactly_one_of be used


woshidan:
Gem versions:

- Rack 2.0.6
- Grape 1.1.0

I defined the API as below and expected to return a 400 error when the memo parameter is String type value instead of a Hash type value, but 500 error returned.

```ruby
require 'grape'

module TestApp
  class API < Grape::API

    params do
      requires :schedule, type: Hash do
        requires :participants, type: Array[String]
      end

      optional :memo, type: Hash do
        optional :text,   type: String
        optional :image_url, type: String

        optional :custom_body, type: Hash, coerce_with: JSON

        exactly_one_of :text, :custom_body
        exactly_one_of :image_url, :custom_body
      end
    end

    post "/" do
      # do something
    end
  end
end

run TestApp::API
```

Here is examples of test request and the responses.

### CASE 1

```sh
curl localhost:9292/ --request POST \
--header "Content-Type: application/json" \
--header "Accept: application/json" \
--data-binary '{ "schedule": { "participants": [ "Tom", "Bob"]}, "memo" : { "text": "HOGEHOGE", "image_url": "http://example.com/images/001.jpg" } }'
```

```
status code: 201, response: null
```

```sh
curl localhost:9292/ --request POST \
--header "Content-Type: application/json" \
--header "Accept: application/json" \
--data-binary '{ "schedule": { "participants": [ "Tom", "Bob"]}, "memo" : {"custom_body": "{ \"xxx\": \"yyy\" }" } }'
```

```
status code: 201, response: null
```

### CASE 2

```sh
curl localhost:9292/ --request POST \
--header "Content-Type: application/json" \
--data-binary '{ "schedule": { "participants": [ "Tom", "Bob"]}, "memo" : {"custom_body": "HOGE"} }'
```

```
status code: 400, response: {"error":"memo[custom_body] is invalid"}
```

### CASE 3

```sh
curl localhost:9292/ --request POST \
--header "Content-Type: application/json" \
--header "Accept: application/json" \
--data-binary '{ "schedule": { "participants": [ "Tom", "Bob"]}, "memo" : { "text": "HOGEHOGE", "image_url": "http://example.com/images/001.jpg", "custom_body": "{ \"xxx\": \"yyy\" }" } }'
```

```
status code: 400, response: {"error":"text, custom_body are mutually exclusive, image_url, custom_body are mutually exclusive"}
```

### CASE 4

```sh
[...]

PR #1822, opened 2018-11-21
Raise validation error when optional Hash type parameter is received String type value and exactly_one_of be used

woshidan:
Add spec and fixes https://github.com/ruby-grape/grape/issues/1821 .

When String type was given instead of Hash type value, the item in scoped_params becomes String type and will be crashed due to not having any? method.


dblock:
Looks good, needs a `rubocop -a ; rubocop --auto-gen-config` and a CHANGELOG entry. Thanks.

woshidan:
@dblock I've applied a rubocop lint and add a entry to CHANGELOG.md.

dblock:
Great branch name! ;)

Issue #1815, opened 2018-11-08
Autoloading issues after the introduction of Remountable API?


andreacfm:
The introduction of this commit https://github.com/ruby-grape/grape/commit/cf14fc79d0100b988fb0d5ccd9182e3135e9ae48 completely broke one of my api's.
I get any possible erro from `guess_for_anonymous` when autoloading modules to missing methods from included module.

My API is mounted as:

```
# api_base.rb
class ApiBase < Grape::API
   mount InternalApi::Base
end

# internal_api/base.rb
module InternalApi
    class Base < Grape::API
        mount InternalApi::V1::Base
        mount InternalApi::V2::Base
....


# Then InternalApi::V1::Base and InternalApi::V2::Base mount the final api files
```

One of the error I get (undefined method paginate) from:

```
module InternalApi
    module V1 
      class Base
          version :v2
           include Grape::Kaminari
           paginate per_page: 30, max_per_page: 200  
```

Is this expected and does exists an upgrade guide?

dblock:
Thanks for trying HEAD out @andreacfm, cc: @myxoh, related to #1803.

It would be helpful to get a spec that can reproduce this, let's fix it or document a workaround before release.

andreacfm:
@dblock here a spec to show the failing in including modules.
https://github.com/ruby-grape/grape/pull/1816

myxoh:
Hi, I'm going to be looking at the issues caused by the PR now. Sorry, it's been a very heptic week for me

rromanchuk:
I haven't spent any time investigation besides just reverting, just sharing what I know

```
/dependencies.rb:203:in `guess_for_anonymous': SlackBot cannot be autoloaded from an anonymous class or module (NameError)
```
in application.rb
`config.autoload_paths << "#{Rails.root}/lib"`

Location of module
`lib/slack_bot.rb`

Somewhere inside `Grape::API`

`include SlackBot`

myxoh:
@rromanchuk I couldn't reproduce that on Rails 5.2. I'll try on 4.2. Perhaps you could answer a couple of questions for me?
Which Rails version are you using?
How are you loading the API?
Is the errors happening on load time or run time?

myxoh:
@andreacfm I also was't able to reproduce autoloading issues adding `helpers V1Helpers` where `V1Helpers` is a module of helpers. It seemed to work just fine

myxoh:
[...]

PR #1823, opened 2018-11-23
Fixes the issue with the autoloader #1815

myxoh:
In order to properly use the autoloader, we do not want to delegate `anonymous?` to the instance.

This should fix most remaining issues with the autoloader in master: https://github.com/ruby-grape/grape/issues/1815

Sorry for the lack of specs. It's really hard to reproduce without setting up a new project.

This was tested using: https://github.com/myxoh/testing-grape/commits/master
(If you run rspec it crashes, when you add this commit, it stops crashing)


dblock:
Merging, thanks. Consider some spec options:

- You could try writing a spec into spec/integration, these use their own Gemfile's.
- You could write a spec that says "does not delegate anonymous? to the instance" and add a link to the issue, so that someone doesn't accidentally remove it.

Issue #1825, opened 2018-11-26
Inheriting from Grape::API yields anonymous class


delner:
With the release of version 1.2.0, the following class definition yields an anonymous class:

```
> class Foo < ::Grape::API; end
=> nil
> Foo
=> #<Class:0x0000560d0a089978>
```

This breaks some code that depends on referencing `Foo` as a constant to return a well-known type, not an anonymous class. A common case would be calling `Foo.name` which previously returned `'Foo'` would now return `nil`.

A workaround for this, [as alluded to by the release notes](https://github.com/ruby-grape/grape/blob/3cf89cf1f253a36c73d4b9e8a6b2aca673f0b915/UPGRADING.md), is to convert all such instances of `::Grape::API` to `::Grape::API::Instance` instead. However this doesn't really address the issue in cases where inheriting from `Grape::API` is still considered acceptable practice.

dblock:
Good catch, I am not sure what the best way to fix it is, but let's make sure to make some specs, cc: @myxoh.

myxoh:
Could you tell me which version of Ruby you are using?

On MRI Ruby 2.3.3

![image](https://user-images.githubusercontent.com/5561861/49079687-2f92af00-f239-11e8-9692-e48874b94506.png)

On MRI Ruby 2.5.1

![image](https://user-images.githubusercontent.com/5561861/49079737-55b84f00-f239-11e8-8ae6-2db07d728290.png)

So

delner:
I was having trouble reliably replicating this the same way you did (irb), using Ruby 2.3.7. However, this "anonymous class with no name" problem can be replicated with the following Minitest, when working with the API class from an ActiveSupport notification:

```
require 'minitest'
require 'minitest/autorun'
require 'rack/test'
require 'grape'

class TestGrapeAPI < Grape::API
  namespace :base do
    desc 'Returns a success message'
    get :success do
      'OK'
    end
  end
end

class GrapeAPITest < MiniTest::Test
  include Rack::Test::Methods

  def setup
    ::ActiveSupport::Notifications.subscribe('endpoint_run.grape', &method(:callback))
  end

  def app
    TestGrapeAPI
  end

  def test_api_name
    get '/base/success'
    assert last_response.ok?
    assert_equal('OK', last_response.body)
  end

  def callback(name, start, finish, id, payload)
    api = payload[:endpoint].options[:for]
    # Expect the API name to not be nil
    refute_nil(api.name)
  end
end
```

[...]

PR #1826, opened 2018-11-27
Exposes the name of the API  via `to_s` (Addresses #1825)

myxoh:
Addresses: https://github.com/ruby-grape/grape/issues/1825

It does so by exposing the base's `to_s`

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -5,6 +5,8 @@ Upgrading Grape
 
 #### Changes in the Grape::API class
 
+##### Patching the class
+
 In an effort to make APIs re-mountable, The class `Grape::API` no longer refers to an API instance,
 rather, what used to be `Grape::API` is now `Grape::API::Instance` and `Grape::API` was replaced
 with a class that can contain several instances of `Grape::API`.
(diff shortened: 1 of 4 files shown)


delner (approved):
Changes look good to me! :+1:

UPGRADING.md
```diff
@@ -29,6 +31,23 @@ class Grape::API::Instance
...
+- Generally: you can access the named class from the instance calling the getter `base`.
+- In particular: If you need the `name`, you can use `.to_s`, which will fallback on: `base`.`name`
+
+**Deprecated**
+```ruby
+  payload[:endpoint].options[:for].name
+```
+**New**
+```ruby
+  payload[:endpoint].options[:for].to_s # or `.base.name`
```
dblock: This will be introduced in 2.1, I think for 1.2.0 it should say `base.name` here, but in 1.2.1 we can recommend `to_s`? Or does `to_s` give you the name before 1.2.0?

spec/grape/named_api_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'spec_helper'
+
+describe 'A named API' do
+  NamedAPI = Class.new(Grape::API) do
```
dblock: put this in a `before` and `remove_const` in an `after` to keep the spec clean.

UPGRADING.md
delner: Prior to 1.2.0, `to_s` on the API object returned the class name.

dblock:
LGTM, merged

PR #1831, opened 2018-12-03
Support security in desc block

fotos:
Adds support for `security` in `desc` block.

Proposing the change since [options hash parameters are marked as deprecated](https://github.com/ruby-grape/grape/blob/v1.2.1/lib/grape/dsl/desc.rb#L58) and the bug in ruby-grape/grape-swagger#706 prevents using both (options are not merged into the block).

```Ruby
desc `Create User` do
  security [
    # security schemes
  ]
  # …
end
````

Includes some minor formatting changes in the spec (in separate commits) and I didn't update the `README.md` since it [already points to `grape-swagger` for more info](https://github.com/ruby-grape/grape#describing-methods).

Related to #1791 and (slightly) to ruby-grape/grape-swagger#729.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1831](https://github.com/ruby-grape/grape/pull/1831): Support security in desc block - [@fotos](https://github.com/fotos).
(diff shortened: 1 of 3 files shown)


dblock:
Looks good, merged.

PR #1832, opened 2018-12-05
Support body_name in desc block 

fotos:
Adds support for `body_name` in `desc` block.

Already [described in `grape-swagger`](https://github.com/ruby-grape/grape-swagger#overriding-the-name-of-the-body-parameter-).

Related to #1791.


fotos:
I made a similar thought but then again this should be a one-off migration from the (deprecated) options hash to the block syntax. I don't know how often these options change and whether making them configurable is justified.

Issue #1775, opened 2018-08-02
Grape::Extensions::Hashie::Mash::ParamBuilder can't be set as a default globally in 1 place


myxoh:
Context: I'm upgrading a massive rails 3.2 API to rails 4. I'm late to the party to realise that Grape changed the default param builder.

Problem: The proposed solution for backwards compatibility: 
`include Grape::Extensions::Hashie::Mash::ParamBuilder` can't be applied globally as far as I could investigate. 
The reason for concern is that we have over 200+ API endpoints, and I'd rather not have to modify the 200 files just for the sake of maintaining backwards compatibility.

Is there anyway of setting this builder as the default builder for Grape? Just re-opening Grape::API or building a custom parent class that included `Grape::Extensions::Hashie::Mash::ParamBuilder` didn't cut it for me.
Using the 
`def self.inherited(base)`
 `  base.include Grape::Extensions::Hashie::Mash::ParamBuilder`
`end`
didn't help either, as in some of our API's the routes hadn't been loaded by then causing some trouble.

Perhaps you've though of an approach, or a new approach must be made?

Thanks in advance for the help

myxoh:
Thank you @dblock that's awesome!

dblock:
reopening since it's still a valid feature

myxoh:
I think this can be closed now. Apologise for the original confusion when creating this issue

PR #1833, opened 2018-12-05
Allows to set a global ParamBuilder

myxoh:
This is related to the issue: https://github.com/ruby-grape/grape/issues/1775

Creates a new module for global configuration (I checked the documentation and couldn't find one available), and added the setting for a global param builder


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Your contribution here.
 * [#1832](https://github.com/ruby-grape/grape/pull/1832): Support `body_name` in `desc` block - [@fotos](https://github.com/fotos).
 * [#1831](https://github.com/ruby-grape/grape/pull/1831): Support `security` in `desc` block - [@fotos](https://github.com/fotos).
+* [#1831](https://github.com/ruby-grape/grape/pull/1833): Allows to set the `ParamBuilder` globally - [@myxoh](https://github.com/myxoh).
```
darren987469: The PR is 1833. Maybe order contributions in desc order?

myxoh: You are right, my bad

dblock:
If we want to add global configuration to Grape lets adopt the Ruby-esque standard like all the other gems. I should be able to do

```ruby
Grape.configure do |config|
   config.x = 
end
```

Take a look at https://github.com/dblock/strava-ruby-client/blob/master/lib/strava/api/config.rb for an example.

myxoh:
> If we want to add global configuration to Grape lets adopt the Ruby-esque standard like all the other gems. I should be able to do
> 
> ```ruby
> Grape.configure do |config|
>    config.x = 
> end
> ```
> 
> Take a look at https://github.com/dblock/strava-ruby-client/blob/master/lib/strava/api/config.rb for an example.

Not particularly a fan of it (which is why it wasn't my first choice), but should be fairly easy to implement, and I see your point 👍 
I'll do so between later today, and early tomorrow.
Other than that, any point you want to add?

dblock:
Ruby is lovely because it's often convention over configuration, and predictable imperfect way of doing things consistently. Long live Ruby :)

Code looks good otherwise so far. I'll re-review thoroughly when you have the config stuff sorted out.

Thanks!

myxoh:
@dblock I've made those changes

README.md
```diff
@@ -541,6 +542,29 @@ end
...
+
+Grape counts with a module `Grape::Config` for some basic configuration at load time.
+Currently the configurable settings are:
+
+* `param_builder`: Sets the default [Parameter Builder](#parameters), by default: `Grape::Extensions::ActiveSupport::HashWithIndifferentAccess::ParamBuilder`
+
+To change a setting value make sure that at some point on load time the code the following code runs
+
+```ruby
+Grape::Config.configure do |config|
```
dblock: It should be `Grape.configure do |config|`, because we're not configuring the configuration, we're configuring Grape.

README.md
```diff
@@ -618,6 +642,8 @@ params do
 end
 ```
 
+Or globally with the [Configuration](#configuration) `Grape::Config.param_builder`
```
dblock: This should be `Grape.config.param_builder`.

dblock (changes requested):
See notes.

spec/grape/request_spec.rb
```diff
@@ -62,6 +62,24 @@ module Grape
       end
     end
 
+    describe 'when the param_builder is set to Hashie' do
+      before do
```
dblock: I think this is overcomplicating things and stubs it to a point where it can actually be buggy and we wouldn't notice. Always better to do the real thing! Try this:

```ruby
before do
   Grape.configure do |config|
     config.praram_builder = Grape::Extensions::Hashie::Mash::ParamBuilder
   end
end

after do
  Grape::Config.reset
end
```

I actually think the `reset` could belong in spec/support/config.rb or something like that and be invoked every spec to make sure any global configuration is cleared for future use.

spec/grape/config_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'spec_helper'
+
+describe '.configure' do
+  let!(:grape_double) { Module.new { extend Grape::Config } }
```
dblock: Get rid of all the doubles and stubbing, just do the real thing.

README.md
```diff
@@ -541,6 +542,29 @@ end
 
 [grape-swagger]: https://github.com/ruby-grape/grape-swagger
 
+## Configuration
+
+Grape counts with a module `Grape.configure` for some basic configuration at load time.
```
dblock: Counts is probably the wrong word. I tend to keep this as short as possible. For example:

_Use `Grape.configure` to setup global settings at load time._

README.md
```diff
@@ -618,6 +642,8 @@ params do
 end
 ```
 
+Or globally with the [Configuration](#configuration) `Grape.configure.param_builder`
```
dblock: Missing a period.

README.md
```diff
@@ -541,6 +542,29 @@ end
 
 [grape-swagger]: https://github.com/ruby-grape/grape-swagger
 
+## Configuration
+
+Grape counts with a module `Grape.configure` for some basic configuration at load time.
+Currently the configurable settings are:
+
+* `param_builder`: Sets the default [Parameter Builder](#parameters), by default: `Grape::Extensions::ActiveSupport::HashWithIndifferentAccess::ParamBuilder`
```
dblock: Sets the Parameter Builder, defaults to `Grape::Extensions::ActiveSupport::HashWithIndifferentAccess::ParamBuilder`

dblock (approved):
Looks good. I made some minor suggestions for language in README. Feel free to make/not make them and merge (I would squash since the change ended up being quite minimal).

dblock:
We do need a green CI. Didn't look at failures.

myxoh:
Opened issue for this: https://github.com/ruby-grape/grape/issues/1839 It seems to be related with coveralls which is no longer maintained

dblock:
👍

Issue #1835, opened 2018-12-06
Memory usage bump after upgrading to the latest version


filipebarros:
After upgrading to the latest version I've noticed the memory consumption keeps going up so I created a dummy application to check if the same happened with the previous version:

```
require 'grape'

class API < Grape::API
  get :heartbeat do
    'OK'
  end
end

run API
```

Here's the memory usage with both `1.2.1` version and `1.1.0` at the start of the process and after 1000 and 10000 calls:

| | 1.2.1 - 1000 | 1.1.0 - 1000 | 1.2.1 - 10000 | 1.1.0 - 10000 |
| --- | --- | --- | --- | --- |
| Start | 25.4Mb | 26.4Mb | 25.2Mb | 25.1Mb | 
| Finish | 45.0Mb | 29.0Mb | 198.5Mb | 29.7Mb |

Here's the script I used to run the API calls.

```
for run in {1..1000} do
  curl -s "http://localhost:9292/heartbeat"
done
```

dblock:
Let's debug this more thoroughly? 

Check out https://github.com/ruby-grape/grape/issues/665 where we debugged a memory leak successfully for next steps.

cc: @myxoh

nijikon:
We are experiencing this as well, reverting to `1.1.0` fixed this issue for us.

myxoh:
Interesting, I wouldn't have thought the changes I've implemented should have impacted the memory consumption other than ever so slightly at load time. I'll take a look into this somewhere between late today and early tomorrow

myxoh:
oh wow. I think I've found the issue. This is unique to mounting the API directly on Rack I believe. `call` and `call!` are being called on the base instance, and hence continously added ot the `setup` of the application.
I already have a fix in place that passes all specs. I'll be creating a PR in a minute now.

Thanks for realizing this!

myxoh:
As a colour note, on my local testing this fix puts the memory footprint of `1.2.2` in the same values as the `1.1.0` .

dblock:
@nijikon @filipebarros Please do try the code from https://github.com/ruby-grape/grape/pull/1836 and confirm whether the problem is fixed?

myxoh:
Also, benchmarking the time for 1,000 calls gives me similar results for both `1.1.0` and `master`

dblock:
Good. I say cut a release.

filipebarros:
[...]

PR #1836, opened 2018-12-06
Addresses memory issues related to new multi-instance approach

myxoh:
As part of the new multi-instance approach, class calls to the API are being intercepted, and stored so that they can be call as part of the setup to all instances. 
However, when you are running the API in Rack, the method `call` is not part of the configuration, but an action. We do NOT want to store every time the method `call` was called for security and memory management issues:

Addresses: https://github.com/ruby-grape/grape/issues/1835

### DIFF
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -6,7 +6,7 @@ module Grape
   # should subclass this class in order to build an API.
   class API
     # Class methods that we want to call on the API rather than on the API object
-    NON_OVERRIDABLE = Class.new.methods.freeze
+    NON_OVERRIDABLE = Class.new.methods.freeze + %i[call call!]
 
     class << self
       attr_accessor :base_instance, :instances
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1836](https://github.com/ruby-grape/grape/pull/1836): Addresses memory issue due to storing in memory and not releasing `call` method calls - [@myxoh](https://github.com/myxoh).
```
dblock: Make this `Fix: memory leak not releasing ...`.

myxoh:
Fxed

lib/grape/api.rb
```diff
@@ -6,7 +6,7 @@ module Grape
   # should subclass this class in order to build an API.
   class API
     # Class methods that we want to call on the API rather than on the API object
-    NON_OVERRIDABLE = Class.new.methods.freeze
+    NON_OVERRIDABLE = Class.new.methods.freeze + %i[call call!]
```
nijikon: I think that this should be `(Class.new.methods + %i[call call!]).freeze`

dblock: I was bored, https://github.com/ruby-grape/grape/pull/1837/files

nijikon: Nice!

PR #1840, opened 2018-12-10
Use coveralls-reborn for rails-edge.

dblock:
Attempt at fixing CI.


myxoh:
Thank you

LeFnord:
@dblock let you run coberalls with that change?

Issue #1841, opened 2018-12-12
File type validator not checking the type of :tempfile


Nyangawa:
Consider following parameter declaration
```ruby
      params do
        requires :file, type: File, desc: 'The file to be uploaded'
      end
```

Currently, the File validator is only validating the existence of key :tempfile
```ruby
# lib/grape/validations/types/file.rb
        def value_coerced?(value)
          # Rack::Request creates a Hash with filename,
          # content type and an IO object. Do a bit of basic
          # duck-typing.
          value.is_a?(::Hash) && value.key?(:tempfile)
        end
```

Rack will set `params[:file][:tempfile]` to a `TempFile` object and `params[:file][:filename]` to a `String` if the end user really uploads a file.

However, the value of `params[:file][:tempfile]` is easily controllable by sending a normal POST request
```
curl -F "file[tempfile]=/etc/passwd" -F "file[filename]=myfile" http://host:port
```

Here the request will pass the check `value.key?(:tempfile)` while it's not a valid file uploading request in fact.

I think this issue is a risk and has caused some security issues (https://nvd.nist.gov/vuln/detail/CVE-2018-18649 for example). So I suggest to patch this validator to validate the type of :tempfile. Here is a [great example](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/api/validations/types/safe_file.rb)

dblock:
This makes sense, appreciate a PR.

Nyangawa:
Sure, PR submitted

dm1try:
PR totally makes sense

but the code below is 😢
```
File.read(attrs[:file][:tempfile]) # instead of attrs[:file][:tempfile].read
```
`File.read(Tempfile.new)`  is even [undocumented](https://ruby-doc.org/core-2.5.0/IO.html#method-c-read)

dblock:
Closed via https://github.com/ruby-grape/grape/pull/1844.

dblock:
@dm1try I think the whole point is that it can be a `String` which would cause you to read a file on the system like `/etc/passwd` and better safe than sorry.

PR #1844, opened 2018-12-13
fix: enforce tempfile to be a Tempfile object

Nyangawa:
#1841


spec/grape/validations/validators/coerce_spec.rb
```diff
@@ -397,6 +397,10 @@ def self.parsed?(value)
         post '/upload', file: 'not a file'
         expect(last_response.status).to eq(400)
         expect(last_response.body).to eq('file is invalid')
+
+        post '/upload', file: { filename: 'fake file', tempfile: '/etc/passwd' }
+        expect(last_response.status).to eq(400)
+        expect(last_response.body).to eq('file is invalid')
```
dblock: I know we're just copy-pasting, but for next time each of these should be a separate spec with a `before` block that sets up the `subject`.

dblock:
Perfect, thanks.

Issue #1847, opened 2018-12-14
Missing nested optional array transformed into hash with declared method


hyunskang:
grape `v1.1.0`
ruby `v2.4.4`

API definition:
``` ruby
params do
    requires :obj, type: Hash, allow_blank: false do
    requires :id, type: Integer, allow_blank: false
    optional :arr, type: Array, allow_blank: true do
      requires :k, type: String, allow_blank: false
      requires :v, type: String, allow_blank: false
    end
  end
end
put '/results' do
  present(declared(params))
end
```

Request body:
``` JSON
{ 
    "obj": { 
        "id": 1 
    } 
}
```

Response body:
``` JSON
{
    "obj": {
        "id": 1,
        "arr": {
            "k": null,
            "v": null
        }
    }
}
```

When nested array is missing from the request, `declared` transforms it into a Hash with `nil` values. 
I would expect it to return `nil` or an empty array. It looks like there was an issue opened for it, but the specs written for it do not seem to be testing the issue. https://github.com/ruby-grape/grape/issues/1642

dnesteryuk:
@kadotami fixed this issue in https://github.com/ruby-grape/grape/pull/2043 :tada:

PR #1848, opened 2018-12-14
update existing spec to test correct behavior of missing nested array

hyunskang:
Modifies existing test for nested arrays to verify that the key is mapped to an Array object. Test fails because `declared` transforms the missing Array params into a Hash.

Related Issue: https://github.com/ruby-grape/grape/issues/1847


dnesteryuk:
@hyunskang thank you for this PR :clap:  A similar test was added in https://github.com/ruby-grape/grape/pull/2043

Issue #1846, opened 2018-12-14
Validator proposal: same_as


glaucocustodio:
Hi folks, I just added a custom validator (`same_as`) to my app and I would like to know whether it would be welcome to grape itself:

```ruby
resources :passwords do
  params do
    requires :current_password, allow_blank: false
    requires :password, allow_blank: false
    requires :password_confirmation, allow_blank: false, same_as: :password
  end
  post do
  end
end
```

Thanks.

myxoh:
Can you think of any other example in which it would be useful other that the `password` `password_confirmation`? 
I've spent some time trying to think of alternative use cases and I'm having a hard time. If there aren't any it seems to be as a bit of an overkill to have a validator for just one use case, especially as there are plenty of libraries that handle this specific validation at the model level (arguably it'd be better at a param level, but seems to niche)

I'll let @dblock make the final call

glaucocustodio:
I can only think in confirmation cases like email and password. Anyway I would say it is worth to have common validators built-in like this one and some others like `length`. Those are very small, isolated (one file) and nice to have.

glaucocustodio:
I am closing because it was implemented on PR #1850.

PR #1850, opened 2018-12-17
Add same_as validator

glaucocustodio:
Implements https://github.com/ruby-grape/grape/issues/1846.


spec/spec_helper.rb
```diff
@@ -20,6 +20,7 @@
   config.include Rack::Test::Methods
   config.include Spec::Support::Helpers
   config.raise_errors_for_deprecations!
+  config.filter_run_when_matching :focus
```
glaucocustodio: This line is out of the scope from this PR,  but I find it to be very useful. Can we keep?

dblock:
Great stuff, merged. You can take this from here and make `greater_than`, `less_than`, etc.?

dblock:
No, because I mean it should be relative to another value, eg.

```ruby
params do
   requires :first
   requires :last, greater_than: :first
end
```

dblock:
Related, it might make sense to extract some of these validators into separate gems at some point. We could have a whole bunch of mathematical validators, password strength validators, etc.

glaucocustodio:
Ok, I thought you were talking about length.. Gt/Lt make sense in this case.

I agree with the idea of extracting validators in another gem in the future..

PR #1852, opened 2018-12-18
fix: validators checking aliased param instead of original one

glaucocustodio:
Hi folks.

I've pointed a scenario and I would like to hear from you whether it is a bug or not.

Please, see the comments next the code.


spec/grape/validations/params_scope_spec.rb
```diff
@@ -155,6 +155,16 @@ def initialize(value)
...
     end
+
+    it do
+      subject.params do
+        requires :foo, as: :bar, allow_blank: false
+      end
+      subject.get('/alias-not-blank-with-value') {}
+      get '/alias-not-blank-with-value', foo: 'any'
+
+      expect(last_response.status).to eq(200)
```
glaucocustodio: I get status 400 here with the `last_response.body` = `foo is empty`.

As the alias (`as`) is called before the `allow_blank`, it deletes the `foo` key from `params` and sets `params[:bar]` with its value.

Given that alias is just for internal usage (inside `get` block I mean), I think this should not fail, cause the `foo` param was given as expected. Do you agree?

myxoh: I'd tend to agree that this looks like a bug to me. Could you open an issue for this?

myxoh: (Thanks for looking at this!)

dblock:
Looks like a bug.

lib/grape/validations/params_scope.rb
```diff
@@ -293,6 +293,8 @@ def validates(attrs, validations)
         # type casted values
         coerce_type validations, attrs, doc_attrs, opts
 
+        move_alias_to_end(validations)
```
dblock: This isn't reused, and we do a ton of things above, so I would just inline the two lines of the method here.

I think `delete` also always makes a copy, and we should assume a `nil` value is a thing, so to be consistent I what we do above:

```ruby
# explain why
validations[:as] = validations.delete(:as) if validations.key?(:as)
```

dblock: Errrr.... re-reading the code we move that `:as` to the "end". I understand why it works, but feels to me that it's wrong to rely on the order in `validations`.

glaucocustodio: I agree that relying on the order is not good, but that was the easier way I found to fix the bug so far.

Another approach would be set a variable to let others validators know about alias, but it seems to be a bigger change and I am not sure whether the outcome would be worth..

Please let me know whether you have more ideas.

dblock: I just don't think we _can_ use on this approach to work reliably. There're no ordering guarantees. I think extracting `as` somehow is the proper way of doing it. Give it a try! We're here to help.

glaucocustodio: Check the new approach now please. Build is failing because of rubocop `lib/grape/endpoint.rb:6:3: C: Class has too many lines. [295/288]`.

dblock: Run `rubocop -a ; rubocop --auto-gen-config` for this.

lib/grape/endpoint.rb
```diff
@@ -319,7 +320,14 @@ def build_helpers
       Module.new { helpers.each { |mod_to_include| include mod_to_include } }
     end
 
-    private :build_stack, :build_helpers
+    def remove_aliased_params
+      return unless route_setting(:aliased_params)
```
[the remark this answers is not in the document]
glaucocustodio: We were removing the original param key inside the `as` validator, but I removed it (https://github.com/ruby-grape/grape/pull/1852/files#diff-edf496041c46c5c44b88a165b3b32206L11) to place the removal after all validators have run.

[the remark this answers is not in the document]
glaucocustodio: I don't think `key?` is necessary..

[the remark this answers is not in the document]
dblock: I mean `key?` is faster, and more ideomatic, too, because it allows for `nil` value and returns `true` in that case. We should never have a `nil` value assigned here, so that seems "better". Either way it's NBD.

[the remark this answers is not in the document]
dblock: I am a little slow :)

Why does removing the original param key _after_ all validators have run fix this issue?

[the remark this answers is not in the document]
dblock: Nevermind, I re-read the comment below, I get it now.

dblock:
Merged, thank you.

PR #1856, opened 2019-01-30
Minor cleanup to README.md in a few places

hmuar:

PR #1857, opened 2019-01-31
CI, relax rules for "rack_edge" builds

dm1try:
test only a single job and allow failures

ref https://github.com/ruby-grape/grape/pull/1856#issuecomment-459153246

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: ruby
 
 sudo: false
 
+# "gemfile" is required for "allow_failures" option,
+# see https://docs.travis-ci.com/user/customizing-the-build/#matching-jobs-with-allow_failures
+gemfile:
+
 matrix:
   include:
     - rvm: 2.5.3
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,6 +2,10 @@ language: ruby
 
 sudo: false
 
+# "gemfile" is required for "allow_failures" option,
+# see https://docs.travis-ci.com/user/customizing-the-build/#matching-jobs-with-allow_failures
+gemfile:
```
dm1try: not sure if this is needed, as our `.travis.yml` already do not follow 
> When matching jobs against the definitions given in allow_failures, all conditions in allow_failures must be met exactly, and all the keys in allow_failures element must exist in the top level of the build matrix (i.e., not in matrix.include).

or I'm missing something :)

dm1try:
it might be worth considering whether we need a tracking issue for the breaking changes that are related to `rack`(see the referenced comment)

.travis.yml
dblock: 🤷‍♂️

dm1try: from the [docs](https://docs.travis-ci.com/user/customizing-the-build/#matching-jobs-with-allow_failures)
> **and all the keys in allow_failures element must exist in the top level of the build matrix (i.e., not in matrix.include).**

according to this information we should have `rvm` key on the top level of build matrix(?) but not in matrix.include

I used this example as basis(see the link to docs):
```yaml
language: php
php:
- 5.6
- 7.0
env: # important!
matrix:
  include:
  - php: 7.0
    env: KEY=VALUE
  allow_failures:
  - php: 7.0
    env: KEY=VALUE
```
>Without the top-level env, no job will be allowed to fail

dm1try: > 🤷‍♂️

@dblock sorry, I'm not certainly sure that you mean. It's something like "I'm not sure what are you talking about" or "I don't know also". If this is the latter then ignore my comment above :)

dblock: I meant I don't know also, trust whatever you do here

PR #1863, opened 2019-02-12
Fix entity data type identification

fotos:
I hit a pretty obscure bug 🐛.

The `grape-swagger` gem uses the `type` documentation parameter in order to [identify the deduce name](https://github.com/ruby-grape/grape-swagger/blob/v0.32.1/lib/grape-swagger/doc_methods/data_type.rb#L50-L60). When the type (class) is stringified `grape-swagger` cannot check for method existance and falls back to taking the [last module as the name](https://github.com/ruby-grape/grape-swagger/blob/v0.32.1/lib/grape-swagger/doc_methods/data_type.rb#L58).

Here is a (simplified) example of the problem:

```Ruby
class API::V1::Entities::Event::Date < Grape::Entity
  expose :id
end

module API
  module V1
    class Events < Grape::API
      class Dates < Grape::API
        resources :events do
          route_param :event_id do
            params do
              requires :event_id, type: String, uuid: true, documentation: { format: 'uuid' }
            end

            resources :dates do
              params do
                requires :date, type: API::V1::Entities::Event::Date
              end
              post do
                # do something meaningful

                present date, with: API::V1::Entities::Event::Date
              end
            end
          end
        end
      end
    end
  end
end
```

If the entity name ends in `Date` then it's regarded as `primitive` and the swagger documentation is generated as a `formData`:

```YAML
post:
  parameters:
    - in: path
      name: event_id
      type: string
      format: uuid
      required: true
    - in: formData
      name: date
      type: string
      format: null
      required: true
```

(where the `date` type is then also marked as a `string` instead of object).

While generating the param documentation `type` attribute, the `type` is converted to a string. The string `type` is only used to check whether the param is an `Array`, which can be accomplished by a class inheritance check as well.

With the changes in this PR and by adding `API::V1::Events::Date.entity_name`

```Ruby
class API::V1::Entities::Event::Date < Grape::Entity
  expose :id

  def self.entity_name
    'EventDate'
  end
end
```

the generated swagger doc is correct:

```YAML
post:
  parameters:
    - in: path
      name: event_id
      type: string
      format: uuid
      required: true
    - name: EventsEventIdDates
      in: body
      required: true
      schema:
        $ref: '#/definitions/postEventsEventIdDates'
```

Everything works as expected, the specs pass and I didn't see any regressions.

I'm not 100% sure if I got this right, please let me know what you think.

### DIFF
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -84,7 +84,7 @@ module Grape
         end
 
         def declared_param_is_array?(declared_param)
-          route_options_params[declared_param.to_s] && route_options_params[declared_param.to_s][:type] == 'Array'
+          route_options_params[declared_param.to_s] && route_options_params[declared_param.to_s][:type] <= Array
         end
 
         def route_options_params
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -241,7 +241,7 @@ module Grape
 
         coerce_type = infer_coercion(validations)
 
-        doc_attrs[:type] = coerce_type.to_s if coerce_type
+        doc_attrs[:type] = coerce_type if coerce_type
 
         desc = validations.delete(:desc) || validations.delete(:description)
         doc_attrs[:desc] = desc if desc
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -2795,10 +2795,10 @@ XML
       subject.get 'method'
 
       expect(subject.routes.map(&:params)).to eq [{
-        'group1'         => { required: true, type: 'Array' },
+        'group1'         => { required: true, type: Array },
         'group1[param1]' => { required: false, desc: 'group1 param1 desc' },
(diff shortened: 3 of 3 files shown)


fotos:
⚠️ this PR may not be correct after all, please do not merge yet 😢

Although no specs are broken by the change, I think I might have triggered a bug similar to ruby-grape/grape-swagger#666 👹. By changing the array "identification" the generated swagger docs for an endpoint that accepts an array of elements (with the hack mentioned in ruby-grape/grape-swagger#666) broke.

I'll investigate a bit more and try to write a few more specs that cover all cases (this and ruby-grape/grape-swagger#666).

Something is fishy with the arrays… 🐟

dblock:
Ok, I do like the general direction of this though. Strong type checking is better than `Array`.

fotos:
@dblock I traced the problem with arrays in `grape-swagger` and proposed a fix in ruby-grape/grape-swagger#742 (previous code included a nasty trick).

I'm not 100% sure about the implications. What do you think is the best way forward?

dblock:
Let's see the maintainer(s) of that repo merge it! Keeping this open.

PR #1864, opened 2019-02-20
Allows for a block of code to always run after the endpoint

myxoh:
https://github.com/ruby-grape/grape/issues/1865

Motive: Sometime we do things like logging, or open sessions, or other actions during an API request and we want to make sure those actions are wrapped up afterwards.

This introduces a DSL way of handling this after every request on an API

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@
     - [Nested mutually_exclusive, exactly_one_of, at_least_one_of, all_or_none_of](#nested-mutually_exclusive-exactly_one_of-at_least_one_of-all_or_none_of)
   - [Namespace Validation and Coercion](#namespace-validation-and-coercion)
   - [Custom Validators](#custom-validators)
-  - [Validation Errors and Rescuing](#validation-errors-and-rescuing)
+  - [Validation Errors](#validation-errors)
   - [I18n](#i18n)
   - [Custom Validation messages](#custom-validation-messages)
     - [presence, allow_blank, values, regexp](#presence-allow_blank-values-regexp)
@@ -109,7 +109,7 @@
   - [Register custom middleware for authentication](#register-custom-middleware-for-authentication)
 - [Describing and Inspecting an API](#describing-and-inspecting-an-api)
 - [Current Route and Endpoint](#current-route-and-endpoint)
-- [Before and After](#before-and-after)
+- [Before, After and Ensure](#before-after-and-ensure)
 - [Anchoring](#anchoring)
 - [Using Custom Middleware](#using-custom-middleware)
   - [Grape Middleware](#grape-middleware)
@@ -1603,7 +1603,7 @@ end
 
 Every validation will have it's own instance of the validator, which means that the validator can have a state.
 
-### Validation Errors and Rescuing
+### Validation Errors
 
 Validation and coercion errors are collected and an exception of type `Grape::Exceptions::ValidationErrors` is raised. If the exception goes uncaught it will respond with a status of 400 and an error message. The validation errors are grouped by parameter name and can be accessed via `Grape::Exceptions::ValidationErrors#errors`.
 
@@ -1654,13 +1654,6 @@ Similarly, no regular expression test will be performed if `:blah` is blank in t
 ```ruby
 params do
   required :blah, allow_blank: false, regexp: /blah/, fail_fast: true
-end
-```
-
-You can ensure a block of code runs after every request (including failures) with `ensure`:
-```ruby
-ensure_block do
-  This.block(of: code) # Will definetely run after every request
 end
 ```
 
@@ -3096,19 +3089,22 @@ class ApiLogger < Grape::Middleware::Base
 end
 ```
 
-## Before and After
+## Before, After and Ensure
 
 Blocks can be executed before or after every API call, using `before`, `after`,
 `before_validation` and `after_validation`.
+If the API fails the `after` call will not be trigered, if you need code to execute for sure
+use the `ensure_block`
 
 Before and after callbacks execute in the following order:
 
 1. `before`
 2. `before_validation`
 3. _validations_
-4. `after_validation`
-5. _the API call_
-6. `after`
+4. `after_validation` (if Validation Successful)
+5. _the API call_ (if Validation Successful)
+6. `after` (if Validation & API Successful)
+7. `ensure_block` (ALWAYS)
 
 Steps 4, 5 and 6 only happen if validation succeeds.
 
@@ -3125,6 +3121,13 @@ For example, using a simple `before` block to set a header.
 ```ruby
 before do
   header 'X-Robots-Tag', 'noindex'
+end
+```
+
+You can ensure a block of code runs after every request (including failures) with `ensure`:
+```ruby
+ensure_block do
+  This.block(of: code) # Will definetely run after every request
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@
   - [Register custom middleware for authentication](#register-custom-middleware-for-authentication)
 - [Describing and Inspecting an API](#describing-and-inspecting-an-api)
 - [Current Route and Endpoint](#current-route-and-endpoint)
-- [Before, After and Ensure](#before-after-and-ensure)
+- [Before, After and Finally](#before-after-and-finally)
 - [Anchoring](#anchoring)
 - [Using Custom Middleware](#using-custom-middleware)
   - [Grape Middleware](#grape-middleware)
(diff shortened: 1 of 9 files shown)


lib/grape/dsl/request_response.rb
```diff
@@ -131,6 +131,11 @@ def rescue_from(*args, &block)
           namespace_stackable(:rescue_options, options)
         end
 
+        def ensure_block(&ensured_block)
```
undersky0: method description and example use are missing

lib/grape/middleware/error.rb
```diff
@@ -46,9 +47,19 @@ def call!(env)
...
+          execute_ensured_block! if should_ensure?
         end
       end
 
+      def should_ensure?
+        options[:ensured]
+      end
+
+      def execute_ensured_block!
+        options[:ensured_block].call
```
undersky0: coudl be block_given? instead of should_ensure check

LeFnord:
is it a different funcionality as this one [#before-and-after](https://github.com/ruby-grape/grape#before-and-after)?

myxoh:
@LeFnord 
It's related to the `before-and-after` (potentially that section would be a better place on the description)
The difference between `after` and `ensure_block` is that `ensure_block` will run even when there has been an exception on the API while `after` will not

lib/grape/middleware/error.rb
myxoh: Thanks @undersky0 , however `block_given?` would be false as this method is not sent a block (only ensure_block is sent the block)

lib/grape/dsl/request_response.rb
myxoh: Thanks! Will address!

LeFnord:
… ok, something like an `around` 😉

myxoh:
Difference would be that this is not run before.
So the cycle would be 
before 
api_call
IF success
  after
END
ensure

LeFnord:
nice … please update above readme section with this, thanks

dm1try:
> is it a different funcionality as this one [#before-and-after](https://github.com/ruby-grape/grape#before-and-after)?
> It's related to the `before-and-after`

I don't think so; the implementation looks like a missing complementary part for `rescue_from` sugar and even lives in the error middleware. 

>Sometime we do things like logging, or open sessions, or other actions during an API request and we want to make sure those actions are wrapped up afterwards.

maybe this can be solved by using an endpoint middleware([the spec](https://github.com/ruby-grape/grape/blob/2633f74411aa3334a4ba18072d6297a403e1d83e/spec/grape/middleware/base_spec.rb#L32-L39))

myxoh:
I've updated the lifecycle clarifying how this will run. I agree that it could be sorted by Middleware @dm1try  but I do feel like having an `after` block that always run should be a feature in Grape regardless. 
Could we move the discussion around the feature to the issue: https://github.com/ruby-grape/grape/issues/1865 and discuss over here only the implementation?

dm1try:
> I've updated the lifecycle clarifying how this will run.

Users can be confused by the current implementation, as the description hints that `ensure_block` is the part of a request lifecycle, though this block is run in the different scope(in the error middleware) which means it does not share the endpoint state.
```ruby
before do
 @state = true
end
ensure_block do
  @state # error
end
```
Moreover according to the ordering in [the current middleware stack](https://github.com/ruby-grape/grape/blob/master/lib/grape/endpoint.rb#L282-L311), `ensure_block` can be called without a preceding `before` block if the error will be raised before an endpoint will run(for example,  versioner middleware can raise an error that will be caught by error middleware); it feels unexpected.

> but I do feel like having an `after` block that always run should be a feature in Grape regardless.

btw, the same behavior that proposed here can be emulated by using [filters middleware](https://github.com/ruby-grape/grape/blob/2633f74411aa3334a4ba18072d6297a403e1d83e/lib/grape/middleware/filter.rb#L6) without providing anything new:
```ruby
use Grape::Middleware::Filter, after: -> { do_something }
```

dblock:
I like the syntactic sugar of `ensure_block`. I would prefer `ensure`, but I presume it won't work as a reserved keyword? naming is hard :)

I think @dm1try has a very valid point (nice catch) about scope of execution and corresponding `before`, I think that needs to be fixed and explained clearly what the scope of execution is

myxoh:
Nice catch @dm1try, I'll give it a go at moving the scope, this was just a simple MVP.

Regarding wording, I went for `ensure` initially but yes, it being a reserved keyword made things nasty.

We could go with things like `always` or `finally` (borrowing from other languages)

dblock:
How do others feel about `finally` vs. `ensure_block`?

myxoh:
after some thought I think I like `finally` better. I've also proceeded to move the logic to the `endpoint` so it shares scope with other callbacks like `before` and the helpers.

LeFnord:
+1 for `finally`

README.md
```diff
@@ -3089,19 +3089,22 @@ class ApiLogger < Grape::Middleware::Base
 end
 ```
 
-## Before and After
+## Before, After and Finally
 
 Blocks can be executed before or after every API call, using `before`, `after`,
 `before_validation` and `after_validation`.
+If the API fails the `after` call will not be trigered, if you need code to execute for sure
+use the `finally`
```
dblock: Missing period.

README.md
```diff
@@ -3089,19 +3089,22 @@ class ApiLogger < Grape::Middleware::Base
...
 1. `before`
 2. `before_validation`
 3. _validations_
-4. `after_validation`
-5. _the API call_
-6. `after`
+4. `after_validation` (if Validation Successful)
+5. _the API call_ (if Validation Successful)
+6. `after` (if Validation & API Successful)
+7. `finally` (ALWAYS)
```
dblock: Nitpicking, let's not SHOUT :) I would make all of the comments lowercase (`upon successful validation`, `always`, etc.)

README.md
```diff
@@ -3121,6 +3124,13 @@ before do
 end
 ```
 
+You can ensure a block of code runs after every request (including failures) with `finally`:
```
dblock: Add a space below.

README.md
```diff
@@ -3121,6 +3124,13 @@ before do
...
+```ruby
+finally do
+  This.block(of: code) # Will definetely run after every request
```
dblock: maybe just `# this code will run after every successful or failed request`, cause `This.block(of: code)` is not ruby

lib/grape/dsl/callbacks.rb
```diff
@@ -43,6 +43,26 @@ def after_validation(&block)
         def after(&block)
           namespace_stackable(:afters, block)
         end
+
+        # Allows you to specify a something that will always be executed after a call
+        # API call. Unlike the `after` block, this code will run even on
+        # unsuccesful requests
```
dblock: Add a period after `requests`.

lib/grape/endpoint.rb
```diff
@@ -245,33 +245,38 @@ def run
...
+          cookies.write(header)
 
-        run_filters afters, :after
-        cookies.write(header)
+          # status verifies body presence when DELETE
+          @body ||= response_object
 
-        # status verifies body presence when DELETE
-        @body ||= response_object
+          # The Body commonly is an Array of Strings, the application instance itself, or a File-like object
```
dblock: lowercase `body`

dblock:
I'm cool with this change if @dmitry is cool with this change?

lib/grape/endpoint.rb
```diff
@@ -245,33 +245,38 @@ def run
...
+          response_object = file || [body]
 
-        # The Body commonly is an Array of Strings, the application instance itself, or a File-like object
-        response_object = file || [body]
-        [status, header, response_object]
+          [status, header, response_object]
+        ensure
+          # An error on the finallies should not prevent a succesful request from executing
+          begin run_filters finallies, :finally end
```
dm1try: `begin/end` looks useless. do you mean 
```ruby
run_filters finallies, :finally rescue nil
```
if so, it is certainly a bad practice.

myxoh: I did mean that. 
And I'd tend to agree is bad practice, however I would be concern about a `finally` breaking an otherwise successful request. 
Do you have an alternative suggestion? Or would you just allow the broken finally to break the request (There's certainly an argument for that)

myxoh:
I'll address your changes @dblock  as soon as @dm1try helps me decide on the situation of a finally raising an error

lib/grape/endpoint.rb
myxoh: (I'm merely moving this around but happy to address this anyway)

lib/grape/endpoint.rb
dblock: In ruby, when an `ensure` block fails, it just re-raises an exception. I definitely don't think we should be swallowing an error here. I expect to at least get a 500 back with the exception details.

dm1try: I agree with @dblock, this should be consistent with `ensure`.

a user should decide if he uses "safe" code in `finally` or he needs some additional guard.
otherwise some unexpected `NameError` in recently-added "finally" callback could stop the running all other "finallies" and no one will know about this.

myxoh:
Addressed

dblock:
Merged. 

I want to thank @dm1try @LeFnord and of course @myxoh for this beautiful example of collaboration, strong, but not stubborn opinions and an excellent feature that will make API developers' work easier. 🍻

PR #1868, opened 2019-03-11
Fix NoMethodError with none Hash params

ksss:
```rb
params do
  requires :foos, type: Array do
    optional :bar
    given :bar do
      requires :baz
    end
  end
end
get '/test' { 'ok' }
```

```
client.get('/test', foos: ['abc'])
#=> NoMethodError (undefined method `with_indifferent_access' for "abc":String)
```

### DIFF
--- a/spec/grape/validations/params_scope_spec.rb
+++ b/spec/grape/validations/params_scope_spec.rb
@@ -703,6 +703,26 @@ describe Grape::Validations::ParamsScope do
     before do
       subject.params do
         requires :foos, type: Array do
+          optional :foo
+          given :foo do
+            requires :bar
+          end
+        end
+      end
+      subject.get('/test') { 'ok' }
+    end
+
+    it 'should pass none Hash params' do
+      get '/test', foos: ['']
+      expect(last_response.status).to eq(200)
+      expect(last_response.body).to eq('ok')
+    end
+  end
+
+  context 'when validations are dependent on a parameter within an array param within #declared(params).to_json' do
+    before do
+      subject.params do
+        requires :foos, type: Array do
           optional :foo_type, :baz_type
           given :foo_type do
             requires :bar
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 #### Fixes
 
-* Your contribution here.
+* [#1868](https://github.com/ruby-grape/grape/pull/1868): Fix NoMethodError with none hash params - [@ksss](https://github.com/ksss).
 
 ### 1.2.3 (2019/01/16)
 
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -52,6 +52,7 @@ module Grape
 
         return true unless @dependent_on
         return params.any? { |param| meets_dependency?(param, request_params) } if params.is_a?(Array)
+        return false unless params.respond_to?(:with_indifferent_access)
         params = params.with_indifferent_access
 
         @dependent_on.each do |dependency|


dblock:
I don't think the fix is correct. In the example case the type of items in `foos` is expected to be a `Hash`, implied by an optional `:bar`. Passing a `String` should yield an error.  I agree the current error is not great and I'd want Grape to do better.

spec/grape/validations/params_scope_spec.rb
```diff
@@ -700,6 +700,26 @@ def initialize(value)
...
+            requires :bar
+          end
+        end
+      end
+      subject.get('/test') { 'ok' }
+    end
+
+    it 'should pass none Hash params' do
+      get '/test', foos: ['']
+      expect(last_response.status).to eq(200)
```
dblock: I believe this should be a 400. The first item in `foos` does not match the expected input of my API.

ksss: I followed this behavior.

```rb
params do
  optional :foo
  given :foo do
    requires :bar
  end
end
get('/test') { 'ok' }
```

```rb
client.get '/test', 'hello'
last_response.status #=> 200
last_response.body #=> "ok"
```

Is this also not intentional?

dblock: Hm. So we have this spec today? I feel that's wrong. Maybe @namusyaka or @dm1try have an opinion?

dm1try: @dblock I do not see any contradiction in the examples above, in the latter case `foo` is not provided, so it's not validated. in the original case `foos` is provided, so it's validated.

according to the right behavior, if we assume that the implicit requirement for `type: Array` **with block** is "array of objects" then you are right, we should return a validation error.

dblock: Ok I think I agree. So what @dm1try is saying is that the spec above that @ksss is suggesting is correct by design, but that the spec in this PR is not because see above. So I'll take back my "i feel that's wrong". 

@ksss does that make sense?

ksss: @dm1try Thank you for reviewing.
@dblock That makes sense.
I expect that scope in this PR is no errors occur on validation step.

dblock: @dm1try Just to confirm, you're saying this PR is good to go as is? Merge if yes.

dm1try: @dblock I think it's fine, at least it's better than `NoMethodError`.

though, in the future this spec should pass:
```ruby
  context 'group values' do
    before do
      subject.params do
        requires :value, type: Hash do
        end
      end
      subject.get('/hash') { }

      subject.params do
        requires :value, type: Array do
        end
      end
      subject.get('/array_of_hashes') { }
    end

    it 'should be consistent' do
      get '/hash', value: ""
      expect(last_response.body).to eq("value is invalid")

      get '/array_of_hashes', value: [""]
      expect(last_response.body).to eq("value is invalid")
    end
  end
```

I've tested the idea to set implicit `Array[Hash]` type [here](https://github.com/ruby-grape/grape/blob/b8b85af12ad4860b6ee4ae2074f864a849400209/lib/grape/validations/params_scope.rb#L163)
```ruby
        if(block && (validations[:type].nil? || validations[:type] == Array))
          validations[:type] = Array[Hash]
        end
```
and while it works the case above, it fails for the cases where we rely on Virtus coerce logic for the hash
```
irb -Ilib -rgrape
irb(main):001:0> Virtus::Attribute.build(Array[Hash]).coerce({just: "hash"})
=> [{:just=>nil, "hash"=>nil}]
irb(main):002:0> Virtus::Attribute.build(Array).coerce({just: "hash"})
=> [[:just, "hash"]]
```

so this needs more investigation

Issue #1871, opened 2019-03-14
`Module#parent` deprecation warning with Rails 6


shlima:
Got deprecations warnings with Rails 6.0.0.beta3:

```bash
DEPRECATION WARNING: `Module#parent` has 
been renamed to `module_parent`. 
`parent` is deprecated and will be removed in Rails 6.1. 
```

The warnings refers to the `desc` block

rromanchuk:
looks like this is probably the offender https://github.com/ruby-grape/grape/blob/2dcef9af4dd483a1a3d56cbfd660056ed8e9fe42/lib/grape/dsl/settings.rb#L144

dm1try:
@rromanchuk this method is from `InheritableSetting`
https://github.com/ruby-grape/grape/blob/51217ea2b2d39ff7268feef93838bec880c6fd35/lib/grape/util/inheritable_setting.rb#L8

it should not be a cause of deprecation warning(if the method is not overloaded somehow)

dm1try:
`Module#parent` is a part of [`active_support/core_ext/module/introspection`](
https://github.com/rails/rails/blob/d8d6bd5e63a9a4a6c06a4dde3c7137ee2be105fd/activesupport/lib/active_support/core_ext/module/introspection.rb#L46-L52) but it does not look like grape uses this patch:
```
ruby -Ilib -rgrape -e 'p Module.method_defined?(:parent)'
false
```
at least on load

Could someone provide a traceback for the warning?

myxoh:
I think this was sorted on: https://github.com/ruby-grape/grape/pull/1876 
If I am mistaken, please feel free to re-open the issue

PR #1876, opened 2019-04-08
Fix const errors being hidden by bug in const_missing

dandehavilland:
I had not included a file and so a constant was missing. I did not receive the correct error as the custom `const_missing` was throwing an error of its own.

This PR resolves that for me.

```bash
NameError:
  undefined local variable or method `parent' for MyApiClass
# /repos/grape/lib/grape/api.rb:107:in `method_missing'
# /repos/grape/lib/grape/api.rb:67:in `const_missing'
# ./lib/routes/my_api_class.rb:24:in `block in <class:MyApiClass>'
# /repos/grape/lib/grape/endpoint.rb:57:in `call'
# /repos/grape/lib/grape/endpoint.rb:57:in `block (2 levels) in generate_api_method'
# /repos/grape/lib/grape/endpoint.rb:56:in `block in generate_api_method'
# /repos/grape/lib/grape/endpoint.rb:263:in `block in run'
# /repos/grape/lib/grape/endpoint.rb:243:in `run'
# /repos/grape/lib/grape/endpoint.rb:318:in `block in build_stack'
# /repos/grape/lib/grape/middleware/base.rb:31:in `call!'
# /repos/grape/lib/grape/middleware/base.rb:24:in `call'
# /repos/grape/lib/grape/middleware/error.rb:37:in `block in call!'
# /repos/grape/lib/grape/middleware/error.rb:36:in `catch'
# /repos/grape/lib/grape/middleware/error.rb:36:in `call!'
# ...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#1876](https://github.com/ruby-grape/grape/pull/1876): Fix const errors being hidden by bug in `const_missing` - [@dandehavilland](https://github.com/dandehavilland).
 * [#1864](https://github.com/ruby-grape/grape/pull/1864): Adds `finally` on the API - [@myxoh](https://github.com/myxoh).
 * [#1869](https://github.com/ruby-grape/grape/pull/1869): Fix issue with empty headers after `error!` method call - [@anaumov](https://github.com/anaumov).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,13 @@
 #### Features
 
 * Your contribution here.
-* [#1876](https://github.com/ruby-grape/grape/pull/1876): Fix const errors being hidden by bug in `const_missing` - [@dandehavilland](https://github.com/dandehavilland).
 * [#1864](https://github.com/ruby-grape/grape/pull/1864): Adds `finally` on the API - [@myxoh](https://github.com/myxoh).
 * [#1869](https://github.com/ruby-grape/grape/pull/1869): Fix issue with empty headers after `error!` method call - [@anaumov](https://github.com/anaumov).
 
 #### Fixes
 
 * [#1868](https://github.com/ruby-grape/grape/pull/1868): Fix NoMethodError with none hash params - [@ksss](https://github.com/ksss).
+* [#1876](https://github.com/ruby-grape/grape/pull/1876): Fix const errors being hidden by bug in `const_missing` - [@dandehavilland](https://github.com/dandehavilland).
 
 ### 1.2.3 (2019/01/16)
 
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -64,8 +64,6 @@ module Grape
       def const_missing(*args)
         if base_instance.const_defined?(*args)
           base_instance.const_get(*args)
-        elsif @base_parent && @base_parent.const_defined?(*args)
-          @base_parent.const_get(*args)
         else
           super
         end
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -64,7 +64,7 @@ module Grape
       def const_missing(*args)
         if base_instance.const_defined?(*args)
           base_instance.const_get(*args)
-        elsif parent && parent.const_defined?(*args)
+        elsif defined?(parent) && parent.const_defined?(*args)
           parent.const_get(*args)
         else
           super
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -64,8 +64,8 @@ module Grape
       def const_missing(*args)
         if base_instance.const_defined?(*args)
           base_instance.const_get(*args)
-        elsif defined?(parent) && parent.const_defined?(*args)
-          parent.const_get(*args)
+        elsif @base_parent && @base_parent.const_defined?(*args)
+          @base_parent.const_get(*args)
         else
           super
         end
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -3883,7 +3883,7 @@ XML
     before { subject.mount mounted => '/const' }
 
     it 'raises an error' do
-      expect { get '/const/missing' }.to raise_error(NameError).with_message(/SomeRandomConst/)
+      expect { get '/const/missing' }.to raise_error(NameError).with_message(/SomeRandomConstant/)
     end
   end
 end
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -3871,16 +3871,19 @@ XML
   end
 
   describe 'const_missing' do
-    class Mounted < Grape::API
-      get '/missing' do
-        SomeRandomConst
+    subject(:grape_api) { Class.new(Grape::API) }
+    let(:mounted) do
+      Class.new(Grape::API) do
+        get '/missing' do
+          SomeRandomConstant
+        end
       end
     end
 
-    before { subject.mount Mounted => '/const' }
+    before { subject.mount mounted => '/const' }
 
     it 'raises an error' do
-      expect { get '/const/missing' }.to raise_error(NameError).with_message('uninitialized constant Mounted::SomeRandomConst')
+      expect { get '/const/missing' }.to raise_error(NameError).with_message(/SomeRandomConst/)
     end
   end
 end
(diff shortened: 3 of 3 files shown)


lib/grape/api.rb
```diff
@@ -64,7 +64,7 @@ def api.inherited(child_api)
       def const_missing(*args)
         if base_instance.const_defined?(*args)
           base_instance.const_get(*args)
-        elsif parent && parent.const_defined?(*args)
+        elsif defined?(parent) && parent.const_defined?(*args)
```
dblock: Can `parent` ever be `nil` here, cause `defined?(nil)` is `"nil"` and `nil` doesn't have a `const_defined` method.

dandehavilland: Should this even be `parent` or should it be `@base_parent`? I don't see any definition of `parent` anywhere.

dandehavilland: Good catch btw :)

dblock: Not sure, but I think we need some specs here. Try something?

dandehavilland:
@dblock I've added a spec for the issue I experienced. It's tied to mounted modules and is resolved by checking `@base_parent` rather than `parent`. Is it possible that `parent` was a typo there? I can't find a reference to it anywhere else.

dblock:
> @dblock I've added a spec for the issue I experienced (fails on master, passes with this fix). It's tied to mounted modules and is resolved by checking `@base_parent` rather than `parent`. Is it possible that `parent` was a typo there? I can't find a reference to it anywhere else.

Yes absolutely.

dblock:
I know it's unrelated, but care to fix the Rails edge problem as well for 💚, please

spec/grape/api_spec.rb
```diff
@@ -3869,4 +3869,18 @@ def before
       expect(grape_api.eql?(MyAPI))
     end
   end
+
+  describe 'const_missing' do
+    class Mounted < Grape::API
```
dblock: This pollutes global scope. Try `Class.new(Grape::API)` like the example above? You can give it a name with `const_set`. Another way to do it is to remove the class in an `after` block.

spec/grape/api_spec.rb
```diff
@@ -3869,4 +3869,18 @@ def before
...
+      get '/missing' do
+        SomeRandomConst
+      end
+    end
+
+    before { subject.mount Mounted => '/const' }
+
+    it 'raises an error' do
+      expect { get '/const/missing' }.to raise_error(NameError).with_message('uninitialized constant Mounted::SomeRandomConst')
```
dblock: I'm glad we have a proper spec for this now!

spec/grape/api_spec.rb
dandehavilland: > Try Class.new(Grape::API)

Yeah, I tried that initially but for some reason it doesn't give the same result.

Does this look right to you?
```ruby
let(:mounted) { Class.new(Grape::API) }
# ...
before { subject.mount mounted => '/const' }
```

dandehavilland:
> I know it's unrelated, but care to fix the Rails edge problem as well for 💚, please

I can take a look if you point me in the right direction - I haven't touched Rails in a long while though so no promises!

dm1try:
> Is it possible that parent was a typo there? I can't find a reference to it anywhere else.

according to https://github.com/ruby-grape/grape/commit/ce4966def374ca0fbfed95e520373578111c43e8, `parent` could be the part of active support module introspection interface(`parent/parent_name` were added; I doubt that it's a typo). it is somehow related to the integration with Rails, see https://github.com/ruby-grape/grape/issues/1871

If this is true then `parent != @base_parent` but it works because the initial implementation does not have specs.

dblock:
The Rails/edge failure seems trivial: `expected NameError with "uninitialized constant Mounted::SomeRandomConst", got #<NameError: uninitialized constant SomeRandomConst> with backtrace:`, so the expectation should just be a little more flexible with the message not containing the namespace. @dandehavilland I think `.to raise_error NameError, /SomeRandomConst/` will do it, give it a try?

This PR seems to be fixing https://github.com/ruby-grape/grape/issues/1871, as it no longer uses `parent`, so I think this kills both birds with one stone. 

Let's see what @myxoh says.

myxoh:
I'm surprised no specs are failing after this change, I would have expected some to be. We probably need to add some coverage there.

I imagine that `ActiveSupport` at some point deprecated the `parent` method on classes altogether. The solution would be to use the new method I'd say

myxoh:
@dandehavilland I see! yes, that sounds to be the likely cause of the issue! Could we add a Travis grid for your scenario?
As I wasn't able to get that spec to fail in any of the existing ones.

If you are right, then the solution would probably to add the reqiurement to that module, or to solve what I was trying to do without ActiveSupport (finding the base module of a class)

dm1try:
> As I wasn't able to get that spec to fail in any of the existing ones.

It's weird. The spec provided in this issue is red in my case, and this is expected behaviour as grape does not load explicitly `active_support/core_ext/module/introspection` in which the needed functional is implemented. 

```bash
bundle exec rspec spec/grape/api_spec.rb:3873
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.
Run options: include {:locations=>{"./spec/grape/api_spec.rb"=>[3873]}}

Grape::API
  const_missing
    raises an error (FAILED - 1)

Failures:

  1) Grape::API const_missing raises an error
     Failure/Error: expect { get '/const/missing' }.to raise_error(NameError).with_message('uninitialized constant Mounted::SomeRandomConst')

       expected NameError with "uninitialized constant Mounted::SomeRandomConst", got #<NameError: undefined local variable or method `parent' for Mounted:Class
       Did you mean?  print> with backtrace:
...
```

myxoh:
I don't know why yesterday I wasn't able to reproduce this, I now have been able to.
I also now understand why no test is failing when you've changed from `parent` to `@base_parent`

It seems to be the case that that branch of the if condition was no longer required, I think the issue that made it required was fixed on commit https://github.com/ruby-grape/grape/commit/532c08d6f3386ae63a5a31c16f281f4d0ba2bbd3

Therefore the proper solution for the issue you are presenting here is just removing the whole `elsif` as it is no longer required. It should look like:

```ruby
      def const_missing(*args)
        if base_instance.const_defined?(*args)
          base_instance.const_get(*args)
        else
          super
        end
      end
```
I've already run the specs (including the one you've introduced) and it seems to be fixed when doing this

myxoh:
@dandehavilland just a friendly pig whether you can update the PR, I think that will also fix the spec that's currently failing

dandehavilland:
Ok will do that later today!

myxoh:
Well the spec is still failing, not sure what to make of it, I think it might be worth updating the spec in this instance as the error message is being generated on a low level, and I don't think this is indicative of a larger issue.

I'll leave this to your criteria

dblock:
The spec failure is trivial:

```
1) Grape::API const_missing raises an error
     Failure/Error: expect { get '/const/missing' }.to raise_error(NameError).with_message('uninitialized constant Mounted::SomeRandomConst')
     
       expected NameError with "uninitialized constant Mounted::SomeRandomConst", got #<NameError: uninitialized constant SomeRandomConst> with backtrace:
```

Note the namespace in the error vs. no namespace in the expected error

Change to `expect { get '/const/missing' }.to raise_error(NameError).with_message(/uninitialized constant .* SomeRandomConst/) or something similar with a regex or even just `SomeRandomConst`

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#1876](https://github.com/ruby-grape/grape/pull/1876): Fix const errors being hidden by bug in `const_missing` - [@dandehavilland](https://github.com/dandehavilland).
```
dblock: This should go into Fixes.

spec/grape/api_spec.rb
```diff
@@ -3869,4 +3869,21 @@ def before
...
+        get '/missing' do
+          SomeRandomConstant
+        end
+      end
+    end
+
+    before { subject.mount mounted => '/const' }
+
+    it 'raises an error' do
+      expect { get '/const/missing' }.to raise_error(NameError).with_message(/SomeRandomConst/)
```
dblock: Spell out SomeRandomConstant (add "ant") just to be consistent with the code above.

dandehavilland:
> Two more minor things please. Apologies for my OCD.

I missed both of those - fixed!

dblock:
Merged, thanks.

Issue #1878, opened 2019-04-18
Mounting instructions not working in Rails 6 beta


wentaoliu:
The upcoming Rails 6.0 has some breaking changes in autoloading.
See https://github.com/rails/rails/issues/35567#issuecomment-473460592
I followed those steps in Rails 6.0.0.beta3 and it works fine.
Should we document this in the Readme file, or wait and see whether it still holds in the final release of Rails 6.0?

dblock:
It would be helpful to document it for Rails 6 Betas sooner than later. We can remove the "beta" part when Rails 6 ships. Appreciate a PR!

myxoh:
I think this issue is now addressed. Thanks for the contribution!

PR #1879, opened 2019-04-18
Update mounting instructions for Rails 5&6 in README

wentaoliu:
For Rails >=5.2, modification of `application.rb` is not necessary.
For Rails 6.0 beta, the new `Zeitwerk` loader brings some breaking changes. `API` in our Twitter::API example now has to be explicitly set as an acronym in `inflections.rb`. 
Issue: https://github.com/ruby-grape/grape/issues/1878


dblock:
Thanks

PR #1881, opened 2019-04-28
Fix #1880 to prevent header case change in some RSpec test scenarios

### LINKED ISSUE #1880: Case insensitive headers
Using Grape 1.2.3 and started experiencing this issue after upgrading. Ruby 4.6.2p47, Rails 4.2.11.1 I've found that headers get lowercased, except the first character 'X', when using RSpec tests so my lookups for the header key fails. For example, when a request is made through my app, the case is converted to "X-Auth-Token", but when access through RSpec the same header appears as "X-auth-token".

I understand that headers are case insensitive, but what is responsible for converting the header case in grape? Ideally, the case is either all caps, all lowercase, or camel case - or there is a method provided in grape to lookup header keys with case insensitive matching. I don't want to write a wrapper as stated here: https://www.ruby-forum.com/t/case-insensitive-hash/51767/2

I'm at a loss on how to best access headers so that they are case insensitive, so any help is appreciated.

Thanks in advance.

Cyclic:
When using Grape with RSpec and some variation of Ruby versions, the case of headers is not maintained, causing header key lookups to fail.

### DIFF
--- a/lib/grape/request.rb
+++ b/lib/grape/request.rb
@@ -32,7 +32,7 @@ module Grape
       env.each_pair do |k, v|
         next unless k.to_s.start_with? HTTP_PREFIX
 
-        k = k[5..-1].split('_').each(&:capitalize!).join('-')
+        k = k[5..-1].split('_').each(&:capitalize).join('-')
         headers[k] = v
       end
       headers


dblock:
Oh wow, this is so strange as `split` should return a copy of a string. I'd like to merge this, just needs a CHANGELOG entry please? Maybe a spec where the original headers are maintained as expected?

Cyclic:
Thank you, @dblock! Just after creating the PR, the Travis CI build revealed some interesting variations on the issue. In addition, I have several specs locally that fail still because of casing, but the 30 some odd instances that failed my specs before are now working. I am tracking down why there are now 3 new failures in the non-spec tests, likely the fact that capitalize is making the keys ALL-UPPERCASE when the specs seem to be converted to Pascal-Case by capitalize. I am still digging in. It seems that either capitalize is getting a different variant based on the server, or something is wonky with my environment using different variations. 

I wonder if it might make sense to use a more reliable, less changing method and standard, such as downcase or upcase. Better yet, maybe it does make sense to have a wrapper class for the headers so that accessing the keys will be case insensitive, but I don't know how compatible any of this will be with other versions as it seems like a big change. It took a very long time to finally track this down, and I wonder if other developers have had similar experiences.

Let me know if you have any insights. I'll keep trying to narrow down why I am left with this last variation. Even Travis CI seems to have its own variation of all uppercase, so it's not so uncommon. 🤯

dblock:
Ok, I should stop talking out of my *ss and start reading my own code. Apologies. Ignore everything I said so far on this issue, I am completely wrong.

The `headers` helper was added in https://github.com/ruby-grape/grape/commit/f6f585ea6cc720e779a37f756b9e9cda4786dad2 and from day 1 it converted the headers case into pascal-case. So we should say it's by design and document it clearly. I would appreciate *that* PR.

This means that in your code you should expect that the case is not maintained. I'll close this PR because we definitely don't want this, and let's find the cause of #1880 there.

PR #1884, opened 2019-05-05
Change misleading naming of alias

### LINKED ISSUE #1875: [docs] "alias" is misleading
Grape can rename parameters via `as:`, so for example:
```ruby
params do
  requires :email_address, as: :email
end
```
will map the input paramter `email_address` to the key `:email` in `params` and`declared(params)`.

I like this feature very much and it does exactly the right thing. But the feature has a quite misleading name:

> ### Alias
>
> You can set an alias for parameters using `as`, which can be useful when [ ...]
 
The term "alias" implies that both parameters co-exist, that they can be used interchangeably. This is clearly not the case here, there's no `params[:email_address]`. Maybe there's some actual aliasing going on within grape, but not on the outside.

I therefore kindly ask you to change the term "alias". Something like "parameter mapping" or "renaming" seems much more appropriate.

tokachev:
Closes #1875. 

This PR is to rename misleading term "alias" to "renaming". Changes cover docs and variables naming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@
   - [Validation of Nested Parameters](#validation-of-nested-parameters)
   - [Dependent Parameters](#dependent-parameters)
   - [Group Options](#group-options)
-  - [Alias](#alias)
+  - [Renaming](#renaming)
   - [Built-in Validators](#built-in-validators)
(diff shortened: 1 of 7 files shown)


dblock:
Great stuff, merging.

Issue #1892, opened 2019-07-03
Inherit from Grape::API::Instance breaks Boolean type


sondnm:
Recently I followed the UPGRADING.md guide to upgrade Grape from 1.0.3 to 1.2.4 and an error occurs due to the use of `type: Boolean` in `params` block.
This happens only when inheriting from `Grape::API::Instance`. I wrote a small rack app to test and the behaviour is the same.

```ruby
# config.ru
require 'grape'

module App
  class API < Grape::API::Instance
    resource :system do
      params do
        optional :hit, type: Boolean
      end
      get :ping do
        "pong"
      end
    end
  end
end

run App::API
```

Error:
```ruby
NameError: uninitialized constant App::API::Boolean
```

I could easily change the `Boolean` to `Virtus::Attribute::Boolean` to fix this issue on my app but since you already supported this Boolean type feature, I think it should be supported in the next releases too.

Thank you.

dblock:
Feel free to PR a fix to this effect. I would also try to track where this got removed. There's probably a good reason since `Boolean` is not a native type in Ruby.

myxoh:
Note: if you are just creating an API you should not be inheriting from `::Instance` rather just from `Grape::API`

In the `Upgrading.md` guide the reference to when you should use `Grape::API::Instance` it is only when you are monkeypatching the class (which is not what you are doing here)

i.e. on this example you should be doing:

```
# config.ru
require 'grape'

module App
  class API < Grape::API
    resource :system do
      params do
        optional :hit, type: Boolean
      end
      get :ping do
        "pong"
      end
    end
  end
end

run App::API
```

sondnm:
Thank you @myxoh for pointing that out. As I can see, the inheritance from `Grape::API` works great for a rack app.

However, on my app where I integrate `grape` with `rails` and using `rspec` together with `rack-test`,  when I run the api specs, it returns 404 if it inherits from `Grape::API`. Notice that the APIs still works normally when running on a real server. Since I want to make sure the tests pass before deployment, I just let them inherit from `Grape::API::Instance` 😄 . That's where I encounter the `Boolean` problem.

FYI, my app's APIs structure would look similar to this.

```ruby
Rails.application.routes.draw do
  mount API::Base, at: "/"
end
```

```ruby
# app/controllers/api/base.rb

class API::Base < Grape::API
  mount API::V1::Base
end
```

```ruby
# app/controllers/api/v1/base.rb

class API::V1::Base < Grape::API
[...]

PR #1893, opened 2019-07-05
Makes sure Grape::API behaves as a Rack::App 

myxoh:
By calling: 'call' on the first mounted instance rather than the base instance (which is never mounted) which will have the environment information as to where it was mounted

This addresses the underlying issue raised by: https://github.com/ruby-grape/grape/issues/1892

### DIFF
--- a/spec/grape/integration/rack_spec.rb
+++ b/spec/grape/integration/rack_spec.rb
@@ -31,4 +31,26 @@ describe Rack do
       input.unlink
     end
   end
+
+  context 'when the app is mounted' do
+    def app
+      @main_app ||= Class.new(Grape::API) do
+        get 'ping'
+      end
+    end
+
+    let!(:base) do
+      app_to_mount = app
+      Class.new(Grape::API) do
+        namespace 'namespace' do
+          mount app_to_mount
+        end
+      end
+    end
+
+    it 'finds the app on the namespace' do
+      get '/namespace/ping'
+      expect(last_response.status).to eq 200
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


spec/grape/integration/rack_spec.rb
```diff
@@ -31,4 +31,26 @@
       input.unlink
     end
   end
+
+  context 'when the app is mounted' do
+    def app
```
dblock: This pollutess the global namespace, put it into a namespace itself?

myxoh: In what way @dblock ? I am under the impression that methods defined in a context are contained to that context?

dblock: Okay, this is news to me. Since when does it work as one would expect?!

```ruby
require 'rspec'

describe 'Context' do
  def m1
    'm1'
  end

  it 'is m1' do
    expect(m1).to eq 'm1'
  end

  context 'c1' do
  	def m2
	  'c1:m2'
  	end

    it 'is m1' do
      expect(m1).to eq 'm1'
    end

    it 'is m2' do
      expect(m2).to eq 'c1:m2'
    end
  end

  context 'c2' do
	it 'is m2' do
	  expect(m2).to eq 'c1:m2'
    end
  end
end
```

```
  1) Context c2 is m2
     Failure/Error: expect(m2).to eq 'c1:m2'
     
     NameError:
       undefined local variable or method `m2' for #<RSpec::ExampleGroups::Context::C2 "is m2" (./spec.rb:27)>
     # ./spec.rb:28:in `block (3 levels) in <top (required)>'
```

My bad. Ignore me.

PR #1898, opened 2019-07-24
Memory profiling fix, Refactor ValidatorFactory

Bhacaz:
While doing some profiling for my organisation Rails app, l have found some quick wins to improve the memory of this gem. 

The major change was to refactor the `Grape::Validations::ValidatorFactory` to use singleton methods instead of instances.

I have done some tests to see how many bytes I can save by params.

```ruby
require 'memory_profiler'

report = MemoryProfiler.report do
  require 'grape'

  class EchoAPI < Grape::API
    version 'v1'
    format :json
    prefix :api

    params do
      requires :message, type: String # I use a loop to try many param
    end
    post :echo do
      params[:message]
    end
  end
end

report.pretty_print(to_file: 'grape.txt')
```

### 1 param
```
# GEM
# allocated memory by gem
# -----------------------------------
#     450750  grape-1.2.4
#
# AFTER REFACTOR
# Total allocated: 22411303 bytes (191577 objects)
# Total retained:  1837180 bytes (14401 objects)
#
# allocated memory by gem
# -----------------------------------
#    428248  grape/lib
#
# Saved allocated: 450750 - 428248 = 22502 bytes (0.022 mb)
```

### 10 params
```
# GEM
# allocated memory by gem
# -----------------------------------
#     493062  grape-1.2.4
#
# AFTER REFACTOR
# allocated memory by gem
# -----------------------------------
#     461488  grape/lib
#
# Saved allocated: 493062 - 461488 = 31574 bytes (0.032 mb)
```
### 1000 params
```
# GEM
# allocated memory by gem
# -----------------------------------
#    5095446  grape-1.2.4
#
# AFTER REFACTOR
# allocated memory by gem
# -----------------------------------
#    4065952  grape/lib
#
# Saved allocated: 5095446 - 4065952 = 1029494 bytes (1.029 mb)
```
The result is ~1010 bytes per param.
mb save per params = **0.00101** * x + 0.0213

### DIFF
--- a/lib/grape/dsl/validations.rb
+++ b/lib/grape/dsl/validations.rb
@@ -27,10 +27,11 @@ module Grape
           setting = description_field(:params)
           setting ||= description_field(:params, {})
           Array(names).each do |name|
-            setting[name[:full_name].to_s] ||= {}
-            setting[name[:full_name].to_s].merge!(opts)
+            full_name_string = name[:full_name].to_s
+            setting[full_name_string] ||= {}
+            setting[full_name_string].merge!(opts)
 
-            namespace_stackable(:params, name[:full_name].to_s => opts)
+            namespace_stackable(:params, full_name_string => opts)
           end
         end
       end
--- a/lib/grape/dsl/validations.rb
+++ b/lib/grape/dsl/validations.rb
@@ -27,11 +27,11 @@ module Grape
           setting = description_field(:params)
           setting ||= description_field(:params, {})
           Array(names).each do |name|
-            full_name_string = name[:full_name].to_s
-            setting[full_name_string] ||= {}
-            setting[full_name_string].merge!(opts)
+            full_name = name[:full_name].to_s
+            setting[full_name] ||= {}
+            setting[full_name].merge!(opts)
 
-            namespace_stackable(:params, full_name_string => opts)
+            namespace_stackable(:params, full_name => opts)
           end
         end
       end
--- a/lib/grape/validations/validator_factory.rb
+++ b/lib/grape/validations/validator_factory.rb
@@ -2,11 +2,11 @@ module Grape
   module Validations
     class ValidatorFactory
       def self.create_validator(**options)
-        options.delete(:validator_class).new(options[:attributes],
-                                             options[:options],
-                                             options[:required],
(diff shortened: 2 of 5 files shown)


lib/grape/dsl/validations.rb
```diff
@@ -27,10 +27,11 @@ def document_attribute(names, opts)
           setting = description_field(:params)
           setting ||= description_field(:params, {})
           Array(names).each do |name|
-            setting[name[:full_name].to_s] ||= {}
-            setting[name[:full_name].to_s].merge!(opts)
+            full_name_string = name[:full_name].to_s
```
dblock: I am going to nitpick, can we just call this variable `full_name` please?

lib/grape/validations/validator_factory.rb
```diff
@@ -1,17 +1,12 @@
...
-      end
-
-      def create_validator
-        @validator_class.new(@options[:attributes],
-                             @options[:options],
-                             @options[:required],
-                             @options[:params_scope],
-                             @options[:opts])
+      def self.create_validator(**options)
+        options.delete(:validator_class).new(options[:attributes],
```
dblock: Are we modifying the value of caller's `options` here? We need to `dup` here before doing that.

Bhacaz: The initializer was doing the [same thing](https://github.com/ruby-grape/grape/pull/1898/files/549e1bd84bd39dd63357d12a90a39d23b900279f#diff-8fe683a7109bdcb8bcea665a775a876cL5).

I don't think there is something to worried here, but I can miss something...

```ruby
@validator_class = options.delete(:validator_class)
```

dblock: I see. We should change it still because that typically has unexpected side effects (we're modifying the caller's hash table).

dblock: This should just be `options[:validator_class].new`, why are we `delete`-ing it at all?

Bhacaz: I don't know `delete` is used... I'll change it just like you suggest it.

dblock:
Tests pass, so let's :shipit:

PR #1903, opened 2019-08-30
Fix Hash params renaming

bikolya:
Previously it wasn't possible to rename attributes that have a block: https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/parameters.rb#L134. Known side effects: renamed keys aren't removed from `params` and cause duplication there, but `declared(params)` keeps only the correct keys.

This PR doesn't fix the problem with duplication of keys in the `params` hash when a deeply nested attribute or an attribute with a block is renamed. Here we remove renamed keys https://github.com/ruby-grape/grape/blob/master/lib/grape/endpoint.rb#L327-L332, but the only place we mark keys for deletion is here: https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/params_scope.rb#L128 and it doesn't cover the case of hash/array attributes with block or even ordinary nested attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#1893](https://github.com/ruby-grape/grape/pull/1893): Allows `Grape::API` to behave like a Rack::app in some instances where it was misbehaving - [@myxoh](https://github.com/myxoh).
 * [#1898](https://github.com/ruby-grape/grape/pull/1898): Refactor `ValidatorFactory` to improve memory allocation - [@Bhacaz](https://github.com/Bhacaz).
 * [#1900](https://github.com/ruby-grape/grape/pull/1900): Define boolean for `Grape::Api::Instance` - [@Bhacaz](https://github.com/Bhacaz).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * [#1893](https://github.com/ruby-grape/grape/pull/1893): Allows `Grape::API` to behave like a Rack::app in some instances where it was misbehaving - [@myxoh](https://github.com/myxoh).
 * [#1898](https://github.com/ruby-grape/grape/pull/1898): Refactor `ValidatorFactory` to improve memory allocation - [@Bhacaz](https://github.com/Bhacaz).
 * [#1900](https://github.com/ruby-grape/grape/pull/1900): Define boolean for `Grape::Api::Instance` - [@Bhacaz](https://github.com/Bhacaz).
+* [#1903](https://github.com/ruby-grape/grape/pull/1903): Allow nested params renaming (Hash/Array) [@bikolya](https://github.com/bikolya).
```
dblock: You're missing a `-`, so Danger bot complains.

Issue #1902, opened 2019-08-29
How to run the compilation at startup rather than first request


viraptor:
I'm having an issue where the first request (or a number of first requests - one per worker) takes much longer than the rest. That extra time is spent running `Grape::API.compile`.

I'd like to find a way to run that process during application startup instead (I'm running grape directly from rack). Doing it before I fork workers would also help a lot.

I've done some obvious attempts without success. Is there some snippet showing how to achieve manual initialisation?

dblock:
If you hold an API instance you can call [instance.compile](https://github.com/ruby-grape/grape/blob/51123cdcc59367b2946b1df70965dcec8ddd8966/lib/grape/api/instance.rb#L50). Is the issue finding the root of your API? What have you tried? 

I tried adding this to [grape-on-rack api.rb](https://github.com/ruby-grape/grape-on-rack/blob/master/app/api.rb) and it worked.

```ruby
puts Acme::API.instance # => nil
Acme::API.compile
puts Acme::API.instance  # => #<#<Class:0x00007f9bc56ae448>:0x00007f9bc4954658>
```

I'd like a PR to this effect in the documentation once you've figured this out? Thanks.

myxoh:
I've generated a very small test, but it seems just compiling doesn't actually reduce the time of the first call.
This is my test:
```ruby
require 'benchmark'
puts Benchmark.measure {
  puts `curl localhost:9292/ping`
}
```
And the results with or without compiling on rackup are 0.015 for the first call, and ~0.0075 for all subsequent calls

this is my config.ru
```ruby
require 'rubygems'
require 'bundler/setup'
Bundler.require(:default)
require './small_api.rb'

SmallAPI.compile
run SmallAPI
```

I've also attempted
```ruby
require 'rubygems'
require 'bundler/setup'
Bundler.require(:default)
require './small_api.rb'
warmup do
  SmallAPI.compile
end

run SmallAPI
```

myxoh:
Note that warming up using Rack mock to actually call the API does the trick:

```ruby
require 'rubygems'
require 'bundler/setup'
Bundler.require(:default)
require './small_api.rb'
warmup do |app|
  client = Rack::MockRequest.new(app)
  client.get('/ping')
end

run SmallAPI
```

That's a workable work around but I'm not too thrilled about it

dblock:
I'm looking forward to someone attaching a profiler to this! :)

myxoh:
[...]

PR #1904, opened 2019-09-03
[AutoLoad] Autoload Fixes

myxoh:
Partially addresses the issues raises on: https://github.com/ruby-grape/grape/issues/1902

The way it does so is it allows a method `compile!` that makes a large percentage of the work to warm up the API on load time.

More work is needed before the API is 100% warmed up.

--------------

Taking suggestions as to how best to test this

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@
 - [Installation](#installation)
 - [Basic Usage](#basic-usage)
 - [Mounting](#mounting)
+  - [All](#all)
   - [Rack](#rack)
   - [ActiveRecord without Rails](#activerecord-without-rails)
   - [Alongside Sinatra (or other frameworks)](#alongside-sinatra-or-other-frameworks)
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ end
 ### All
 
 
-By default Grape will compile the routes on the first route, it is possible to pre-load routes using the `compile!` method
+By default Grape will compile the routes on the first route, it is possible to pre-load routes using the `compile!` method.
 
 ```ruby
 Twitter::API.compile!
--- a/README.md
+++ b/README.md
@@ -261,12 +261,30 @@ end
 
 ## Mounting
 
+### All
+
+
+By default Grape will compile the routes on the first route, it is possible to pre-load routes using the `compile!` method
+
+```ruby
+Twitter::API.compile!
+```
+
+This can be added to your `config.ru` (if using rackup), `application.rb` (if using rails), or any file that loads your server.
+
 ### Rack
 
 The above sample creates a Rack application that can be run from a rackup `config.ru` file
 with `rackup`:
 
 ```ruby
+run Twitter::API
+```
+
+(With pre-loading you can use)
+
+```ruby
+Twitter::API.compile!
 run Twitter::API
 ```
 
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -112,6 +112,7 @@ module Grape
       end
 
       private
+
       def instance_for_rack
         if never_mounted?
           base_instance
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -261,6 +262,17 @@ end
 
 ## Mounting
 
+### All
+
+
+By default Grape will compile the routes on the first route, it is possible to pre-load routes using the `compile!` method
```
dblock: Add a period

lib/grape/api.rb
```diff
@@ -111,8 +106,21 @@ def method_missing(method, *args, &block)
...
+        instance_for_rack.compile! # See API::Instance.compile!
+      end
+
       private
 
+      def instance_for_rack
+        if never_mounted?
+          base_instance
+        else
+          mounted_instances.first
```
dblock: Should this be all instances and the method to return an array?

lib/grape/api/instance.rb
```diff
@@ -79,9 +79,13 @@ def cascade(value = nil)
           end
         end
 
+        def compile!
+          LOCK.synchronize { compile } unless instance
```
dblock: Shouldn't you be checking for instance inside the synchronize?

```ruby
def compile!
    return if instance
    LOCK.syncrhonize { compile unless instance } 
 end
```

myxoh: that does make sense, even though it's not what was happening on the original method

lib/grape/api.rb
myxoh: hmm. I can't remember why this should ever return one of the mounted instances, as you probably want to always be interacting with rack on a predictable instance. I'll have to see what happens if I have this always be the `base_instance`

myxoh: OK so this change was initially introduced in: https://github.com/ruby-grape/grape/pull/1893/files 
The reason why this is needed is that we used to support - before remounting was introduced -  the ability to pre-mount the API  you will run in a namespace (and I was maintaing backwards compatibility).

Basically you are allowed to do
```ruby
class ToBeMountedinRack < Grape::API
  get 'endpoint'
end

class Namespace < Grape::API
  namespace 'namespace' do 
    mount ToBeMountedinRack
   end
end

run ToBeMountedinRack
```

And the expectation is that you will be able to get `/namespace/endpoint` 
I am not sure I want to maintain this behaviour forever, as it doesn't really play nicely with re-mounting an endpoint and I don't find it very sensible (I would find it sensible to `run Namespace` if you want the above behavour, rather than mounting `ToBeMountedinRack`)

Nonetheless, I don't think this is the appropiate PR to introduce this change

dblock: Agreed. We can deprecate this separately.

lib/grape/api/instance.rb
```diff
@@ -79,9 +79,13 @@ def cascade(value = nil)
           end
         end
 
+        def compile!
+          LOCK.synchronize { compile unless instance }
```
dblock: I think we want to also `return if instance` here to avoid taking the lock every time?

myxoh: Forgot about it, but have now resolved

dblock:
Ok, I'm good with this. Let's add a test? Something that calls `compile!` explicitly?

dblock:
I merged this.

spec/grape/api_spec.rb
```diff
@@ -885,6 +885,40 @@ class DummyFormatClass
...
 
+  describe '.compile!' do
+    it 'requires the grape/eager_load file' do
+      expect(app).to receive(:require).with('grape/eager_load') { nil }
+      app.compile!
+    end
+
+    it 'compiles the instance for rack!' do
+      stubbed_object = double(:instance_for_rack)
+      allow(app).to receive(:instance_for_rack) { stubbed_object }
```
darren987469: @myxoh I am curious about the purpose of this test case. It seems to test nothing?

PR #1909, opened 2019-09-27
Use defined? to correct not initialized warnings

nbeyer:
The class instances variables `@versions` and `@namespace_description` may not be initialized when used and Ruby will report their access as a warning. Using the defined? keyword will correct that. There are plenty of alternative changes could be made to avoid this, but I just picked the most minimal change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 #### Fixes
 
+* Your contribution here.
 * [#1909](https://github.com/ruby-grape/grape/pull/1909): Use defined? to correct not initialized warnings - [@nbeyer](https://github.com/nbeyer).
 * [#1893](https://github.com/ruby-grape/grape/pull/1893): Allows `Grape::API` to behave like a Rack::app in some instances where it was misbehaving - [@myxoh](https://github.com/myxoh).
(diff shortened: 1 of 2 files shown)


dblock:
I was never able to see these. How do you reproduce the warnings?

Needs a CHANGELOG entry. There's also something about the octokit plugin & danger going south, can you please take a look and make sure the build is green? https://github.com/intridea/hashie/pull/486 was having a similar problem.

nbeyer:
From the command-line, warnings can be turned on by adding the '-w' switch.

From RSpec, warnings can be turned on in the config (i.e. spec_helper.rb), like this:
```ruby
RSpec.configure do |config|
  # ...
  config.warnings = true
  # ...
end
```

nbeyer:
@dblock I have the builds running cleanly now. Let me know if you'd like me to add the warnings enablement to the spec_helper.rb.

Note - this method redefinition causes a lot of warnings - https://github.com/ruby-grape/grape/blob/master/lib/grape/api.rb#L57

dblock:
@nbeyer Are you saying we can have a clean test run without warnings? If so we should turn on warnings as errors (the answer to your question is "yes, please").

dblock:
@nbeyer want to finish this?

nbeyer:
Abandoning this for #1995

PR #1911, opened 2019-10-01
AtLeastOneOfValidator properly treats nested params in errors

### LINKED ISSUE #1849: AtLeastOneOfValidator doesn't provide the full name of attribute in errors
Currently, AtLeastOneOfValidator only provides the minimal attribute name on a validation error. This is different from other validators where we use the full name. As an example, let's consider the following parameters:

```ruby
  requires :needed_attr
  optional :attr_1
  optional :attr_2
  at_least_one_of :attr1, :attr2
end
```
If `address[needed_attr]` isn't set, the hash key for that attribute's errors is `address[needed_attr]`. On the other hand, the hash key for errors on `address[attr1]`  and `address[attr2]` is `['attr1', 'attr2']`. This is inconsistent. Is this expected behavior? If not, I'd be willing to submit a pull request to fix it.

dnesteryuk:
Unlike other validators `Grape::Valiations::AtLeastOneOfValidator` didn't wrap params into a scope (namespace) for errors when params were nested.

Closes #1849

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1911](https://github.com/ruby-grape/grape/pull/1911): Make sure `Grape::Valiations::AtLeastOneOfValidator` properly treats nested params in errors - [@dnesteryuk](https://github.com/dnesteryuk).
 * [#1893](https://github.com/ruby-grape/grape/pull/1893): Allows `Grape::API` to behave like a Rack::app in some instances where it was misbehaving - [@myxoh](https://github.com/myxoh).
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,7 +18,7 @@
 
 #### Features
 
-* [#1888](https://github.com/ruby-grape/grape/pull/1888): Makes the `configuration` hash widly available - [@myxoh](https://github.com/myxoh).
+* [#1888](https://github.com/ruby-grape/grape/pull/1888): Makes the `configuration` hash widely available - [@myxoh](https://github.com/myxoh).
```
dblock: Thanks for this.

dblock:
👍 merged

PR #1912, opened 2019-10-01
Upgrade rack-test to 1.1.0

### LINKED ISSUE #1656: Upgrade to rack-test 0.7.0

dnesteryuk:
As it was mentioned in #1658 empty arrays aren't ignored anymore, so specs should be updated.

Is a note in the changlelog required for this PR? It isn't a new feature, neither a bugfix :thinking:  

Closes #1656


dblock:
Merged. For the future it's OK to mention this in CHANGELOG.

PR #1913, opened 2019-10-04
Fix error messages for nested multiple params validations

bikolya:
Previously multiple params validators returned messages without considering the scope.
For example, in case of hash params:
`{ "param_1,param_2": 'error message' }`
instead of:
`{ "hash[param_1],hash[param_2]": 'error message' }`

Or in case of arrays, indexes were ignored completely:
`{ "param_1,param_2": 'error message' }`
instead of:
`{ "array[1][param_1],array[1][param_2]": 'error message' }`

This PR fixes it by using `AttributesIterator` in a similar way to `Grape::Validations::Base`.

Also the specs are refactored to work in a similar way to the ordinary single-parameter validations.

### DIFF
--- a/lib/grape/validations/attributes_iterator.rb
+++ b/lib/grape/validations/attributes_iterator.rb
@@ -5,12 +5,11 @@ module Grape
 
       attr_reader :scope
 
-      def initialize(validator, scope, params, multiple_params: false)
+      def initialize(validator, scope, params)
         @scope = scope
         @attrs = validator.attrs
         @original_params = scope.params(params)
         @params = Array.wrap(@original_params)
-        @multiple_params = multiple_params
       end
 
       def each(&block)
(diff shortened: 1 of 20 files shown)


lib/grape/validations/attributes_iterator.rb
```diff
@@ -5,11 +5,12 @@ class AttributesIterator
 
       attr_reader :scope
 
-      def initialize(validator, scope, params)
+      def initialize(validator, scope, params, multiple_params: false)
```
dblock: How about writing another attribute iterator, `MutlipleParamsAttributeIterator` instead of this additional parameter?

lib/grape/validations/validators/multiple_params_base.rb
```diff
@@ -1,26 +1,32 @@
...
+          rescue Grape::Exceptions::Validation => e
+            array_errors << e
+          end
+        end
 
-      def scope_requires_params
-        @scope.required? || scoped_params.any? { |param| param.respond_to?(:any?) && param.any? }
+        raise Grape::Exceptions::ValidationArrayErrors, array_errors if array_errors.any?
       end
+      # rubocop:enable HashEachMethods
```
dblock: `rubocop -a ; rubocop --auto-gen-config` to prevent Rubocop from being annoying, don't skip this here.

dblock:
This is solid work. See above for my comments. Needs a green build, please.

lib/grape/validations/attributes_iterator.rb
bikolya: Great idea, thanks! I changed this part, please check it out. Also the build is green now.

dblock:
Perfect, merged.

PR #1914, opened 2019-10-04
Run specs in random order

splattael:
This PR enables spec randomization by default. It also fixes a couple of failures discovered after enabling `--order=rand` :nerd_face:

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --color
 --format=documentation
+--order=rand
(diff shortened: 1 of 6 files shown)


dblock:
This is great, thank you. Do add yourself to CHANGELOG, this is important work. Squash, I'll merge.

splattael:
I also wanted to add a `I18nSupport` and a RSpec tag `:i18n` to streamline locale specs but this can be a follow-up PR :+1: 

Thanks a lot @dblock :bowing_man:

PR #1916, opened 2019-10-13
Drop old appraisals

NikolayRys:
Rails 3 and 4 appraisals have been excluded from travis build for a year: https://github.com/ruby-grape/grape/commit/9c942f4814b4e0e95bb721a59f97b8a4a37b2347#diff-354f30a63fb0907d4ad57269548329e3

The recent update to rack-test: https://github.com/ruby-grape/grape/pull/1912/files#diff-0bd0610801e75105f6818e38cc85c172 made the appraisals for rails_3 and rails_4 harder to fix, because they use old rack-spec of their own.

Probably it's time to get rid of them. It will allow to run cleanly run `appraisal rake spec`.
Please, let me know what you think.

UPD: also removing a work-around specific for rails_3 in `spec/grape/integration/rack_spec.rb`  More details: https://github.com/ruby-grape/grape/commit/3115c6ec513d4b36a57c943a7cc5f22e05535472


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1916](https://github.com/ruby-grape/grape/pull/1916): Drop old appraisals - [@NikolayRys](https://github.com/NikolayRys).
```
dblock: Nitpick: be explicit? How about "drop Rails 3 and 4 support"? If we're no longer testing, we aren't supporting.

NikolayRys: No, probs, I reword it, but initially my assumption was that you don't support for a long time(since you have removed them from the travis build), so what I did was just a clean-up.

dblock: Oh I see. I'm good with this.

PR #1917, opened 2019-10-13
Update access to rack constant

NikolayRys:
A small fix to keep up with the new changes in rack: https://github.com/rack/rack/commit/6746515453a44cc6268c46d525158278cf96307d#diff-7c1a24d5b2fe58a6f925c7cacc6c55e7L558-R558

The format of the object with the possible statuses has changed in a recent release, which breaks the rack-edge appraisal, so this PR fixes it.

### DIFF
--- a/spec/grape/middleware/formatter_spec.rb
+++ b/spec/grape/middleware/formatter_spec.rb
@@ -213,7 +213,13 @@ describe Grape::Middleware::Formatter do
   context 'no content responses' do
     let(:no_content_response) { ->(status) { [status, {}, ['']] } }
 
-    Rack::Utils::STATUS_WITH_NO_ENTITY_BODY.each do |status|
+    STATUSES_WITHOUT_BODY = if Gem::Version.new(Rack.release) >= Gem::Version.new('2.1.0')
+                              Rack::Utils::STATUS_WITH_NO_ENTITY_BODY.keys
+                            else
+                              Rack::Utils::STATUS_WITH_NO_ENTITY_BODY
+                            end
+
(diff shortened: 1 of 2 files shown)


spec/grape/middleware/formatter_spec.rb
```diff
@@ -213,7 +213,13 @@ def to_xml
   context 'no content responses' do
     let(:no_content_response) { ->(status) { [status, {}, ['']] } }
 
-    Rack::Utils::STATUS_WITH_NO_ENTITY_BODY.each do |status|
+    STATUSES_WITHOUT_BODY = if Rack::RELEASE >= '2.1.0'
```
dblock: Use `Gem::Version.new` to compare, otherwise we're going to be in trouble with version 11.

https://medium.com/little-programming-joys/how-to-compare-version-strings-in-ruby-ced0916a4b71

NikolayRys: Sure, I'll fix it. 
You did in this way here: https://github.com/ruby-grape/grape/blob/master/spec/grape/integration/rack_spec.rb#L23 So I've assumed it's how you would want it here too.

dblock: That link doesn't go anywhere interesting ;) If you see it done elsewhere, please also fix!

NikolayRys: It did until you've merged my other PR :)
But I'll fix this today, thank you

NikolayRys:
Updated as suggested.

PR #1918, opened 2019-10-16
Helper method to access controller context in middleware

NikolayRys:
This is a feature suggestion.

Recently, working with Grape and writing some `rescue_from` handlers, I've encountered a need to access the request params, headers, and some other stuff how it was in the original controller method.

It was possible, but there was no documented and elegant way to do it. And I needed to go through the Grape source code to figure out a way to reach it. 

With this change, we can save effort and time for other clients by preventing the need to rediscover this. It is useful not only for the error handling, but also for every piece of the middleware, including formatters, and any custom middleware inherited from Grape::Middleware::Base.

This PR adds the actual helper method for accessing the context, tests, and documentation for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2497,6 +2497,17 @@ class Twitter::API < Grape::API
 end
 ```
 
+Inside the `rescue_from` block, the environment of the original controller method(`.self` receiver) is accessible through the `#context` method.
+
+```ruby
+class Twitter::API < Grape::API
+  rescue_from :all do |e|
+    user_id = context.params[:user_id]
+    error!("error for #{user_id}")
+  end
+end
+```
+
 #### Rescuing exceptions inside namespaces
 
 You could put `rescue_from` clauses inside a namespace and they will take precedence over ones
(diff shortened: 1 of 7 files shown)


dblock:
This introduces yet another way to access the same thing, so I think I am against it. I think you want to be able to do `params[:x]` or `request.params[:x]` directly like inside any API. What do you think?

dblock:
Ok, I buy this. A module would be helpful, I'd merge that.

NikolayRys:
Awesome, I'll adjust the PR

NikolayRys:
As requested, moved it out to a common module.

README.md
```diff
@@ -3121,6 +3132,8 @@ end
 
 Use [Doorkeeper](https://github.com/doorkeeper-gem/doorkeeper), [warden-oauth2](https://github.com/opperator/warden-oauth2) or [rack-oauth2](https://github.com/nov/rack-oauth2) for OAuth2 support.
 
+You can access the controller context with the `#context` method inside any auth middleware inherited from `Grape::Middlware::Auth::Base`.
```
myxoh: Can we clarify here why you'd want to use the `#context`. For example
```
you can access the controller params, headers, and helpers through the context with the `#context` method inside any auth middleware inherited from `Grape::Middlware::Auth::Base`.
```

myxoh: (Unless I'm mistaken and you can't access the helpers)

NikolayRys: Yes, the helpers are accessible :) 
I'll improve the wording

NikolayRys: Done

dblock:
Perfect.

PR #1920, opened 2019-10-18
Replace Virtus with dry-types

### LINKED ISSUE #1347: Switch from virtus to dry-types
I don't really know what you do with `virtus` (I think it's coercion) but I think it could be a good idea to replace `virtus` by `dry-types` (http://dry-rb.org/gems/dry-types/) for many reasons: 
1. This reddit post from Solnic, the virtus creator: https://www.reddit.com/r/ruby/comments/3sjb24/virtus_to_be_abandoned_by_its_creator/
2. Performance (from the dry-types doc): "roughly 10-12x faster than Virtus"

dnesteryuk:
As we know Virtus isn't maintained anymore, so it makes sense to use something different for coercing. There is a [dry-types](https://dry-rb.org/gems/dry-types/1.2/) lib which was created as a replacement for Virtus (at least a part of it). Although, it is only about coercion.

This change gets rid of Virtus and adds dry-types to the stack. Almost, all tests pass. However, there are a few tests which were modified, actually, they didn't make any sense. Yeah, Virtus was providing
some "magic" to make them green, but I think they were unreal.

The structure inside didn't change match. I tried to integrate dry-types without huge refactoring.

If people want to use Virtus for custom types, it is possible after implementing a parse method which initializes a model.

```ruby
class User
  include Virtus.model

  attribute :id, Integer
  attribute :name, String

  def self.parse(attrs)
    new(attrs)
  end
end
```

This PR is kind of "work in progress", I would like to gather feedback then proceed. If somebody can add this change to their project and launch tests, it will be awesome, the goal is to avoid breaking changes.

Missing parts:

 - [x] changelog
 - [x] upgrade instruction (at least about Virtus models as custom type)
 - [x] solve an issue with `Hashie::Mash`
 - [x] Grape offences 
 - [x] Deprecate Ruby 2.3

It closes #1347.


myxoh:
Following this, shout if you get stuck or need some help

dblock:
👀

Gemfile
```diff
@@ -9,6 +9,7 @@ group :development, :test do
   gem 'hashie'
   gem 'rake'
   gem 'rubocop', '0.51.0'
+  gem 'virtus'
```
dblock: This can't be used at all in the final PR.

UPGRADING.md
```diff
@@ -1,6 +1,39 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.2.5
+
+[Virtus](https://github.com/solnic/virtus) isn't used for coercing anymore. If your project depends on Virtus, you have to add it to your `Gemfile`. Also, if Virtus is used as a custom type
```
dblock: Simlify and shorten, "is no longer used for coercing parameters" and just ", add it to your Gemfile"

UPGRADING.md
```diff
@@ -1,6 +1,39 @@
...
+
+  attribute :id, Integer
+  attribute :name, String
+
+  def self.parse(attrs)
+    new(attrs)
+  end
+end
+```
+
```
dblock: I think we want to explain how to rewrite a custom type with dry-types here, first. If I am upgrading I'd rather not use Virtus unless I can't.

lib/grape/validations/types/custom_type_coercer.rb
```diff
@@ -160,10 +122,15 @@ def enforce_symbolized_keys(type, method)
...
-                new_value.map do |item|
-                  item.is_a?(Hash) ? symbolize_keys(item) : item
+              # if the value cannot be parsed, return InvalidValue
+              begin
+                method.call(val).tap do |new_val|
+                  new_val.map do |item|
+                    item.is_a?(Hash) ? symbolize_keys(item) : item
+                  end
                 end
+              rescue StandardError => _e
```
dblock: This is too broad of a catch-all, I think should be a parser error only.

lib/grape/validations/types/dry_type_coercer.rb
```diff
@@ -0,0 +1,36 @@
+require 'dry-types'
+
+module DryTypes
+  include Dry.Types()
```
dblock: Remove `()`, or maybe just use `Dry.Types....` everywhere. It's liberally used below as either.

Gemfile
dnesteryuk: There are a few tests which use Virtus to verify that it can be used as a custom type, it means I need to remove the :thinking:

UPGRADING.md
dnesteryuk: Dry-types cannot be used in place of Virtus for custom types, [dry-schema](https://dry-rb.org/gems/dry-schema/1.4/) might be. So, if users need something like Virtus they need to keep  it or migrate to dry-schema. I guess it is up to them to decide, the goal is to avoid breaking changes, so if they use Virtus, they can still use it until they have time to migrate.

lib/grape/validations/types/custom_type_coercer.rb
dnesteryuk: We don't know which error might be raised by a custom type. Users define it. :slightly_smiling_face:

lib/grape/validations/types/dry_type_coercer.rb
dnesteryuk: I would love to do that, but it is how dry-types [should be used](https://dry-rb.org/gems/dry-types/1.2/getting-started/). After calling `Dry.Types()` [registered types get added to a container](https://github.com/dry-rb/dry-types/blob/master/lib/dry/types.rb#L160-L212), it involves adding [constants](https://github.com/dry-rb/dry-types/blob/master/lib/dry/types/module.rb#L102-L117) and some other meta-programming.

Gemfile
dblock: I worry that we'll leave something that works in tests but not in prod. Maybe those can become integration tests with their own Gemfile?

UPGRADING.md
dblock: It sounds like you're saying that if you want a custom type you *have* to use Virtus? Or are you saying you *can* use Virtus? Either way let's make this very clear in the upgrading/readme doc.

lib/grape/validations/types/custom_type_coercer.rb
dblock: That's why it's a problem. Catch all could be something you actually don't expect and there's no way to know if you catch all. We should not be catching all here at all or having an explicit contract with a coercer. If you want the latter, introduce an exception such as CoercionError and rescue that. Document that coercers should throw this or a child of this exception.

lib/grape/validations/types/dry_type_coercer.rb
dblock: Ok, I think I am missing something. I am staring at dry-types types.rb code and I don't get it. How does it make `Dry.Types` different from `Dry.Types()`?

Gemfile
dnesteryuk: You are right, Virtus isn't needed anymore. I removed it and adjusted tests testing custom types :slightly_smiling_face:

UPGRADING.md
dnesteryuk: No, Virtus isn't required for custom types. As it was before, any class implementing the `parse` method might be used. :slightly_smiling_face: I've added an additional sentence to the upgrading doc.

lib/grape/validations/types/custom_type_coercer.rb
dnesteryuk: You are right, this `begin/rescue` isn't required. Most likely I added to fix some broken tests, after refactoring, it became useless.

lib/grape/validations/types/dry_type_coercer.rb
dnesteryuk: [Dry.Types](https://github.com/dry-rb/dry-types/blob/master/lib/dry/types.rb#L210) is a method which returns an instance of a module to be included. :slightly_smiling_face:  Here is [that module](https://github.com/dry-rb/dry-types/blob/master/lib/dry/types/module.rb#L20).

dnesteryuk:
Dry-types requires Ruby 2.4 or greater :disappointed:

dblock:
> Dry-types requires Ruby 2.4 or greater 😞

Deprecate Ruby 2.3 with a note, changelog, readme change, etc.

dnesteryuk:
Hi @dblock I made recommended changes, please, have a look again. :relaxed: Thanks for your patience! :smile:

README.md
```diff
@@ -167,6 +167,8 @@ The current stable release is [1.2.4](https://github.com/ruby-grape/grape/blob/v
 
 ## Installation
 
+After **1.2.5**, Ruby 2.4.0 or newer is required.
```
dblock: This should go into UPGRADING, for this doc just state that Ruby 2.4 or newer is required.

UPGRADING.md
```diff
@@ -1,6 +1,47 @@
...
+  attribute :name, String
+end
+
+# somewhere in your API
+params do
+  requires :user, type: User
+end
+```
+
+you need to add a class-level `parse` method to the model:
```
dblock: Start with a cap, "Add a ..."

UPGRADING.md
```diff
@@ -1,6 +1,47 @@
...
+    new(attrs)
+  end
+end
+```
+
+Custom types which don't depend on Virtus don't require any changes.
+
+#### Ruby
+
+After adding dry-types, Ruby 2.4.0 or newer is required.
```
dblock: Let's put this section on top of the list?

UPGRADING.md
```diff
@@ -1,6 +1,47 @@
...
+  attribute :id, Integer
+  attribute :name, String
+
+  def self.parse(attrs)
+    new(attrs)
+  end
+end
+```
+
+Custom types which don't depend on Virtus don't require any changes.
```
dblock: Add a "For more information see ..." and link to this PR.

dnesteryuk:
@dblock I've made changes to `README` and `UPGRADING`. Please, have a look.

dblock:
I'm good with this! @myxoh want to confirm and merge?

dblock:
Let's get this build to green? There's some legit looking failure with rack-edge.

UPGRADING.md
```diff
@@ -1,6 +1,49 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.2.5
```
dblock: Let's make this 1.3.0.

dnesteryuk:
@dblock [that commit](https://github.com/rack/rack/commit/e702d31335c1a820e99c3acdd9d3368ac25da010#diff-20bf6f597c10712d63e5c160a19c9a79) broke Rack :disappointed:

spec/grape/api_spec.rb
```diff
@@ -1133,7 +1133,7 @@ class DummyFormatClass
 
       subject.use Rack::Chunked
       subject.get('/stream') { stream test_stream }
-      get '/stream', {}, 'HTTP_VERSION' => 'HTTP/1.1'
+      get '/stream', {}, 'HTTP_VERSION' => 'HTTP/1.1', 'SERVER_PROTOCOL' => 'HTTP/1.1'
```
dnesteryuk: @dblock Here is how I fixed the problem with Rack

dblock: That's totally fine, we just want to make sure Grape is compatible.

dblock:
I merged this, thank you so much for doing this work @dnesteryuk, it's a big one. I suggest you email the mailing list describing this change and asking people to give the master branch a try, especially if they do custom type checking.

FunkyloverOne:
@dblock I've just tried to run a test suite on one of my projects against your master branch, and all green :grin: There's no custom type checking though.

dnesteryuk:
@dblock where is the mailing list? :grimacing:

dblock:
> @dblock where is the mailing list? 😬

http://groups.google.com/group/ruby-grape

nbulaj:
Any reason why we tied to version 1.1.x of dry-types? @dnesteryuk 

I tried Grape from the `master` branch with Ruby 2.7 and see **a lot** of deprecation messages like:

```
/.../ruby-2.7.0/gems/dry-types-1.1.1/lib/dry/types/any.rb:18: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

Even if it's already fixed or would be fixed in the future versions of Dry-types we can't update to it due to version restrictions :(

dblock:
> Any reason why we tied to version 1.1.x of dry-types? @dnesteryuk I know it changes even with patch version changes, but ... I don't think it's a good solution btw

We should relax it, please PR.

Issue #1901, opened 2019-08-21
Request Headers section in README seems misleading


FunkyloverOne:
Perhaps it is related to #1770.

So the README says that an example below will be  normalized:
```
curl -H "secret_PassWord: swordfish" ...
```
But actually - it won't. If a header name contains an underscore - in `env` it will remain as is and without `HTTP_` prefix, and it won't appear in `headers` helper at all.

dblock:
Please fix?

myxoh:
I am adding some integration specs, not sure whether we want to merge them, but they do demonstrate a different behaviour.

Test here: https://github.com/ruby-grape/grape/pull/1927

FunkyloverOne:
Interesting, I have such issues in a Rails app, with Grape API mounted in `routes.rb`. And code like `env.to_json` does not even work there, it ends up with an infinite loop.

Perhaps that's some new `activesupport-5.2.4` bug though.

However, I can try `env.keys` instead, and my initial issue is still reproducing. Perhaps there's something else is wrong with the project. I gotta dive deeper into it to figure it out.

PR #1927, opened 2019-10-30
Actually mounts and curls an API

myxoh:
This is a way of doing and end-end test of grape with a sample API and actually performing API calls against it. 

I am open to suggestions.

Tests the issue described in: https://github.com/ruby-grape/grape/issues/1901

### DIFF
--- a/spec/integration/real_api/a_real_api_spec.rb
+++ b/spec/integration/real_api/a_real_api_spec.rb
@@ -1,8 +1,8 @@
 describe 'Mounting a real API' do
   let!(:server_thread) do
     Thread.new do
-      `fuser -k 9292/tcp`
-      `rackup spec/integration/real_api/example_api/config.ru`
+      `fuser -k 9292/tcp  2>/dev/null`
+      system("rackup spec/integration/real_api/example_api/config.ru 2>/dev/null")
     end
   end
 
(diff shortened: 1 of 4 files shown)


dblock:
I like it! Would make sure those command line errors are handled, I believe `system` doesn't properly error out, so you want to be checking for exit code.

myxoh:
I've got rid of the logs. Not sure whether there was something else you were thinking about. Were you thinking about sig trapping? 
```ruby
  Signal.trap("TERM") do
    shutdown()
  end
```

spec/integration/real_api/a_real_api_spec.rb
```diff
@@ -0,0 +1,49 @@
+describe 'Mounting a real API' do
+  let!(:server_thread) do
+    Thread.new do
+      `fuser -k 9292/tcp  2>/dev/null`
+      system("rackup spec/integration/real_api/example_api/config.ru 2>/dev/null")
```
dblock: I meant that this call will fail without an exception if, for example, `rackup` cannot run. You need to handle the return from it.

myxoh: :+1:

dblock:
Rebase/fixup?

myxoh:
Right. I need to spend some time getting this right, but it's been a very busy month.
Let me close it for now and get back to it

PR #1931, opened 2019-11-14
[LazyBlock] Allows using complex expressions with configuration

myxoh:
This PR is motivated by a bug I've introduced on our system.
The problem, when you are trying to waterfall a configuration, is that `LazyValue` doesn't play nicely inside an expression so this code
 ```ruby
value = configuration[:attribute] || 'default_value' 
```
Is evaluated when the class is load as: configuration[:attribute] => LazyValue || 'default_value' => LazyValue
Which means that even when you mount the API with `attribute: nil` the value will equal `nil` rather than `'default_value'` as a developer might expect.

To mitigate this, I've introduced a lazy block, so you are able to put that expression within a `mounted` block which will only be evaluated when the API is mounted, and furthermore, will prevent methods using the response of the expression to running before the API is mounted

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,6 +485,42 @@ class ConditionalEndpoint::API < Grape::API
 end
 ```
 
+More complex results can be achieved by using mounted as an expression within which the `configuration` is already evaluated as a Hash.
+
+```ruby
+class ExpressionEndpointAPI < Grape::API
+  get(mounted { configuration[:route_name] || 'default_name' }) do
+    # some logic
+  end
+end
+```
+
+```ruby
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -485,6 +485,42 @@ class ConditionalEndpoint::API < Grape::API
 end
 ```
 
+More complex results can be achieved by using mounted as an expression within which the `configuration` is already evaluated as a Hash.
```
dblock: Quote `mounted`.

dblock:
👍

dblock:
Merged

PR #1937, opened 2019-11-29
Fix bloat in released gem.

### LINKED ISSUE #1936: Gem bloat
Version 1.2.x has unnecessary files in the gem:

```
ls -lah ./vendor/bundle/ruby/2.6.0/gems/grape-1.2.4/pkg/
total 3M     
drwxr-xr-x    2 root     root        4.0K Nov 28 10:33 .
drwxr-xr-x    7 root     root        4.0K Nov 28 10:33 ..
-rw-rw-r--    1 root     root      638.0K Nov 28 10:33 grape-1.2.0.gem
-rw-rw-r--    1 root     root      851.0K Nov 28 10:33 grape-1.2.1.gem
-rw-rw-r--    1 root     root        1.7M Nov 28 10:33 grape-1.2.3.gem
```

dblock:
Closes #1936.

### DIFF
--- a/grape.gemspec
+++ b/grape.gemspec
@@ -19,7 +19,9 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'rack-accept'
   s.add_runtime_dependency 'virtus', '>= 1.0.0'
 
-  s.files         = Dir['**/*'].keep_if { |file| File.file?(file) }
+  s.files         = %w[CHANGELOG.md CONTRIBUTING.md README.md grape.png UPGRADING.md LICENSE]
+  s.files        += %w[grape.gemspec]
+  s.files        += Dir['lib/**/*']
   s.test_files    = Dir['spec/**/*']
   s.require_paths = ['lib']
 end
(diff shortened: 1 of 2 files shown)


grape.gemspec
```diff
@@ -19,7 +19,9 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'rack-accept'
   s.add_runtime_dependency 'virtus', '>= 1.0.0'
 
-  s.files         = Dir['**/*'].keep_if { |file| File.file?(file) }
+  s.files         = %w[CHANGELOG.md CONTRIBUTING.md README.md UPGRADING.md LICENSE]
+  s.files        += %w[Rakefile grape.gemspec]
```
dm1try: it seems to me that those two are not needed for the distribution

grape.gemspec
```diff
@@ -19,7 +19,9 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'rack-accept'
   s.add_runtime_dependency 'virtus', '>= 1.0.0'
 
-  s.files         = Dir['**/*'].keep_if { |file| File.file?(file) }
+  s.files         = %w[CHANGELOG.md CONTRIBUTING.md README.md UPGRADING.md LICENSE]
```
dm1try: I would add `grape.png` as `README.md` has a local link to it

grape.gemspec
dblock: Don't you need .gemspec? How does it know dependencies otherwise?

dblock: You do need .gemspec for dependencies. That's the actual declaration of the gem, and you would get "Could not find gem 'grape' in source at `vendor/grape-1.2.5`." if you `gem unpack` what you build without it and try to `bundle install`.

dm1try: I do not need it, for sure :) this file is not a requirement for the gem distribution, though the workflow with "gem unpack" you mentioned above requires it because `path` option expects either [`gemspec` file or explicit `version` of the gem](https://bundler.io/man/gemfile.5.html#PATH)(the mentioned issue could be fixed by using `gem 'grape', '1.2.5', path: 'vendor/grape-1.2.5'`).

Anyway, I'm okay with that

PR #1938, opened 2019-12-01
Add project metadata to the gemspec

orien:
Add `bug_tracker_uri`, `changelog_uri`, `documentation_uri`, and `source_code_uri` to the gemspec metadata.

These [project metadata](https://guides.rubygems.org/specification-reference/#metadata) will facilitate easy access to project information. The URI will be available on the [Rubygems project page](https://rubygems.org/gems/grape), via the rubygems API, and the `gem` and `bundle` command-line tools with the next release.

### DIFF
--- a/grape.gemspec
+++ b/grape.gemspec
@@ -11,6 +11,12 @@ Gem::Specification.new do |s|
   s.summary     = 'A simple Ruby framework for building REST-like APIs.'
   s.description = 'A Ruby framework for rapid API development with great conventions.'
   s.license     = 'MIT'
+  s.metadata    = {
+    'bug_tracker_uri'   => 'https://github.com/ruby-grape/grape/issues',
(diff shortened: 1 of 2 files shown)


dm1try:
@orien thanks!

grape.gemspec
```diff
@@ -11,6 +11,12 @@ Gem::Specification.new do |s|
   s.summary     = 'A simple Ruby framework for building REST-like APIs.'
   s.description = 'A Ruby framework for rapid API development with great conventions.'
   s.license     = 'MIT'
+  s.metadata    = {
+    'bug_tracker_uri'   => 'https://github.com/ruby-grape/grape/issues',
```
dblock: Do these have to be strings? or `bug_tracker_uri: 'http...'` which would read a bit cleaner. I am totally nitpicking.

orien:
@dblock: Apparently they need to be strings:

> All keys and values must be Strings

https://guides.rubygems.org/specification-reference/#metadata

PR #1940, opened 2019-12-03
get rid of a needless step in HashWithIndifferentAccess

dnesteryuk:
There was an extra operation in `Grape::Extensions::ActiveSupport::HashWithIndifferentAccess::ParamBuilder` it is unnecessary, people already use Rails 4, 5 or 6. Devs who are on Rails 3 less likely will upgrade to latest Grape.

After removing that operation Grape consumes less memory.

Before: 36080 bytes
After:  32472 bytes


dblock:
So effectively by doing this we're removing support for Rails 3? Which we didn't really talk about before or test against. Maybe we want to be extra cautious and write it in UPGRADING or CHANGELOG? I'll merge this as is though.

Issue #1945, opened 2019-12-17
"`base_instance?': undefined method `base_instance' for nil:NilClass (NoMethodError)" when inheriting from Grape::API::Instance for v 1.2.5


H3JFC:
## Background
My team is working on an application that depends on Grape working with rails. We tried upgrading from 1.2.2 to 1.2.5 and couldn't get the app to start. 

We have relied heavily on inheriting from Grape::API::Instance instead of Grape::API to D.R.Y. our app up significantly with factory methods (details below). In our last upgrade we followed the guidelines regarding the usage of Grape::API::Instance instead of Grape::API [Upgrading.md](https://github.com/ruby-grape/grape/blob/master/UPGRADING.md#upgrading-to--120). This feels like a bug as there have only been 3 patch versions since this worked for us. We've tried compiling / mounting in different ways per the suggestions in the README and haven't been able to get it to work.

## Details:
[Link to Small Sample Project](https://github.com/H3JFC/rails-grape-api)
[Link to Diff (below)](https://github.com/H3JFC/rails-grape-api/commit/9b1f473a41416ca0dfad84b706631d0061367831)
### app/api/sample/api.rb
```
module Sample
  class Api < Grape::API::Instance
    format :json
    prefix :api

    def self.factory(text)
      get :test do
        text
      end
    end
    resource :test do
      factory 'hello world'
    end
  end
end
```
### config/routes.rb
```
Rails.application.routes.draw do
  mount Sample::Api => '/'
  # For details on the DSL available within this file, see https://guides.rubyonrails.org/routing.html
end
```

## Error Log
```
=> Booting Puma
=> Rails 6.0.2 application starting in development 
=> Run `rails server --help` for more startup options
Exiting
/Users/dev_user/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/grape-1.2.5/lib/grape/api/instance.rb:33:in `base_instance?': undefined method `base_instance' for nil:NilClass (NoMethodError)
	from /Users/dev_user/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/grape-1.2.5/lib/grape/api/instance.rb:122:in `evaluate_as_instance_with_configuration'
	from /Users/dev_user/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/grape-1.2.5/lib/grape/api/instance.rb:104:in `nest'
[...]

PR #1946, opened 2019-12-17
Fixes issue by checking if there's a base is there at all

nicolashopin:
Addresses issue raised here: https://github.com/ruby-grape/grape/issues/1945

The cause of the issue is the introduction of the `LazyBlock` where blocks would only be excecuted on properly mounted instances, the side effect is for people using just `Instance` they had no base at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 #### Fixes
 
-* [#1931](https://github.com/ruby-grape/grape/pull/1946): Fixes issue when using namespaces in Grape::API::Instance mounted directly - [@myxoh](https://github.com/myxoh)
+* [#1931](https://github.com/ruby-grape/grape/pull/1946): Fixes issue when using namespaces in `Grape::API::Instance` mounted directly - [@myxoh](https://github.com/myxoh).
 
 ### 1.2.5 (2019/12/01)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,7 +13,7 @@
 
 #### Fixes
 
-* Your contribution here.
+* [#1931](https://github.com/ruby-grape/grape/pull/1946): Fixes issue when using namespaces in Grape::API::Instance mounted directly - [@myxoh](https://github.com/myxoh)
```
dblock: Needs a period. Quote `Grape::API::Instance`.

lib/grape/api/instance.rb
```diff
@@ -121,7 +121,7 @@ def evaluate_as_instance_with_configuration(block, lazy: false)
             self.configuration = value_for_configuration
             response
           end
-          if base_instance? && lazy
+          if base && base_instance? && lazy
```
dblock: Should we be doing this check inside `base_instance?`

myxoh: I don't think it makes sense, as `base_instance?` usually means: "This instance is not used anywhere and is just a template for other instances" while, if an API has no `base`, it will most likely be used directly

dblock:
Thanks for a quick fix!

PR #1948, opened 2019-12-26
Relax dry-types dependency version

nbulaj:
From the PR discussion: https://github.com/ruby-grape/grape/pull/1920#issuecomment-569010623

Ruby 2.7.0 was released December 25, 2019, and has some noisy [deprecation messages about keyword arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/). 

Thanks to PR #1920 we migrated from Virtus to dry-types, but set too strict version for it - `~> 1.1.1`, so new versions of the dry-types gem couldn't be used with Grape (like current 1.2.2 from December 14, 2019). Also 1.1.x versions of dry-types produces a lot of deprecation warnings with Ruby 2.7 that can be too noisy.

I relaxed dependency version to `>= 1.1` and tested it against latest dry-types version on rubygems.org and everything seems to be good.

### RSpec results with dry-types 1.2.2

image

### DIFF
--- a/grape.gemspec
+++ b/grape.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency 'activesupport'
   s.add_runtime_dependency 'builder'
-  s.add_runtime_dependency 'dry-types', '~> 1.1.1'
+  s.add_runtime_dependency 'dry-types', '>= 1.1'
   s.add_runtime_dependency 'mustermann-grape', '~> 1.0.0'
   s.add_runtime_dependency 'rack', '>= 1.3.0'
   s.add_runtime_dependency 'rack-accept'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#1948](https://github.com/ruby-grape/grape/pull/1948): Relax `dry-types` dependency version - [@nbulaj](https://github.com/nbulaj).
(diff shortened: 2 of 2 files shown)


grape.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency 'activesupport'
   s.add_runtime_dependency 'builder'
-  s.add_runtime_dependency 'dry-types', '~> 1.1.1'
+  s.add_runtime_dependency 'dry-types', '~> 1.1'
```
dblock: Do we need to lock it at all? I think unless there's a specific reason that version x.y.z doesn't work we should leave this blank.

nbulaj: Actually  there were cases with dry-types when it has some incompatibility issues between versions, at least it was like this before 1.0 release. As for now it must be stable enough. But if somebody has an old version - Grape just wouldn't work for him. 

I tested it now with 1.0.0 and 3 specs are failed. The same is for 1.0.1. But everything is OK starting from 1.1.0

dblock: Should this be >= 1.1 then?

nbulaj: Only if we don't afraid of breaking changes in 2.0. If not - I'll change :)

dblock: Not afraid, let's make it `>= 1.1`, again I generally am of the opinion we should only be restricting for known issues. In this case you've identified that < 1.1 doesn't work, so it makes perfect sense. But we don't need to be afraid of future incompatibility and I would rather find sooner than later by a breaking build.

nbulaj: Makes sense for me :+1: 
Done

PR #1949, opened 2019-12-27
Add support for Ruby 2.7

nbulaj:
Add support for Ruby 2.7 both in Travis and in Grape itself

---

Ruby 2.7.0 was released December 25, 2019, and has some noisy [deprecation messages about wrong usage of keyword arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/).

Grape gem already correctly uses double-splat operator in some places, but not all over the code-base. That must be fixed.

### So, what's done here:

* Add Ruby 2.7 to Travis matrix, use 2.6.x for Gemfile and rails 5 as it's already done for previous versions.
* Fix method signatures and calls that produce deprecation messages for kwargs used.

To see a ton of deprecation messages you can just install Ruby 2.7, checkout current `master` branch and run test suite. There are still some messages not related to the gem itself that comes from `mustermann`, `multi_json` and a few other gems that must be fixed themselves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,27 +8,31 @@ gemfile:
 
 matrix:
   include:
-    - rvm: 2.6.5
+    - rvm: 2.7.0
       script:
         - bundle exec danger
-    - rvm: 2.6.5
+    - rvm: 2.7.0
       gemfile: Gemfile
-    - rvm: 2.6.5
+    - rvm: 2.7.0
       gemfile: gemfiles/rack_edge.gemfile
-    - rvm: 2.6.5
+    - rvm: 2.7.0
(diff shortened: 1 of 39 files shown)


anatol:
Is there any plan to release a gem version that supports ruby2.7 (i.e. with the ruby27 fixes above)?

dblock:
> Is there any plan to release a gem version that supports ruby2.7 (i.e. with the ruby27 fixes above)?

Yes, but at the very least we need to check #1957. Maybe you can help?

dblock:
https://github.com/ruby-grape/grape/issues/1958

PR #1951, opened 2019-12-31
Small Ruby 2.7 + changelog fixes

nbulaj:
As from the title: small fix for kwargs in `Grape::Exceptions::Validation` class and typo fix in changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 #### Features
 
 * Your contribution here.
-* [#1948](https://github.com/ruby-grape/grape/pull/1949): Add support for Ruby 2.7 - [@nbulaj](https://github.com/nbulaj).
+* [#1949](https://github.com/ruby-grape/grape/pull/1949): Add support for Ruby 2.7 - [@nbulaj](https://github.com/nbulaj).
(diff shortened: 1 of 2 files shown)


dblock:
Is this something that fixes a bug? I mean is there a test that can be written that fails without this change?

nbulaj:
No-no, it's not a bugfix :) It's one more fix for Ruby 2.7 / 3.0 kwargs deprecatiob messages that eas somehow lost in #1949 . Without it you'll just see a warning in the stdout. And a typo fix in the changelog. So no additional specs required, just regression that already exist

anatol:
We are looking at ruby2.7 issues in gitlab and one we found is https://gitlab.com/gitlab-org/gitlab/issues/194348  Could it be related to the fix above? Or you think the issue is somewhere else?

dblock:
Maybe. I would check against HEAD, but also note #1957.

nbulaj:
@anatol I checked GitLab issue and don't sure I understood the source of the issue. I mean I can't find in the logs who exactly produced an error you faced (GitLab internals? rack oauth2? Grape?).

This PR and #1949 add double-splat operator in places where Grape uses kwargs in method definition. It could be a source of the issue, but first it would be great to find the line of the code which produces it

PR #1953, opened 2020-01-04
delete a reversible stackable values class

dnesteryuk:
The reversible stackable values object was initialized for every endpoint, however, it is only needed for keeping rescue handlers. The idea is simple, handlers defined "closer" to an endpoint have higher priority. [That test](https://github.com/ruby-grape/grape/blob/master/spec/grape/api_spec.rb#L3215-L3232) demonstrates how it works.

In our project rescue handlers are defined at the top level, so almost every endpoint keeps the unused object.

The mentioned behavior is easy to achieve with the stackable values object and the `reverse_each` method. Thus, endpoints keeps less objects and have less code to be maintained.

Besides that, there are a few other simple performance optimizations.


lib/grape/validations/types/build_coercer.rb
```diff
@@ -83,7 +83,11 @@ def self.cache_instance(type, method, strict, &_block)
       end
 
       def self.cache_key(type, method, strict)
-        [type, method, strict].compact.map(&:to_s).join('_')
+        [type, method, strict].each_with_object(+'') do |val, memo|
```
dblock: Should be `_` to avoid `fo_obar` and `foo_bar` type situations.

lib/grape/dsl/settings.rb
```diff
@@ -171,7 +168,7 @@ def within_namespace(&_block)
       # the superclass's :inheritable_setting.
       def build_top_level_setting
         Grape::Util::InheritableSetting.new.tap do |setting|
-          if defined?(superclass) && superclass.respond_to?(:inheritable_setting) && superclass != Grape::API
+          if defined?(superclass) && superclass.respond_to?(:inheritable_setting) && superclass != Grape::API::Instance
```
dblock: Is there a bug fix here?

lib/grape/validations/types/build_coercer.rb
dnesteryuk: I guess joining '_' is unnecessary since the key is meant for "machines" here. Do you think it is necessary? :slightly_smiling_face:

lib/grape/dsl/settings.rb
dnesteryuk: Oh, I forgot to mention this one. [Grape::API](https://github.com/ruby-grape/grape/blob/master/lib/grape/api.rb) doesn't make use of `Grape::Util::InheritableSetting`, but [Grape::API::Instance does](https://github.com/ruby-grape/grape/blob/master/lib/grape/api/instance.rb#L10).

```ruby
irb(main):006:0> Grape::API.respond_to?(:inheritable_setting)
=> false
irb(main):007:0> Grape::API::Instance.respond_to?(:inheritable_setting)
=> true
```

Thus, there is no reason to expect `Grape::API`. 

The condition was added when [Grape::API worked](https://github.com/ruby-grape/grape/blob/6244ff8b5cdac97bf949aa0f09d9d13967b1ac8e/lib/grape/api.rb#L7) with `Grape::Util::InheritableSetting`.

lib/grape/validations/types/build_coercer.rb
dblock: Are we 100% sure that type/method/strict don't have the situation like `type=fo` + `method=obar` vs. `type=foo` + `method=bar`, it would generate the same key. So I think better safe than sorry.

lib/grape/dsl/settings.rb
dblock: I understand that part, but without digging into details it's a code change has a side effect. So either we're missing a spec or the entire `if` is unnecessary?

lib/grape/validations/types/build_coercer.rb
dnesteryuk: Good point! I've changed it.

lib/grape/dsl/settings.rb
dnesteryuk: It didn't cause big harm, however, a test for catching the regression makes sense :+1: Please, have a look! :slightly_smiling_face:

dblock:
Merged, 👍

nbulaj:
I updated Grape from git (from d9cf755e0b265b22a109b172d063278d36f6683b to current with this PR merged) and all my test suites began to fail :( Somewhere with Grape params validation messages, in other places with other errors. If I revert Grape to d9cf755e0b265b22a109b172d063278d36f6683b - everything works

dblock:
@nbulaj Please try writing specs for any regressions?

nbulaj:
Oh, it could be hard, but I'll try to

nbulaj:
Ah, I see. It fails with the test that must check this part:

```ruby
rescue_from ActiveRecord::RecordNotFound do |e|
  logger.warn("[#{e.class.name}] #{e.message}")
  error!({ message: "Not found" }, 404)
end
```

Instead of rescuing the error it just throws it and I have 

>  ActiveRecord::RecordNotFound:
>    Couldn't find Task with [WHERE "tasks"."deleted_at" IS NULL AND "tasks"."key" = $1]

in the test

nbulaj:
Ok, so this is the script to reproduce an breaking change:

**Gemfile:**

```ruby
source " https://rubygems.org" 

# With just :
#    gem "grape", "~> 1.2" 
#  or before this PR merged everything works

gem "grape", git: "https://github.com/ruby-grape/grape"
gem "activerecord"
gem "puma"
```

**Grape API:**

```ruby
require 'bundler/setup'
Bundler.setup

require 'grape'
require 'active_record'

module API
  class Endpoint < Grape::API
    desc "Test"
    get :test do
      # Simulate User.find(10)
      raise ::ActiveRecord::RecordNotFound.new("test", "name", "key", [])
    end
  end

  class Root < Grape::API
    version :v1, using: :path
    format :json

    content_type :json, "application/json; charset=UTF-8"

      error!({ messages: e.errors.map { |k, v| { fields: k, errors: v.map(&:message) } } }, 422)
    end

    rescue_from ::ActiveRecord::RecordNotFound do |e|
      error!({ message: "Not found" }, 404)
    end

    rescue_from StandardError do |e|
      raise e
    end

    mount API::Endpoint
  end
end

run API::Root
```

Run the API (`bundle exec rackup test_app.ru`)  and curl `http://localhost:9292/v1/test.json`

**Expected result:**

* See JSON with `message: "Not found"`

**Actual result:**

* See unhandled exception with full stacktrace :(

Sorry, can't write a RSpec test right now - don't have a time. But would be great if @dnesteryuk  could take a look

dblock:
I opened https://github.com/ruby-grape/grape/issues/1957 to track. Please someone take a look.

dnesteryuk:
@dblock @nbulaj I will take a look next weekend. It looks like this case was missing in tests :disappointed:

dm1try:
added the failing spec(#1959) for testing functional behavior
though, I'm out of the context for helping to fix this right now)

PR #1956, opened 2020-01-08
Align implementation with Rack SPEC

ioquatix:
And update specs/expand testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ matrix:
     - rvm: 2.7.0
       gemfile: Gemfile
     - rvm: 2.7.0
+      gemfile: gemfiles/rack1.gemfile
+    - rvm: 2.7.0
+      gemfile: gemfiles/rack2-0.gemfile
+    - rvm: 2.7.0
       gemfile: gemfiles/rack_edge.gemfile
     - rvm: 2.7.0
       gemfile: gemfiles/rails_edge.gemfile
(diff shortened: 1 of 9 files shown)


ioquatix:
I think this is okay to merge.

dblock:
👍 merged, although I do worry about `.to_a` causing unexpected regressions (eg. streaming response)

ioquatix:
Rack middleware must return an array of `[status, headers, body]` and all the `.to_a` method is doing is forcing that. It doesn't change any of the individual fields, e.g. `body` is still `body`. You shouldn't be returning a rich object in order to support streaming.

That being said, I understand your concern.

nbulaj:
Actually this changes broked our app after upgrade, check this one: https://github.com/ruby-grape/grape/issues/2030#issuecomment-624676579

ghiculescu:
Should we add something about this to https://github.com/ruby-grape/grape/blob/master/CHANGELOG.md#130-20200111

I ended up here looking into https://github.com/ruby-grape/grape/issues/1966 and it'd be great if the changelog called that out somehow

dnesteryuk:
@ghiculescu of course, go ahead! :slightly_smiling_face:

ghiculescu:
> @ghiculescu of course, go ahead! 🙂

https://github.com/ruby-grape/grape/pull/2059

Issue #1957, opened 2020-01-08
Regression? 


dblock:
See https://github.com/ruby-grape/grape/pull/1953#issuecomment-572064622

dnesteryuk:
Unfortunately, it isn't easy to fix. I would recommend to drop my commit from the master and do

```
git push origin -f
```

or I can open another PR where I will restore the ReverseStackableValues. :disappointed:

dm1try:
> or I can open another PR where I will restore the ReverseStackableValues. 😞

👍, an explicit revert is more suitable

dblock:
Yeah, someone please PR a revert.

@dnesteryuk Care to explain the problem?

dm1try:
closed via #1961

dnesteryuk:
When rescue handlers get added, every handler is added to a separate hash.

```ruby
[{ RuntimeError => handler }, { StandardError => handler }]
```

Even though they are defined in one endpoint

```ruby
rescue_from RuntimeError do
 # catch
end

rescue_from StandardError do
 # catch
end
```

So, after reverting this array, we got the wrong order. If handlers defined in an endpoint were in one hash

```ruby
[{ RuntimeError => handler, StandardError => handler }]
```

after adding inherited values it would look like this:

```ruby
[
  { RuntimeError => inheritedHandler }, 
  { RuntimeError => handler, StandardError => handler }
]
```

my solution would've worked. However, the [stackable values](https://github.com/ruby-grape/grape/blob/master/lib/grape/util/stackable_values.rb) doesn't work this way. Actually, any structure doesn't work this way.

I tried a few other ideas, but they didn't work because of [namespaces](https://github.com/ruby-grape/grape/blob/9feefdff949078a266784e0d2d6f74213f7cf729/lib/grape/dsl/settings.rb#L157-L166) and [copies of settings](https://github.com/ruby-grape/grape/blob/9feefdff949078a266784e0d2d6f74213f7cf729/lib/grape/util/inheritable_setting.rb#L67-L81), there are lots of things happening with settings. Grape really shuffles them a lot.

I am sorry guys for troubles :disappointed:

dblock:
> I am sorry guys for troubles 😞

No trouble at all. Bug was caught before release. A test case was added (thanks @dmitry!) that ensures rescue handlers happen in predictable order, which was clearly a miss and potentially a serious regression in the future because of some other change. I call this a win!

PR #1959, opened 2020-01-09
add failing spec for .rescue_from regression

dm1try:
see #1957


dblock:
I guess I shouldn't be lazy, https://github.com/ruby-grape/grape/pull/1961 :)

dm1try:
#1961 includes the changes

PR #1961, opened 2020-01-09
Revert "delete a reversible stackable values class"

dblock:
Reverts ruby-grape/grape#1953

Includes #1959

PR #1962, opened 2020-01-09
Re-apply the micro-optimizations and parent fix from #1961.

dblock:
from #1961


dblock:
@dm1try care to review & merge?

dm1try:
@dblock lgtm, thanks)

PR #1963, opened 2020-01-10
The values validator must properly work with booleans

### LINKED ISSUE #1577: Boolean validation ignores values
```
params do
  requires model_name do
      requires :params1, type: Boolean,allow_blank: false, desc: "params1",values: [true]
  end
end
```
When this API is called with params1 value as false should return an exception. But grape doesn't returns exception and the API  is successfully executed.

dnesteryuk:
Fixes #1577 .

When we check whether a value is given or not, it is safer to look for `nil`, there are cases when `false` is ok value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


dblock:
Thanks, yikes. If you have a moment write a spec for a non-boolean value here. Should be rejected obviously as well.

PR #1965, opened 2020-01-10
Fixes typos in README

davidalee:
Fixes typos in README: `Middlware` => `Middleware`


dblock:
Thank you! For README changes we don't need changelog, but it's NBD.

Issue #1967, opened 2020-01-11
Broken detection of Proc


ioquatix:
https://github.com/ruby-grape/grape/blob/d58dc0ab7a0b51625217deedd8110d1030be7cf7/lib/grape/middleware/stack.rb#L80-L84

This code is buggy. The correct solution is have a separate field for `&block`.

https://bugs.ruby-lang.org/issues/16500

cc @dblock

ioquatix:
Caused initially by something here:

https://github.com/ruby-grape/grape/blob/d58dc0ab7a0b51625217deedd8110d1030be7cf7/lib/grape/dsl/middleware.rb#L12-L53

anatol:
Note that this bug breaks gitlab https://gitlab.com/groups/gitlab-org/-/epics/2380

dblock:
Can someone please help with a failing spec for this? I'd be happy to try and fix.

dblock:
Note that I tried. There's a bunch of specs around https://github.com/ruby-grape/grape/blob/master/spec/grape/api_spec.rb#L1308 and I couldn't get anything to fail.

dm1try:
> Can someone please help with a failing spec for this? I'd be happy to try and fix.
> Note that I tried. There's a bunch of specs around https://github.com/ruby-grape/grape/blob/master/spec/grape/api_spec.rb#L1308 and I couldn't get anything to fail.

according to the description `Stack#merge_with` should be tested
I'm going to install 2.7.0 and check this)

dm1try:
@dblock added the spec based on the context @ioquatix provided above, see ref

dm1try:
guys, take a look https://github.com/ruby-grape/grape/pull/1968/commits/dec3e1ff5dbf3215a714565e62b12bd2ef6b0ddb
is it an acceptable fix?)

mame:
All versions of Ruby seems to have a bug in combination with rest and block arguments.

```
def foo(*ary)
  p ary
end
ary = [1, 2, 3, proc {}]
foo(*ary, &ary.pop) #=> [1, 2, 3], expected: [1, 2, 3, #<Proc:...>]
```

I believe it should be fixed on the side of the interpreter.  https://github.com/ruby-grape/grape/commit/dec3e1ff5dbf3215a714565e62b12bd2ef6b0ddb looks good to me.  It will work even after the interpreter is fixed.

https://bugs.ruby-lang.org/issues/16500 might be a different issue from this.

[...]

PR #1968, opened 2020-01-11
fix proc forwarding in Middleware::Stack#merge_with for ruby 2.7.0

dm1try:
ref #1967

~it fails on ruby 2.7.0 because of an "extra" argument(the proc) is forwarded~(FIXED)

`public_send(operation, *args, &args.pop)` works differently in ruby 2.7.0, `args` is not modified by `pop` if placed in the method call before the `pop`; though it looks more expected to me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#1968](https://github.com/ruby-grape/grape/pull/1968): Fix args forwarding in Grape::Middleware::Stack#merge_with for ruby 2.7.0 - [@dm1try](https://github.com/dm1try).
 
 ### 1.3.0 (2020/01/11)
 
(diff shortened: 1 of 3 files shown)


dblock:
👍

anatol:
Oh, did you release 1.3.0 already? Is there a chance to get 1.3.1 soon enough? It will help gitlab with ruby 2.7 migration. https://gitlab.com/groups/gitlab-org/-/epics/2380

cc @dnesteryuk

dblock:
@anatol Have you tested against HEAD? Want to make sure nothing else is known that needs to be fixed.

anatol:
Testing gitlab with grape HEAD is not possible at this point as it requires this ticket to be resolved first https://gitlab.com/gitlab-org/gitlab/issues/195960

anatol:
I patched gitlab's grape-1.1.0 `merge_with` function with the fix above and things seems work fine now.

dnesteryuk:
@anatol There are a few more bugs (regressions) which need to be fixed before releasing 1.3.1. Let's see what we get this week :wink:

Issue #1969, opened 2020-01-13
Coercing to BigDecimal and Numeric broken in v1.30.0


FlickStuart:
When creating attempting to coerce a param to `BigDecimal`, Grape `1.3.0` will return an HTTP 400. This is still possible in versions prior to `1.3.0`.

I am able to replicate this by adding the following test to `spec/grape/validations/validators/coerce_spec.rb` just after `context 'coerces' do` at line 157:
```
      it 'BigDecimal' do
        subject.params do
          requires :bigdecimal, coerce: BigDecimal
        end
        subject.get '/bigdecimal' do
          params[:bigdecimal].class
        end

        get '/bigdecimal', bigdecimal: '45'
        expect(last_response.status).to eq(200)
        expect(last_response.body).to eq("BigDecimal")
      end
```
Which fails with
```
Grape::Validations::CoerceValidator coerce coerces BigDecimal
     Failure/Error: expect(last_response.status).to eq(200)
     
       expected: 200
            got: 400
     
       (compared using ==)
     # ./spec/grape/validations/validators/coerce_spec.rb:168:in `block (4 levels) in <top (required)>
```

Stepping into the test, `last_response` is:
```
last_response
=> #<Rack::MockResponse:0x00007fe123fd0758
 @block=nil,
 @body=["bigdecimal is invalid"],
 @buffered=true,
 @cookies={},
 @errors="",
 @header={"Content-Type"=>"text/plain"},
 @length=0,
 @original_headers={"Content-Type"=>"text/plain"},
 @status=400,
 @writer=#<Method: Rack::MockResponse(Rack::Response::Helpers)#append>>
```

Stepping into `DryTypeCoercer.call ` in `lib/grape/validations/types/dry_type_coercer.rb`, `@coercer` is `BigDecimal`, running `@coercer` raises the error:
```
NoMethodError: undefined method `[]' for BigDecimal:Class
```

If we add a `BigDecimal` mapping to `PrimitiveCoercer` in `lib/grape/validations/types/primitive_coercer.rb` like so:
```
class PrimitiveCoercer < DryTypeCoercer
        MAPPING = {
          Grape::API::Boolean => DryTypes::Params::Bool,

          # unfortunatelly, a +Params+ scope doesn't contain String
          String              => DryTypes::Coercible::String,
          BigDecimal          => DryTypes::Coercible::Decimal
        }.freeze
```
[...]

PR #1971, opened 2020-01-13
BigDecimal coercion

FlickStuart:
This PR attempts to fix coercion of params to BigDecimal as per https://github.com/ruby-grape/grape/issues/1969

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Fixes
 
-* Your contribution here.
+* [#1971](https://github.com/ruby-grape/grape/pull/1971): Fix BigDecimal coercion - [@FlickStuart](https://github.com/FlickStuart).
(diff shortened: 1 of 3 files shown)


FlickStuart:
> Please, extend the `STRICT_MAPPING` constant which is used for checking types when custom coercers are given

Extending `STRICT_MAPPING` looks simple enough... I am concerned that no tests failed without the change being made already and I don't think understand enough of the internals of Grape to write an effective test for this.

dblock:
Def want a test for that STRICT_MAPPING change. I’m ok merging without and opening an issue. Deferring to @dnesteryuk.

tlconnor:
@dblock this bug is preventing some users from moving to Grape 1.3.0. Would you be able to release Grape 1.3.1 soon? Thanks!

dblock:
> @dblock this bug is preventing some users from moving to Grape 1.3.0. Would you be able to release Grape 1.3.1 soon? Thanks!

I opened https://github.com/ruby-grape/grape/issues/2008, add your +1.

PR #1974, opened 2020-01-14
more tests for Grape::Validations::Types::PrimitiveCoercer

dnesteryuk:
Unit tests better focus on corner cases in this case than integration tests which might be too high level.

I guess changelog isn't required for this.


dblock:
Thanks for picking this up.

PR #1976, opened 2020-01-15
Be sure classes/modules listed for autoload really exist

### LINKED ISSUE #1972: cannot load such file -- grape/extensions/deep_hash_with_indifferent_access after update to 1.3.0
Looks like `Grape::Extensions::DeepHashWithIndifferentAccess` was removed here:
https://github.com/ruby-grape/grape/commit/7ed75987bf530b6ccb02f40c1f11fa7d5707c7fd#diff-0beb8afd3f95a465d9cdf19b4bb7aa59

but it's still declared to autoload here:
https://github.com/ruby-grape/grape/blob/v1.3.0/lib/grape.rb#L87

dnesteryuk:
Fixes #1972

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#1976](https://github.com/ruby-grape/grape/pull/1976): Ensure classes/modules listed for autoload really exist - [@dnesteryuk](https://github.com/dnesteryuk).
 * [#1971](https://github.com/ruby-grape/grape/pull/1971): Fix BigDecimal coercion - [@FlickStuart](https://github.com/FlickStuart).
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#1976](https://github.com/ruby-grape/grape/pull/1976): Be sure classes/modules listed for autoload really exist - [@dnesteryuk](https://github.com/dnesteryuk).
```
dblock: Be sure -> Ensure

spec/spec_helper.rb
```diff
@@ -10,6 +10,26 @@
 require 'bundler'
 Bundler.require :default, :test
 
+# Grape uses autoload https://api.rubyonrails.org/classes/ActiveSupport/Autoload.html.
```
dblock: Let's put this in https://github.com/ruby-grape/grape/tree/master/spec/support, call it `eager_load.rb` and call this function `eager_load!` to follow the Ruby pattern.

dnesteryuk:
@dblock thanks for the review, please, have a look again :wink:

PR #1977, opened 2020-01-16
Skip validation for a file if it is optional and nil

### LINKED ISSUE #1970: Optional file param validation
I have the following code:

```ruby
params do
  optional :attachment, type: File
end

post do 
end
```

## Expectation

With grape < 1.2.5 `POST` with empty `attachment` param returns 200

## Reality

`POST` action with empty `attachment` returns 400 `{"error":"attachment is invalid"}`

dnesteryuk:
Fixes #1970

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


dblock:
👍 needs rebase, merge on green

PR #1978, opened 2020-01-18
more unit tests for coercion

dnesteryuk:
What do you think about unit tests for such cases? :slightly_smiling_face:


dblock:
I like me tests.

Issue #1975, opened 2020-01-14
Change behaviors of rescue_from to not depend on mount order


wowinter13:
Hi, found an interesting case when `rescue_from :all` should override outer `rescue_from :error_class`
```ruby
class PublicApi < Grape::API
  rescue_from ActiveRecord::RecordNotFound do
    binding.pry
    error_response(status: 403)
  end
  mount AuthController
end
class AuthController < Grape::API
  rescue_from :all do 
    error_response(status: 403)
  end

  post :request do
    User.find_by!(phone: phone)
  end
end
```
```bash
15:  rescue_from ActiveRecord::RecordNotFound do
15:     binding.pry
16:     error_response(status: 403)
17:  end
```
For some reason, it seems to not working correctly. I suppose `AuthController rescue_from :all` to be called first.
But if we mount AuthController above `rescue_from ActiveRecord::RecordNotFound` it would work correctly.
So, does it look like a bug or additional info to documentation?

dblock:
Can you turn it into a spec? Looks incorrect to me.

wowinter13:
@dblock Should I create a PR with spec? Or gist in the discussion would be enough?

dblock:
> @dblock Should I create a PR with a spec? Or gist in the discussion would be enough?

Yes. And don't hesitate to try a fix!

jcagarcia:
Hey guys! 👋 

I've been examining this issue and I'm uncertain if it qualifies as a bug. As far as I can tell, Grape seems to be functioning as intended.

In your `PublicApi < Grape::API` class, you've specified a `rescue_from` for the specific class `ActiveRecord::RecordNotFound`. Following this, you're mounting the `AuthController < Grape::API` class, which has a declared `rescue_from :all`.

What's happening here is that upon mounting, you essentially "merge" the `AuthController` into the `PublicApi` context, inheriting all the defined `rescue_from` methods. Nevertheless, your primary class remains `PublicApi`.

Consequently, when an error occurs, the first error handlers considered are those within the class. If none apply, it then resorts to the `:all` handlers. 

https://github.com/ruby-grape/grape/blob/3901bf4ddc85b83bf6df9c26b6850f2d6117a663/lib/grape/middleware/error.rb#L43-L46

[...]

PR #1982, opened 2020-01-21
Added failing spec for nested rescue_from declarations

wowinter13:
See: #1975


dblock:
Doesn't look like it's failing?

wowinter13:
Sorry, a bit of a misunderstanding

dblock:
Cool, now that we have a spec all we need is a fix! Give it a shot.

wowinter13:
@dblock, I have some ideas on how to fix it, but want to discuss due to lack of knowledge.

Somehow we need to [use](https://github.com/ruby-grape/grape/blob/cf57d250c3d77a9a488d9f56918d62fd4ac745ff/lib/grape/middleware/error.rb#L46) `rescue_handler_for_any_class` before any other handler, BUT only in case of nested rescue blocks. For now I don't see an obvious solution without stacking all errors (like :rescue_all,  :rescue_grape_exceptions etc) to some namespace. Maybe there is any better realization like using reverse_stacking from descendants to the ancestor? Or..?

dblock:
Feels like maybe the tree of rescue handlers should be traversed as a tree?

PR #1987, opened 2020-02-04
Re-add exactly_one_of mutually exclusive error message

ZeroInputCtrl:
Breaking change at revision 96e4e1044e73e64414af709352b270ab4f88c864
Before the breaking change specs only tested a `Grape::Exceptions::Validation` error was returned rather than the message itself so this fell through when changed over to the new `MultipleParamsBase` code.

### DIFF
--- a/lib/grape/validations/validators/exactly_one_of.rb
+++ b/lib/grape/validations/validators/exactly_one_of.rb
@@ -7,7 +7,8 @@ module Grape
     class ExactlyOneOfValidator < MultipleParamsBase
       def validate_params!(params)
         return if keys_in_common(params).length == 1
-        raise Grape::Exceptions::Validation.new(params: all_keys, message: message(:exactly_one))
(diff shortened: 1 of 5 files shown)


lib/grape/validations/validators/exactly_one_of.rb
```diff
@@ -7,7 +7,8 @@ module Validations
     class ExactlyOneOfValidator < MultipleParamsBase
       def validate_params!(params)
         return if keys_in_common(params).length == 1
-        raise Grape::Exceptions::Validation.new(params: all_keys, message: message(:exactly_one))
+        raise Grape::Exceptions::Validation.new(params: all_keys, message: message(:exactly_one)) if keys_in_common(params).length.zero?
```
dblock: We do a lot of `keys_in_common` calls, let's fix this?

```ruby
_keys_in_common = keys_in_common(params)
return if _keys_in_common.length == 1
raise ... if _keys_in_common.none?
raise 
```

ZeroInputCtrl: of course

dblock:
Merged, thank you.

Issue #1954, opened 2020-01-06
Ruby 2.5 issue with private method full_message 


danillos:
[Ruby 2.5.0](https://blog.bigbinary.com/2018/03/13/ruby-2-5-adds-exception-full_message-method.html) adds the public method `full_message` and Grape is overriding this method as private here:

https://github.com/ruby-grape/grape/blob/d58dc0ab7a0b51625217deedd8110d1030be7cf7/lib/grape/exceptions/validation_errors.rb#L51

dblock:
Sounds like we should at least refactor our way out of this. PRs welcome.

dblock:
Closed via https://github.com/ruby-grape/grape/pull/1988

PR #1988, opened 2020-02-08
refactored the full_messages method and stop overriding full_message

hosseintoussi:
Hi, Thank you for your work here 🙏.

This PR is an attempt to address: https://github.com/ruby-grape/grape/issues/1954. 

As mentioned in the issue, Ruby 2.5 add a `full_message` to Exception (https://ruby-doc.org/core-2.5.0/Exception.html#full_message-method). Grape overrides this and to stop that this PR removes the `full_message` private method.


dblock:
Thanks

PR #1994, opened 2020-02-18
Fix typo in README

bellmyer:
I found a typo in the README ("trigered" instead of "triggered") so I thought I'd be helpful. After that I fixed another typo so hopefully this is merge-worthy :)


dblock:
Thanks!

PR #1995, opened 2020-02-19
Address some Ruby warnings

nbeyer:
* Add warnings enablement to spec_helper.rb to see future warnings
* Use instance_variable_defined? to clean up not initialized warnings
* Add 'inherited' to NON_OVERRIDABLE list
* Add multi_xml to test dependencies in Gemfile to address spec failure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 * Your contribution here.
 
 #### Fixes
+
+* [#1995](https://github.com/ruby-grape/grape/pull/1995): Address some Ruby warnings - [@nbeyer](https://github.com/nbeyer).
 * [#1994](https://github.com/ruby-grape/grape/pull/1993): Fix typos in README - [@bellmyer](https://github.com/bellmyer).
 * [#1993](https://github.com/ruby-grape/grape/pull/1993): Lazy join allow header - [@ericproulx](https://github.com/ericproulx).
 * [#1987](https://github.com/ruby-grape/grape/pull/1987): Re-add exactly_one_of mutually exclusive error message - [@ZeroInputCtrl](https://github.com/ZeroInputCtrl).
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -5,13 +5,16 @@
 * Your contribution here.
 
 #### Fixes
+
+* [#1995](https://github.com/ruby-grape/grape/pull/1995): Address some Ruby warnings - [@nbeyer](https://github.com/nbeyer).
```
dblock: Let's spell out which warnings?

nbeyer: updated at 122fb205e3d753df2c270ce0fedac0ed43240bcf

nbeyer:
@dblock This PR has been rebased.

Gemfile
```diff
@@ -28,6 +28,7 @@ group :test do
   gem 'grape-entity', '~> 0.6'
   gem 'maruku'
   gem 'mime-types'
+  gem 'multi_xml'
```
dblock: We have integration tests with this, and we don't want to load multi_xml for all tests, why was it required?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 * Your contribution here.
 
 #### Fixes
+
+* [#1995](https://github.com/ruby-grape/grape/pull/1995): Address some Ruby warnings (undefined instance variables, method redefined) - [@nbeyer](https://github.com/nbeyer).
```
dblock: Or just `Fix: "undefined instance variables" and "method redefined" warnings`?

Gemfile
nbeyer: How does that work in practice? When I checkout the code in master, run bundle install, then run rspec, I get these test failures ...

```
  1) ActiveSupport::XmlMini uses multi_xml
     Failure/Error: expect(Grape::Xml).to eq(::MultiXml)
     
     NameError:
       uninitialized constant MultiXml
       Did you mean?  MultiJson
     # ./spec/integration/multi_xml/xml_spec.rb:7:in `block (2 levels) in <top (required)>'
```

CHANGELOG.md
nbeyer: Sure. Changed at 4fd9c2c8048163a0cfdb3036e4c50c5629e08e28

Gemfile
nbeyer: Nevermind, I re-read the [contributing document](https://github.com/ruby-grape/grape/blob/master/CONTRIBUTING.md) and noticed I was running the tests incorrectly. I ran things through rake and it works without it.

I remove the gem dependency here 18c0310. Sorry about that.

dblock: No worries. It would be nice if we could make it so `rspec` runs successfully by default, since that's a common pattern.

dblock:
Merged, thank you.

dblock:
Now that we run with `--warnings`, how far are we from making tests fail on unexpected warnings?

nbeyer:
I've not found anything great for failing tests based on unexpected warnings. It's possible to plug into [Ruby's Warning module](https://ruby-doc.org/core-2.6/Warning.html), but it would require some framework building, as it seems likely that warnings from dependencies are going to pop-up, which would be out of the control of this project.

PR #2001, opened 2020-02-28
Params: optional arrays within hashes/JSON objects broken

dleavitt:
See failing tests below for more specifics.

To reproduce
In a params block, define an array param within a Hash or JSON param:
```ruby
params do
  optional :data, type: Hash do
    optional :array, type: Array do
      optional :name, type: String
    end
  end
end
```
When the request body looks like this: `{ data: {} }`
I'd expect that calling `declared(params)` within the route would return something like this:
`{ data: { array: [] } }`
or maybe this:
`{ data: {} }`

What's actually returned is this:
`{ data: { array: { name: nil } } }`

A hash rather than an array.


myxoh:
Makes sense, thanks for the test!

dblock:
@dleavitt now that we have a test (thanks), try fixing it?

spec/grape/validations/params_scope_spec.rb
```diff
@@ -395,6 +395,60 @@ def initialize(value)
...
+            optional :name, type: String
+          end
+        end
+      end
+      subject.post('/nesty') { declared(params).to_json }
+
+      post '/nesty', { data: { } }
+      expect(last_response.status).to eq 201
+      # {"data"=>{"json"=>{"name"=>nil}}}
+      expect(JSON.parse(last_response.body)['data']['json']).not_to be_a Hash
```
dnesteryuk, line 448: I guess here we should expect the hash. @dleavitt could you verify this test?

numbata:
I ran these tests and only params with nested JSON's seem to be broken. And it is fixable by calling `declared(params, include_missed: false)` instead of `declared(params)`.

Also, looking at the rest of the specs from `lib/grape/validations/params_scope.rb` file, it looks like expected behavior and not a bug.

dleavitt:
Feel free to close if you're cleaning house!

PR #2003, opened 2020-03-01
Update rubocop

ericproulx:
Updated rubocop to latest version.
I also added rubocop-performance


Gemfile
```diff
@@ -10,7 +10,8 @@ group :development, :test do
   gem 'bundler'
   gem 'hashie'
   gem 'rake'
-  gem 'rubocop', '0.51.0'
+  gem 'rubocop'
```
dblock: We don't commit Gemfile.lock, so you should specify a version here or future builds will break for no reason as newer versions get released.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#2003](https://github.com/ruby-grape/grape/pull/2003): Updated rubocop - [@ericproulx](https://github.com/ericproulx).
```
dblock: Upgraded RuboCop to ...

dblock:
👍

PR #2004, opened 2020-03-04
Rubocop todo fixes

ericproulx:
Hello,

Here are some fixes based on rubocop todo:
- Style/HashTransformValues
- Style/RescueStandardError
- Lint/RedundantRequireStatement
- Performance/RegexpMatch
- Style/MutableConstant
- Lint/DisjunctiveAssignmentInConstructor

### DIFF
--- a/lib/grape/validations/validators/coerce.rb
+++ b/lib/grape/validations/validators/coerce.rb
@@ -76,7 +76,7 @@ module Grape
         converter.call(val)
 
       # Some custom types might fail, so it should be treated as an invalid value
-      rescue StandardError
+      rescue
         Types::InvalidValue.new
       end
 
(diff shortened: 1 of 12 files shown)


lib/grape/validations/validators/coerce.rb
```diff
@@ -76,7 +76,7 @@ def coerce_value(val)
         converter.call(val)
 
       # Some custom types might fail, so it should be treated as an invalid value
-      rescue
+      rescue StandardError
```
dblock: This is actually a behavior change. It's the right one though. Let's add it to UPGRADING?

ericproulx: Separate PR ?

dblock: Probably best. Remove it from this one?

dblock:
👍

PR #2005, opened 2020-03-04
Content types registrable

ericproulx:
I found that `Grape::ContentTypes ::CONTENT_TYPE` was mutable but had a `rubocop disable` comment.

Since you have `Util::Registrable`, I found that it would be better to use that pattern instead of a mutable const. 

Also, I moved `content_type` in grape folder since it wasn't `module Util`


dblock:
👍

PR #2006, opened 2020-03-05
Upgrade coerce behavior when rescuing error.

ericproulx:
Fix Style/RescueStandardError

As discussed @dblock


ericproulx:
After reading this [article](http://rubylearning.com/satishtalim/ruby_exceptions.html), when `rescue` has not arguments, its actually rescuing StandardError implicitly.

Based on [rubocop doc](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RescueStandardError), default is explicit so that's why it appeared in the todo.

dblock:
Oh, my bad. Made you do the work for nothing. Rebase and merge.

PR #2014, opened 2020-03-16
Reduced array allocation

ericproulx:
```log
# Still based on our app

# current Grape master
allocated memory by class
  15834248  Array
  15730608  Hash
   6726012  String
   3079724  Regexp
    202016  MatchData

# next_version
allocated memory by class
  15730608  Hash
   6726012  String
   5870664  Array
   3079724  Regexp
    202016  MatchData
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ### 1.3.2 (Next)
 
 #### Features
-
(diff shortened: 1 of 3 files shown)


dblock:
Nice. Although I thought the old code was at least more interesting :)

Issue #2013, opened 2020-03-16
Array[String] no longer works with strings with commas


stanhu:
This passed in v1.1.0 but fails in v1.3.0: Reproduction step:

```diff
diff --git a/spec/grape/validations/validators/coerce_spec.rb b/spec/grape/validations/validators/coerce_spec.rb
index dbf57c9..c47f925 100644
--- a/spec/grape/validations/validators/coerce_spec.rb
+++ b/spec/grape/validations/validators/coerce_spec.rb
@@ -396,6 +396,19 @@ describe Grape::Validations::CoerceValidator do
         expect(JSON.parse(last_response.body)).to eq(%w[a b c d])
       end
 
+      it 'parses Array[String] type with strings with commas' do
+        subject.params do
+          requires :values, type: Array[String]
+        end
+        subject.get '/strings' do
+          params[:values]
+        end
+
+        get '/strings', values: 'a, b'
+        expect(last_response.status).to eq(200)
+        expect(JSON.parse(last_response.body)).to eq(['a, b'])
+      end
+
       it 'parses parameters with Array[Integer] type' do
         subject.params do
           requires :values, type: Array[Integer], coerce_with: ->(val) { val.split(/\s+/).map(&:to_i) }
```

```
Failures:

  1) Grape::Validations::CoerceValidator coerce using coerce_with parses Array[String] type with no coercion
     Failure/Error: expect(last_response.status).to eq(200)

       expected: 200
            got: 400

       (compared using ==)
     # ./spec/grape/validations/validators/coerce_spec.rb:408:in `block (4 levels) in <top (required)>'

Finished in 0.17659 seconds (files took 1.83 seconds to load)
47 examples, 1 failure

Failed examples:

rspec ./spec/grape/validations/validators/coerce_spec.rb:399 # Grape::Validations::CoerceValidator coerce using coerce_with parses Array[String] type with no coercion
```

dblock:
Looks like a regression, but certainly a weird one. I would dig up what made it work in the first place and then we can decide whether we should fix it or not.

stanhu:
I think it's a difference in  `Dry::Types::Coercions::Params.to_ary` vs `Virtus::Attribute`:

```ruby
(byebug) @coercer
#<Dry::Types[Constructor<Array fn=Dry::Types::Coercions::Params.to_ary>]>
(byebug) val
"tag1, tag2"
(byebug) @coercer[val]
[...]

PR #2016, opened 2020-03-16
Add upgrade notice for Array[String]

stanhu:
If `Array[String]` is used without a `coerce_with` block, existing APIs may
quietly fail without explanation. This change to the upgrade
documentation makes that clear.

Closes https://github.com/ruby-grape/grape/issues/2013

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -42,6 +42,26 @@ end
 
 Custom types which don't depend on Virtus don't require any changes.
 
+#### Ensure that Array[String] types have explicit coercions
+
+Unlike Virtus, dry-types does not perform any implict coercions. If you
+have any uses of `Array[String]`, be sure they use a `coerce_with`
+block. For example:
+
+```ruby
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -42,6 +42,26 @@ end
...
+```
+
+It's quite common to pass a comma-separated list, such as `tag1,tag2` as
+`values`. Previously Virtus would implicitly coerce this to
+`Array(values)` so that `["tag1,tag2"]` would pass the type checks, but
+with `dry-types` the values are no longer coerced for you. To fix this,
+you might do:
+
+```ruby
+requires :values, type: Array[String], coerce_with: ->(val) { val.split(',').map(:strip) }
```
stanhu: This isn't quite right. Let me see.

stanhu:
```suggestion
requires :values, type: Array[String], coerce_with: ->(val) { val.split(',').map(&:strip) }
```

PR #2019, opened 2020-03-17
Avoid coercing parameter with multiple types to an empty Array

### LINKED ISSUE #2018: Optional values with multiple types get coerced to an empty Array
This breaks certain APIs that expect a parameter passed in with `nil` to remain `nil`. To reproduce the problem:

```diff
diff --git a/spec/grape/validations/validators/coerce_spec.rb b/spec/grape/validations/validators/coerce_spec.rb
index dbf57c9..cb2cca5 100644
--- a/spec/grape/validations/validators/coerce_spec.rb
+++ b/spec/grape/validations/validators/coerce_spec.rb
@@ -647,6 +647,19 @@ describe Grape::Validations::CoerceValidator do
         expect(last_response.body).to eq('String')
       end

+      it 'respects nil values' do
+        subject.params do
+          optional :a, types: [File, String]
+        end
+        subject.get '/' do
+          params[:a].class.to_s
+        end
+
+        get '/', a: nil
+        expect(last_response.status).to eq(200)
+        expect(last_response.body).to eq('NilClass')
+      end
+
       it 'fails when no coercion is possible' do
         subject.params do
           requires :a, types: [Boolean, Integer]
```

This is the test failure:

```
  2) Grape::Validations::CoerceValidator coerce multiple types respects nil values
     Failure/Error: expect(last_response.body).to eq('NilClass')

       expected: "NilClass"
            got: "Array"

       (compared using ==)
     # ./spec/grape/validations/validators/coerce_spec.rb:660:in `block (4 levels) in <top (required)>'
```

It looks like this happens because of this code block: https://github.com/ruby-grape/grape/blob/8deebb5b6dcc10db1e7eed0706db6749d38218c8/lib/grape/validations/validators/coerce.rb#L68-L71

I'm not quite sure I understand why the `type.is_a?(Array)` check needs to be there. This allows tests to pass:

```diff
$ git diff
diff --git a/lib/grape/validations/validators/coerce.rb b/lib/grape/validations/validators/coerce.rb
index 561ec1b..b2b4cbf 100644
--- a/lib/grape/validations/validators/coerce.rb
+++ b/lib/grape/validations/validators/coerce.rb
@@ -68,9 +68,10 @@ module Grape
         # define default values for structures, the dry-types lib which is used
         # for coercion doesn't accept nil as a value, so it would fail
         if val.nil?
-          return []      if type == Array || type.is_a?(Array)
+          return []      if type == Array
           return Set.new if type == Set
           return {}      if type == Hash
+          return val
         end

         converter.call(val)
```

Also, if we need to accept `nil` as a coerced value, there is https://github.com/dry-rb/dry-types/blob/bd75d556e70c2d3bef1f96140a8f33fb44c31f9d/docsite/source/optional-values.html.md.

stanhu:
If an optional parameter has multiple types (e.g. `[File, String]`) and a
`nil` parameter is passed, Grape would previously coerce the value to an
empty `Array`. This can break certain APIs that treat nil values
differently from an empty Array.

To fix this, we refactor the nil handling into a `coerce_nil` method and
remove the special case check for handling an Array of types.

Closes #2018

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#2019](https://github.com/ruby-grape/grape/pull/2018): Avoid coercing parameter with multiple types to an empty Array - [@stanhu](https://github.com/stanhu).
 
 ### 1.3.1 (2020/03/11)
 
--- a/lib/grape/validations/validators/coerce.rb
+++ b/lib/grape/validations/validators/coerce.rb
@@ -65,19 +65,19 @@ module Grape
       end
 
       def coerce_value(val)
-        # define default values for structures, the dry-types lib which is used
-        # for coercion doesn't accept nil as a value, so it would fail
-        if val.nil?
-          return []      if type == Array || type.is_a?(Array)
-          return Set.new if type == Set
-          return {}      if type == Hash
-        end
-
-        converter.call(val)
-
+        val.nil? ? coerce_nil(val) : converter.call(val)
       # Some custom types might fail, so it should be treated as an invalid value
       rescue StandardError
         Types::InvalidValue.new
+      end
+
+      def coerce_nil(val)
+        # define default values for structures, the dry-types lib which is used
+        # for coercion doesn't accept nil as a value, so it would fail
+        return []      if type == Array
+        return Set.new if type == Set
+        return {}      if type == Hash
+        val
       end
 
       # Type to which the parameter will be coerced.
--- a/spec/grape/validations/validators/coerce_spec.rb
+++ b/spec/grape/validations/validators/coerce_spec.rb
@@ -180,6 +180,23 @@ describe Grape::Validations::CoerceValidator do
         expect(last_response.body).to eq(integer_class_name)
       end
 
+      it 'String' do
+        subject.params do
+          requires :string, coerce: String
+        end
+        subject.get '/string' do
+          params[:string].class
+        end
+
+        get '/string', string: 45
+        expect(last_response.status).to eq(200)
+        expect(last_response.body).to eq('String')
+
+        get '/string', string: nil
+        expect(last_response.status).to eq(200)
+        expect(last_response.body).to eq('NilClass')
+      end
+
       it 'is a custom type' do
         subject.params do
           requires :uri, coerce: SecureURIOnly
@@ -645,6 +662,19 @@ describe Grape::Validations::CoerceValidator do
         get '/', a: 'anything else'
         expect(last_response.status).to eq(200)
         expect(last_response.body).to eq('String')
+      end
+
+      it 'respects nil values' do
+        subject.params do
+          optional :a, types: [File, String]
+        end
+        subject.get '/' do
+          params[:a].class.to_s
+        end
+
+        get '/', a: nil
+        expect(last_response.status).to eq(200)
+        expect(last_response.body).to eq('NilClass')
       end
 
       it 'fails when no coercion is possible' do


lib/grape/validations/validators/coerce.rb
```diff
@@ -68,9 +68,10 @@ def coerce_value(val)
         # define default values for structures, the dry-types lib which is used
         # for coercion doesn't accept nil as a value, so it would fail
         if val.nil?
-          return []      if type == Array || type.is_a?(Array)
+          return []      if type == Array
           return Set.new if type == Set
           return {}      if type == Hash
+          return val     if type.is_a?(Array)
```
stanhu: I'm not sure about this `type.is_a?(Array)` check. I assume that passing a `nil` value to a parameter with multiple types might lead to ambiguity?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 #### Features
 * Your contribution here.
+* [#2019](https://github.com/ruby-grape/grape/pull/2018): Avoid coercing parameter with multiple types to an empty Array - [@stanhu](https://github.com/stanhu).
```
dblock: This is a fix, move it out of here.

dblock:
Looks like a bug, thanks for the fix! Move the changelog entry and I'll merge.

lib/grape/validations/validators/coerce.rb
dblock: Should this be returning `val` (a.k.a. `nil`) in all cases here?

stanhu: I had that originally. I didn't see any specs fail. I was trying to minimize changes here and wasn't sure if there was a reason why the original code was checking `type.is_a?(Array)`.

stanhu: I added a test for coercion of `String`. I didn't see any issue with dry-types in the current form. Happy to change this to a `return val` if you think that's the right thing here.

dblock: I think that's the right thing to do. We shouldn't have code that we don't understand ;) Let's find out the hard way if we broke something.

dblock: Try to refactor out a `coerce_nil` method if you think it makes sense?

stanhu: Done!

dblock:
See rubocop offense, I'll merge on green.

lib/grape/validations/validators/coerce.rb
```diff
@@ -67,11 +67,7 @@ def valid_type?(val)
       def coerce_value(val)
         # define default values for structures, the dry-types lib which is used
         # for coercion doesn't accept nil as a value, so it would fail
-        if val.nil?
-          return []      if type == Array || type.is_a?(Array)
-          return Set.new if type == Set
-          return {}      if type == Hash
-        end
+        return coerce_nil(val) if val.nil?
```
dblock: val.nil? ? coerce_nil(val) : converter.call(val)

lib/grape/validations/validators/coerce.rb
```diff
@@ -80,6 +76,13 @@ def coerce_value(val)
         Types::InvalidValue.new
       end
 
+      def coerce_nil(val)
+        return []      if type == Array
+        return Set.new if type == Set
+        return {}      if type == Hash
+        return val
```
dblock: just `val`

dblock: this is a good place for a `case`

```ruby
case type
when Array
when ...
```

stanhu: That doesn't actually work in this case because we need to distinguish between `type.is_a?(Array)` and `type == Array` (https://stackoverflow.com/questions/3908380/ruby-class-types-and-case-statements).

lib/grape/validations/validators/coerce.rb
ericproulx: Unfortunately, it was necessary
```ruby
    it 'handles default value to array' do
      subject.params do
        optional :a, type: Array[Integer], default: []
      end
      subject.put('/test') { declared(params).to_json }
      put '/test', { a: nil }
      expect(last_response.body).to eq('{"a":[]}')
    end
```
That test fails because `type != array` but it's `type.is_a?(Array)
I'll make a PR to fix it

PR #2021, opened 2020-03-17
Report endpoint_call.grape which includes the whole stack

### LINKED ISSUE #2010: `endpoint_run.grape` callback is called without middlewares
Hello. I'm trying to add the monitoring by subscribing to `'endpoint_run.grape'` via `ActiveSupport::Notifications.subscribe`. 

It works well until an exception occurs inside API. I have a bunch of `rescue_from` blocks to return different status codes depending on exceptions, such as

    rescue_from ActiveRecord::RecordNotFound do
      error!({ error: :not_found }, 404)
    end

The problem is when I call an endpoint which raises ActiveRecord::RecordNotFound, it really returns 404 status code, but in my callback `data[:endpoint].status` is still 200 (however data[:exception_object] is not null). 

A quick look at the code shows, that callback wraps only the main call, without middlewares. I believe, it should wrap middlewares too because:
- middlewares are part of the business logic. It's not good to have 404 returned to a client and 200 reported to the monitoring.
- timing will be more accurate because the middlewares can eat time too.

Kukunin:
Fixes #2010 

To keep the backward compatibility, I introduced a new event. The integration works fine in my project with this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3810,9 +3810,17 @@ Grape has built-in support for [ActiveSupport::Notifications](http://api.rubyonr
 
 The following are currently supported:
 
+#### endpoint_call.grape
+
+The main execution of an endpoint, includes filters, rendering and all middlewares.
+
+* *endpoint* - The endpoint instance
+* *response* - A typical Rack response, for example `[404, {"Content-Type"=>"application/json; charset=UTF-8"}, ["{\"error\":\"not_found\"}"]]`
+
 #### endpoint_run.grape
 
-The main execution of an endpoint, includes filters and rendering.
+Similar to `endpoint_call.grape` but does not include middlewares, thus might be inaccurate.
+One more difference that it returns original exceptions, before `rescue_from` handlers.
 
 * *endpoint* - The endpoint instance
 
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -226,7 +226,10 @@ module Grape
     def call!(env)
       env[Grape::Env::API_ENDPOINT] = self
       @env = env
-      @app.call(env)
+      context = {endpoint: self, env: @env}
+      ActiveSupport::Notifications.instrument('endpoint_call.grape', context) do
+        @app.call(env).tap { |response| context[:response] = response }
+      end
     end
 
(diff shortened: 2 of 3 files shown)


Kukunin:
Rubocop limit was increased previously too https://github.com/ruby-grape/grape/commit/28d34ee19bdf70cb8471535b4c76312d440a63a6#diff-74423563ec7c74f4cb7f74689b49e7c1R47, and I believe to refactor it, a person should a better insight about grape than I have.

So I increased limit by 3 lines too

README.md
```diff
@@ -3810,9 +3811,17 @@ Grape has built-in support for [ActiveSupport::Notifications](http://api.rubyonr
...
+
+The main execution of an endpoint, includes filters, rendering and all middlewares.
+
+* *endpoint* - The endpoint instance
+* *response* - A typical Rack response, for example `[404, {"Content-Type"=>"application/json; charset=UTF-8"}, ["{\"error\":\"not_found\"}"]]`
+
 #### endpoint_run.grape
 
-The main execution of an endpoint, includes filters and rendering.
+Similar to `endpoint_call.grape` but does not include middlewares, thus might be inaccurate.
```
dblock: I think "inaccurate" is incorrect. I think here we want to highlight what this does, as opposed to what it doesn't. Maybe "Execution of the endpoint, excluding filters, middlewares or execution inside rescue_from blocks." + explain what happens during an exception?

README.md
```diff
@@ -3810,9 +3811,17 @@ Grape has built-in support for [ActiveSupport::Notifications](http://api.rubyonr
...
+One more difference that it returns original exceptions, before `rescue_from` handlers.
```
dblock: Should be part of the text above. But this is unclear. What does "returns" mean here?

Kukunin:
Thanks for the review. Will try to fix the issues later today or tomorrow

dblock:
@Kukunin bumping myself on top of your list!

dblock:
> hi @dblock. do you mean that it'd be good to finalize and get this PR merged? If yes, then it's something I can do

Yes, please! Rebase & al. I know it's old but I think we still want it.

PR #2025, opened 2020-03-19
Fix Decimal type category

kdoya:
It may be more useful by using `DryTypes::Params::Decimal`.  Because `DryTypes::Coercible::Decimal` fails coercing `45.1` for json reqeust.


dblock:
Not sure what happened to the bot, but this fixes a bug and needs a changelog entry, please.

kdoya:
Thanks

PR #2026, opened 2020-03-20
Fix a regression in coerce_with when coercion returns nil

misdoro:
Previously nil was handled as valid value for coercion for integer parameters.

With refactoring introduced in 1.3.0 integer parameter coerced with a custom method can't coerce to nil any more.

This change reverts back to previous behaviour where coercion to nil was allowed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#2026](https://github.com/ruby-grape/grape/pull/2026): Fix a regression in coerce_with when coercion returns nil - [@misdoro](https://github.com/misdoro).
 * [#2025](https://github.com/ruby-grape/grape/pull/2025): Fix Decimal type category - [@kdoya](https://github.com/kdoya).
 * [#2019](https://github.com/ruby-grape/grape/pull/2019): Avoid coercing parameter with multiple types to an empty Array - [@stanhu](https://github.com/stanhu).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 #### Fixes
 
 * Your contribution here.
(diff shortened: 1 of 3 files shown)


spec/grape/validations/validators/coerce_spec.rb
```diff
@@ -525,6 +525,43 @@ def self.parsed?(value)
         expect(last_response.body).to eq('3')
       end
 
+      it 'allows coerce_with to return nil' do
```
dblock: It would be nice to refactor all those `get`'s below in their own test. Minor.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#2026](https://github.com/ruby-grape/grape/pull/2026): Fix a regression in coerce_with when coercion returns nil - [@misdoro](https://github.com/misdoro).
```
dblock: Quote `coerce_with` and `nil`

spec/grape/validations/validators/coerce_spec.rb
misdoro: Updated. Although many other tests are written in a similar way here.

CHANGELOG.md
misdoro: Right, fixed.

misdoro:
I think I would want a sanity check from @dnesteryuk for this PR.

If this is not a regression but a breaking change was introduced by design, it may well just make sense to update the `UPGRADING.md`

spec/grape/validations/validators/coerce_spec.rb
dblock: Yeah and that sucks. Did you refactor all of them? :) Don't hold back.

dblock:
I err on the side of caution and would update UPGRADING anyway.

@dnesteryuk all yours

PR #2031, opened 2020-03-29
Special types provided by Grape are custom ones

### LINKED ISSUE #1986: Array[File] doesn't validate correctly.
If you have a parameter of type `Array[File]`, validation always fails because the primitive type coercer is trying to do `@coercer[val]`, and `@coercer` is `::File`.

```ruby
optional(:nested, type: Hash, default: {}) do
  optional :files, type: Array[File]
end
```

Version 1.2.5 doesn't have this issue, so 1.3.0 broke parameter type coercion.

dnesteryuk:
Users tries to use types provided by Grape as an example how to write custom types. So, it makes sense to treat them as custom.

Besides that, it fixes #1986 where a collection of a custom type wasn't coerced properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 #### Fixes
 
 * Your contribution here.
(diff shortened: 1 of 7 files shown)


dblock:
👍

PR #2043, opened 2020-04-27
Modify declared for nested array and hash

kadotami:
I fixed bugs of declared method.

key of options[:route_options][:params] is not considered for nested params.

I changed to these.
* if nested array is nil, declared method return [].
* if nested hash is nil, declared method return {}.

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -296,9 +296,12 @@ describe Grape::Endpoint do
               optional :seventh
             end
           end
+          optional :nested_arr, type: Array do
+            optional :eighth
+          end
         end
-        optional :nested_arr, type: Array do
-          optional :eighth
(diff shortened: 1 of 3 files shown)


spec/grape/endpoint_spec.rb
```diff
@@ -422,6 +425,7 @@ def app
         get '/declared?first=present'
         expect(last_response.status).to eq(200)
         expect(JSON.parse(last_response.body)['nested']).to be_a(Hash)
+        expect(JSON.parse(last_response.body)['nested']).to eq({})
```
dnesteryuk, line 428: what do you think about parsing the response only once?

kadotami: sorry! I'll modify soon.

dblock:
👍

PR #2045, opened 2020-04-29
Add tests for coercing bad values to a boolean

DanielHeath:
I was going to open an issue, but thought that including code would help illustrate, so it's a PR instead.

Currently, if you supply "" for a boolean parameter, it does not get coerced at all - instead, it gets passed through.

I think that's confusing and inconsistent; it would be better to return `Nil`. Is that a sensible change?

### DIFF
--- a/spec/grape/api/defines_boolean_in_params_spec.rb
+++ b/spec/grape/api/defines_boolean_in_params_spec.rb
@@ -19,14 +19,35 @@ describe Grape::API::Instance do
       DefinesBooleanInstanceSpec::API
     end
 
-    let(:expected_body) do
-      { class: 'TrueClass', value: true }.to_s
-    end
-
-    it 'sets Boolean as a type' do
+    it 'sets "true" as a Boolean type' do
       post '/echo?message=true'
       expect(last_response.status).to eq(201)
-      expect(last_response.body).to eq expected_body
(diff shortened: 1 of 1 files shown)


dnesteryuk:
I would prefer to coerce an empty string to `false` in this case :thinking: The reason of the current behavior is [there](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/types/primitive_coercer.rb#L40).

DanielHeath:
IMO an empty string is an invalid value to assign a boolean and should be treated the same as (eg) passing a string to a numeric value.

myxoh:
I tend to agree I would expect `nil` here as well. Thanks for adding the test!

dblock:
I think I agree, I'd want this to be `nil`. Would take a PR. What else breaks?

dnesteryuk:
Dry-types treats an empty string as invalid for the boolean type :thinking: 

```
Dry::Types::CoercionError ( cannot be coerced to false)
```

dblock:
> Dry-types treats an empty string as invalid for the boolean type 🤔
> 
> ```
> Dry::Types::CoercionError ( cannot be coerced to false)
> ```

Interesting. I don't know ;)

dnesteryuk:
@DanielHeath thanks for the test, the request was implemented :wink:

PR #2049, opened 2020-05-03
Coerce an empty string to nil in case of the bool type

dnesteryuk:
Satisfies a pending test in #2045

### DIFF
--- a/lib/grape/validations/types/primitive_coercer.rb
+++ b/lib/grape/validations/types/primitive_coercer.rb
@@ -36,7 +36,7 @@ module Grape
 
         def call(val)
           return InvalidValue.new if reject?(val)
-          return nil if val.nil?
+          return nil if val.nil? || treat_as_nil?(val)
           return '' if val == ''
 
           super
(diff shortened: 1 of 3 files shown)


lib/grape/validations/types/primitive_coercer.rb
```diff
@@ -46,7 +46,7 @@ def call(val)
 
         attr_reader :type
 
-        # This method maintaine logic which was defined by Virtus. For example,
+        # This method maintain logic which was defined by Virtus. For example,
```
dblock, line 49: If we fixing this, let's actually fix it with "This method maintains" :)

lib/grape/validations/types/primitive_coercer.rb
```diff
@@ -55,6 +55,12 @@ def reject?(val)
             (val.is_a?(String) && type == Hash) ||
             (val.is_a?(Hash) && type == String)
         end
+
+        # Dry-Types treats an empty string as invalid. However, Grape community wants to have
```
dblock, line 59: I think we should explain why and link to the issue. How about: "However, Grape considers an empty string as absence of value and coerces it into `nil`."?

PR #2050, opened 2020-05-03
Refactor route public_send to AttributeTranslator

ericproulx:
Hi,

I transformed `route` public send to actual methods in `attribute_translator`.  Makes more sense than falling back on `method_missing`.

I also removed `Any` in router and use pure `AttributeTranslator`.


dblock:
:shipit:

PR #2002, opened 2020-02-29
Lazy lookup and lazy object

ericproulx:
Hello,

I'm still working on improving string allocations and this PR follows my work.

- I added some lazy lookup tables at some places for a better string allocations.
- I added a lazy object which kind of a proxy for executing code when needed. It's mainly used when building headers for request. I also added a lazy object const that contains converted well known http headers.
- I fixed a small leak in stackable_values for empty arrays. It's now generating empty_array only when needed
- I replaced some %w by actual const

Here's some stats based on our app:
###
Overall Grape object allocations (development environment)
###
```log
# master

Total allocated: 47115631 bytes (437131 objects)
Total retained:  6323963 bytes (16355 objects)

# this PR
Total allocated: 41291815 bytes (300978 objects)
Total retained:  5841323 bytes (11641 objects)
```
### 
benchmark/simple.rb ruby 2.4.9
###
```log
# master
eproulx@Erics-Macbook grape % git checkout master
Switched to branch 'master'
eproulx@Erics-Macbook grape % ruby benchmark/simple.rb
Warming up --------------------------------------
              simple   522.000  i/100ms
Calculating -------------------------------------
              simple      5.420k (± 5.8%) i/s -     27.144k in   5.025995s
eproulx@Erics-Macbook grape % ruby benchmark/simple.rb
Warming up --------------------------------------
              simple   531.000  i/100ms
Calculating -------------------------------------
              simple      5.490k (± 4.7%) i/s -     27.612k in   5.041801s
eproulx@Erics-Macbook grape % git checkout lazy_lookup_lazy_object
Switched to branch 'lazy_lookup_lazy_object'
eproulx@Erics-Macbook grape % ruby benchmark/simple.rb            
Warming up --------------------------------------
              simple   605.000  i/100ms
Calculating -------------------------------------
              simple      5.871k (±10.0%) i/s -     29.040k in   5.000857s
eproulx@Erics-Macbook grape % ruby benchmark/simple.rb
Warming up --------------------------------------
              simple   606.000  i/100ms
Calculating -------------------------------------
              simple      6.409k (± 4.0%) i/s -     32.118k in   5.019708s
```

Thanks

PR #2053, opened 2020-05-14
Fix broken multiple mounts.

Jack12816:
**- What is it good for**

There is a nasty bug on the `Grape::API::Instance#add_head_not_allowed_methods_and_options_methods` method which causes a `FrozenError` when an API class is mounted multiple times. I'm not quite sure why the mounts happens multiple times at my setup. My application includes:
```
1.3.2  grape
0.8.0  grape-entity
1.3.0  grape-jwt-authentication
2.1.0  grape-route-helpers
1.1.0  grape-swagger
0.3.4  grape-swagger-entity
0.3.1  grape-swagger-rails
```

This is the actual stack trace of the issue:

```
FrozenError (can't modify frozen Array):

grape (1.3.2) lib/grape/api/instance.rb:208:in `block (2 levels) in add_head_not_allowed_methods_and_options_methods'
grape (1.3.2) lib/grape/api/instance.rb:199:in `each'
grape (1.3.2) lib/grape/api/instance.rb:199:in `block in add_head_not_allowed_methods_and_options_methods'
grape (1.3.2) lib/grape/api/instance.rb:197:in `each'
grape (1.3.2) lib/grape/api/instance.rb:197:in `add_head_not_allowed_methods_and_options_methods'
grape (1.3.2) lib/grape/api/instance.rb:156:in `initialize'
grape (1.3.2) lib/grape/api/instance.rb:52:in `new'
grape (1.3.2) lib/grape/api/instance.rb:52:in `compile'
grape (1.3.2) lib/grape/api/instance.rb:85:in `block in compile!'
grape (1.3.2) lib/grape/api/instance.rb:85:in `synchronize'
grape (1.3.2) lib/grape/api/instance.rb:85:in `compile!'
grape (1.3.2) lib/grape/api/instance.rb:65:in `call'
grape (1.3.2) lib/grape/api.rb:68:in `call'  
```

**- What I did**

The offending line at the `Grape::API::Instance` just assigns a frozen array constant to the routing settings. When using `CONST.dup` we can simply unfreeze it for later changes.

```ruby
# Causes FrozenError's when later touched by +#<<+ for example
route_settings[:methods] = Grape::Http::Headers::SUPPORTED_METHODS
```

Download


lib/grape/api/instance.rb
```diff
@@ -209,7 +211,7 @@ def add_head_not_allowed_methods_and_options_methods
             route_settings[:endpoint] = route.app
 
             # using the :any shorthand produces [nil] for route methods, substitute all manually
-            route_settings[:methods] = Grape::Http::Headers::SUPPORTED_METHODS if route_settings[:methods].include?('*')
+            route_settings[:methods] = Grape::Http::Headers::SUPPORTED_METHODS.dup if route_settings[:methods].include?('*')
```
dnesteryuk, line 214: I would prefer to duplicate it only when `route_settings[:methods]` gets modified. There are lots of projects which don't use the mount feature, they will "pay" for this duplication without any benefits. This line was changed as a part of https://github.com/ruby-grape/grape/pull/2002 which decreased memory consumption. So, it would be cool to keep it. :slightly_smiling_face: 

Lazy duplication is a way to go :wink:

Jack12816: I just dropped the duplication in favor of some logic. There is no need to fiddle around with the array when the current route method is `*` or there is already an `*` in the methods list. Then we can directly replace it with all the supported methods from the constant. Then there is no cost, like before :)

lib/grape/api/instance.rb
```diff
@@ -191,6 +191,9 @@ def cascade?
       # with a list of HTTP methods that can be called. Also add a route that
       # will return an HTTP 405 response for any HTTP method that the resource
       # cannot handle.
+      #
+      # rubocop:disable Metrics/AbcSize because of the complex workflow
```
dblock, line 195: Disable these in .rubocop.yml or just run `rubocop -a ; rubocop --auto-gen-config`.

lib/grape/api/instance.rb
```diff
@@ -239,6 +243,8 @@ def add_head_not_allowed_methods_and_options_methods
           end
         end
       end
+      # rubocop:enable Metrics/AbcSize
```
dblock, line 246: Remove.

spec/grape/api/instance_spec.rb
```diff
@@ -51,4 +51,48 @@ def app
...
+        end
+      end
+      Class.new(Grape::API) do
+        mount first
+        mount first
+        mount second
+      end
+    end
+
+    it 'does not raise a FrozenError on first instance' do
```
dblock, line 78: I understand that the bug is a frozen error raised, but let's rewrite this as a positive test of what you expect the response to be. Otherwise in the future when the behavior changes but no exceptions are raised we won't catch this issue.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * [#2049](https://github.com/ruby-grape/grape/pull/2049): Coerce an empty string to nil in case of the bool type - [@dnesteryuk](https://github.com/dnesteryuk).
 * [#2043](https://github.com/ruby-grape/grape/pull/2043): Modify declared for nested array and hash - [@kadotami](https://github.com/kadotami).
 * [#2040](https://github.com/ruby-grape/grape/pull/2040): Fix a regression with Array of type nil - [@ericproulx](https://github.com/ericproulx).
+* [#2050](https://github.com/ruby-grape/grape/pull/2053): Fix broken multiple mounts - [@Jack12816](https://github.com/Jack12816).
```
dblock, line 16: Let's be more explicit so that the reader can identify with this change, how about:

_Fix `FrozenError` when mounting an API multiple times._

spec/grape/api/instance_spec.rb
```diff
@@ -51,4 +51,48 @@ def app
       expect(an_instance.top_level_setting.parent).to be_nil
     end
   end
+
+  context 'with multiple moutes' do
+    let(:root_api) do
+      first = Class.new(Grape::API::Instance) do
```
dblock, line 57: You could `let(:first) { }` and `let(:second)` making it a bit cleaner.

dblock:
You still have to remove the manually added `# rubocop:enable...` in the code, I am afraid `-a` doesn't get rid of them.

spec/grape/api/instance_spec.rb
Jack12816: Correct me if I'm getting you wrong, but there is a positive and a negative test?
```
it 'responds the correct body at the first instance' 
it 'does not raise a FrozenError on first instance'
```

Could you explain this in more detail please? :)

dblock: The first covers the second in terms of functionality.

The second standalone is a misleading test. It would make sense if the call in some cases was supposed to raise FrozenError and we explicitly wanted to ensure that in certain conditions that's not the case.

dblock:
I merged this, thank you.

Jack12816:
Thank you!

Issue #1998, opened 2020-02-26
Array[Array[X]]] and Array[Hash[X]]] return wrong indices in errors


senhalil:
Basically the indices of nested arguments are not kept correctly during validation process. It happens both for Hashes and Arrays declared inside another Array. 

In the case of `Array[Array[X]]`  the inner index keeps increasing (in the error messages) suggesting there might be something wrong in the recursion logic of validators -- i.e., it acts as a global count.

Created the following spec to explain the issue (disclaimer: first spec file, not sure if I got it right).

```ruby
# frozen_string_literal: true

require 'spec_helper'

describe Grape::Validations::DefaultValidator do
  describe '#validate!' do
    subject(:validate) { post path, params }

    module ValidationsSpec
      module DefaultValidatorSpec
        class API < Grape::API
          rescue_from Grape::Exceptions::ValidationErrors do |e|
            error!(e.errors.transform_keys! { |key| key.join(',') }, 400)
          end

          params do
            requires :points, type: Array do
              # Following optional can be replaced with requires
              optional :location, type: Hash, allow_blank: false do
                requires :lat, type: Float, allow_blank: false
                requires :lon, type: Float, allow_blank: false
              end
            end
          end
          post '/nested-array-hash' do
          end


          params do
            requires :points, type: Array do
              # Following optional can be replaced with requires
              optional :location, type: Array, allow_blank: false do
                requires :id, type: Fixnum, allow_blank: false
              end
            end
          end
          post '/nested-array-array' do
          end

        end
      end
    end

    def app
      ValidationsSpec::DefaultValidatorSpec::API
    end

      context 'when params are nested inside hash which is inside an array' do
        let(:path) { '/nested-array-hash' }
        let(:params) {
          {
            points: [
              { location: nil },
              { location: { lat: nil, lon: 0.0 } },
              { location: { lat: "string", lon: 0.0 } },
              { location: { lat: 0.0, lon: "string" } },
              { location: { lat: 0.0, lon: nil } },
              { location: nil },
            ]
          }
        }

        it 'does return a validation error with correct array indices' do
          validate
          expect(JSON.parse(last_response.body)).to eq(
            'points[0][location]'      => ['is empty'],
            'points[1][location][lat]' => ['is empty'],
            'points[2][location][lat]' => ['is invalid'],
            'points[3][location][lon]' => ['is invalid'],
            'points[4][location][lon]' => ['is empty'],
            'points[5][location]'      => ['is empty'],
          )
        end
[...]

PR #2055, opened 2020-05-20
Add failing test for multiple nested params error indices

infiniteluke:
This reproduces the issue described in #1998. The `AttributesIterator` seems to control what the a given scopes index is set to. I believe this bug is due to index not iterating in sync with each param being validating in an array.

You can see here that the scope is set to 2 before we even start validating the `value` param:
image

### DIFF
--- a/lib/grape/exceptions/validation_errors.rb
+++ b/lib/grape/exceptions/validation_errors.rb
@@ -23,6 +23,7 @@ module Grape
       end
 
       def each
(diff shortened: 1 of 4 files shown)


StefanHagen:
Hey @infiniteluke ! This is exactly the issue I'm having currently, and I see the issue is still open. Did you find a workaround for this problem? Thanks for making that spec!

PR #2062, opened 2020-05-26
Add note about an undocumented regression in bool coercion

ghiculescu:
This pattern wasn't documented, but it was all over our codebase, which means we probably saw it online somewhere and it worked fine.

Adding a note to upgrading guide for future readers (/future me) trying to debug tests breaking in this update.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -125,10 +125,10 @@ requires :true_value, type: TrueClass
 requires :bool_value, types: [FalseClass, TrueClass]
 ```
 
-These will now crash. Instead you should do:
+This is no longer supported, if you do this, your values will never be valid. Instead you should do this:
 
 ```ruby
-requires :true_value, type: Boolean # in your endpoint you should validate if this actually `true`
+requires :true_value, type: Boolean # in your endpoint you should validate if this is actually `true`
 requires :bool_value, type: Boolean
 ```
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -121,15 +121,15 @@ end
 Previous Grape versions allowed this, though it wasn't documented:
 
 ```ruby
-requires :valueTrue, type: FalseClass
-requires :valueBool, types: [FalseClass, TrueClass]
+requires :true_value, type: TrueClass
+requires :bool_value, types: [FalseClass, TrueClass]
 ```
 
 These will now crash. Instead you should do:
 
 ```ruby
-requires :valueTrue, type: Boolean # in your endpoint you should validate if this actually `false`
-requires :valueBool, type: Boolean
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -116,6 +116,22 @@ params do
...
+#### Coercing to `FalseClass` or `TrueClass` no longer works
+
+Previous Grape versions allowed this, though it wasn't documented:
+
+```ruby
+requires :true_value, type: TrueClass
+requires :bool_value, types: [FalseClass, TrueClass]
+```
+
+These will now crash. Instead you should do:
```
dblock, line 128: Crash? What actually happens?

dblock:
Why do we feel that this is not just a regression/bug?

UPGRADING.md
ghiculescu: https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/types/dry_type_coercer.rb#L65 <- this line crashes, because `@coercer` is `false` or `true`, rather than an instance of a coercer.

ghiculescu:
I think it's an issue on our end because https://github.com/ruby-grape/grape#supported-parameter-types says `Boolean` is supported but nothing special about `FalseClass` and `TrueClass`.

dblock:
> I think it's an issue on our end because https://github.com/ruby-grape/grape#supported-parameter-types says `Boolean` is supported but nothing special about `FalseClass` and `TrueClass`.

I don't see why we shouldn't support `TrueClass` and `FalseClass`. I'll merge this PR because this is indeed a regression and it's better to have it documented, but I'd like this to be (re)implemented as a feature - maybe take a stab at it?

ghiculescu:
Comment addressed [here](https://github.com/ruby-grape/grape/pull/2062#discussion_r430563117) unless you want more detail in the upgrade guide?

ghiculescu:
@dblock i'm happy to have a go at that but in a separate PR. wanna get our upgrade done first :)

UPGRADING.md
dblock: I just mean we need to say what users get instead of saying "Crash", ie. explain this to the users. It should say something like this: "The example above will now cause `error bla bla bla`." To work around this, use a `Boolean` and check for its value.

ghiculescu: I clarified the wording. The exception is actually handled on the next line, so it's not a crash, it's just that the value is never considered valid.

dblock:
Merged. 

I think an API that uses types that cannot be coerced into, such as in this case, should not mount. I would take a PR that does that for even a handful of primitive types like `TrueValue` and `FalseValue`.

Issue #1992, opened 2020-02-13
Ruby 2.7 deprecation warning with middleware/base


skarger:
Ruby 2.7 changes the way keyword args work:
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

This causes the following deprecation warning:

```
gems/rack-2.2.2/lib/rack/builder.rb:158: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
gems/2.7.0/gems/grape-1.3.0/lib/grape/middleware/base.rb:17: warning: The called method `initialize' is defined here
```

Here is the relevant line in rack 2.2.2:
https://github.com/rack/rack/blob/ffa6dc789ca2f35f39acab15467321b1f9f1ca75/lib/rack/builder.rb#L158

I'm not completely clear on how grape should fix. This may help:
https://eregon.me/blog/2019/11/10/the-delegation-challenge-of-ruby27.html

dblock:
Can you please take a look at fixes in #1949 and #1951, we probably need something in the same lines. Feel free to PR.

dim:
I would create a PR, but I am not entirely sure how where exactly this warning originates. Any ideas?

skarger:
@dim I've held off too because I wasn't sure about the fix, but I gave it a shot here:
https://github.com/ruby-grape/grape/pull/2064

@dblock I took the opposite approach of the PRs you mentioned because the call that prompts the warning happens in Rack.

dim:
@skarger I commented on the PR, I think the `options` argument must remain optional

dblock:
Closed via https://github.com/ruby-grape/grape/pull/2064.

dim:
Hmm, I am still seeing this:

```
../gems/rack-2.2.3/lib/rack/builder.rb:158: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
../gems/grape-1.4.0/lib/grape/middleware/error.rb:30: warning: The called method `initialize' is defined here
```

skarger:
@dim that's coming from a different class, `Middleware::Error` vs. `Middleware::Base`.

Fixed form: https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/base.rb#L17
Remaining warning: https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/error.rb#L30

The Error case may be fixable in a similar way to #2064.

Issue #2030, opened 2020-03-28
Release v. Next


dblock:
Ongoing issue. If you feel like you need a new release because of changes in master, please add your +1. One of the maintainers will do it when they have time. A great way to accelerate this process is to send one of them a pony. Or coffee.

stanhu:
@dblock Any chance you can get around to this soon?

dblock:
This weekend unless someone beats me to it.

dnesteryuk:
I will take care of that on this weekend, most likely, today :smile: 

![grape   RubyGems org   your community gem host](https://user-images.githubusercontent.com/128024/79039641-5b72c280-7beb-11ea-86b9-241506412fc2.png)

stanhu:
Thank you @dnesteryuk and @dblock. 1.3.2 looks great. I think we can close this out.

dnesteryuk:
I guess we can keep this thread, so people can request next release :wink:

nbulaj:
There are some changes inroduced in 1.3.0 (and still valid for 1.3.2) that broke our production app after update.

We have app1 (Rails + Grape), app2 (Sinatra), frontend app. Frontend app communicates with app2 which in turn proxy requests to app1 using `curb` gem. When we used Grape 1.2.5 - everything was OK. But when we just changed version number in Gemfile and run the app with 1.3.0 - frontend app started to show infinite loader for requests. While debugging I found that Grape API start to respond to `curb` with Transfer-Encoding `chunked` O_o

I'll try to create a reproducable script for this, but don't sure how long it ill take. Maybe maintainers have an idea what could face such issue? I reviewed all the changes in git between 1.2.5 and 1.3.0 and didn't find the culprit :( Maybe I'll also try "by-commit" strategy to investigate

dblock:
> And _app2_ starts to receive response headers with Transfer-Encoding `chunked` and never finish the request. Should I create an issue for that? I don't know if could create a reproducable example..

Thanks for bringing this up and for hunting down what changed. You should open a new issue either way. I think we need two things:

1. Find a workaround. You should figure out why your app stopped working and how to fix the app - IMO transfer-encoding and request not finishing are just symptoms, why would it break a client?
1. A new paragraph in UPGRADING once we figure out what the issue is.

nbulaj:
OK I created a reproducable example with small readme on how to test it: https://gist.github.com/nbulaj/bd89ba366290033bf71c0f9b60927709

[...]

PR #2064, opened 2020-05-28
Treat Middleware::Base#initialize options as Hash.

skarger:
Addresses https://github.com/ruby-grape/grape/issues/1992.

Before this PR, `Grape::Middleware::Base` is treating the `options` param as keyword args, but it's actually meant as a `Hash` positional argument.

https://github.com/ruby-grape/grape/blob/536622a38ae5ed266477689cdd7afabd92046ed2/lib/grape/middleware/base.rb#L17

The former style, `**options`, prompts a deprecation warning in Ruby 2.7.

`gems/2.7.0/gems/rack-2.2.2/lib/rack/builder.rb:158: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`
`gems/2.7.0/gems/grape-1.3.3/lib/grape/middleware/base.rb:17: warning: The called method 'initialize' is defined here`.

For a middleware subclass of `Grape::Middleware::Base`, Rack's `use` method will instantiate it with `middleware.new(app, *args, &block)`, passing the `options` parameter in `*args`:

https://github.com/rack/rack/blob/a5e80f01947954af76b14c1d1fdd8e79dd8337f3/lib/rack/builder.rb#L158

My assumption is that, contrary to the suggestion by the deprecation warning, it's not feasible to change Rack to add `**` to the call where it passes `*args`. Plus, Grape's code implies that `options` really is meant as a positional Hash argument, so I changed it to be one.

The tests pass after this change, including
https://github.com/ruby-grape/grape/blob/v1.3.3/spec/grape/middleware/base_spec.rb.

The main risk I see here is if there's some behavior or usage style that the `**options` version supported that will now break.

### DIFF
--- a/lib/grape/middleware/base.rb
+++ b/lib/grape/middleware/base.rb
@@ -14,9 +14,9 @@ module Grape
 
       # @param [Rack Application] app The standard argument for a Rack middleware.
       # @param [Hash] options A hash of options, simply stored for use by subclasses.
-      def initialize(app, **options)
+      def initialize(app, options = {})
         @app = app
-        @options = default_options.merge(**options)
+        @options = default_options.merge(options)
         @app_response = nil
       end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 #### Fixes
 
 * Your contribution here.
-* [](): Fix Ruby 2.7 deprecation warning in `Grape::Middleware::Base#initialize` - [@skarger](https://github.com/skarger).
+* [2064](https://github.com/ruby-grape/grape/pull/2064): Fix Ruby 2.7 deprecation warning in `Grape::Middleware::Base#initialize` - [@skarger](https://github.com/skarger).
 
 ### 1.3.3 (2020/05/23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 #### Fixes
 
 * Your contribution here.
-* [2064](https://github.com/ruby-grape/grape/pull/2064): Fix Ruby 2.7 deprecation warning in `Grape::Middleware::Base#initialize` - [@skarger](https://github.com/skarger).
+* [#2064](https://github.com/ruby-grape/grape/pull/2064): Fix Ruby 2.7 deprecation warning in `Grape::Middleware::Base#initialize` - [@skarger](https://github.com/skarger).
 
 ### 1.3.3 (2020/05/23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 #### Fixes
 
 * Your contribution here.
+* [](): Fix Ruby 2.7 deprecation warning in `Grape::Middleware::Base#initialize` - [@skarger](https://github.com/skarger).
 
 ### 1.3.3 (2020/05/23)
 


dblock:
Thanks! Can we please 1) add a test that makes sure options are properly assigned 2) fix the changelog

dblock:
It would be nice if tests failed on warnings, too. If you find out how to enable that part of this PR, I'd appreciate it!

lib/grape/middleware/base.rb
```diff
@@ -14,9 +14,9 @@ class Base
 
       # @param [Rack Application] app The standard argument for a Rack middleware.
       # @param [Hash] options A hash of options, simply stored for use by subclasses.
-      def initialize(app, **options)
+      def initialize(app, options = {})
```
dim, line 17: I think the backwards-compatible workaround solution would be:

```ruby
def initialize(app, options = nil)
  @app = app
  @options = default_options.dup
  @options.update(options) if options.is_a?(Hash)
  @app_response = nil
end
```

This retains the optional argument character of `options` while resolving the deprecation warning.

dim:
Actually, I think this needs to be fixed in Rack::Builder. I will submit an issue.

dim:
Looking at https://github.com/dim/rack/commit/8dad94973dd783fa2ab9246b4dcfcaeb3368061a, I think just using https://github.com/ruby/ruby2_keywords in Middleware::Base may also solve this issue

lib/grape/middleware/base.rb
dblock: And do we need it to be backwards compatible here?

skarger: Defaulting the `options` parameter to an empty hash is backwards compatible. When declaring the parameter with `**options`, the `options` variable inside the method will default to an empty hash. 

See this example, run with Ruby 2.7.

```
def example_method(positional, **options)
  puts "  positional: #{positional}"
  puts "  options class: #{options.class}"
  puts "  options: #{options}\n"
end
```

```
Given positional argument only:
example_method("arg1")
  positional: arg1
  options class: Hash
  options: {}

Given positional argument followed by hash argument:
example_method("arg1", a_key: "a value")
  positional: arg1
  options class: Hash
  options: {:a_key=>"a value"}
```

Like the existing `initialize` method in `Grape::Middleware::Base`, the fact that `example_method` has `options` with the `**` in front declares it a keyword arguments, not as a (positional) hash argument. However the logic inside `Grape::Middleware::Base#initialize` treats `options` as a hash - it just merges it with `default_options`.

When passing a hash as a positional argument, Ruby 2.7 prints the deprecation warning because it's treating the given hash as keyword arguments in accordance with the method signature:

```
Given positional argument followed by positional hash:
example_method("arg1", { a_key: "a value" })
test.rb:15: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
test.rb:1: warning: The called method `example_method' is defined here
  positional: arg1
  options class: Hash
  options: {:a_key=>"a value"}
```

Changing the signature to `def example_method(positional, options = {})` makes the deprecation warning go away, but otherwise matches the same behavior for all of these examples. This is what I've done in this PR.

skarger:
> Can we please 1) add a test that makes sure options are properly assigned

@dblock the existing tests do that, and they still pass after this change:
https://github.com/ruby-grape/grape/blob/v1.3.3/spec/grape/middleware/base_spec.rb#L127

I'm happy to add another test, if there's some detail not covered in the existing tests that you're thinking about?

dblock:
The CHANGELOG line is missing a `#` in the PR number. Danger is nitpicky ;)

dblock:
I'm satisfied with this, will merge on 🍏

skarger:
@dblock fixed, thanks for looking.

dim:
Any chance for a v1.3.4 release?

dblock:
Please add a +1 to #2030

Issue #2066, opened 2020-06-03
Date coercion discrepancy between 1.2.5 and 1.3.0


petekinnecom:
Thanks for the gem/project.

In upgrading from 1.2.5 to 1.3.0, I've noticed that the way an empty string is coerced for a Date param has changed. Previously, it would be returned as `nil`, now it is returned as `""`.  Based on [Upgrading.md](https://github.com/ruby-grape/grape/blob/master/UPGRADING.md#upgrading-to--130) it's not clear to me whether this change is desired or would be considered a bug.  

Here is a script to demonstrate the discrepancy:

```
require 'bundler/inline'

gemfile do
  # gem 'grape', git: 'git@github.com:ruby-grape/grape' # returns { class: "String" }
  # gem 'grape', '1.3.0' # returns { class: "String" }
  # gem 'grape', '1.2.5' # returns { class: "NilClass" }

  gem 'rack', '2.0.6'
end

class API < Grape::API
  params do
    requires :mydate, type: Date
  end
  get '/' do
    { class: params.fetch(:mydate).class.inspect }
  end
end

Thread.new { Rack::Server.start(app: API, Port: 9292) }

sleep 2 # adjust as desired :P

require 'net/http'
puts Net::HTTP.get('localhost', '/?mydate=', 9292)
```

Thanks.

dblock:
You should never get a String when you want a Date, so I think this is a bug. Care to write a spec?

petekinnecom:
Yeah, I'll open a PR

petekinnecom:
https://github.com/ruby-grape/grape/pull/2067

PR #2067, opened 2020-06-03
Coerce empty string to nil for all Primitive types except String

petekinnecom:


petekinnecom:
Here's a reproduction of the behavior on v.1.2.5: https://github.com/petekinnecom/grape/commit/ec99d8cec24760934b3b0adc5a7ca46a139f0dfe#diff-d4c132540f49911bbd1c73264c2c738bR12

dblock:
This does modify external behavior of interacting with an API. I think we need:

- A spec as comprehensive as https://github.com/petekinnecom/grape/commit/ec99d8cec24760934b3b0adc5a7ca46a139f0dfe#diff-d4c132540f49911bbd1c73264c2c738bR12
- A API-level spec describing the behavior in https://github.com/ruby-grape/grape/issues/2066

WDYT?

dnesteryuk:
@dblock do you mean more integration tests? What is wrong with unit tests? :thinking: Personally, I like more unit tests, they are more focused (point to code under tests).

dblock:
> @dblock do you mean more integration tests? What is wrong with unit tests? 🤔 Personally, I like more unit tests, they are more focused (point to code under tests).

Yes. This behavior affects what happens when you call an API, and there are no changes in tests for that external behavior. We often rely on those integration tests when we're not sure what's supposed to happen and still keep introducing regressions. Better safe than sorry. @petekinnecom if you aren't adding any I'm OK with merging this though. Just let us know?

petekinnecom:
I've pushed a commit that has integration tests that mirror the tests found for `nil` (they are found just above in the same file I edited). The behavior for primitive types is all that's been updated here, so for other types, this just captures existing behavior.  For the exceptions here the params are not parsed as an empty string but rather cause a `400` response:

https://github.com/ruby-grape/grape/pull/2067/commits/619db09483194f27f87719d840ef23e72e7de565#diff-d4c132540f49911bbd1c73264c2c738bR533
https://github.com/ruby-grape/grape/pull/2067/commits/619db09483194f27f87719d840ef23e72e7de565#diff-d4c132540f49911bbd1c73264c2c738bR550

dblock:
Perfect, merged. Thank you.

PR #2072, opened 2020-06-12
Fix eager loading due to ServeStream rename

stanhu:
ServeFile was renamed to ServeStream in #1520. Calling Grape.eager_load!
would fail with:

```
uninitialized constant Grape::ServeFile (NameError)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,10 @@ matrix:
     - rvm: 2.7.1
       gemfile: gemfiles/rails_6.gemfile
     - rvm: 2.7.1
+      gemfile: Gemfile
+      script:
+        - bundle exec rspec spec/integration/eager_load
+    - rvm: 2.7.1
       gemfile: gemfiles/multi_json.gemfile
       script:
         - bundle exec rake
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#2072](https://github.com/ruby-grape/grape/pull/2072): Fix eager loading due to ServeStream rename - [@stanhu](https://github.com/stanhu).
 * [#1520](https://github.com/ruby-grape/grape/pull/1520): Un-deprecate stream-like objects - [@urkle](https://github.com/urkle).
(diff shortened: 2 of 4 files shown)


dblock:
We can add an integration spec for this, can't we? Requiring the file should raise an error rn.

dblock:
Bump @stanhu want to finish this?

stanhu:
@dblock Ok, done.

.travis.yml
```diff
@@ -29,11 +29,13 @@ matrix:
...
       script:
         - bundle exec rake
+        - bundle exec rspec spec/integration/eager_load
         - bundle exec rspec spec/integration/multi_json
     - rvm: 2.7.1
       gemfile: gemfiles/multi_xml.gemfile
       script:
         - bundle exec rake
+        - bundle exec rspec spec/integration/eager_load
         - bundle exec rspec spec/integration/multi_xml
```
dblock, line 39: Let's not mix multi_json and multi_xml tests with eager loading tests. Add a new section to use the default gemfile and run `rspec spec/integration/eager_load`?

.travis.yml
```diff
@@ -25,6 +25,11 @@ matrix:
       gemfile: gemfiles/rails_5.gemfile
     - rvm: 2.7.1
       gemfile: gemfiles/rails_6.gemfile
+    - rvm: 2.7.1
+      gemfile: Gemfile
+      script:
+        - bundle exec rake
```
dblock, line 31: No need to do this since we already run tests for 2.7.x elsewhere.

dblock:
Errr... now looking into this I have more questions.

- does this even need to be an integration spec? I can easily fail this with just adding this spec as is
- what is this file supposed to do? we don't document using eager loading like this - it came from https://github.com/ruby-grape/grape/pull/1904, @myxoh

dnesteryuk:
This issue also happens after calling:

```ruby
API.compile!
```

So, we might use it in the integration test.

dblock:
@stanhu Want to finish so we can include this in the release? Let's include `API.compile!` too.

stanhu:
@dblock I'm not sure if this is what you wanted, but take a look.

dblock:
Made some changes and merged via https://github.com/ruby-grape/grape/commit/b3adbb4060f81a14f1cba35cabe796d35a442bff.

PR #2076, opened 2020-06-26
Populate the env route information when calling the automatically added endpoints

anakinj:
Noticed that calling the `route` method in a `before` hook causes an `NoMethodError` when the hook is invoked for the automatically generated Method Not Allowed endpoints. The reason seems to be that the route information dictionary is not populated in these cases.

These changes will populate the route information before calling the method not allowed endpoint. Also included some minor refactoring for the methods involved in all this.

### DIFF
--- a/lib/grape/api/instance.rb
+++ b/lib/grape/api/instance.rb
@@ -192,37 +192,15 @@ module Grape
       # will return an HTTP 405 response for any HTTP method that the resource
       # cannot handle.
       def add_head_not_allowed_methods_and_options_methods
-        routes_map = {}
-
-        self.class.endpoints.each do |endpoint|
-          routes = endpoint.routes
-          routes.each do |route|
-            # using the :any shorthand produces [nil] for route methods, substitute all manually
-            route_key = route.pattern.to_regexp
-            routes_map[route_key] ||= {}
-            route_settings = routes_map[route_key]
-            route_settings[:pattern] = route.pattern
-            route_settings[:requirements] = route.requirements
-            route_settings[:path] = route.origin
-            route_settings[:methods] ||= []
-            if route.request_method == '*' || route_settings[:methods].include?('*')
-              route_settings[:methods] = Grape::Http::Headers::SUPPORTED_METHODS
-            else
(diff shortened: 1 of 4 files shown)


lib/grape/api/instance.rb
```diff
@@ -198,19 +198,21 @@ def add_head_not_allowed_methods_and_options_methods
...
             # using the :any shorthand produces [nil] for route methods, substitute all manually
-            route_key = route.pattern.to_regexp
-            routes_map[route_key] ||= {}
-            route_settings = routes_map[route_key]
-            route_settings[:pattern] = route.pattern
-            route_settings[:requirements] = route.requirements
-            route_settings[:path] = route.origin
+            route_settings = routes_map[route.pattern.to_regexp] ||= {}
+
+            route_settings.merge!(options: {},
```
dblock, line 203: This is a bit hard to grok IMO. Feels like we should decide whether we want to use `merge!` or assign fields. Are you trying to optimize because `merge` can do things in bulk? I would `merge` everything or assign fields one by one. I also really dislike the side effect of `||=` and then taking a `route_settings` reference on that.

anakinj: Totally agree. I've initially re-wrote it to try to understand what it does also RuboCop had some issues with the length of the method when I added that one `:options` key into the hash.

So now as a second attempt to understand what it does i re-wrote it again :).

anakinj:
The more I look into this the more I feel that the automatically generated endpoints do not belong into the core (was not even aware of this functionality before the before hook started to give us a hard time). Was thinking that maybe it could be possible to extract the features into a separate GEM or something.

dblock:
I merged the change, thank you.

> The more I look into this the more I feel that the automatically generated endpoints do not belong into the core (was not even aware of this functionality before the before hook started to give us a hard time). Was thinking that maybe it could be possible to extract the features into a separate GEM or something.

Generally unless we feel that nobody wants a feature and we feel that it was a mistake, we try not to introduce backwards incompatible changes. Either way open an issue (or a PR) and describe what you think we should do? We can discuss it there?

PR #2077, opened 2020-06-27
Simplify logic for defining declared params

dnesteryuk:
Now, route settings get declared params when an endpoint gets initialized, the `Grape::Validations::ParamsScope` doesn't need to worry about that anymore.

Since the endpoint takes care of finalizing declared params, the `declared_params` method works with a structure which doesn't require any processing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [#1520](https://github.com/ruby-grape/grape/pull/1520): Un-deprecate stream-like objects - [@urkle](https://github.com/urkle).
 * [#2060](https://github.com/ruby-grape/grape/pull/2060): Drop support for Ruby 2.4 - [@dblock](https://github.com/dblock).
 * [#2060](https://github.com/ruby-grape/grape/pull/2060): Upgraded Rubocop to 0.84.0 - [@dblock](https://github.com/dblock).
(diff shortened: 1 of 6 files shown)


dblock:
👍

dblock:
This caused https://github.com/ruby-grape/grape-swagger/pull/801, at the very least this is not expected. 

Got some time to look at it, @dnesteryuk?

dnesteryuk:
@dblock yeah, I will take a look this week

LeFnord:
@dnesteryuk … think it's not realy needed, cause `declared_params` will not be used anymore in grape-swagger

dnesteryuk:
@LeFnord thanks for you fix :+1:

mobilutz:
In a project I am working on, we used the `@declared_params` attribute to get data for a StrictHash Validator.

Is there an alternative way now to get `declared_params` within a Validator?

This is the validator pseudo code:
```
          actual_keys = params[attr_name].keys.map(&:to_sym)
          errors = (actual_keys - allowed_keys).map { |key| error(key, attr_name) }
          
          raise ::Grape::Exceptions::ValidationArrayErrors, errors if errors.any?
        end

        def allowed_keys
          # NOTE: This of course does not return the real keys, just shows that we used the now `nil` @declared_params
          @allowed_keys ||= @scope.instance_variable_get(:@declared_params)
        end
```

Issue #2079, opened 2020-07-06
Empty arrays (as passed by her/faraday) are parsed as [nil]


PragTob:
:wave: 

And first thanks for all the work on the project! :green_heart: 

We have a grape API and applications talking to each other, for which we use [her](https://github.com/remi/her) which uses [faraday](https://github.com/lostisland/faraday).

At some point the faraday behavior changed for sending empty arrays https://github.com/lostisland/faraday/pull/801

The problem occurs when we use her to query resoruces `Resources::Something.where(param_name: [])`. It used to be that they were omitted (at least as URL params, unsure) but now they are sent as `param_name[]` in the URL, which is a bugfix. Grape parses this as `[nil]` which is not what I'd expect (I'd expect `[]`) and also then fails the parameter validation (we have a type of `Array[String]`). `allow_blank: true` doesn't help because an array with nil isn't blank (and we'd still have the wrong value to deal with).

We're running the latest grape release (1.3.3).

I'll work around the issue with `coerce_with` or defining my own type or something. Still thought it was worth opening an issue about it.

dblock:
There was a bunch of work related to this in 1.4.0 (HEAD). I suggest writing a spec and seeing if master has fixed this?

PragTob:
@dblock thanks for letting me know :) I have a bunch of specs for this. Let me see if it's fixed on main.

PragTob:
@dblock just tried it out with 197d5b463b83c9417381837186d6dce7c6fd4ef0 - still exactly the same behavior for us.

dnesteryuk:
@PragTob Could you add a spec for this bug, please? 

I've tried this in tests

```ruby
get '/required?cats[]='
```

`declared(params)` gave:

```ruby
{"cats"=>[""]}
```

PragTob:
Morning @dnesteryuk !

I believe the difference is `=` vs. no `=` aka in our case we don't have the `=`/the way faraday/grape queries.

I'll add a PR with a spec (at least) just have to check whether I can do it on company time or if it's gotta be my own :)

PragTob:
Failing spec etc. sits at #2080 now

PragTob:
Moved to https://github.com/rack/rack/issues/1696 as it seems to me it's rather in rack.

PR #2054, opened 2020-05-15
coercing of nested arrays

dnesteryuk:
It is a regression which was introduced after migrating to DryTypes. Now, it is again possible to define an array of arrays as a type.

```ruby
params do
  requires :skills, type: Array[[String]]
end
```

Fixes #2041

PR #2080, opened 2020-07-07
Document empty array parsing bug for #2079

PragTob:
Was unsure where to put the specs, there's probably a better
place but here seemed fine for now.

Added some other array related specs for good measure.

Happy to also take a dig at resolving this bug if someone pointed me the right direction :)

For #2079

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -589,6 +589,40 @@ describe Grape::Endpoint do
     end
   end
 
+  describe 'array handling' do
+    let(:json) { JSON.parse(last_response.body) }
+    let(:arr) { json.fetch 'arr' }
+
+    before do
+      subject.format :json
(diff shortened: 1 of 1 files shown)


dblock:
> Happy to also take a dig at resolving this bug if someone pointed me the right direction :)

Please do. I would look at other array coercing fixes, e.g. https://github.com/ruby-grape/grape/pull/2054 for inspiration.

PragTob:
Moved to https://github.com/rack/rack/issues/1696 as it seems to me it's rather in rack.

richkettle:
Im also experiencing this issue. A fix that worked for me is to change

`put "/api/url/food/bar", params: params, headers: headers`

to:

`put "/api/url/food/bar", params: params.to_json, headers: headers.merge("CONTENT_TYPE" => "application/json")`

PR #2082, opened 2020-07-09
introduce a finalize! method

dnesteryuk:
There are a few variables (settings) which are only needed while defining API, once everything is set up those variables
could be cleaned up to save memory.

So, the `finalize!` method was added to the API and Instance classes, I found only few variables which might be cleaned up, over time there might be more.

The `finalize!` method could be a part of the `compile!` method, however, `compile!` must be called on a child level (for example, `Twitter::API.compile!`) while `finalize!` must be called on `Grape::API` to find all inheritors and clean them up.

This PR isn't finished, I would like to get early feedback before going deeper.

## TODO:
 - [ ] better tests
 - [ ] document this feature in README.md
 - [ ] add a section to UPGRADING.md
 - [x] update changelog

### DIFF
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -8,7 +8,7 @@ module Grape
   # should subclass this class in order to build an API.
   class API
     # Class methods that we want to call on the API rather than on the API object
-    NON_OVERRIDABLE = (Class.new.methods + %i[call call! configuration compile! inherited]).freeze
+    NON_OVERRIDABLE = (Class.new.methods + %i[call call! configuration compile! inherited finalize!]).freeze
 
     class << self
       attr_accessor :base_instance, :instances
@@ -33,6 +33,7 @@ module Grape
         @setup = []
         @base_parent = base_instance_parent
         @base_instance = mount_instance
+        @finalized = false
       end
 
       # Redefines all methods so that are forwarded to add_setup and be recorded
@@ -128,6 +129,20 @@ module Grape
         instance_for_rack.compile! # See API::Instance.compile!
       end
 
+      # Finds all inheritors and cleans up variables which are only needed while setting up API.
+      # This method must be called on the +Grape::API+ class. Otherwise, all heirs won't be
+      # cleaned up.
+      def finalize!
+        @setup = nil
+
+        # find heirs and call +finalize!+ on them
+        ObjectSpace.each_object(Class).each { |klass| klass.finalize! if klass < self }
+
+        (instances || []).each(&:finalize!)
+
+        @finalized = true
+      end
+
       private
 
       def instance_for_rack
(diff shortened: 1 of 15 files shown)


lib/grape/api.rb
```diff
@@ -141,7 +156,7 @@ def instance_for_rack
       # Adds a new stage to the set up require to get a Grape::API up and running
       def add_setup(method, *args, &block)
         setup_step = { method: method, args: args, block: block }
-        @setup << setup_step
+        @setup << setup_step unless @finalized
```
dblock, line 159: When does this happen? Should this be an exception?

Add `finalized?` and `unless finalized?`.

dm1try:
> There are a few variables (settings) which are only needed while defining API

The whole `namespace_stackable/inheritable` thing is needed only while defining API.
After the setup you should have a router and bunch of configured API handlers(they should be "flat" in memory/there is no need for inheritance/stack/tree structure in runtime). The "alternative inheritance" mechanism provided by DSL is only needed  for proper configuration of those handlers on setup.

In that sense, I would re-write the framework from scratch =)

lib/grape/api.rb
dnesteryuk: @dblock Which exception do you mean? Here we stop adding method calls to the @setup, for example:

```ruby
Api.get '/' do
  # logic here
end
```

It won't be added after calling `finalize!`

dnesteryuk:
I've measured RAM before and after the change. A following tiny API was used:

```ruby
class API < Grape::API
  version 'v1', using: :path

  helpers do
    params :address_params do
        requires :street, type: String
        requires :postal_code, type: Integer
        optional :city, type: String
        optional :tags, type: Array[String]
      end
    end
  end

  namespace :addresses do
    params do
      use :address_params
    end

    post '/' do
      # some logic
    end

    patch '/' do
      # some logic
    end
  end
end
```

### Before calling `finalize!`

```
Total allocated: 380742 bytes (3676 objects)
Total retained:  63946 bytes (494 objects)
```

### After calling `finalize!`

```
Total allocated: 381118 bytes (3681 objects)
Total retained:  62658 bytes (483 objects)
```

_1,25 KB was saved in this case_

dblock:
How do we really feel about introducing a new API for something so minor in terms of memory allocation gain? In many ways I feel that we're doing GC's work. 

Should we be finding a way to finalize automatically in the first request?

myxoh:
A good time to call finalize! might be after mounting the API / Endpoint. Probably we should be cleaning up the set up objects after the endpoints are mounted (I think to remember we had a callback for on_mount in a few of these places)

dnesteryuk:
@myxoh There are cases when an endpoint is a top one. In that case, setup steps will stay as they are. Actually, after measuring memory, I am not sure whether it is worth. :thinking:

dnesteryuk:
The profit of this change is too low. I am not going to continue.

Issue #2087, opened 2020-07-10
Grape v1.4.0 regression: Cache-Control is always set to no-cache


stanhu:
In Grape v1.4.0 https://github.com/ruby-grape/grape/pull/1520/files#diff-8e15cddb80f42ca7d43872846d80ecbeR275 broke our test suite because `Cache-Control` is always set to `no-cache` in https://github.com/ruby-grape/grape/blob/84d68bd3f92a3cffcb3bced8b387bf98779d61ef/lib/grape/dsl/inside_route.rb#L333. 

We set `Cache-Control` to `max-age=0, private, must-revalidate, no-store, no-cache`.

Maybe we should append`no-cache` only if isn't present? Or maybe the headers shouldn't be touched if `nil` is returned?

stanhu:
I'm also not sure if overriding the `Content-Length` and `Transfer-Encoding` by default creates other issues.

jarthod:
Got the exact same problem. Good to see it's already fixed :clap: waiting for the next release then ;)

PR #2088, opened 2020-07-10
Set Cache-Control only for streamed responses

stanhu:
The pull request #1520 introduced a regression that always caused the
`Cache-Control` HTTP header to be set to `no-cache`, even if the
 response wasn't a stream. To fix this, we only set HTTP headers if there
 is an actual stream.

Closes https://github.com/ruby-grape/grape/issues/2087

### DIFF
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -328,6 +328,8 @@ module Grape
       # * https://github.com/rack/rack/blob/99293fa13d86cd48021630fcc4bd5acc9de5bdc3/lib/rack/chunked.rb
       # * https://github.com/rack/rack/blob/99293fa13d86cd48021630fcc4bd5acc9de5bdc3/lib/rack/etag.rb
       def stream(value = nil)
+        return if value.nil? && @stream.nil?
+
         header 'Content-Length', nil
         header 'Transfer-Encoding', nil
         header 'Cache-Control', 'no-cache' # Skips ETag generation (reading the response up front)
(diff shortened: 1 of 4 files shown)


lib/grape/dsl/inside_route.rb
```diff
@@ -328,6 +328,8 @@ def sendfile(value = nil)
       # * https://github.com/rack/rack/blob/99293fa13d86cd48021630fcc4bd5acc9de5bdc3/lib/rack/chunked.rb
       # * https://github.com/rack/rack/blob/99293fa13d86cd48021630fcc4bd5acc9de5bdc3/lib/rack/etag.rb
       def stream(value = nil)
+        return if value.nil? && !instance_variable_defined?(:@stream)
```
stanhu, line 331: This isn't enough because `@stream = nil` is already set in the `initialize`.

lib/grape/dsl/inside_route.rb
```diff
@@ -328,6 +328,8 @@ def sendfile(value = nil)
       # * https://github.com/rack/rack/blob/99293fa13d86cd48021630fcc4bd5acc9de5bdc3/lib/rack/chunked.rb
       # * https://github.com/rack/rack/blob/99293fa13d86cd48021630fcc4bd5acc9de5bdc3/lib/rack/etag.rb
       def stream(value = nil)
+        return if value.nil? && @stream.nil?
```
dblock, line 331: This fixes a bug where headers re reset if you just do `.stream`. Write a spec for it please?

Issue #2090, opened 2020-07-15
 Ruby 2.7 deprecation warning with middleware/error


dim:
Addressing issue #1992 via #2064 has fixed some of the warnings but not all of them, I am still seeing this:

```
/gems/rack-2.2.3/lib/rack/builder.rb:158: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/gems/grape-1.4.0/lib/grape/middleware/error.rb:30: warning: The called method `initialize' is defined here
```

dblock:
Closed via https://github.com/ruby-grape/grape/pull/2091

PR #2091, opened 2020-07-15
Fix Ruby 2.7 keyword deprecations

dim:
Another attempt to address https://github.com/ruby-grape/grape/issues/2090 properly

### DIFF
--- a/lib/grape/middleware/stack.rb
+++ b/lib/grape/middleware/stack.rb
@@ -6,11 +6,12 @@ module Grape
     # It allows to insert and insert after
     class Stack
       class Middleware
-        attr_reader :args, :block, :klass
+        attr_reader :args, :opts, :block, :klass
 
-        def initialize(klass, *args, &block)
+        def initialize(klass, *args, **opts, &block)
           @klass = klass
           @args = args
+          @opts = opts
           @block = block
         end
 
@@ -90,8 +91,13 @@ module Grape
       def build(builder = Rack::Builder.new)
         others.shift(others.size).each(&method(:merge_with))
         middlewares.each do |m|
-          opts = m.args.extract_options!
-          m.block ? builder.use(m.klass, *m.args, **opts, &m.block) : builder.use(m.klass, *m.args, **opts)
+          if RUBY_VERSION >= '2.7'
+            m.block ? builder.use(m.klass, *m.args, **m.opts, &m.block) : builder.use(m.klass, *m.args, **m.opts)
+          else
+            args = m.args
+            args += [m.opts] unless m.opts.empty?
+            m.block ? builder.use(m.klass, *args, &m.block) : builder.use(m.klass, *args)
+          end
         end
         builder
       end
--- a/lib/grape/middleware/stack.rb
+++ b/lib/grape/middleware/stack.rb
@@ -22,8 +22,8 @@ module Grape
         end if use_ruby2_keywords
 
         def use_in(builder)
-          args  = m.args
-          args += [m.opts] unless m.opts.empty?
+          args  = self.args
+          args += [opts] unless opts.empty?
           block ? builder.use(klass, *args, &block) : builder.use(klass, *args)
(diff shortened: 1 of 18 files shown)


dblock:
Not sure what's up with CI, force push?

lib/grape/middleware/stack.rb
```diff
@@ -90,8 +91,13 @@ def merge_with(middleware_specs)
       def build(builder = Rack::Builder.new)
         others.shift(others.size).each(&method(:merge_with))
         middlewares.each do |m|
-          opts = m.args.extract_options!
-          m.block ? builder.use(m.klass, *m.args, **opts, &m.block) : builder.use(m.klass, *m.args, **opts)
+          if RUBY_VERSION >= '2.7'
```
dblock, line 94: This won't work after version 10.

```ruby
2.6.6 :001 > RUBY_VERSION 
 => "2.6.6" 
2.6.6 :002 > RUBY_VERSION > '2.7.7'
 => false 
2.6.6 :003 > RUBY_VERSION > '2.6.1'
 => true 
2.6.6 :004 > RUBY_VERSION > '16.1'
 => true 
```

Lets find a way to avoid this `if`?

dblock: You could use https://github.com/dblock/with-version, but I am not sure we want a dependency for syntax sugar, so use something like https://metaredux.com/posts/2018/10/28/a-better-way-to-compare-versions-in-ruby.html if you must.

If we don't have another way, take the `if` outside of the execution part, and declare separate method implementations for different versions. This way we don't re-evaluate the same thing again and again during execution.

dim: Apologies, fixed this a while ago but forgot to push :facepalm:

lib/grape/middleware/stack.rb
```diff
@@ -6,14 +6,27 @@ module Middleware
     # It allows to insert and insert after
     class Stack
       class Middleware
-        attr_reader :args, :block, :klass
+        use_ruby2_keywords = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7')
```
dblock, line 9: Since we only use this once, let's avoid an unnecessary variable and just do 

```ruby
if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7')
...
else
...
end
```

lib/grape/middleware/stack.rb
```diff
@@ -6,14 +6,27 @@ module Middleware
...
+
+        def initialize(klass, *args, **opts, &block)
           @klass = klass
-          @args = args
+          @args  = args
+          @opts  = opts
           @block = block
         end
 
+        def use_in(builder)
```
dblock, line 20: Should `use_in` be private?

lib/grape/middleware/stack.rb
dim: I didn't want to do the comparison/computation on every call. Happy to add a memoized static method instead.

lib/grape/middleware/stack.rb
dim: makes sense, will update

dim: actually, no, it's called from `Stack#build`

lib/grape/middleware/stack.rb
dblock: Right, so put it around method definition.

```ruby
if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7')
  def use_in
  end
else
  def use_in
  end
end
```

dblock: If you do want to leave it as a method and be memoized, put it into a separate module, e.g. `Grape::Ruby::Version` ala https://github.com/dblock/with-version (we don't want to take a dependency I think for such seldom usage).

dblock:
Much thanks.

PR #2097, opened 2020-08-20
Skip to set default value unless `meets_dependency?`

wanabe:
Hello,

`Grape::Validations::Base#validate!` calls `#validate_param!` only if `@scope.meets_dependency?(val, params)` is truthy.
But `Grape::Validations::DefaultValidator#validate!` ignores the cope dependency now.

It doesn't matter in most cases because of `@scope.should_validate?`, but this is not the case for an array that contains both matching and non-matching conditions.
So I guess `Grape::Validations::DefaultValidator#validate!` should the check scope dependency as same as `Grape::Validations::Base#validate!`.

How about it?


dblock:
Thanks for this.

PR #2098, opened 2020-08-25
Add nested array coercion spec

braktar:
The coerce seems to be performed correctly, nevertheless, once we want to match the obtained value to the expected type it fails.

Seems related to : https://github.com/ruby-grape/grape/pull/2054


braktar:
My current assumption is that the type_check provided by the [infer_type_check](https://github.com/ruby-grape/grape/blob/192a2a2bbe59da0dac4ea435916ea5576d866155/lib/grape/validations/types/custom_type_coercer.rb#L94) method can't evaluate the returned array of array.

braktar:
Some progress here, recursive objects were not recursively verified.

Nevertheless, When parameters are expected to be coerced, the rack_params returned by the ParamBuilder is flatten
['a', 'c', 'b'] instead of [['a','c'],['b']]

[[]] return [[""]] which seems a bit odd.

[['a', { bar: 0 }], ['b']] is currently considered as incorrect, but not for the good reason. It is currently wrong due to the flatten.

Any hint about it ?

dblock:
Try POSTing JSON in the spec?

braktar:
Thanks  !

dblock:
🚀

dblock:
If you feel like doing more work, there's a [number of issues that have "coercion"](https://github.com/ruby-grape/grape/issues?q=is%3Aissue+is%3Aopen+coercion) in the title that could be in the same code path ;)

PR #2099, opened 2020-08-31
Add Truffleruby head to CI

gogainda:


gogainda:
https://github.com/oracle/truffleruby/issues/2084

gogainda:
#https://github.com/oracle/truffleruby/issues/2084

gogainda:
https://github.com/oracle/truffleruby/issues/2084

dblock:
I am open to including this into the set of builds that allow failures if someone (you?) wants to help out with truffleruby-related issues. We haven't had much success in the past with anyone caring about things like rbx.

gogainda:
@eregon I filled out an issue , but I am not sure when it's going to be fixed. Since tests itself are passing prb I can mark it as allow failures for now

eregon:
One potential workaround is to run RuboCop only on MRI, it seems redundant to run it on TruffleRuby anyway.
This could be done by changing the Rakefile: https://github.com/ruby-grape/grape/blob/master/Rakefile

Otherwise the error should be solved by https://github.com/rubocop-hq/rubocop/pull/8602, and then we need to wait the next RuboCop release and make sure it's used by this repo.

dblock:
I would take a change to run RuboCop once for the entire suite.

eregon:
OK, then the easiest way to do that is to remove it as a dependency of the Rakefile default task, and to add an extra `script: rubocop` in TravisCI.
@gogainda Can you try that? Otherwise I can give it a shot.

gogainda:
@eregon yep, will do this

gogainda:
@dblock now I run rubocop once and truffleruby tests are green

gogainda:
@dblock @eregon the issue with rubocop is fixed in 0.90 but I think we should leave upgrading rubocop version outside of this PR

dblock:
1. Let's leave Rakefile alone and run `rake spec` in Travis instead, so that developers don't forget that Rubocop offenses need to be fixed during development.
2. I'd like us to add truffle-ruby to `allow_failures` even though it's not failing, as I don't personally have the bandwidth to support multiple runtimes, similarly to rbx, and others.

gogainda:
@dblock sure, but I am not really familiar with travis yml syntax, if you have an example it would be great. Regarding allow_failures - I can add it

.travis.yml
```diff
@@ -11,6 +11,9 @@ matrix:
     - rvm: 2.7.1
       script:
         - bundle exec danger
+    - rvm: 2.7.1
+      script:
+        - rubocop
```
dblock, line 16: This should be `bundle exec rubocop` to pin to the version in Gemfile.

dblock:
> @dblock sure, but I am not really familiar with travis yml syntax, if you have an example it would be great. Regarding allow_failures - I can add it

`script: bundle exec rake spec`

gogainda:
@dblock addressed your comments

.travis.yml
```diff
@@ -35,12 +34,12 @@ matrix:
     - rvm: 2.7.1
       gemfile: gemfiles/multi_json.gemfile
       script:
-        - bundle exec rake
+        - bundle exec rubocop
```
dblock, line 37: This line should be removed and default to `rake spec`, we're running all the usual tests, not rubocop.

.travis.yml
```diff
@@ -35,12 +34,12 @@ matrix:
...
         - bundle exec rspec spec/integration/multi_json
     - rvm: 2.7.1
       gemfile: gemfiles/multi_xml.gemfile
       script:
-        - bundle exec rake
+        - bundle exec rubocop
```
dblock, line 42: same

gogainda:
@dblock done

dblock:
Merged, thanks for this.

gogainda:
@eregon

Issue #2101, opened 2020-09-11
Breaking change for nested params in `declared` in Grape 1.3.3


tlconnor:
A PR was merged back in April and included in Grape `1.3.3` that has changed the way `declared` works.
https://github.com/ruby-grape/grape/pull/2043

Before this PR was merged, the `declared` helper would guarantee that the entire params structure would be recreated, including deeply nested parameters. Any leaf params not provided would have `nil` values.

With this merged PR, any missing `Hash` params would be evaluated as `{}`, even if they has deeper nested params defined.

Here is an example from `endpoint_spec.rb`. The `params` are defined as:

```ruby
subject.params do
  requires :first
  optional :second
  optional :third, default: 'third-default'
  optional :nested, type: Hash do
    optional :fourth
    optional :fifth
    optional :nested_two, type: Hash do
      optional :sixth
      optional :nested_three, type: Hash do
        optional :seventh
      end
    end
    optional :nested_arr, type: Array do
      optional :eighth
    end
  end
  optional :arr, type: Array do
    optional :nineth
  end
end
```

Given a test that calls `get '/declared?first=present'`, the result of `declared(params)` would be:

In Grape `1.3.2`:

```json
{
  "first": "present",
  "second": null,
  "third": "third-default",
  "nested": {
    "fourth": null,
    "fifth": null,
    "nested_two": {
      "sixth": null,
      "nested_three": {
        "seventh": null
      }
    }
  },
  "nested_arr": []
}
```

In Grape `1.3.3`:

```json
{
  "first": "present",
  "second": null,
  "third": "third-default",
  "nested": {},
  "arr": []
}
```

It is not clear which of these is the correct behavior. The merged PR https://github.com/ruby-grape/grape/pull/2043 would suggest that the latter is correct, however it is actually very useful to have the previous behavior from 1.3.2.

The README isn't clear about what the behavior should be for deeply nested params.

dblock:
Ouch, this keeps biting us, cc: @kadotami and @dnesteryuk.

Naively reading our own README it feels like if you say `include_missing: true`, we should get the entire structure. Do you want to attempt a change/fix for this and see whether we are breaking any specs? Reading #2043 it feels that we simply didn't consider this, but I could be wrong.

tlconnor:
[...]

PR #2103, opened 2020-09-17
Ensure complete declared params structure is present

tlconnor:
Back in April 2020 a fix was merged that ensured that missing Hash / Array params would be initialized as `[]` / `{}` during `declared(params, include_missing: true)`.

https://github.com/ruby-grape/grape/pull/2043

This fix introduced a bug that would result in deeply nested param structures being replaced with an empty Hash.
https://github.com/ruby-grape/grape/issues/2101

This PR fixes the issue by ensuring that missing hashes are only initialized as `{}` if they don't have any declared nested params.

Fixes https://github.com/ruby-grape/grape/issues/2101


dblock (changes requested):
This is great. Thank you.

Let's beef up the specs a little bit: make one without deep nesting, make a separate spec for deep nested hashes, deep nested arrays, multiple levels of deep nesting. 

And let's go an extra mile in README to explain these behaviors?

UPGRADING needs a paragraph, too.

tlconnor:
@dblock I have improved the specs and updated the README / CHANGELOG / and UPGRADING guide.

It is not clear to me what we should do with missing `Set` params. Do you think a missing `Set` / `Set[Anything]` should be resolved as `Set.new`, or do you think they should remain as `nil`?

UPGRADING.md
```diff
@@ -1,6 +1,19 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.4.1
+
+Previous to 1.3.3, the `declared` helper would always return the complete params structure if `include_missing=true` was set. In 1.3.3 the behavior was changed, so a missing Hash with or without nested parameters would always resolve as `{}`.
```
dblock, line 6: Previous -> prior

UPGRADING.md
```diff
@@ -1,6 +1,19 @@
...
+
+In 1.4.1 this behavior is reverted, so the whole params structure will always be available via `declared`, regardless of whether any params are passed.
+
+The following rules now apply to the `declared` helper when params are missing and `include_missing=true`:
+
+* Hash params with childern will resolve as a hash with keys for each declared child.
+* Hash params with no children will resolve as `{}`.
+* Array params will resolve as `[]`.
+* All other params will resolve as `null`.
```
dblock, line 15: Add a simple before/after example.

Add a link to the PR, `For more information see [#2103](https://github.com/ruby-grape/grape/pull/2103).`. We've been forgetting those, would be nice to re-add where missing below.

dblock (approved):
This LGTM with minor comments for UPGRADING.

For `Set`, I think it should be `Set.new` because `[]` and `{}` are just shortcuts for `Array.new` and `Hash.new`.

```irb
2.6.6 :001 > Hash.new
 => {} 
2.6.6 :002 > {}.class
 => Hash 
2.6.6 :003 > Hash.new == {}
 => true 
2.6.6 :004 > Array.new
 => [] 
2.6.6 :005 > [].class
 => Array 
2.6.6 :006 > Array.new == []
 => true 
```

dblock:
If you don't mind feeding my OCD, squash your last two commits (amend the second one that makes the actual code change), and keep the move of the spec separate, so we can continue seeing the diff. NBD otherwise.

tlconnor:
@dblock all comments have now been addressed. I have also added support for `Set`, and have included test cases for `Array[Something]` and `Set[Something]`.

UPGRADING.md
```diff
@@ -1,6 +1,45 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.4.1
+
+Prior to 1.3.3, the `declared` helper would always return the complete params structure if `include_missing=true` was set. In 1.3.3 the behavior was changed, so a missing Hash with or without nested parameters would always resolve to `{}`.
```
dblock, line 6: was changed -> a regression was introduced that ...

UPGRADING.md
```diff
@@ -1,6 +1,45 @@
...
+
+In 1.4.1 this behavior is reverted, so the whole params structure will always be available via `declared`, regardless of whether any params are passed.
```
dblock, line 8: Let's make this 1.5, and please bump the version in version.rb.

dblock (approved):
I'm good with this with a 🍏 build and a 1.5 version bump.

dblock:
Change other references to 1.4.1 (in CHANGELOG, etc.) to 1.5.

lib/grape/dsl/inside_route.rb
```diff
@@ -70,57 +70,44 @@ def declared_hash(passed_params, options, declared_params, params_nested_path)
...
+          key += '[' + params_nested_path[1..-1].join('][') + ']' if params_nested_path.size > 1
+
+          route_options_params = options[:route_options][:params] || {}
+          type = route_options_params.dig(key, :type)
+          has_children = route_options_params.keys.any? { |k| k != key && k.start_with?(key) }
+
+          if type == 'Hash' && !has_children
             {}
-          elsif should_be_empty_array?(passed_children_params, params_nested_path)
+          elsif type == 'Array' || type&.start_with?('[')
```
dblock, line 98: This seems iffy but I'll let it slide. I wish we were comparing a non-String type...

tlconnor: I agree, it is a bit messy. It was also pre-existing, so not entirely my fault 😄

tlconnor:
> Change other references to 1.4.1 (in CHANGELOG, etc.) to 1.5.

Not sure why I missed those. Now fixed.

dblock:
Merged, thanks for your excellent work here, @tlconnor! You might want to email the mailing list asking people to test this out before we release v. next.

tlconnor:
@dblock please hold off from releasing the new version, through testing against my app I believe I have found a newly introduced bug. I will work on a fix ASAP.

PR #2105, opened 2020-09-28
Lock rubocop-ast to < 0.7

tlconnor:
The latest versions of rubocop-ast (>= 0.7) are incompatible with older versions of rubocop. Since grape currently use rubocop 0.84 we need to lock rubocop-ast to a compatible version

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ group :development, :test do
   gem 'hashie'
   gem 'rake'
   gem 'rubocop', '0.84.0'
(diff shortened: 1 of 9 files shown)


tlconnor:
@dblock master (and all other branches) are now failing CI due to a new incompatible version of rubocop-ast being published.

https://github.com/rubocop-hq/rubocop-ast/blob/master/CHANGELOG.md#changes

dblock:
Thanks for fixing this.

PR #2107, opened 2020-09-30
Get rid of read_chunks.

dblock:
This is just silly.


dblock:
Well, this can't quite be done because rack returns Rack::Response and in later versions an Array. So we have to call `each`.

PR #2109, opened 2020-10-01
Fix bug with handling array params.

tlconnor:
Fixes an issue introduced in 7e432153f08d4985b0df625d11aef28a38beb4fe that results in Array leaf params being ignored.

For example, given the following API and Request:

```
class Api < Grape::API
  params do
    optional :array, type: Array
  end
  post 'example' do
    declared(params, include_missing: true)
  end
end

POST /example
{ "array": ["value"] }
```

Expected Response Body:

```
{ "array": ["value"] }
```

Observed Response Body:

```
{ "array": [] }
```


dblock:
Thanks. Consider beefing up more specs for other types for similar situations.

PR #2111, opened 2020-10-05
Overwriting a parameter default with nil does not work

Jack12816:
Given a shared parameter block `:basic`:
```ruby
        params :basic do
          with(type: Float, default: 5.0) do
            optional :distance
          end
        end
```

When we use the parameter block and want to overwrite the default like this:
```ruby
      params do
        # optional :distance, default: nil # does not work here
        use :basic
        optional :distance, default: nil # does not work here, too
      end
      get do
        # ...
      end
```

Then the expected default would be `nil` but it is `5.0`. When we change the default overwrite to
something else than `nil`, like `1.0` it works as expected.

NOTE: This is currently just a failing spec. I would like to fix it, but currently, I just found out that the `DefaultValidator` is setup twice where I would expect a single one with the latest defined value. Can someone give me a hint on this?

### DIFF
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -75,8 +75,7 @@ module Grape
 
               params_nested_path_dup = params_nested_path.dup
               params_nested_path_dup << declared_param.to_s
-
-              memo[memo_key] = handle_passed_param(params_nested_path_dup) do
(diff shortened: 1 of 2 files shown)


dblock:
I think we just don't allow multiple declarations of parameters, the first or the last one wins? Or some mix of two? Does this work without `use:` but inlining parameters? Try to extend the spec in that direction to start?

PR #2112, opened 2020-10-05
field using type turned into array when declared(params)

braktar:
It seems that the recent works on declared change the interaction between 'declared' and the 'types' field

### DIFF
--- a/spec/grape/validations/params_scope_spec.rb
+++ b/spec/grape/validations/params_scope_spec.rb
@@ -91,6 +91,43 @@ describe Grape::Validations::ParamsScope do
     end
   end
 
+  context 'when using multiple types' do
+    it 'coerces the parameter via the type\'s parse method' do
+      subject.params do
+        requires :bar, type: Array do
+          optional :foo, types: [String, Float, Integer], coerce_with: ->(c) {
+              if c
+                if /([0-9]+):([0-9]+):([0-9]+)/ =~ c.to_s
+                  'HH:MM:SS format'
+                elsif /\A[0-9]+\.{0,1}[0-9]*\z/ =~ c.to_s
+                  'Float or Integer format'
+                else
+                  'Invalid Time value'
+                end
+              end
+            }
+          end
+      end
+      subject.post('/types') { declared(params)[:bar].first[:foo] }
+
+      post '/types', bar: [{ foo: '00:00:01' }]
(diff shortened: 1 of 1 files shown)


dblock:
what was the behavior in 1.3.x, 1.4.x?

either way looks like a bug, another one... /cc: @tlconnor

dblock:
Related https://github.com/ruby-grape/grape/pull/2103

braktar:
Before the integration of #2103 the spec passed successfully

braktar:
Here is a fix which avoid the misinterpretation of the type

braktar:
That's strange that the type from [empty_typed_arr](https://github.com/ruby-grape/grape/blob/02d7113d09eb9fcb4264c841d1fdd305e3e8adb5/spec/grape/endpoint/declared_spec.rb#L32) turned from Array[String] to [String] when arrived into `handle_passed_param`

dblock:
Let's extend tests with `Hash` and `Set` along with Array?

lib/grape/dsl/inside_route.rb
```diff
@@ -91,10 +91,9 @@ def handle_passed_param(params_nested_path, has_passed_children = false, &_block
           route_options_params = options[:route_options][:params] || {}
           type = route_options_params.dig(key, :type)
           has_children = route_options_params.keys.any? { |k| k != key && k.start_with?(key) }
-
           if type == 'Hash' && !has_children
             {}
-          elsif type == 'Array' || type&.start_with?('[')
+          elsif type == 'Array' || type && type.start_with?('[') && !type.include?(',')
```
tlconnor, line 96: You are right @dblock that working with Strings here is not ideal. I don't know enough to know if there is a better alternative.

braktar:
> Let's extend tests with `Hash` and `Set` along with Array?

I don't really see what should exhibit these tests, as the only ambiguity is between `types : [x, y]` and `type: Array[x, y]`

dblock:
> > Let's extend tests with `Hash` and `Set` along with Array?
> 
> I don't really see what should exhibit these tests, as the only ambiguity is between `types : [x, y]` and `type: Array[x, y]` which both give type: [x, y] while processed.
> 
> It seems to require a deeper investigation

I mean what happens if `requires :bar, type: Hash do ...`?

dblock:
@stanhu isn't this the same?

PR #2117, opened 2020-10-09
Add test for declared params with multiple types

stanhu:
Adds tests from https://github.com/ruby-grape/grape/pull/2112

### DIFF
--- a/spec/grape/validations/params_scope_spec.rb
+++ b/spec/grape/validations/params_scope_spec.rb
@@ -91,6 +91,43 @@ describe Grape::Validations::ParamsScope do
     end
   end
 
+  context 'when using multiple types' do
+    it 'coerces the parameter via the type\'s parse method' do
+      subject.params do
+        requires :bar, type: Array do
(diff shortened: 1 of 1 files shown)


stanhu:
> my question was also whether it matters if we put `Hash` in `requires :bar, type: Array do`, so whether we need to have this test try both Hash and Array?

@dblock If I use `requires :bar, type: Set`, I get:

```
     Failure/Error: raise Grape::Exceptions::UnsupportedGroupTypeError.new unless Grape::Validations::Types.group?(type)
     
     Grape::Exceptions::UnsupportedGroupTypeError:
       group type must be Array, Hash, JSON or Array[JSON]
```

It seems that `Hash` and `Array` are not valid group types: https://github.com/ruby-grape/grape/blob/f3073e6123a55e4ec9faa268b741d28502d2dcbc/lib/grape/validations/types.rb#L114

Do we need a test for this?

stanhu:
Oh, you were asking about `Hash`. Let me look at that.

dblock:
@stanhu want to finish this?

Issue #2122, opened 2020-10-16
2.7 deprecation in middleware/stack.rb


Legogris:
This seems to be broken since 2.7.0:

```
grape-1.5.0/lib/grape/middleware/stack.rb:87: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
grape-1.5.0/lib/grape/middleware/stack.rb:11: warning: The called method `initialize' is defined here
```

Constructor:
https://github.com/ruby-grape/grape/blob/a28d37a9dc480266431e855cf670f038c54ef1d2/lib/grape/middleware/stack.rb#L11

Problematic calls (AFAICT):
https://github.com/ruby-grape/grape/blob/a28d37a9dc480266431e855cf670f038c54ef1d2/lib/grape/middleware/stack.rb#L75
https://github.com/ruby-grape/grape/blob/a28d37a9dc480266431e855cf670f038c54ef1d2/lib/grape/middleware/stack.rb#L87

dblock:
Closed via https://github.com/ruby-grape/grape/pull/2123

PR #2123, opened 2020-10-16
Fix 2.7 deprecation warning in middleware/stack

Legogris:
Attempt at fixing #2122 

See https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/ (`Handling argument delegation`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * Your contribution here.
 * [#2115](https://github.com/ruby-grape/grape/pull/2115): Fix declared_params regression with multiple allowed types - [@stanhu](https://github.com/stanhu).
+* [#2123](https://github.com/ruby-grape/grape/pull/2123): Fix 2.7 deprecation warning in middleware/stack - [@Legogris](https://github.com/Legogris).
 
 ### 1.5.0 (2020/10/05)
 
--- a/lib/grape/middleware/stack.rb
+++ b/lib/grape/middleware/stack.rb
@@ -70,9 +70,9 @@ module Grape
         middlewares[i]
       end
 
-      def insert(index, *args, &block)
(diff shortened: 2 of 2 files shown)


dblock:
Thanks, I think this was the last one.

Legogris:
> Thanks, I think this was the last one.

Thanks for merging! Almost ;)
https://github.com/ruby-grape/grape/pull/2121

eregon:
This is unfortunately not correct, `*args, **opts`-delegation is not correct in some cases on 2.7.
The only correct way to do delegation on Ruby 2.7 is using `ruby2_keywords` + `*args`-delegation.
This is also what the Rails ActionDispatch::MiddlewareStack does (https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/stack.rb, which is very similar).

See https://eregon.me/blog/2019/11/10/the-delegation-challenge-of-ruby27.html for why it's broken on 2.7,
and as you mention above https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/
> A compatible delegation that works on Ruby 2.6, 2.7 and Ruby 3
> In short: use Module#ruby2_keywords again.

FWIW, I'm coming from https://github.com/oracle/truffleruby/issues/2159

dblock:
Interesting. Can we get rid of all of this complexity?

eregon:
Yes, if you use `ruby2_keywords` you only need a few `ruby2_keywords` calls on methods accepting `*args`.

dblock:
@eregon care to PR a fix?

eregon:
https://github.com/ruby-grape/grape/pull/2132

Legogris:
@eregon Derp, great that you caught this!

Issue #2100, opened 2020-09-09
Performance degradation (~6x) when using given to model a dependent relation


senhalil:
When change one of the parameters of out API as dependent to another one, the post duration increased from ~30 seconds to ~600 seconds. Is this normal?

Our model is deeply nested but both the dependee and depender are on the first level. 

```ruby
params do
  optional :matrix
  given matrix: ->(val) { val.nil? || val.empty? } do
    requires :vehicle, type: Hash do
      requires :router_mode, type: String
    end
  end
  given matrix: ->(val) { val.any? } do
    requires :vehicle, type: Hash do
      optional :matrix_id, type: String
      optional :router_mode, type: String
      exactly_one_of :matrix_id, :router_mode
    end
  end
end
```

dblock:
I bet it's fixed by https://github.com/ruby-grape/grape/pull/2096. Try HEAD?

senhalil:
Unfortunately, the performance degradation still exists. Actually, I work together with @braktar and we tested on his computer with the fix,

dblock:
Put this into a benchmark and lets fix it.

dnesteryuk:
@senhalil are you going to work on the benchmark?

senhalil:
@dnesteryuk  this issue fell in the cracks apparently, thanks for the ping. 

I don't think I will be able to work on this issue since we gave up using `given`. The performance issue is still there and can be observed by making one of the parameters dependent in the existing large_model benchmark https://github.com/ruby-grape/grape/blob/master/benchmark/large_model.rb

dnesteryuk:
I wrote this simple benchmark to compare params without dependencies and with them:

```ruby
require 'grape'
require 'benchmark/ips'

class GivenAPI < Grape::API
  prefix :api
  version 'v1', using: :path

  params do
    optional :matrix
    given matrix: ->(val) { val.nil? || val.empty? } do
      requires :vehicle, type: Hash do
        requires :router_mode, type: String
      end
    end
    given matrix: ->(val) { val.any? } do
      requires :vehicle, type: Hash do
        optional :matrix_id, type: String
        optional :router_mode, type: String
        exactly_one_of :matrix_id, :router_mode
      end
    end
  end
  post '/' do
    'hello'
  end
end

class SimpleAPI < Grape::API
  prefix :api
  version 'v1', using: :path

  params do
    optional :matrix

    requires :vehicle, type: Hash do
      optional :matrix_id, type: String
[...]

PR #2127, opened 2020-10-27
fix a performance issue with dependent params

dnesteryuk:
Fixes https://github.com/ruby-grape/grape/issues/2100

The reason was in `ActiveSupport::HashWithIndifferentAccess`, it is super expensive.

When users use a `Grape::Extensions::ActiveSupport::HashWithIndifferentAccess::ParamBuilder` or `Grape::Extensions::Hashie::Mash::ParamBuilder` parameter builder there is no change. However, users who use `Grape::Extensions::Hash::ParamBuilder` must make sure a parameter to be dependent on must be a symbol.

```ruby
given :matrix do
  # block here
end
```

Benchmark after this fix:

```
               Given     1.000  i/100ms
              Simple     1.000  i/100ms
               Given      0.804  (± 0.0%) i/s -     49.000  in  61.186831s
              Simple      0.855  (± 0.0%) i/s -     52.000  in  60.926097s

Comparison:
              Simple:        0.9 i/s
               Given:        0.8 i/s - 1.06x  slower
```

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,28 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.5.1
+
+#### Dependent params
+
+If you use [dependent params](https://github.com/ruby-grape/grape#dependent-parameters) with
+`Grape::Extensions::Hash::ParamBuilder`, make sure a parameter to be dependent on is set as a Symbol.
+If a String is given, a parameter that other parameters depend on won't be found even if it is present.
+
+_Correct_:
+```ruby
+given :matrix do
+  # dependent params
+end
+```
+
+_Wrong_:
+```ruby
+given 'matrix' do
+  # dependent params
+end
+```
+
 ### Upgrading to >= 1.5.0
 
 Prior to 1.3.3, the `declared` helper would always return the complete params structure if `include_missing=true` was set. In 1.3.3 a regression was introduced such that a missing Hash with or without nested parameters would always resolve to `{}`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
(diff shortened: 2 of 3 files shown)


UPGRADING.md
```diff
@@ -1,6 +1,19 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.5.1
+
+#### Dependent params
+
+If you use [dependent params](https://github.com/ruby-grape/grape#dependent-parameters) with
+`Grape::Extensions::Hash::ParamBuilder`, make sure a parameter to be dependent on is set as a Symbol
```
dblock, line 9: Explain what won't work? In this case a `String`? Also add a period at the end of the sentence or maybe a before/after section?

dblock (approved):
This makes sense. Let's beef up UPGRADING to explain in which case it might not work?

UPGRADING.md
dnesteryuk: > Also add a period at the end of the sentence or maybe a before/after section?

I am not sure what you mean here :blush: Could you explain?

dblock: a Symbol. (it's a full sentence, just missing a period)

dblock: Regarding before/after, you're saying make sure it's a Symbol. What happens if it's not? Explain that.

dnesteryuk:
@dblock please, have a look again. thanks.

UPGRADING.md
```diff
@@ -1,6 +1,28 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.5.1
+
+#### Dependent params
+
+If you use [dependent params](https://github.com/ruby-grape/grape#dependent-parameters) with
+`Grape::Extensions::Hash::ParamBuilder`, make sure a parameter to be dependent on is set as a Symbol.
+If a String is given, a parameter that other parameters depend on won't be found even if it is present.
```
dblock, line 10: So is this a bug? Should we also fix this (in another PR)?

PR #2128, opened 2020-11-04
Fix validation error for nested array when  `requires` => `optional` => `requires`

dwhenry:
This bug is due to the params parsing code return `{}` when the value is not found.
This is fine in most cases, however in the instance that the value was optional
and has nested required values.

I was not able to find a way to confine the changes to just the parameter parsing
as this resulted in the indexing being incorrect if any of the other array objects
had an error.

I have used a class to indicate that an Optional Value was missing as this avoid issues
with the value actually being in the response.

### DIFF
--- a/lib/grape/validations/single_attribute_iterator.rb
+++ b/lib/grape/validations/single_attribute_iterator.rb
@@ -17,7 +17,6 @@ module Grape
       # are the parameter parsing stage as they are required to ensure
       # the correct indexing is maintained
       def skip?(val)
-        # return false
         val == Grape::DSL::Parameters::EmptyOptionalValue
       end
 
--- a/lib/grape/validations/single_attribute_iterator.rb
+++ b/lib/grape/validations/single_attribute_iterator.rb
@@ -7,8 +7,18 @@ module Grape
 
       def yield_attributes(val, attrs)
         attrs.each do |attr_name|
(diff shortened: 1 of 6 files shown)


dblock:
Looks like there are some legit spec failures here ... 

cc: @stanhu @dm1try @dnesteryuk who dealt with params recently to add some comments

lib/grape/validations/single_attribute_iterator.rb
```diff
@@ -7,10 +7,20 @@ class SingleAttributeIterator < AttributesIterator
...
         end
       end
 
+
+      # This is a special case so that we can ignore tree's where option
+      # values are missing lower down. Unfortunately we can remove this
+      # are the parameter parsing stage as they are required to ensure
+      # the correct indexing is maintained
+      def skip?(val)
+        # return false
```
dblock, line 20: remove

dblock:
Try writing more specs with other combinations of required parameters inside optional ones? I suspect we have similar problems with hash parameters and combinations of arrays and hashes.

dwhenry:
@dblock added tests for all the edge cases that I believe this would affect and they all pass with this change. Let me know if you can think of anything else it would be worth testing.

dblock:
i'm merging this

Issue #1966, opened 2020-01-10
Router#cascade? undefined method `[]' for nil:NilClass


delner:
Grape is raising an `NoMethodError` when implementing a really basic API endpoint.

## What I expect

Basic requests to not raise errors.

## What actually happens

It raises `NoMethodError`. Sample output from Minitest (described below):

```
  1) Error:
GrapeTest#test_api_success:
NoMethodError: undefined method `[]' for nil:NilClass
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/router.rb:164:in `cascade?'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/router.rb:95:in `transaction'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/router.rb:72:in `identity'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/router.rb:57:in `block in call'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/router.rb:137:in `with_optimization'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/router.rb:56:in `call'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/api/instance.rb:165:in `call'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/api/instance.rb:69:in `call!'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/api/instance.rb:64:in `call'
    /usr/local/bundle/gems/grape-1.2.5/lib/grape/api.rb:66:in `call'
    /usr/local/bundle/gems/rack-test-1.1.0/lib/rack/mock_session.rb:29:in `request'
    /usr/local/bundle/gems/rack-test-1.1.0/lib/rack/test.rb:266:in `process_request'
    /usr/local/bundle/gems/rack-test-1.1.0/lib/rack/test.rb:129:in `custom_request'
    /usr/local/bundle/gems/rack-test-1.1.0/lib/rack/test.rb:58:in `get'
    /usr/local/lib/ruby/2.3.0/forwardable.rb:202:in `get'
    /app/test/contrib/grape/request_test.rb:25:in `test_api_success'
```

## Steps to reproduce

Dependencies from Gemfile.lock

```
grape (1.2.5)
      activesupport
      builder
      mustermann-grape (~> 1.0.0)
      rack (>= 1.3.0)
      rack-accept
      virtus (>= 1.0.0)
rack (2.1.0)
[...]

PR #2130, opened 2020-11-17
Mention rack 2.1.0 support in UPGRADING.md

maxkerp:
I'm currently upgrading a fairly big rails app I did not write myself which uses grape. Since I unfortunately never used grape before (I want to though :wink:), I was in unknown territory and it took me quite I while to figure out that grape added support for rack >= 2.1.0 first on 1.3.0.

First thing I did when running into [https://github.com/ruby-grape/grape/issues/1966](this error) was checking the Upgrade.md file for any signs of rack incompatibility, but I did not find any. I wish I did, since it took me a few hours to find the mentioned issue. (I really have to up my google-fu, I know :see_no_evil:)

But for any poor soul in the future, who is just as bad as googling as I am, it might be helpful to mention this in the Upgrade guide :)

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -189,6 +189,8 @@ end
 
 ### Upgrading to >= 1.3.0
 
+#### Added Rack 2.1.0 support
+
 #### Ruby
 
 After adding dry-types, Ruby 2.4 or newer is required.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -189,7 +189,7 @@ end
 
 ### Upgrading to >= 1.3.0
 
-#### Added Rack 2.1.0 support
+You will need to upgrade to this version if you depend on `rack >= 2.1.0`
 
 #### Ruby
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -189,7 +189,7 @@ end
 
 ### Upgrading to >= 1.3.0
 
-You will need to upgrade to this version if you depend on `rack >= 2.1.0`
+You will need to upgrade to this version if you depend on `rack >= 2.1.0`.
 
 #### Ruby
 


dblock:
Thanks! Explain what the developer that's upgrading has to do though? Does one have to upgrade and therefore add a specific version of Rack to Gemfile? Link to the issue?

maxkerp:
Well that got merged fast :grin: Always feels good to contribute to oss, even when it's a tiny bit :relaxed: 

I had the option to either upgrade to `grape` >= 1.3.0 or ensure that we are using a compatible version of `rack`.

Since I want to wait with the upgrade of `grape` til we are done with the upgrade to rails `5.0` I opted to pin the version of `rack` to `~> 2.0.0` so it will be below version `2.1.0`.

Since `rack` isn't part of a Gemfile in rails project per se I needed to add it altogether like this:

```ruby
gem 'rack', '~> 2.0.0'
``` 
I will remove this line from my Gemfile once I'm done upgrading to `grape` >= 1.3.0 :)

Does this answer your questions?

dblock:
First, thanks. But I am looking for a sentence of instructions below the line that you're adding to the user UPGRADING. So I am confused to what we're trying to tell the user. UPGRADING instructions should tell users what to do, CHANGELOG says what changed and feels like we're mixing the two.

maxkerp:
Know I understand and I think this is kind of tricky. The information I thought I might find here is the incompatibility of `grape` versions with other "big" gems/dependencies. When I looked at the file I wasn't looking for instructions of _how_ I upgrade to a specific version, but rather of _why_ I would want to.

The line I added tells me that versions <= 1.3.0 don't work with `rack` >= 2.1.0 and from there I can decide what to do next. This is probably very project specific. Pinning the `rack` version worked for me, but might not for someone else. 

So if you are looking for instructions it might be something like

In the section "Upgrading to >= 1.3.0":

"You will need to upgrade to this version if you depend on `rack` >= 2.1.0"

Or in the section "Upgrading to >= 1.2.5":

"If something prevents you from upgrading to version >= 1.3.0 it might be helpful to pin `rack` to `~> 2.0.0` in your Gemfile, since version 1.2.5 does not work with `rack` >= 2.1.0"

What do you think?

dblock:
Your suggestions sound good! Go for it.

UPGRADING.md
```diff
@@ -189,6 +189,8 @@ end
 
 ### Upgrading to >= 1.3.0
 
+You will need to upgrade to this version if you depend on `rack >= 2.1.0`
```
dblock, line 192: It's missing a period.

dblock (changes requested):
Thanks. Can I nitpick?

maxkerp:
Sure, you can. Added it :v:

PR #2131, opened 2020-11-17
Fix Ruby 2.7 keyword deprecation warning in validators/coerce

K0H205:
Fix these deprecation warnings
```
grape/lib/grape/validations/validators/coerce.rb:21: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
grape/lib/grape/validations/validators/base.rb:16: warning: The called method `initialize' is defined here
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#2131](https://github.com/ruby-grape/grape/pull/2131): Fix Ruby 2.7 keyword deprecation warning in validators/coerce - [@K0H205](https://github.com/K0H205).
 
 ### 1.5.1 (2020/11/15)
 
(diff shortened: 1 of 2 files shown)


dblock:
Is there any way we can fail CI when we have these warnings?

K0H205:
> Is there any way we can fail CI when we have these warnings?

Umm, I can't suggest enough ways to make CI fail with ruby warnings.

eregon:
> Is there any way we can fail CI when we have these warnings? Maybe a Danger plugin?

The [warning gem](https://github.com/jeremyevans/ruby-warning) could be use to turn warnings into exceptions, that might be one way.

dblock:
@eregon You'll be my hero if you can make it work.

dblock:
Btw, I had tried in `spec_helper` to override `warn`, which worked. The problem was that I couldn't have both specs that expect warnings and always fail on unexpected warnings.

eregon:
That gem seems to allow filtering warnings like `:keyword_separation` so that might be helpful.

PR #2143, opened 2020-12-26
Enable GitHub Actions with updated Rubocop and Danger

anakinj:
Did this migration for the ruby-jwt gem a few weeks back and it has been great. The speed compared to Travis CI is super and having a CI run on forks really give you the chance to polish everything before opening a PR.

Took the liberty of trying to improve the developer experience for this gem also. Unfortunately this was not that straightforward and noticed some issues in the process.

What was done:

- GitHub actions for all the gemfile+ruby combinations that were present in the travis.yml
- Updated to use the latest from the 2.7 branch instead of the pinned 2.7.1
- Dropped rbx-3 test (Did not see any value in investigating how to get that working for the Actions)
- **Re-enabled** RuboCop on CI. It has not been running since #2099 was merged!!!
- For some reason RuboCop was having issues running locally and in the actions so just updated the gem to the latest version to fix those issues.
- A lot of exceptions were added to .rubocop_todo.yml due to the update and because it had been disabled on CI for a few months.
- The Danger related components were updated to a version that supports the Actions
- ~~Needed to disable one test related to XML generation. I suspect the TravisCI setup had some extra dependencies that made it work.~~

Still think this needs a little tuning related to the Coveralls integration, some secrets needs to be added to the repository secrets and maybe some ENV variables for the action that is in charge of pushing the results to Coveralls.


anakinj:
Tried this out on my own fork if you're interested in how it looks:

https://github.com/anakinj/grape/pull/1

LeFnord:
IMHO, I'm not a big fan moving all to GH (Microsoft) and see no reason to do it.

Travis is supporting OSS since years, then Microsoft see it as evil, maybe you remember,  
so why should it be moved now? And possible destroying other companies and make M$ even richer?

anakinj:
I think since TravisCI got bought the support for OSS projects has not been that good. Just looking at the [queues](https://www.traviscistatus.com/#month) for OSS CI jobs does not look that good.

Waiting 5 hours to get feedback on your changes is not really motivating.

anakinj:
Also travis-ci.org is getting deprecated to be replaced with travis-ci.com. Think this is happening when the year changes . The new travis-ci.com is supporting OSS but only some thousand minutes of CI time a month for free, [this](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) is what is happening.

I don't think Idera is any better than M$ when it comes to money.

spec/grape/api_spec.rb
```diff
@@ -3817,7 +3817,9 @@ def serializable_hash
 </hash>
 XML
       end
-      it 'array' do
+      # Temporary skipped
+      # XML generation for an array behaves differently when Rails (or something else) is loaded
+      xit 'array' do
```
dblock, line 3822: We should fix this. How differently?

dblock (changes requested):
I'd like to merge this because builds are taking forever on Travis-CI these days, GH actions are blazing fast, have better UX, etc. We can always go back if needed in some future. Right now a bunch of other projects I am involved with are moving for the reasons @anakinj has outlined above.

Can we also please delete .travis.yml in the same CR, and make sure the badges are updated?

Good catch on Rubocop, thank you.

spec/grape/api_spec.rb
anakinj: I will try to figure out what causes this change in behaviour. Probably some 3rd party dependency when RuboCop and friends got updated.

anakinj:
I will continue on this a little later today.

spec/grape/api_spec.rb
anakinj: I think i've figured it out. The problem is that grape is not directly requiring `active_support/core_ext/array/conversions` that adds the `Array#to_xml` method. It has worked out of luck because activesupport internally loads the file when `active_support/time` is loaded.

Now in the latest 6.1 version of activesupport changes this internal loading, more precisely https://github.com/rails/rails/commit/12959dee0f19a8e050dd1236b031c2c690729905#diff-f805b1a36e26831e5eee78fb55549de44fb842f78443b178d5d5438105152017

I will revert this change and fix the issue in a separate PR

anakinj:
Did what was suggested:

- Reverted change on the spec. The test is failing but should get fixed by #2144
- Removed `.travis.yml`
- Did my best guess on what the url is for the CI badge for the actions

Related to Coveralls something needs to be done to get the token in use for the workflow. Added to the repository and somehow passed as a env variable to the action.

anakinj:
If seen valuable we could add a daily build for the master branch to ensure the compatibility with 3rd party dependencies. 

When working on a gem you tend to have a local `Gemfile.lock` from whenever you did the initial `bundle install` and then when CI fails you immediately start blaming energy the changes you did.

dblock:
> If seen valuable we could add a daily build for the master branch to ensure the compatibility with 3rd party dependencies.
> 
> When working on a gem you tend to have a local `Gemfile.lock` from whenever you did the initial `bundle install` and then when CI fails you immediately start blaming the changes you did even if the issue could be in the differences in dependencies.

Let's do that separately, feature creep ;)

PR #2144, opened 2020-12-28
Explicitly require active_support/core_ext/array/conversions

anakinj:
Grape is not explicitly loading the activesupport parts that are required for Array#to_xml. Because of this the Array serialization to XML breaks when activesupport 6.1 is taken into use.

The commit that deletes what previously was requiring the parts that Grape relies on is this https://github.com/rails/rails/commit/12959dee0f19a8e050dd1236b031c2c690729905#diff-f805b1a36e26831e5eee78fb55549de44fb842f78443b178d5d5438105152017


dblock:
👍

Issue #2142, opened 2020-12-22
Ruby 3 support?


pbstriker38:
When setting up a simple server with ruby `3.0.0-rc1` I get the following error:
```shell
ArgumentError: wrong number of arguments (given 2, expected 1)
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/bundler/gems/grape-0f250dc0f574/lib/grape/middleware/base.rb:18:in `initialize'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/gems/rack-2.2.3/lib/rack/builder.rb:158:in `new'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/gems/rack-2.2.3/lib/rack/builder.rb:158:in `block in use'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/gems/rack-2.2.3/lib/rack/builder.rb:235:in `block in to_app'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/gems/rack-2.2.3/lib/rack/builder.rb:235:in `each'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/gems/rack-2.2.3/lib/rack/builder.rb:235:in `inject'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/gems/rack-2.2.3/lib/rack/builder.rb:235:in `to_app'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/bundler/gems/grape-0f250dc0f574/lib/grape/endpoint.rb:323:in `build_stack'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/bundler/gems/grape-0f250dc0f574/lib/grape/endpoint.rb:355:in `block in lazy_initialize!'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/bundler/gems/grape-0f250dc0f574/lib/grape/endpoint.rb:351:in `synchronize'
  /.rbenv/versions/3.0.0-rc1/lib/ruby/gems/3.0.0/bundler/gems/grape-0f250dc0f574/lib/grape/endpoint.rb:351:in `lazy_initialize!'
[...]

PR #2145, opened 2020-12-28
Ruby 3.0.0 compatibility

ericproulx:
Just trying grape with ruby 3.0. I might have to update rubocop also


anakinj:
Before you waste any time in investigating failures on CI, the failing test is addressed in #2144

ericproulx:
Fix #2142

dblock:
We're moving to GH actions in #2143, which I just merged, update?

anakinj:
This Danger failure is interesting. Think that Danger is not able to interact with it's old entires because it uses a different GitHub token than before. We could either live over the transitiontime or reuse the old token from TravisCI for the danger integration.

Maybe it also could work if a maintainer would delete the old danger entry created from the Travis run, ping @dblock

ericproulx:
Most of the work is :
- Add { } or ** when needed
- Change ** by *

The last one is quite particular. For the middlewares, I figured that [Rack::Builder](https://github.com/rack/rack/blob/8d2e02abba44ac7e9d0ed2a4af9beb6063a81296/lib/rack/builder.rb#L148) is instantiating them like this
```ruby
middleware.new(app, *args, &block)
```
It's the same within the [Grape::Middleware::Stack](https://github.com/ruby-grape/grape/blob/d70a51e4c7841ec34255b176bdf16bdecec5d6f3/lib/grape/middleware/stack.rb#L66). Therefore, I replace the `**options` by `*options` and shifting the first argument if any.

Same thing happened with the `lib/grape/validations/validators/base.rb`.  When having [and_call_original](https://github.com/rspec/rspec-mocks/blob/e882dd0b00a6916337882957b699cdcdea40ec8f/lib/rspec/mocks/message_expectation.rb#L101), the class is actually instantiated :
``` ruby
original.call(*args, &block)
```

I haven't figured how to make a call properly when having a double splat param with a `.new(*args)`

```logs
eproulx@laptop grape % irb      
irb(main):001:0> def test(app, **options); end
=> :test
irb(main):002:0> args = [:app, {opt: [1,2]}]
=> [:app, {:opt=>[1, 2]}]
irb(main):003:0> test(*args)
Traceback (most recent call last):
        5: from /Users/eproulx/.rbenv/versions/3.0.0/bin/irb:23:in `<main>'
        4: from /Users/eproulx/.rbenv/versions/3.0.0/bin/irb:23:in `load'
        3: from /Users/eproulx/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/irb-1.3.0/exe/irb:11:in `<top (required)>'
        2: from (irb):3:in `<main>'
        1: from (irb):1:in `test'
ArgumentError (wrong number of arguments (given 2, expected 1))
```
That's why I made theses changes

ericproulx:
Instead of having `**options` we could go back to `options = {}` which will work.

dblock:
I'm down with whatever solution. Can merge on green - needs a CHANGELOG.

anakinj:
> I'm down with whatever solution. Can merge on green -needs a CHANGELOG. Looks like danger didn't rerun for the PR when it was updated? It should have picked up on the changelog. Did we configure this wrong?

Check my previous comment, think its because the token for danger has changed and it cannpt interact with previous entries

dblock:
I merged this, YOLO

PR #2149, opened 2021-01-14
Update doc ConnectionManagement without rails

oslivan:
In Rails 5, `ActiveRecord::ConnectionAdapters::ConnectionManagement` is removed, please look [the discuss](https://github.com/rails/rails/issues/26947)

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,24 +317,20 @@ Grape will also automatically respond to HEAD and OPTIONS for all GET, and just 
 If you want to use ActiveRecord within Grape, you will need to make sure that ActiveRecord's connection pool
 is handled correctly.
 
+Rails 4
+
 The easiest way to achieve that is by using ActiveRecord's `ConnectionManagement` middleware in your
 `config.ru` before mounting Grape, e.g.:
+```ruby
+use ActiveRecord::ConnectionAdapters::ConnectionManagement
+```
+
+Rails 5+
+
+Due to ActiveRecord's `ConnectionManagement` was removed in Rails 5, so you can use [otr-activerecord](https://github.com/jhollinger/otr-activerecord), and add like below:
 
 ```ruby
-# Rails 4 and less
-use ActiveRecord::ConnectionAdapters::ConnectionManagement
-
-run Twitter::API
-
-# Rails 5 and later
-app = run(Twitter::API)
-
-ActiveRecord::QueryCache.install_executor_hooks
-ActiveRecord::Base.class_eval do
-  self.connection_handlers = { writing_role => default_connection_handler }
-end
-
-run lambda { |env| ActiveSupport::Executor.wrap { app.call(env) } }
+use OTR::ActiveRecord::ConnectionManagement
 ```
 
 ### Alongside Sinatra (or other frameworks)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -321,9 +321,20 @@ The easiest way to achieve that is by using ActiveRecord's `ConnectionManagement
 `config.ru` before mounting Grape, e.g.:
 
 ```ruby
```
dblock, line 323: I'd like to nitpick here. Let's make sections to keep things organized.

#### Rails 4

```ruby
use ...
```

#### Rails 5+

```ruby
...
```

dblock (changes requested):
This seems like a lot. What's the code that's generated by rails by default, or is there a doc on what replaces `ConnectionManagement` that we can point to? Is everybody really putting this code in with Grape and Rails 5+ (haven't done this in a while)?

dblock:
I read the SO and the discussion carefully. I think for Grape we should rather recommend using https://github.com/ioquatix/activerecord-rack instead of rolling out our own like here, and point to the SO question in the README. What do you think?

oslivan:
I tried install `activerecord-rack`, and not compatible with rails 6, if need to introduce third-party gem, I recommend `otr-activerecord`.

dblock:
> I tried install `activerecord-rack`, and not compatible with rails 6, if need to introduce third-party gem, I recommend `otr-activerecord`.

That's actually spot on, thanks for reminding us of it. I've used it [here](https://github.com/slack-ruby/slack-ruby-bot-server#activerecord). Is it an alternative to what this PR is currently proposing? Want to update?

oslivan:
> > I tried install `activerecord-rack`, and not compatible with rails 6, if need to introduce third-party gem, I recommend `otr-activerecord`.
> 
> That's actually spot on, thanks for reminding us of it. I've used it [here](https://github.com/slack-ruby/slack-ruby-bot-server#activerecord). Is it an alternative to what this PR is currently proposing? Want to update?

Thanks for review, I have changed, please checkout.

README.md
```diff
@@ -317,13 +317,20 @@ Grape will also automatically respond to HEAD and OPTIONS for all GET, and just
 If you want to use ActiveRecord within Grape, you will need to make sure that ActiveRecord's connection pool
 is handled correctly.
 
+Rails 4
```
dblock, line 320: Can you make this a `####` section pls?

README.md
```diff
@@ -317,13 +317,20 @@ Grape will also automatically respond to HEAD and OPTIONS for all GET, and just
...
 `config.ru` before mounting Grape, e.g.:
-
 ```ruby
 use ActiveRecord::ConnectionAdapters::ConnectionManagement
+```
 
-run Twitter::API
+Rails 5+
+
+Due to ActiveRecord's `ConnectionManagement` was removed in Rails 5, so you can use [otr-activerecord](https://github.com/jhollinger/otr-activerecord), and add like below:
```
dblock, line 330: It's not Grape's place to explain what happened in AR :) Just say `Use otr-activerecord as follows.`

dblock (changes requested):
I have more comments, thanks for hanging in here!

README.md
oslivan: Nice suggestion

dblock:
@anakinj Danger failed here to post a note (TOC out of date), https://github.com/ruby-grape/grape/pull/2149/checks?check_run_id=1725006392, did we rotate the token?

dblock:
Merged via https://github.com/ruby-grape/grape/pull/2153

PR #2153, opened 2021-01-19
Updated Danger token

dblock:
merging #2149

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,8 +1,8 @@
 ---
-name: Danger
+name: danger
 on:
   pull_request:
-    ypes: [opened, reopened, edited, synchronize]
+    types: [opened, reopened, edited, synchronize]
 jobs:
   danger:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


anakinj:
@dblock I think the public api token for Danger interaction needs to be added to the CI yml. The secrets for the repository only seem to be available for collaborators.

An example on how Danger does it: https://github.com/danger/danger/blob/master/.github/workflows/CI.yml

Also thing the API key for this repo is already in wild: https://travis-ci.org/github/ruby-grape/grape/jobs/751852078#L184

anakinj:
This worked using the repository secrets because I tested this on my local fork on where i was the owner. But for public pull requests the secrets will be hidden. The UI for the repository secrets even state `Secrets are environment variables that are encrypted. Anyone with collaborator access to this repository can use these secrets for Actions.`

dblock:
> @dblock I think the public api token for Danger interaction needs to be added to the CI yml. The secrets for the repository only seem to be available for collaborators.
> 
> An example on how Danger does it: https://github.com/danger/danger/blob/master/.github/workflows/CI.yml
> 
> Also think the API key for this repo is already in wild: https://travis-ci.org/github/ruby-grape/grape/jobs/751852078#L184

Thanks. I spoke to @orta and it looks like hard-coding the token is the easiest way. Seems like a miss from Github to not allow clear tokens for scenarios like these.

anakinj:
Seems to be working now, great job.

Issue #2093, opened 2020-08-05
Custom error for a custom type - coercion


kubaracek:
I'm having trouble returning a custom error when using custom coercion. My type is a bit more advanced hash which on its own holds `type` that can be of a few values and based on this value I'm doing validations of the rest of the hash.

The problem I'm experiencing is that when simply doing `fail "this attribute must be X"` in my class, grape returns default `X is invalid`. I'm catching the error before it reaches the user with `rescue_from Grape::Exceptions::ValidationErrors` but I don't seem to have the error in there to handle it properly.

My end goal is to be able to post a JSON as in the following example and have custom validations based on the type of the object, I've tried to do that with a `given` attribute, but that has it's own issues (which I think have been already reported here, it seems like the second `given` overrides the first)

```json
"foos": [
   { "type": "file",
     "fileName": "nameOfFile",
     "file64": "base64"
   },
   { "type": "note",
     "title": "hello world",
     "text": "Grape is good"
   }
]
```

my params block looks currently as

```ruby
params do
    requires :foos, type: Array, coerce_with: GrapeApi::V1::Types::Foo do
       requires :type, type: String, values: ["file", "note"]
    end
end
```

When I tried to change that to `given` without the coercion which would look as the following example, the validations are broken

```ruby
params do
    requires :foos, type: Array do
       requires :type, type: String, values: ["file", "note"]
       given type: ->(val) { val == "file"} do
          requires :fileName
          requires :file64
       end
       
       given type: ->(val) { val == "note"} do
          requires :title
          requires :text
       end
    end
end
```

What's the best way for me to implement this? the coercion kinda works, but I would like to communicate back to the user that for example the `title` is a required argument when the type is `note`. Currently what he gets is `foos is Invalid` which doesn't say much. Thank you

kubaracek:
I'm just trying a custom type instead of the coercion but the `fail "hello world"` still produces the default `foo is Invalid`

```ruby
class Foo
	attr_reader :value
	def initialize(value)
		@value = value
	end
	def self.parse(value)
		fail "hello world"
		new(value)
	end
end

params do                            
  optional :foo, type: Array[Foo]  
end                                  
                                     
post :foo do                         
  params.inspect                     
end
```

kubaracek:
Ok, found the reason here https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/coerce.rb#L73 the error handling does not care about the original error and raises the default coercion error. I might create a PR for that

[...]

PR #2157, opened 2021-01-28
Custom types can set a message to be used in the response when invalid

dnesteryuk:
Just return an instance of `Grape::Types::InvalidValue` with the message:

```ruby
class Color
  def self.parse(value)
    return value if %w[blue red green].include?(value)

    Grape::Types::InvalidValue.new('Invalid color')
  end
end
```

Any raised exception will be treated as an invalid value as it was before.

Implements an idea listed in https://github.com/ruby-grape/grape/issues/2093

### DIFF
--- a/README.md
+++ b/README.md
@@ -1183,7 +1183,8 @@ Aside from the default set of supported types listed above, any class can be
 used as a type as long as an explicit coercion method is supplied. If the type
 implements a class-level `parse` method, Grape will use it automatically.
 This method must take one string argument and return an instance of the correct
-type, or raise an exception to indicate the value was invalid. E.g.,
+type, or return an instance of `Grape::Types::InvalidValue` which optionally
+accepts a message to be returned in the response.
 
 ```ruby
 class Color
@@ -1193,8 +1194,9 @@ class Color
   end
 
   def self.parse(value)
-    fail 'Invalid color' unless %w(blue red green).include?(value)
-    new(value)
+    return new(value) if %w[blue red green]).include?(value)
+
+    Grape::Types::InvalidValue.new('Unsupported color')
   end
 end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -1193,8 +1194,9 @@ class Color
   end
 
   def self.parse(value)
-    fail 'Invalid color' unless %w(blue red green).include?(value)
-    new(value)
+    return new(value) if %w[blue red green]).include?(value)
+
+    Grape::Types::InvalidValue.new('Unsupported color')
```
dblock, line 1199: I feel like this should be a `raise` that gets handled, not a return.

dblock (commented):
This is great, but I think we should rely on an exception to separate valid returns from errors in the flow. We could similarly rescue `StandardError` and convert it into something with more detail like a `ParseError` that would help the caller identify the actual parameter that could not be parsed. WDYT?

README.md
dnesteryuk: How will we know that an exception message should be returned to the user if we rescue `StandardError`? I would like to avoid guessing :slightly_smiling_face:  We already return the invalid value [there](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/types/custom_type_coercer.rb#L58) and [there](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/types/dry_type_coercer.rb#L67) and [there](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/types/array_coercer.rb#L41).

In general, if there is an easy way to return a value indicating a failure, I prefer to do that. It is cheaper:

```ruby
require 'benchmark/ips'

def return_value
  :invalid
end

def raise_exception
  raise StandardError
end

Benchmark.ips do |x|
  x.report('return value') do
    return_value
  end
  x.report('raise error') do
    begin
      raise_exception
    rescue StandardError
    end
  end

  x.compare!
end
```

```
        return value     1.990M i/100ms
         raise error   149.703k i/100ms
        return value     19.948M (± 0.4%) i/s -    101.470M in   5.086841s
         raise error      1.513M (± 0.5%) i/s -      7.635M in   5.044907s

Comparison:
        return value: 19947800.7 i/s
         raise error:  1513417.9 i/s - 13.18x  (± 0.00) slower
```

dblock: Ok, I clearly didn't do my homework, especially on the existing code.

PR #2158, opened 2021-02-02
Spelling corrections

matiasasis:
Just some spelling corrections


matiasasis:
It was not clear for me If I should add something to changelog, I thought as this is just comments corrections nothing should be added.

dblock:
Great thanks! No need for CHANGELOG

PR #2160, opened 2021-02-09
Fix testing to be compatible with Rails 7 dependencies

anakinj:
Rails 7 is dropping pre Ruby 2.7 support. So limiting the Rails edge tests to only be executed on 2.7


anakinj:
This can probably be merged without any changelog?

PR #2161, opened 2021-02-09
Handle EOFError raised by Rack

bschmeck:
In v2.2.0, Rack started raising an EOFError when given an empty body with a multipart upload: https://github.com/rack/rack/pull/1572.  Previously, Rack swallowed this error and execution continued normally.  This PR translates that error to an `InvalidMessageBody` exception, letting Grape return a 400 class error, instead of a 500 class error.

Note that the error raised by Rack will change to `Rack::Multipart::EmptyContentError` in v3 - https://github.com/rack/rack/pull/1688


lib/grape/request.rb
```diff
@@ -15,6 +15,8 @@ def initialize(env, **options)
 
     def params
       @params ||= build_params
+    rescue EOFError
+      raise Grape::Exceptions::InvalidMessageBody, 'multipart/form-data'
```
dblock, line 19: I am concerned that this is hardcoding `multipart/form-data`, input doesn't have to be this way, right?

dblock (changes requested):

lib/grape/request.rb
bschmeck: I don't know when Rack would raise this error other than when given a `multipart/form-data`, but I'm not a Rack expert.  I've pushed a commit to use the object's `content_type` method instead.

This error class seemed the most appropriated, but it requires a content type.  I'm happy to create a different error class if that would be preferable.

dblock: I'm ok with reusing this error, it makes sense to me.

spec/grape/endpoint_spec.rb
```diff
@@ -420,6 +420,22 @@ def app
         expect(last_response.status).to eq(201)
         expect(last_response.body).to eq('Bob')
       end
+
+      it 'returns a 400 if given an invalid multipart body' do
+        # Rack swallowed this error until v2.2.0
+        major, minor, _patch = Rack.release.split('.').map(&:to_i)
```
dblock, line 426: The "correct" way to do this is to use `Gem::Version.new(Rack.release) >= ...`, see https://code.dblock.org/2020/07/16/comparing-version-numbers-in-ruby.html if it's helpful. 

Let's also externalize the `if` as a spec condition so we see it properly skipped.

lib/grape/request.rb
dblock: On a second thought, should we create a EmptyMessageBody error?

dblock (changes requested):
Let's also add a gemfile and run it as part of the matrix explicitly with this version of Rack, ala https://github.com/ruby-grape/grape/blob/master/gemfiles/rack2.gemfile

spec/grape/endpoint_spec.rb
bschmeck: Thanks!  I figured there had to be a better way.

bschmeck:
I've cleaned up how the test is skipped for older versions of Rack, added a custom `EmptyMessageBody` error and explicitly added Rack 2.2 to the test matrix.

I think that addresses all of your comments, but let me know if I missed something.

dblock:
Thanks for hanging in here, great job. Merged.

dblock:
Related, we had #2130, maybe you care to resolve it one way or another? I do think a note on all Rack-related issues could be useful somewhere.

PR #2162, opened 2021-02-19
Corrected a hash modification while iterating issue.

Jack12816:
**- What is it good for**

There is a nasty bug on edge case usage of nested mounted APIs which use a shared class-helper method to abstract the creation of a route. With this fix in place, no errors are raised. Otherwise a `RuntimeError: can't add a new key into hash during iteration` is raised. This is caused by [an inline modification of the @setup set inside the Grape::API class](https://github.com/ruby-grape/grape/blob/v1.5.2/lib/grape/api.rb#L144). This error is caused by the definition of the helper method in a specific way. The following form causes the issue:

```ruby
def uniqe_id_route
  params do
    use :unique_id
  end
  route_param(:id) do
    yield
  end
end  
```

while this form works as expected:

```ruby
def uniqe_id_route(&block)
  params do
    use :unique_id
  end
  route_param(:id, &block)
end
```

**- What I did**

I just converted the inline set modification into a reassignment of a freshly built set and added a test that proves the bugfix.

images

### DIFF
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -4056,4 +4056,81 @@ XML
       expect { get '/const/missing' }.to raise_error(NameError).with_message(/SomeRandomConstant/)
     end
   end
+
+  describe 'custom route helpers on nested APIs' do
+    let(:shared_api_module) do
+      Module.new do
+        # rubocop:disable Style/ExplicitBlockArgument because this causes
+        #   the underlying issue in this form
+        def uniqe_id_route
+          params do
+            use :unique_id
+          end
+          route_param(:id) do
+            yield
+          end
+        end
+        # rubocop:enable Style/ExplicitBlockArgument
+      end
+    end
+    let(:shared_api_definitions) do
+      Module.new do
+        extend ActiveSupport::Concern
+
+        included do
+          helpers do
+            params :unique_id do
+              requires :id, type: String,
+                            allow_blank: false,
+                            regexp: /\d+-\d+/
+            end
+          end
+        end
+      end
+    end
+    let(:orders_root) do
+      shared = shared_api_definitions
+      find = orders_find_endpoint
+      Class.new(Grape::API) do
+        include shared
+
+        namespace(:orders) do
+          mount find
+        end
+      end
+    end
+    let(:orders_find_endpoint) do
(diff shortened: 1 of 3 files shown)


spec/grape/api_spec.rb
```diff
@@ -4056,4 +4056,81 @@ def before
       expect { get '/const/missing' }.to raise_error(NameError).with_message(/SomeRandomConstant/)
     end
   end
+
+  fdescribe 'custom route helpers on nested APIs' do
```
dblock, line 4060: remove `f`

dblock (changes requested):
Doesn't this start adding duplicates into `@setup`? It's a `Set`.

Jack12816:
> Doesn't this start adding duplicates into `@setup`? It's a `Set`.

No, it does not add duplicates, because it's a set. Example:

```ruby
block1 = proc { false }
block2 = proc { true }

setup = Set.new
setup << { method: :default_format, args: [:json], block: block1 }
setup << { method: :default_format, args: [:text], block: block2 }
setup << { method: :default_format, args: [:text], block: block2 }
# =>
# #<Set: {{:method=>:default_format,
#   :args=>[:json],
#   :block=>#<Proc:0x000055bbcbe3a8f0 (irb):5>},
#  {:method=>:default_format,
#   :args=>[:text],
#   :block=>#<Proc:0x000055bbcb83dce0 (irb):6>}}>

setup = Set.new
setup += [{ method: :default_format, args: [:json], block: block1 }]
setup += [{ method: :default_format, args: [:text], block: block2 }]
setup += [{ method: :default_format, args: [:text], block: block2 }]
# #<Set: {{:method=>:default_format,
#   :args=>[:json],
#   :block=>#<Proc:0x000055bbcbe3a8f0 (irb):5>},
#  {:method=>:default_format,
#   :args=>[:text],
#   :block=>#<Proc:0x000055bbcb83dce0 (irb):6>}}>
```

dblock:
Looks good, thank you.

marcusg:
@dblock can you estimate when a new version can be released including this patch?

dblock:
I can do it this week unless someone gets it before me, add your +1 to https://github.com/ruby-grape/grape/issues/2163

PR #2164, opened 2021-03-02
coerce_with should be called for params with nil value

braktar:
Regarding the expected behavior pointed by @dblock 
> So should we adjust the default validator behavior ?

Yes, I believe so. A `nil` value is a `nil` value, not the absence of a parameter.

_Originally posted by @dblock in https://github.com/ruby-grape/grape/issues/2040#issuecomment-615868085_

Here is a first attemp to fix the behavior applied to the CustomTypeCoercer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 * [#2161](https://github.com/ruby-grape/grape/pull/2157): Handle EOFError from Rack when given an empty multipart body - [@bschmeck](https://github.com/bschmeck).
 * [#2162](https://github.com/ruby-grape/grape/pull/2162): Corrected a hash modification while iterating issue - [@Jack12816](https://github.com/Jack12816).
+* [#2164](https://github.com/ruby-grape/grape/pull/2164): Fix: `coerce_with` is now called for params with `nil` value - [@braktar](https://github.com/braktar).
 
 ### 1.5.2 (2021/02/06)
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,33 @@
 Upgrading Grape
 ===============
 
+
+### Upgrading to >= 1.5.3
+
+### Nil value and coercion
+
+Prior to 1.2.5 version passing a `nil` value for a parameter with a custom coercer would invoke the coercer, and not passing a parameter would not invoke it.
+This behavior was not tested or documented. Version 1.3.0 quietly changed this behavior, in such that `nil` values skipped the coercion. Version 1.5.3 fixes and documents this as follows:
+
+```ruby
+class Api < Grape::API
+  params do
+    optional :value, type: Integer, coerce_with: ->(val) { val || 0 }
+  end
+
+  get 'example' do
+     params[:my_param]
+  end
+  get '/example', params: { value: nil }
+  # 1.5.2 = nil
+  # 1.5.3 = 0
+  get '/example', params: {}
+  # 1.5.2 = nil
+  # 1.5.3 = nil
+end
+```
+See [#2164](https://github.com/ruby-grape/grape/pull/2164) for more information.
+
 ### Upgrading to >= 1.5.1
 
 #### Dependent params
--- a/README.md
+++ b/README.md
@@ -1228,6 +1228,7 @@ params do
   end
 end
 ```
+Note that, a `nil` value will call the custom coercion method, while a missing parameter will not.
 
 Example of use of `coerce_with` with a lambda (a class with a `parse` method could also have been used)
 It will parse a string and return an Array of Integers, matching the `Array[Integer]` `type`.
--- a/lib/grape/validations/types/custom_type_coercer.rb
+++ b/lib/grape/validations/types/custom_type_coercer.rb
@@ -52,8 +52,6 @@ module Grape
         #   this should always be a string.
         # @return [Object] the coerced result
         def call(val)
-          return if val.nil?
-
           coerced_val = @method.call(val)
 
           return coerced_val if coerced_val.is_a?(InvalidValue)
--- a/spec/grape/validations/validators/coerce_spec.rb
+++ b/spec/grape/validations/validators/coerce_spec.rb
@@ -754,6 +754,44 @@ describe Grape::Validations::CoerceValidator do
         expect(last_response.body).to eq('3')
       end
 
+      context 'Integer type and coerce_with should' do
+        before do
+          subject.params do
+            optional :int, type: Integer, coerce_with: (lambda do |val|
+              if val.nil?
+                0
+              else
+                val.to_i
+              end
+            end)
+          end
+          subject.get '/' do
+            params[:int].class.to_s
+          end
+        end
+
+        it 'coerce nil value to integer' do
+          get '/', int: nil
+
+          expect(last_response.status).to eq(200)
+          expect(last_response.body).to eq('Integer')
+        end
+
+        it 'not coerce missing field' do
+          get '/'
+
+          expect(last_response.status).to eq(200)
+          expect(last_response.body).to eq('NilClass')
+        end
+
+        it 'coerce integer as integer' do
+          get '/', int: 1
+
+          expect(last_response.status).to eq(200)
+          expect(last_response.body).to eq('Integer')
+        end
+      end
+
       context 'Integer type and coerce_with potentially returning nil' do
         before do
           subject.params do
--- a/spec/grape/validations/validators/coerce_spec.rb
+++ b/spec/grape/validations/validators/coerce_spec.rb
@@ -706,6 +706,44 @@ describe Grape::Validations::CoerceValidator do
         expect(JSON.parse(last_response.body)).to eq([1, 1, 1, 1])
       end
 
+      context 'Array type and coerce_with should' do
+        before do
+          subject.params do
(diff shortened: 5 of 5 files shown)


senhalil (approved):
It looks good to me

dblock:
This needs a CHANGELOG entry please and an UPGRADING note. The behavior has changed, even though it was undocumented. 

Do we think version should be bumped?

senhalil:
For me the current behaviour was surprising -- i.e., coerce not being called for a declared value -- and it immediately registered as a bug.

This new behaviour got introduced silently somewhere between 1.4.1 and 1.5.0 so from my point of view, it is more of a bug fix (to go back to the usual behaviour) then a behaviour change per se. 


If you see this as a backwards (original behaviour) compatible bug fix, semver says [patch level](https://semver.org/#spec-item-6 ) shall be bumped. 

If you see it as correcting the backwards compatibility broken by a previous release then it says [minor version](https://semver.org/#what-do-i-do-if-i-accidentally-release-a-backwards-incompatible-change-as-a-minor-version).

After saying all that, I don't think I am really at the position to suggest one way or another vis-a-vis version change so you have been warned.

dblock:
> This new behaviour got introduced silently somewhere between 1.4.1 and 1.5.0 so from my point of view, it is more of a bug fix (to go back to the usual behaviour) then a behaviour change per se.

Makes sense, bug fix it is. Looking forward to all the README/UPGRADING/CHANGELOG parts. Thank you.

braktar:
The .md files are now updated. + I've added spec in case of an array

UPGRADING.md
```diff
@@ -1,6 +1,31 @@
...
+  end
+  get '/example', params: { value: nil }
+  # 1.5.2 = nil
+  # 1.5.3 = 0
+  get '/example', params: {}
+  # 1.5.2 = nil
+  # 1.5.3 = nil
+end
+```
+
```
dblock, line 28: Add `See [#2164](https://github.com/ruby-grape/grape/pull/2164) for more information.`

UPGRADING.md
```diff
@@ -1,6 +1,31 @@
 Upgrading Grape
 ===============
 
+
+### Upgrading to >= 1.5.3
+
+### Nil value and coercion
+
+A parameter with a value set to `nil` call custom coercion method, but the coercion is not triggered for missing parameters.
```
dblock, line 9: Let's find when the wrong behavior was introduced, try against 1.4.0 and up and let's make sure we are clear how it behaved and that we're _fixing_ this now. I think the text should say something like this:

Prior to version 1.x.0 passing a `nil` value for a parameter with a customer coercer would invoke the coercer, and not passing a value would not invoke it. This behavior was not tested or documented. Version 1.z.0 quietly changed this behavior, in such that ... . Version 1.5.3 fixes and documents this as follows: .... 

Makes sense?

dblock (changes requested):
Good. I still would like clearer UPGRADING help.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 * [#2161](https://github.com/ruby-grape/grape/pull/2157): Handle EOFError from Rack when given an empty multipart body - [@bschmeck](https://github.com/bschmeck).
 * [#2162](https://github.com/ruby-grape/grape/pull/2162): Corrected a hash modification while iterating issue - [@Jack12816](https://github.com/Jack12816).
+* [#2164](https://github.com/ruby-grape/grape/pull/2164): coerce_with is called for params with nil value - [@braktar](https://github.com/braktar).
```
dblock, line 13: Say `Fix: coerce_with` is now called ...` ?

UPGRADING.md
dm1try: mm, I think this behavior was introduced in the migration to `dry-types`, see https://github.com/ruby-grape/grape/commit/9b352a88ff3a4970af7bfa3c633c09e956c2e8ea#diff-6158291f8f15938bb4d6460b274b5106132113841ede636d29d20ef3d01f21b7R53

so it should be reproducible starting from 1.3.0

dm1try:
aside note
there is no specification for passing a `nil` value using HTTP query string, so in the examples above we are relying on the fact that [`rack` interprets](https://github.com/rack/rack/blob/5ce5b2ccb151c62652e25b4711218ede11497cc3/test/spec_utils.rb#L100-L102) `foo` without `=` sign as a key with a `nil` value, while `foo=` is interpreted as a key with an empty string. but not all client implementations follow the same rules.

dm1try:
another aside note

I feel like `coerce_with` is  something redundant. it looks like all requirements can be solved by providing a custom abstraction for a type(just a class) and allowing using lambda in `type`
```ruby
`type`: ->(val) { }
```
I think providing different parsing rules for something should lead to creating a new type as using the same rules will lead to inconsistency from the client perspective, not sure though :) 
```ruby
type: `Integer` # default rules
type: `Integer`, coerce_with: MyInt #another 
type: `Integer`, corce_with: ->() #another
```

UPGRADING.md
braktar: @dm1try thanks for the hint. I confirm that the skip of `nil` value was introduced in v1.3.0, and version v1.2.5 met the expected behavior.

dblock:
Any idea what the truffleruby problem is here? @gogainda @eregon

UPGRADING.md
```diff
@@ -1,6 +1,33 @@
 Upgrading Grape
 ===============
 
+
+### Upgrading to >= 1.5.3
+
+### Nil value and coercion
+
+Prior to 1.2.5 version passing a `nil` value for a parameter with a custom coercer would invoke the coercer, and not passing a parameter would not invoke it.
+This behavior was not tested or documented. Version 1.3.0 quietly changed this behavior, in such that `nil` values skipped the coercion. Version 1.5.3 fixes and documents this as follows :
```
dblock, line 10: nit, extra space between `follows` and `:` ;)

dblock (approved):
I'm good with this, but need to figure out what's up with CI.

UPGRADING.md
braktar: Nice catch !

eregon:
https://github.com/ruby-grape/grape/pull/2164/checks?check_run_id=2039817421
```
     ArgumentError:
       wrong number of arguments (given 0, expected 1)
     # ./lib/grape/exceptions/method_not_allowed.rb:6:in `initialize'
     # <internal:core> core/truffle/exception_operations.rb:18:in `new'
     # <internal:core> core/truffle/exception_operations.rb:18:in `build_exception_for_raise'
     # <internal:core> core/kernel.rb:694:in `raise'
     # ./lib/grape/endpoint.rb:258:in `block in run'
     # ./vendor/bundle/truffleruby/21.1.0-dev-ffeea561/gems/activesupport-6.1.3/lib/active_support/notifications.rb:205:in `instrument'
     # ./lib/grape/endpoint.rb:247:in `run'
     # ./lib/grape/endpoint.rb:322:in `block in build_stack'
```

So https://github.com/ruby-grape/grape/blob/a43fea98617465578b99e84c681bb6ce0dbed91b/lib/grape/endpoint.rb#L258 calls `raise` and it does not give a message String. It does pass a positional Hash argument though.
And https://github.com/ruby-grape/grape/blob/a43fea98617465578b99e84c681bb6ce0dbed91b/lib/grape/exceptions/method_not_allowed.rb#L6 expects to receive that Hash.
It started failing most likely because of https://github.com/oracle/truffleruby/commit/8348283a5efd5de1e1ce8e0468015a74efaa6537, cc @bjfish

The code in `grape` seems fine, I think we need to find another way to extract the `cause:` keyword argument of `raise` (e.g., `kwargs = undefined`) so we don't lose other keys in the Hash.
In Ruby 3 this should no longer be a problem and we could restore the current code, because then the positional Hash would not be considered for keyword arguments.

dblock:
CI should be skipping truffleruby failures (we have this experimental=true in there, don't we?). Feel free to "fix" this by removing it if needed, and someone else can deal with it, I'll merge on green. I don't want to hold up a release either.

eregon:
@dblock That's due to the confusing `continue-on-error` of GitHub, see https://github.com/actions/toolkit/issues/399#issuecomment-738700569.
The latest build as a whole is green, see https://github.com/ruby-grape/grape/actions/runs/624485357
But the PR UI is really bad at showing that.

I think you could switch to `truffleruby` instead of `truffleruby-head` to avoid this issue, and it should also be more stable.
OTOH it's very useful that we see this error, so thank you for reporting it.

bjfish:
I currently see an issue with:
https://github.com/ruby-grape/grape/blob/a43fea98617465578b99e84c681bb6ce0dbed91b/lib/grape/endpoint.rb#L258

The second to parameter to `Kernel#raise` should be a `string` as indicated by the documentation here:
https://ruby-doc.org/core-2.7.2/Kernel.html#method-i-raise

Here is a good example on how to pass custom data to an exception:
https://www.honeybadger.io/blog/ruby-custom-exceptions/

I will also work on fixing the TruffleRuby failure here as it is an incompatibility.

dblock:
Good call @bjfish. Want to help fix the Grape implementation?

bjfish:
@dblock Yes, I've created a PR here with all of these types of usages that I found:
https://github.com/ruby-grape/grape/pull/2165

PR #2165, opened 2021-03-05
Initialize error data using constructors directly

bjfish:
This avoids using the `Kernel#raise` string message parameter to initialize custom data for errors in favor of initializing with the data directly as described here:
https://github.com/ruby-grape/grape/pull/2164#issuecomment-791597922

This is also consistent with other usages like:
https://github.com/ruby-grape/grape/blob/e0f412ca73f6188c233ac9c35661027d28f38a54/lib/grape/middleware/versioner.rb#L29
https://github.com/ruby-grape/grape/blob/e0f412ca73f6188c233ac9c35661027d28f38a54/lib/grape/endpoint.rb#L129

### DIFF
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -255,7 +255,7 @@ module Grape
           run_filters befores, :before
 
           if (allowed_methods = env[Grape::Env::GRAPE_ALLOWED_METHODS])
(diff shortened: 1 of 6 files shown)


dblock:
Anything changes for users that may be relying on this that deserves a CHANGELOG/README/UPGRADING?

eregon (commented):
This look good to me. It should work exactly the same or better, no user visible changes (except maybe slightly faster to build those exceptions).

PR #2167, opened 2021-03-12
handle the changes in errors method in rails 6.1

abo-elleef:
solve #2168
as active_record_instance#errors metho is not returning a hash object anymore. as a part of rails 6.1 upgrade
the JSON error formatter should consider this change and check if the object type is `ActiveModel::DeprecationHandlingMessageHash`

### DIFF
--- a/lib/grape/error_formatter/json.rb
+++ b/lib/grape/error_formatter/json.rb
@@ -22,7 +22,7 @@ module Grape
         private
 
         def wrap_message(message)
-          if message.is_a?(Exceptions::ValidationErrors) || message.is_a?(Hash)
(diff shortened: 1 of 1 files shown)


dblock:
Thanks. At the very least this needs an integration test. 

I don't like that we're checking against a Rails type. Maybe these objects `respond_to?(:something)` which can make this cleaner. Or we need to extend the implementation with a proper Railtie.

dblock:
Another thought, since `User.first.errors.messages` is producing the exception, maybe the "right way" is to avoid the exception altogether?

dblock (changes requested):
Want to finish this @abo-elleef ?

Issue #2173, opened 2021-04-06
After 1.5.0 , 'route_setting :custom , key: :value' is broken with 'mount'.


neocoin:
env 

```
Ruby 3.0.0
Grape 1.5.3
Rails 6.1
```

```ruby
      def route_setting(key, value = nil)
        get_or_set :route, key, value
      end
```

'value' work like frozen hash key. You can reproduce below code. 

```ruby

module V1
  class Status < Grape::API
    desc 'time'
    route_setting :custom, key: :value
    get '/status/time' do
      { time: Time.zone.now }
    end

    desc 'time2'
    route_setting :custom, key: :value
    get '/status/time2' do
      { time: Time.zone.now }
    end
  end
  
  class Mount < Grape::API
    mount ::V1::Status
  end
end
```

Second , 'route_setting' method calling doesn't work. 

```bash
[1] pry(main)> ap V1::Status.routes.map{ [_1.path, _1.settings]}
[
    [0] [
        [0] "/:version/status/time(.json)",
        [1] {
            :description => {
                :description => "time"
            },
                 :custom => {
                :key => :value
            }
        }
    ],
    [1] [
        [0] "/:version/status/time2(.json)",
        [1] {
            :description => {
                :description => "time2"
            }
        }
    ]
]
=> nil
```

Anybody, can explain this situation? 

route_setting work fine at ~> 1.4.0 on ruby 2.7

dblock:
Probably a bug. Turn it into a failing spec and PR that?

neocoin:
I write spec and pr on #2174 . 
But I don't know a solution yet.

neocoin:
@dim 

This problem start after 6deb8f46b9c4057be59006c85cf317b5d64c6180  
Fix Ruby 2.7 keyword deprecations (#2091)	6deb8f46

neocoin:
@dblock 

I find problem point.

Fix retaining setup blocks when remounting APIs	416a7e15	James Lamont <james@getvero.com>	Sep 18, 2020 12:24 PM
https://github.com/ruby-grape/grape/commit/416a7e15bdfda29cd4f9b335a911bb59a416be60#diff-74e08e6432db25a22b629a2f8031d24d01f8c9435bda532d36e22971d0b1c9a0R33

cc @jylamont

stephannv:
I was about to open an issue about it, I even created a file to test it.
```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

[...]

PR #2174, opened 2021-04-06
[WIP] Fix same custom route_setting value ignoreing problem

neocoin:
#2173 After 1.5.0 , 'route_setting :custom , key: :value' is broken with 'mount'.

check plz. @dblock 

ps. I always thank you for grape.

### DIFF
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -30,7 +30,7 @@ module Grape
       # an instance that will be used to create the set up but will not be mounted
       def initial_setup(base_instance_parent)
         @instances = []
-        @setup = Set.new
+        @setup = []
         @base_parent = base_instance_parent
         @base_instance = mount_instance
       end
--- a/spec/grape/api/describing_and_inspecting_spec.rb
+++ b/spec/grape/api/describing_and_inspecting_spec.rb
@@ -6,7 +6,6 @@ require 'spec_helper'
 #
 # @see https://github.com/ruby-grape/grape/issues/2173
 module DescribingAndInspectingSpec
-
   class CustomKeyExample < Grape::API
     desc 'Includes custom settings.'
     route_setting :custom, key: 'value'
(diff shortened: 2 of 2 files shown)


dblock:
Correct rubocop with `rubocop -a`

dblock:
Next step is to bisect down to the commit that broke it, then we can narrow down a fix.

neocoin:
Oh I didn't read convention rules. 
Passed rubocop and add another success spec.

dblock:
> Oh I didn't read convention rules.
> Passed rubocop and add another success spec.

There are no convention rules! I'm just slowly nudging you to find the issue and write a fix :)

lib/grape/api.rb
```diff
@@ -30,7 +30,7 @@ def inherited(api, base_instance_parent = Grape::API::Instance)
       # an instance that will be used to create the set up but will not be mounted
       def initial_setup(base_instance_parent)
         @instances = []
-        @setup = Set.new
```
neocoin, line 33: Check plz @jylamont

dblock:
Because https://github.com/ruby-grape/grape/pull/2102 actually did address a memory leak I think we need to find another solution :(

neocoin:
@dblock  
I almost understand about problem. 

hm.. bad situation. I try to apply some solutions but I failed.

Basically, '@setup = Set.new' break stackable setting concept of grape

I can't find solution that classify always stackable DSL( desc, route_setings, get by code context ) and non-stackable DSL (logger with nil arguments, top_level_setting etc) at runtime after declarative initilization.

on https://github.com/ruby-grape/grape/blob/master/lib/grape/api.rb#L42

Do you have any idea?

dblock:
You probably undertstand more than me now ;) Maybe we can split setup that needs to be retained and setup that doesn't? Anyone else has any ideas?

neocoin:
Too old. Sorry I changed interface to GraphQL. I don't maintain this branch

PR #2177, opened 2021-05-04
Fix ordering issues between as: and default: validations

### LINKED ISSUE #2094: Issue with the order of as: and default: options on param declarations
Am fairly new to Grape, so apologies if this has come up elsewhere already but doing a quick scan I couldn't see anything covering this issue I was having with using both `as` and `default` param options.

If I use the following to set an `endDate` param with a default value and convert it to `end_date` then the default value behaviour works as expected.

```ruby
params do
  optional :endDate, type: Date, default: (Date.current + 50.days).end_of_month, as: :end_date
end
```
Would result in the default param value being set when no param was sent:

```sh
(byebug) params
{"end_date"=>Wed, 30 Sep 2020}
```
However if I use following for the same request, and specify the `as` option first, the default value is never set on the params, even when using `declared` (although not sure this would make a difference anyway).

```ruby
params do
  optional :endDate, as: :end_date, type: Date, default: (Date.current + 50.days).end_of_month
end
```
Would result in:

```sh
(byebug) params
{}
(byebug) declared(params)
{"end_date"=>nil}
```

I'm assuming this might be a bug with how the param options are being parsed by Grape but wanted to also make sure I wasn't doing something dumb.

Catsuko:
Fixes #2094 where it was found that defining `as:` before `default:` would fail to add the default whereas `as:` after `default:` would work fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 #### Fixes
 
-* [#2176](https://github.com/ruby-grape/grape/pull/2176): Fixes issue-2175 - options call failing if matching all routes - [@myxoh](https://github.com/myxoh).
+* [#2176](https://github.com/ruby-grape/grape/pull/2176): Fix: OPTIONS fails if matching all routes - [@myxoh](https://github.com/myxoh).
+* [#2177](https://github.com/ruby-grape/grape/pull/2177): Fix: `default` validator fails if preceded by `as` validator - [@Catsuko](https://github.com/Catsuko).
 * Your contribution here.
 ### 1.5.3 (2021/03/07)
 
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -300,9 +300,15 @@ module Grape
         # type casted values
         coerce_type validations, attrs, doc_attrs, opts
 
+        as = validations.delete(:as)
+
         validations.each do |type, options|
           validate(type, options, attrs, doc_attrs, opts)
         end
+
+        # Validate as last so other validations are applied to
+        # renamed param
+        validate(:as, as, attrs, doc_attrs, opts) if as
       end
 
       # Validate and comprehend the +:type+, +:types+, and +:coerce_with+
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -289,11 +289,7 @@ module Grape
         opts = derive_validator_options(validations)
 
         # Validate for presence before any other validators
-        if validations.key?(:presence) && validations[:presence]
-          validate('presence', validations[:presence], attrs, doc_attrs, opts)
-          validations.delete(:presence)
-          validations.delete(:message) if validations.key?(:message)
-        end
+        validates_presence(validations, attrs, doc_attrs, opts)
 
         # Before we run the rest of the validators, let's handle
         # whatever coercion so that we are working with correctly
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 #### Fixes
 
 * [#2176](https://github.com/ruby-grape/grape/pull/2176): Fixes issue-2175 - options call failing if matching all routes - [@myxoh](https://github.com/myxoh).
+* [#2177](https://github.com/ruby-grape/grape/pull/2177): Fixes issue-2094 - Issue with the order of `as:` and `default:` options - [@Catsuko](https://github.com/Catsuko).
```
dblock, line 10: Not sure how we inherited this pattern, but we don't usually have one :), change this to `Fix: actual specific problem between as: and default: validators`, and for the line above `Fix: OPTIONS fails if matching all routes`. Thanks.

lib/grape/validations/params_scope.rb
```diff
@@ -289,20 +289,22 @@ def validates(attrs, validations)
...
-          validations.delete(:message) if validations.key?(:message)
-        end
+        validates_presence(validations, attrs, doc_attrs, opts)
 
         # Before we run the rest of the validators, let's handle
         # whatever coercion so that we are working with correctly
         # type casted values
         coerce_type validations, attrs, doc_attrs, opts
 
+        as = validations.delete(:as)
```
dblock, line 299: The `validations` arg that is being passed into the function should not be modified to avoid side effects, make a copy.

lib/grape/validations/params_scope.rb
```diff
@@ -464,6 +466,13 @@ def derive_validator_options(validations)
           fail_fast:   validations.delete(:fail_fast) || false
         }
       end
+
+      def validates_presence(validations, attrs, doc_attrs, opts)
+        return unless validations.key?(:presence) && validations[:presence]
+        validate(:presence, validations[:presence], attrs, doc_attrs, opts)
+        validations.delete(:presence)
```
dblock, line 473: Same as above, don't alter validates, make a copy. The caller may be removing things if needed.

dblock (changes requested):
Close! Thank you.

CHANGELOG.md
Catsuko: sounds good, I reworded it so hopefully it describes it better

lib/grape/validations/params_scope.rb
Catsuko: I refactored my changes to avoid mutating the arg however it is still being done within the method outside of my changes:
```ruby
 desc = validations.delete(:desc) || validations.delete(:description)
# ...
 doc_attrs[:documentation] = validations.delete(:documentation) if validations.key?(:documentation)
```
I'd like to try refactoring however the scope seems a little scary so perhaps as part of its own issue.

dblock: Agreed. And thank you, do give it a shot.

dblock:
Thanks for fixing this, merged. It could be an overkill, but maybe we should introduce some kind of Rack-style listing/mounting of validators so that we don't have to do special ordering gymnastics at runtime. Just food for thought.

Catsuko:
@dblock 

I think it'd make the behaviour more consistent too, perhaps there are other validators that don't play nicely with having `:default` before or after. I'd like to give it a go after I do more reading\thinking about it.

PR #2179, opened 2021-07-08
Reproducible spec to showcase `API.inherited` issue

yogeshjain999:
We're trying to extend the `Grape::API` in one of our gem and found one issue where `Grape::API.inherited` hook doesn't get called when overridden via module composition. It only works when overridden in class itself.

For example,

```ruby
module Inherited
  # This doesn't get called when extended in `API` specific classes
  def inherited(api)
    super
    # ...
  end
end

class RootAPI < Grape::API
  extend Inherited
end

class ChildAPI < RootAPI
end
```

We identified this happens because `inherited` isn't calling `super`. I'm creating this PR in order to showcase reproducible spec. I can create another PR with the fix.

### DIFF
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -4081,6 +4081,49 @@ XML
     end
   end
 
+  describe '.inherited' do
+    context 'overriding within class' do
+      let(:root_api) do
+        Class.new(Grape::API) do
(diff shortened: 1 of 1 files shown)


yogeshjain999:
Closing this wrt #2180 merge.

PR #2180, opened 2021-07-08
Call `super` in `API.inherited`

yogeshjain999:
`API.make_inheritable` was overriding `.inherited` hook on given instance itself. But it was not calling `super` which is required to bubble up the inheritance chain.

This PR removes `.make_inheritable` and calls `super` instead.

Fixes PR with reproducible spec #2179.

### DIFF
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -20,10 +20,11 @@ module Grape
 
       # When inherited, will create a list of all instances (times the API was mounted)
       # It will listen to the setup required to mount that endpoint, and replicate it on any new instance
-      def inherited(api, base_instance_parent = Grape::API::Instance)
-        api.initial_setup(base_instance_parent)
+      def inherited(api)
+        super
+
+        api.initial_setup(Grape::API == self ? Grape::API::Instance : @base_instance)
         api.override_all_methods!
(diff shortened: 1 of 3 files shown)


lib/grape/api.rb
```diff
@@ -68,15 +69,6 @@ def call(*args, &block)
         instance_for_rack.call(*args, &block)
       end
 
-      # Allows an API to itself be inheritable:
-      def make_inheritable(api)
```
apotonick, line 72: Why this change? Couldn't we just call `super` in `#inherited` and leave the rest as it is? :beers:

yogeshjain999: Calling this method with `super` was creating warnings about method redefinations as `override_all_methods! ` gets called twice.

I guess `make_inheritable ` was defined in order to pass extra argument, instead of `super`.

dblock:
Given that tests are passing I'm inclined to merge this. WDYT @apotonick (and thanks for jumping in to review!)?

apotonick:
Now looking great! We need this `super` call to hook into your inheritance in [trb-endpoint](https://github.com/trailblazer/trailblazer-endpoint/), since we want to support the winning HTTP framework! :trophy: :beers:

Issue #2182, opened 2021-07-12
Changing the order of Rails middlewares in grape


onurkucukkece:
We are using a custom middleware for authenticating the grape API requests and also using `rack-attack` gem in the Rails application. We need `Rack::Attack` middleware to be called after our custom middleware in grape. Would this somehow be possible?

dblock:
Do you register your auth middleware [like this](https://github.com/ruby-grape/grape#register-custom-middleware-for-authentication)?

dnesteryuk:
This works for us:

```ruby
# config/application.rb
config.middleware.insert_after ActionDispatch::Static, Rack::Deflater
```

@onurkucukkece Have you tried?

dnesteryuk:
@onurkucukkece did you try to require a file where your middleware is defined?

onurkucukkece:
Yes I tried it and it didn't help. This time I got `“No such middleware to insert before` while starting the server. In our application the API part is loaded in a different module than the `Application` class which is inherited from `Rails::Application` and I think it's the reason why I cannot use both suggestions mentioned here. I will try to look into this later on. Thank you both for your help.

dblock:
Let's reopen this though until there's a solution.

onurkucukkece:
I managed to set the order of middlewares by registering it in our `Application` class which is inherited from `Rails::Application`

```ruby
class Company::Application < Rails::Application
  config.middleware.insert_before(Rack::Attack, Middleware::API::Auth)
end
```

dblock:
@onurkucukkece Care to document this in README?

onurkucukkece:
Yes I'd be glad to. I will create a pull request in the coming days

dblock:
Thanks for the readme, let's close this.

PR #2183, opened 2021-08-01
Update readme about registering middleware in custom Rails applications

onurkucukkece:
This PR updates readme for registering a new middleware in custom Rails applications as discussed in issue #2182


dblock:
Thanks!

Issue #2185, opened 2021-08-06
`values` validator and `allow_blank`


maciejpankanin:
Consider this params definition:
```
params do
  requires :id, type: String
  optional :some_param, type: String, values: ["value1", "value2"]
end
```

Recently I spent couple of hours figuring out why `null` is valid value for `some_param`.

I think there should be a few words in documentation about `values` validator saying that `allow_blank` defaults to `true` and it is stronger than supplied `values` options.

dblock:
Make sense, please contribute! Also double check that we have specs for this exact behavior.

dblock:
Closed via https://github.com/ruby-grape/grape/pull/2186

flosacca:
I was confused by the fact that the empty string is not a special case that is allowed by default. I've tested the following example mentioned in `README` and an empty string cannot pass.
```ruby
params do
  requires :state, type: Symbol, values: [:active, :inactive]
end
```

There may be a misunderstanding in the original problem. When `optional` is used, `null` would be allowed unless `allow_blank` is explicitly specified to `false`.

I found that `allow_blank: false` enables `AllowBlankValidator` while `allow_blank: true` actually does nothing. So, whether an empty string is allowed by default depends totally on other factors.

maciejpankanin:
Ops, you are right, I think it should be:
```ruby
params do
  optional :state, type: Symbol, values: [:active, :inactive]
end
```

But still, isn't it misbehavior? If I understand it well, when it is `optional` and I send `null` or `''` validation will pas but if it was `requires` I would get a validation error.

It behaves like `null` or empty strings weren't values.

dblock:
Current behavior looks suspicious to me: I certainly don't expect an optional value to allow an empty string. Try writing a spec for this, fixing it and seeing if any other specs fail?

PR #2186, opened 2021-08-12
Update README.md about values validator used with allow_blank

maciejpankanin:
See #2185.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1530,6 +1530,14 @@ end
 
 While Procs are convenient for single cases, consider using [Custom Validators](#custom-validators) in cases where a validation is used more than once.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1530,6 +1530,16 @@ end
 
 While Procs are convenient for single cases, consider using [Custom Validators](#custom-validators) in cases where a validation is used more than once.
 
+Notice that [allow_blank](#allow_blank) validator is still relevant while using `values`. Having defined params like:
```
dblock, line 1533: Nit, let's use "note" like everywhere else, and make this into a single sentence?

_Note that [allow_blank](#allow_blank) validator applies while using `:values`. In the following example the absence of `:allow_blank` does not prevent `:state` from receiving blank values because `:allow_blank` defaults to `true`._

dblock (changes requested):

README.md
dblock: Don't hesitate to fix the others ;)

PR #2189, opened 2021-09-10
Corrected the endpoint parameter name generation.

Jack12816:
**- What is it good for**

I've expected bad parameter documentation with grape-swagger when renaming with `as: ...`. The expected behavior is that the original parameter name is used for documentation. Let's look at this example:

```ruby
optional :address, type: Hash, as: :address_attributes do
  optional :street, type: String
  optional :street_addition, type: String
  optional :city, type: String
end 
```

I want my API clients to send the `address` parameter, and in my endpoint definition I want to update the entity (AR, `accepts_nested_attributes_for` enabled) by using `user.update!(declared(params, include_missing: false))`. Sure, it's just convenient to rename the parameter for internal use, but that's how it is documented. The resulting Swagger definition looks like this: 

This is clearly the internal name that should not be sent nor be documented. This pull request addresses the documentation issue.

**- What I did**

I just added the original parameter name when a renaming was done which is then used as the parent element name. Additionally, I added a bunch of examples to get this right. 

**- Open questions**

As you might see in the specs I encountered some cases that should behave differently. This boils down to the ability to bypass parameter validation for renamed scalar parameters when the client knows/guesses the internal name. The renaming is implemented right now as a validator which just adds the renamed parameter name into the params hash. When we then use the `declared(params, include_missing: false)` method in order to get only the declared and validated parameters we can be tricked. 

I was wondering why the renaming is not just part of the declared method implementation. This would prevent validation bypassing, while also allows us to ignore parameters we do not expect. So we could ensure the following scenarios:

```ruby
# (1) Rename a to b, while sending a
optional :a, type: Integer, as: :b
params = { a: 1 }
declared(params, include_missing: false)
# expected => { b: 1 }
# actual   => { b: 1 }

# (2) Rename a to b, while sending nothing
optional :a, type: Integer, as: :b
params = { }
declared(params, include_missing: false)
# expected => { }
# actual   => { }

# (3) Rename a to b, while sending b
optional :a, type: Integer, as: :b, values: [1, 2, 3]
params = { b: '5' }
declared(params, include_missing: false)
# expected => { }
# actual   => { b: '5' } (uncasted, unvalidated)
```

Is there something I'm missing here? I would like to address this too if the issue is reasonable.

images


Jack12816:
I spend some time on the parameter renaming just inside the `#declared(params)` method which works now quite fine. I put it in a derived branch from this at https://github.com/Jack12816/grape/tree/renaming. In this branch, the `AsValidator` is just a marker and does nothing anymore.

dblock:
I think you're spot on, and should fix the intended behavior as you described. Your proposed solution is also quite simple. Happy to review on green!

Jack12816:
> I think you're spot on, and should fix the intended behavior as you described. Your proposed solution is also quite simple.

Okay, great. I merged my renaming solution into this branch :) 

> Happy to review on green!
> test / test (ruby-head, true) (pull_request) Failing after 22s — test (ruby-head, true) 

This fails on master, too. [expected NoMethodError with "undefined method `not_exist_method'", got #<NoMethodError: undefined method `not_exist_method'](https://github.com/ruby-grape/grape/pull/2189/checks?check_run_id=3584396407#step:4:3420)

lib/grape/dsl/inside_route.rb
```diff
@@ -48,15 +48,20 @@ def declared_array(passed_params, options, declared_params, params_nested_path)
...
+          renamed_params = route_setting(:renamed_params) || {}
+
           declared_params.each_with_object(passed_params.class.new) do |declared_param, memo|
             if declared_param.is_a?(Hash)
               declared_param.each_pair do |declared_parent_param, declared_children_params|
                 params_nested_path_dup = params_nested_path.dup
                 params_nested_path_dup << declared_parent_param.to_s
                 next unless options[:include_missing] || passed_params.key?(declared_parent_param)
 
+                rename_path = params_nested_path.dup + [declared_parent_param.to_s]
```
dblock, line 60: Do we need to `.dup` it here? The `+` makes a copy I believe.

lib/grape/dsl/inside_route.rb
```diff
@@ -65,13 +70,13 @@ def declared_hash(passed_params, options, declared_params, params_nested_path)
             else
               # If it is not a Hash then it does not have children.
               # Find its value or set it to nil.
-              has_renaming = route_setting(:renamed_params) && route_setting(:renamed_params).find { |current| current[declared_param] }
-              param_renaming = has_renaming[declared_param] if has_renaming
+              next unless options[:include_missing] || passed_params.key?(declared_param)
 
-              next unless options[:include_missing] || passed_params.key?(declared_param) || (param_renaming && passed_params.key?(param_renaming))
+              rename_path = params_nested_path.dup + [declared_param.to_s]
```
dblock, line 75: Same

CHANGELOG.md
```diff
@@ -9,7 +9,9 @@
 * [#2176](https://github.com/ruby-grape/grape/pull/2176): Fix: OPTIONS fails if matching all routes - [@myxoh](https://github.com/myxoh).
 * [#2177](https://github.com/ruby-grape/grape/pull/2177): Fix: `default` validator fails if preceded by `as` validator - [@Catsuko](https://github.com/Catsuko).
 * [#2180](https://github.com/ruby-grape/grape/pull/2180): Call `super` in `API.inherited` - [@yogeshjain999](https://github.com/yogeshjain999).
+* [#2189](https://github.com/ruby-grape/grape/pull/2189): Corrected the endpoint parameter name generation - [@Jack12816](https://github.com/Jack12816).
```
dblock, line 12: Let's try something clearer? Maybe something like "Fix: rename parameters when using`:as`"?

lib/grape/validations/params_scope.rb
```diff
@@ -23,13 +25,14 @@ class ParamsScope
...
       def initialize(opts, &block)
-        @element      = opts[:element]
-        @parent       = opts[:parent]
-        @api          = opts[:api]
-        @optional     = opts[:optional] || false
-        @type         = opts[:type]
-        @group        = opts[:group] || {}
-        @dependent_on = opts[:dependent_on]
+        @element          = opts[:element]
+        @rename           = opts[:rename]
```
dblock, line 29: Rename is a strange variable since it's a verb. How about just `name`?

dblock (changes requested):
This is excellent work. thank you.

I have some nits. Plus this changes behavior that I believe we need to document in UPGRADING, please?

lib/grape/validations/params_scope.rb
Jack12816: Yeah, but `name` is ambiguous with `element` because `element` holds the actual name of the param scope container name. So let's call it `element_renamed`, this causes less confusion I would argue.

lib/grape/dsl/inside_route.rb
Jack12816: You're right. This is a leftover of fiddling around and I wanted to be safe to not touch any state here. Fixed it.

Jack12816:
> This is excellent work. thank you.

Thanks :) 

> I have some nits. Plus this changes behavior that I believe we need to document in UPGRADING, please?

I took care of the nits and added an upgrading section. I would not recommend releasing this change as a patch version because it breaks current behavior, even though it's a bug fix in the narrower sense.

UPGRADING.md
```diff
@@ -1,10 +1,64 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.5.4
```
dblock, line 4: Unwrap lines, we github markdown here.

UPGRADING.md
```diff
@@ -1,10 +1,64 @@
...
+renaming](https://github.com/ruby-grape/grape#renaming) with `:as` was directly
+touching the request payload
+([`#params`](https://github.com/ruby-grape/grape#parameters)) while duplicating
+the old and the new key to be both available in the hash. This allowed clients
+to bypass any validation in case they knew the internal name of the parameter.
+Unfortunately, in combination with
+[grape-swagger](https://github.com/ruby-grape/grape-swagger) the internal name
+(name set with `:as`) of the parameters were documented.
+
+From now on the parameter renaming is not anymore done automatically in the
```
dblock, line 18: While it does change behavior, I think the old behavior is broken. Let's say "This behavior was fixed. Parameter renaming is now done when ..."

UPGRADING.md
```diff
@@ -1,10 +1,64 @@
...
+
+# (2) Rename a to b, while client sends +b+
+optional :a, type: Integer, as: :b, values: [1, 2, 3]
+params = { b: '5' }
+declared(params, include_missing: false)
+# expected => { }        (>= 1.5.4)
+# actual   => { b: '5' } (uncasted, unvalidated, <= 1.5.3)
+```
+
+Another implication of this is the dependent parameter resolution. Prior to
```
dblock, line 41: of this change ...

dblock (approved):
Agreed on the minor bump. Let's bump the version in this PR to 1.6 as well, please. Don't forget to re-update the numbers in UPGRADING.

UPGRADING.md
Jack12816: Will do if you like - I'm a 80 chars person :D

Jack12816:
> Let's bump the version in this PR to 1.6 as well, please. Don't forget to re-update the numbers in UPGRADING.

Did it :)

dblock:
Merged, thank you!

Jack12816:
What do you think about a release in the next few days? :)

dblock:
> What do you think about a release in the next few days? :)

I'll get to it sooner than later unless another maintainer can beat me to it.

Jack12816:
ping @dblock - is there anything blocking a release? :)

dblock:
Released 1.6.0

PR #2190, opened 2021-09-26
upgrade Rubocop & drop Ruby 2.4.x support

dnesteryuk:
To use latest Rubocop we need to drop support for Ruby 2.4.x.

Also, this change fixes some offenses with auto correction.


dblock:
Should we bump version to 1.6.0?

PR #2192, opened 2021-10-04
Memoize the result of Grape::Middleware::Base#response.

Jack12816:
**- What is it good for**

When using the `Grape::Middleware::Base#response` method multiple times from within a custom middleware, the method generates a `Rack::Response` instance each (multiple) time.

[Do not convert Rack::Response to Rack::Response](https://github.com/ruby-grape/grape/commit/7ad647a873b666ad90ebd512c58d037db2f6148e) was missing the memoization I think, because [Don't create Grape::Request multiple times](https://github.com/ruby-grape/grape/commit/22b3aefda6ef31cda2379b3d80ab8299adddd27a#diff-0575f1465bc0e57934dfcda6d3c6b81846e771a1bfce6774b1302787a8e68468) already added memoization for `#request`.

**- What I did**

While writing a simple middleware that should append a response header I hit an unexpected behavior while using the `Grape::Middleware::Base#response` helper. Example time:

```ruby
class ResponseRoute < Grape::Middleware::Base
  # (1) Works with a Grape generated response,
  #     but not with a Rails generated response (looks like they aren't
  #     sharing the Rack::Utils::HeaderHash instance)
  def after
    response['X-Debug'] = 'true'
    response
  end

  # (2) Works with a Grape and Rails generated response 
  def after
    response.tap do |res|
      res['X-Debug'] = 'true'
    end
  end
end
```

After my fix, both versions will work as expected.

images

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#2192](https://github.com/ruby-grape/grape/pull/2192): Memoize the result of Grape::Middleware::Base#response - [@Jack12816](https://github.com/Jack12816).
 * Your contribution here.
 
 ### 1.6.0 (2021/10/04)
--- a/lib/grape/middleware/base.rb
+++ b/lib/grape/middleware/base.rb
@@ -60,7 +60,7 @@ module Grape
       def response
         return @app_response if @app_response.is_a?(Rack::Response)
 
-        Rack::Response.new(@app_response[2], @app_response[0], @app_response[1])
+        @app_response = Rack::Response.new(@app_response[2], @app_response[0], @app_response[1])
       end
 
       def content_type_for(format)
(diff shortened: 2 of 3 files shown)


Jack12816:
@dblock could you add the `hacktoberfest-accepted` label? :upside_down_face:

dblock:
> @dblock could you add the `hacktoberfest-accepted` label? 🙃

What is this? :)

Jack12816:
It's https://hacktoberfest.digitalocean.com/ - a monthlong "hackathon" where the people are encouraged to contribute to open-source projects. :) 

Yeah, and I thought you could add Grape as a participating project or label my PR's individually to count for me :)

dblock:
Done!

Jack12816:
Awesome! Thank you :relaxed:

PR #2196, opened 2021-11-01
allow set passwords_hashed option for digest auth

lHydra:
Rack Digest Auth [allows](https://github.com/rack/rack/blob/e0993d47096d36535dbe725091963a7e0f5aab52/lib/rack/auth/digest/md5.rb#L123) to check the user's credentials with ha1 digest hash instead of check with a plain password. But at the moment this functionality is not available because the condition from `Rack::Auth::Digest::MD5` is always false

```ruby
if opaque.nil? and realm.respond_to? :values_at
  realm, opaque, @passwords_hashed = realm.values_at :realm, :opaque, :passwords_hashed
end
```
(in the grape auth dsl we constantly set the option `opaque` - `options[:opaque] ||= 'secret'`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3297,9 +3297,17 @@ http_basic do |username, password|
 end
 ```
 
+Digest auth supports clear-text passwords and password hashes.
+
 ```ruby
 http_digest({ realm: 'Test Api', opaque: 'app secret' }) do |username|
   # lookup the user's password here
+end
+```
+
+```ruby
+http_digest(realm: { realm: 'Test Api', opaque: 'app secret', passwords_hashed: true }) do |username|
+  # lookup the user's password hash here
 end
 ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#2196](https://github.com/ruby-grape/grape/pull/2196): Add support for `passwords_hashed` param for `digest_auth` - [@lHydra](https://github.com/lhydra).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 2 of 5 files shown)


dblock:
Needs a `rubocop -a`, sorry ;)

lHydra:
> Needs a `rubocop -a`, sorry ;)

fixed

README.md
```diff
@@ -3297,11 +3297,23 @@ http_basic do |username, password|
...
+
+* Compare with the pure password
+
+    ```ruby
+    http_digest({ realm: 'Test Api', opaque: 'app secret' }) do |username|
+      # lookup the user's password here
+    end
+    ```
+
+* If you want to storing the ha1 hash: `MD5(username:realm:password)` and compare with it
```
dblock, line 3310: `to storing` -> `to store`
`ha1` -> `sha` 

The example uses MD5 and not a SHA1. I would remove the bullets and just provide the two examples with my suggested text above.

README.md
```diff
@@ -3297,11 +3297,23 @@ http_basic do |username, password|
...
 ```
 
-```ruby
-http_digest({ realm: 'Test Api', opaque: 'app secret' }) do |username|
-  # lookup the user's password here
-end
-```
+For Digest Auth
+
+* Compare with the pure password
```
dblock, line 3302: `pure` -> `clear text`

"Digest auth supports clear-text passwords and password hashes."

... then just provide the two examples without any other text?

dblock (changes requested):
Typos in README.

README.md
lHydra: I agree with you, it sounds better =)

dblock (approved):

PR #2199, opened 2021-11-23
Expand test coverage of undocumented functionality

jacobherrington:
In an interview recently, I was given the Grape repository sans the
content of this method as a technical screening question. The task was
to discover what had been removed from the project and re-implement the
missing code using only the test suite.

It was actually fairly difficult to re-implement this method because
there are a couple of untested responsibilities. I think this is a
fairly integral component of the framework, so it makes sense to fully
test its functionality.

### DIFF
--- a/lib/grape/dsl/headers.rb
+++ b/lib/grape/dsl/headers.rb
@@ -3,8 +3,11 @@
 module Grape
   module DSL
     module Headers
-      # Set an individual header or retrieve
-      # all headers that have been set.
+      # This method has four responsibilities:
+      # 1. Set a specifc header value by key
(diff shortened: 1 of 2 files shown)


jacobherrington:
I see this as improving documentation, but I'm happy to add a changelog entry if necessary.

dblock:
Thanks for the test!

dblock:
Pretty wild interview test idea if you ask me, btw! How did you do @jacobherrington?

jacobherrington:
@dblock I actually really liked it! I did well enough, but joined another company! (check my account 😉)

PR #2200, opened 2021-12-05
Add validators module to all validators

ericproulx:
Hello,

While looking at the "validations/validators" folder, I found that 'validators" is not a declared module and I was wondering why. I looked at the "validations/types" and "types" is a declared module. Anyway, I thought that for consistency, it had to be done.

There were also a lot of [LeakyConstantDeclaration](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/LeakyConstantDeclaration) inside validators specs. I figured it would be better to use AnonymousClasses instead. Unfortunately, declaring an anonymous Grape::API for every single test was slowing down everything and I thought using `let_it_be` from [test-prof](https://github.com/test-prof/test-prof/blob/master/lib/test_prof/recipes/rspec/let_it_be.rb) would be ideal.

I think rubocop-rspec could be a great addition for the spec. I could open a PR if you want.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 * [#2193](https://github.com/ruby-grape/grape/pull/2193): Fixed the broken ruby-head NoMethodError spec - [@Jack12816](https://github.com/Jack12816).
 * [#2192](https://github.com/ruby-grape/grape/pull/2192): Memoize the result of Grape::Middleware::Base#response - [@Jack12816](https://github.com/Jack12816).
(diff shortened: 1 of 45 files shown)


dblock:
Good stuff.

+1 on rubocop-rspec, please

dm1try:
hmm, it's a breaking change for custom validators so `UPGRADING.md` should be updated.

Jack12816:
breaking change packaged as patch version. :(

dblock:
> hmm, it's a breaking change for custom validators so `UPGRADING.md` should be updated.

Lessons learned here, we should have made a major version bump or built backwards compat, see #2244

PR #2201, opened 2021-12-05
Replace let_it_be by before :all

ericproulx:
Hello,

`let_it_be` seams fleaky in the test suite. I removed it and use before :all instead


dblock:
Looks like a legit build failure in https://github.com/ruby-grape/grape/runs/4422845217?check_suite_focus=true

ericproulx:
I can reproduce every time with the same seed. Another test is changing something. I'll find it. Do you want to keep TestProf ? Seems more ruby than using `before :all` with instance variable.

ericproulx:
It feels like the global settings somehow is not resetted even with new instance.

I solved the problem by calling ::Logger.new instead of double
Screen Shot 2021-12-05 at 17 21 25

ericproulx:
Looks like truffle-ruby doesnt like unicode in instance_variable

https://github.com/ruby-grape/grape/runs/4423127949?check_suite_focus=true

"@😸" is let_it_be prefix.

dblock:
I don't care as much about truffle ruby, but this does need to be rebased and any other legit build failure fixed.

Btw, no need to include in CHANGELOG any changes that were added/removed within the release. Like when you add a feature, then fix a bug in that feature in the same release there's no need to changelog that you fixed a bug as it was never shipped.

PR #2207, opened 2021-12-25
Validations Autoload

ericproulx:
Hello,

I saw that almost all classes/modules were autoloaded but the `Validations` tree wasn't. Anyway, here's a PR with the autoloading. Merry Xmas! 🎄


dblock:
Add to CHANGELOG please?

There's a couple of bugs that mention auto-load, does any of this fix that? https://github.com/ruby-grape/grape/issues?q=is%3Aissue+validations+auto-load+is%3Aopen

ericproulx:
> Thank you! Add to CHANGELOG please?
> 
> There's a couple of bugs that mention auto-load, does any of this fix that? https://github.com/ruby-grape/grape/issues?q=is%3Aissue+validations+auto-load+is%3Aopen

I'm not sure it will fix the issues. I think the best way to load any custom validators would be through the `Grape::Validations.register_validator` function which is used within specs. I've discovered that function through this PR and it's not mention in the documentation. 

Maybe having a lazy load hook would help to clarify when to load custom validators. Something like:

```ruby
# /initializers/grape.rb
ActiveSupport.on_load(:grape). do # would need a 
  require 'my_custom_validator'
  Grape::Validations.register_validator('my_custom_validator', MyCustomValidator)
end
```

dblock:
@ericproulx I think the JRuby failure is related to this:

```
Run bundle exec rake spec
/home/runner/.rubies/jruby-head/bin/jruby -I/home/runner/work/grape/grape/vendor/bundle/jruby/3.0.0/gems/rspec-core-3.10.1/lib:/home/runner/work/grape/grape/vendor/bundle/jruby/3.0.0/gems/rspec-support-3.10.3/lib /home/runner/work/grape/grape/vendor/bundle/jruby/3.0.0/gems/rspec-core-3.10.1/exe/rspec --pattern spec/\*\*/\*_spec.rb  --exclude-pattern spec/integration/\*\*/\*_spec.rb
/home/runner/.rubies/jruby-head/lib/ruby/stdlib/time.rb:3: warning: parentheses after method name is interpreted as an argument list, not a decomposed argument

An error occurred while loading ./spec/grape/api/custom_validations_spec.rb. - Did you mean?
                    rspec ./spec/grape/exceptions/validation_spec.rb
                    rspec ./spec/grape/dsl/validations_spec.rb
                    rspec ./spec/grape/validations_spec.rb***
```

Check out the truffleruby and irb head failures too, maybe easy to fix.

dm1try:
> Something like:
Grape::Validations.register_validator('my_custom_validator', MyCustomValidator)

I don't recommend doing this while we have the same call [in `inherited` callback in Base class](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/base.rb#L73-L77)

PR #2209, opened 2021-12-27
Autoload Types

ericproulx:
Hello

This PR is Like #2207 but for Types

Thanks


dblock:
Would you please check the CI failures on those optional rubies? I think the jruby one is caused by #2207

ericproulx:
@dblock I've installed JRuby-9.3.2.0 locally and everything works fine. Current JRuby build is failing. It might be related.

dblock:
I opened https://github.com/dry-rb/dry-container/issues/81 to track the dry-types/dry-container problem.

PR #2213, opened 2021-12-28
require 'active_support/concern'

ericproulx:
Since `require 'active_support/concern'` was called many times, I thought it would be cleaner to just require it once in the grape.rb. I also renamed `MissingGroupType` and `UnsupportedGroupType` to avoid including path in autoload and I moved `require 'active_support/core_ext/string/output_safety'` in grape.rb since now all `active_support` requires are in this file.


dblock:
Looks good. Needs CHANGELOG and possibly en entry in UPGRADING if someone relies on these exceptions?

dblock:
@ericproulx Want to finish this?

ericproulx:
I will reopen the PR, it will less work

PR #2216, opened 2021-12-29
remove eager_load from test env

dm1try:
as we have [a separated spec](https://github.com/ruby-grape/grape/blob/master/spec/integration/eager_load/eager_load_spec.rb) for this case

partially reverts 752bf4c0a

ref [#2214](https://github.com/ruby-grape/grape/issues/2214#issuecomment-1002608359)


dblock:
Would love an explanation of why this fixes that and a spec that shows that it does, please?

dm1try:
> Would love an explanation of why this fixes that and a spec that shows that it does, please?

it does not fix anything it just shows that the lib does not load some files when `eager_load` is disabled(what's usually the case in dev environment)
it is also shows that a separated spec has a false positive result, as we eager load the whole lib code in the spec helper.

so, the pr is just the reply for your https://github.com/ruby-grape/grape/issues/2214#issuecomment-1002608359

dblock:
Closed via https://github.com/ruby-grape/grape/pull/2222

Issue #2221, opened 2021-12-31
undefined method `multiple?' for Grape::Validations::Types:Module


cristofer:
Hey guys,

`build(deps): bump grape from 1.6.0 to 1.6.1`

This PR https://github.com/ruby-grape/grape/pull/2209 introduced this check https://github.com/ruby-grape/grape/pull/2209/files#diff-e97e1900db45e8b2dbf58e1a31c80ca18d47fe7d7a2ae27a6c98a076c6936b18R177, which breaks in our CI:

```
/home/circleci/repo/vendor/bundle/ruby/2.6.0/gems/grape-1.6.1/lib/grape/validations/params_scope.rb:359:in `infer_coercion'
/home/circleci/repo/vendor/bundle/ruby/2.6.0/gems/grape-1.6.1/lib/grape/validations/params_scope.rb:274:in `validates'
/home/circleci/repo/vendor/bundle/ruby/2.6.0/gems/grape-1.6.1/lib/grape/validations/params_scope.rb:195:in `validate_attributes'
/home/circleci/repo/vendor/bundle/ruby/2.6.0/gems/grape-1.6.1/lib/grape/dsl/parameters.rb:135:in `requires'
```

I went through the loads specified here https://github.com/ruby-grape/grape/pull/2209/files#diff-b56aab6c58cef370018ab45d697b38247b363bf034afcd2e4b4d9fff05c3de0eR230 , and indeed couldnt find anything that could make a reference to `multiple?`

Please let me know if you need anything else in order to reproduce the problem.

Thank you for your great work!

cristofer:
NVM, I just saw the fix here https://github.com/ruby-grape/grape/pull/2219, thanks for the quick reaction to that

dblock:
@dm1try should we yank 1.6.1 from rubygems? is everybody running into this?

dm1try:
@dblock maybe, it's a new experience :)
though,  the upgrade to 1.6.2 should work for most users.

Issue #2218, opened 2021-12-30
Issues after upgrading to 1.6.1 - `undefined method 'multiple?'`


sharq1:
After upgrading to 1.6.1 we started getting this when hen starting the Rails server:
`undefined method `multiple?' for Grape::Validations::Types:Module`

<details>
  <summary>Full stack trace</summary>
  
  ```ruby
Traceback (most recent call last):
        126: from bin/rails:9:in `<main>'
        125: from /home/sharq/.rbenv/ruby/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:31:in `require'
        124: from /home/sharq/.rbenv/ruby/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require_with_bootsnap_lfi'
        123: from /home/sharq/.rbenv/ruby/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/loaded_features_index.rb:100:in `register'
        122: from /home/sharq/.rbenv/ruby/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `block in require_with_bootsnap_lfi'
        121: from /home/sharq/.rbenv/ruby/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `require'
        120: from /home/sharq/.rbenv/ruby/2.7.0/gems/railties-6.1.4.4/lib/rails/commands.rb:18:in `<main>'
        119: from /home/sharq/.rbenv/ruby/2.7.0/gems/railties-6.1.4.4/lib/rails/command.rb:48:in `invoke'
        118: from /home/sharq/.rbenv/ruby/2.7.0/gems/railties-6.1.4.4/lib/rails/command/base.rb:69:in `perform'
        117: from /home/sharq/.rbenv/ruby/2.7.0/gems/thor-1.1.0/lib/thor.rb:392:in `dispatch'
        116: from /home/sharq/.rbenv/ruby/2.7.0/gems/thor-1.1.0/lib/thor/invocation.rb:127:in `invoke_command'
        115: from /home/sharq/.rbenv/ruby/2.7.0/gems/thor-1.1.0/lib/thor/command.rb:27:in `run'
[...]

PR #2222, opened 2022-01-01
Fix autoload types and validators

ericproulx:
Hi, 

I've found a solution for the #2218 AND #2221. This PR adds a `lazy_find_validator` function that loads Grape's validators when needed. That way, we don't have to require all of them since the autoload doesn't trigger the `inherited` function.

Also, `ArrayCoercer` and `SetCoercer` were registering a `DryTypeCoercer` collection through inheritance. I had to find another way to register the collection. Therefore, I applied the same technique.

Finally, that way, we can remove safely the `eager_load!` #2216  in the `spec_helper`. I ran several specs in our stack and I haven't found any issues.

Happy New Year 🍾


lib/grape/validations.rb
```diff
@@ -9,22 +9,32 @@
...
       validators[short_name] = klass
     end
 
-    def self.deregister_validator(short_name)
+    def deregister_validator(short_name)
       validators.delete(short_name)
     end
+
+    # Find a validator and if not found will try to load it
+    def lazy_find_validator(short_name)
```
dm1try, line 31: I would name it `load_validator` or `require_validator`

lib/grape/validations.rb
```diff
@@ -9,22 +9,32 @@
...
+      str_name = short_name.to_s
+      validators.fetch(str_name) do
+        register_validator(str_name, "Grape::Validations::Validators::#{str_name.camelize}Validator".constantize)
```
dm1try, line 34: `register_validator` will be called twice for this constant(an additional call will be in `inherited` callback)

instead of it we can use a lookup for the constant like
```ruby
validators.fetch(str_name) do
  Grape::Validations::Validators.const_get("{str_name.camelize}Validator")
end

```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#2222](https://github.com/ruby-grape/grape/pull/2222): Fix autoload types and validators - [@ericproulx](https://github.com/ericproulx).
```
dm1try, line 9: "Fix" is not needed here.

This is more like "a second attempt of autoload types and validators" :) but it would be enough just remove "Fix"

lib/grape/validations/types.rb
```diff
@@ -22,56 +13,37 @@ module Validations
     # and {Grape::Dsl::Parameters#optional}. The main
     # entry point for this process is {Types.build_coercer}.
     module Types
-      # Types representing a single value, which are coerced.
-      PRIMITIVES = [
```
dm1try, line 26: While this is for "internal use" it was available as a public constant, so it might be a breaking change for someone.

dm1try (commented):
Do I understand right that the original PRs #2207 #2209 solve some "consistency problem" where some part of the code is autoloaded and some part is not? without some numbers, it is unclear whether this is a problem at all.

but I like the part with autoloading validators based on `:short_name`, it echoes to the first part of #2204(though it's not enough for proper integration with Rails for example)

dblock:
> Do I understand right that the original PRs #2207 #2209 solve some "consistency problem" where some part of the code is autoloaded and some part is not? without some numbers, it is unclear whether this is a problem at all.

i don’t think it’s a problem either way; if my memory serves me correctly there were issues trying to move validators to auto loading and so this was left alone - sounds like we don’t have any issues with going one way or another, I vote to move to auto loading for consistency

dm1try:
> I vote to move to auto loading for consistency to auto loading for consistency
 
 that works for me

spec/spec_helper.rb
```diff
@@ -23,8 +23,6 @@ def rollback_transaction; end
   require file
 end
 
-eager_load!
```
dm1try, line 26: the related file in which the implementation for this method lives should be removed too

dm1try: see #2216

dm1try (changes requested):
So, take a look to the comments and rebase the branch(we should not have an intermediate "merge master" commits here). Thanks!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#2222](https://github.com/ruby-grape/grape/pull/2222): autoload types and validators - [@ericproulx](https://github.com/ericproulx).
```
dblock, line 9: Capitalize Autoload

dblock (changes requested):

dm1try (approved):
The changes looks good to me.
though, the commits history looks sloppy :) I can help to fix it(later tonight).

dblock:
> The changes looks good to me. though, the commits history looks sloppy :) I can help to fix it(later tonight).

I just squash-merged, thanks @ericproulx for the code, and @dm1try for a thorough code review.

PR #2229, opened 2022-01-04
do not collect params in route settings

dnesteryuk:
While investigating another issue I discovered that params were documented twice then [merged there](https://github.com/ruby-grape/grape/blob/5053b797429279d91818e2116034ba40636338bb/lib/grape/dsl/routing.rb#L132-L134).

Params are stored in `namespace_stackable` there is no reason to keep them in `route_setting(:description)` too.

I checked this change against our app, it still works :wink:

### DIFF
--- a/benchmark/nested_params.rb
+++ b/benchmark/nested_params.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', 'lib'))
 require 'grape'
 require 'benchmark/ips'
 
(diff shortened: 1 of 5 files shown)


dm1try:
maybe some external gem uses this interface(like `grape-swagger` for example)?

benchmark/nested_params.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', 'lib'))
```
dblock, line 3: Is this necessary? :)

dblock (changes requested):

dnesteryuk:
@dm1try we use `grape-swagger` in our project, it works as it worked before :slightly_smiling_face:

benchmark/nested_params.rb
dnesteryuk: unfortunately, yes. I took it from https://github.com/ruby-grape/grape/blob/master/benchmark/compile_many_routes.rb

dm1try (approved):

dblock (approved):

PR #2233, opened 2022-01-08
a setting for disabling documentation to internal APIs

dnesteryuk:
Our application has 3 APIs, 2 of them are internal and don't require documentation. However, there has not been any option to prevent Grape from documenting parameters for internal APIs.

This change adds a `do_not_document!` setting which instructs Grape to not document parameters, thus needless objects allocation is avoided.

```ruby
class Api < Grape::API
  do_not_document!
end
```

The logic for documenting parameters was moved to a separate class, the `Grape::Validations::ParamsScope` class has to many responsibilities, it would be better to split it.

Our API has lots of endpoints like this:

```ruby
resource :addresses do
  params do
    requires :street, type: String
    requires :postal_code, type: Integer
    optional :city, type: String
  end
  post '/' do
    declared(params, include_missing: false)
    'Done'
  end
end
```

**With documentation:**

```
Total allocated: 342247 bytes (3186 objects)
Total retained:  70915 bytes (594 objects)
```

**Without documentation:**

```
Total allocated: 341183 bytes (3178 objects)
Total retained:  69091 bytes (580 objects)
```

1,78 Kb is saved.


dblock (approved):
I'm good with this, @dmitry care to CR too?

dm1try (approved):
LGTM
though, some of those changes are refactoring not related to documentation, I would prefer to have separate commits for that next time.

PR #2234, opened 2022-01-11
Remove non-UTF-8 characters from format before generating JSON error

bschmeck:
If we receive a request with an unsupported format, that user-provided format is included in the generated error message.  If the format includes non-UTF-8 characters and the app is configured to generate JSON error messages, a `JSON::GeneratorError` is raised.

To work around this error, this PR uses `String#encode`'s ability to replace invalid and undefined UTF-8 characters before calling `Grape::Json::dump`.


dm1try (approved):

bschmeck:
Thanks for the quick merge, and all of your work on Grape!

Issue #2230, opened 2022-01-04
1.6.0 -> 1.6.1 contains breaking change


ro-savage:
Our automated tests are showing

```
rake aborted!
NameError: uninitialized constant Grape::Validations::Base
```

It appears this commit https://github.com/ruby-grape/grape/commit/937b2fc8a7081c5eb74ff8a4a760e30ebabbd780 changed the path from `Grape::Validations::Base` to `Grape::Validations::Validators:Base`

dm1try:
yeah, sorry for that. 

sadly, after the release I [noticed](https://github.com/ruby-grape/grape/pull/2200#issuecomment-1003264156) that the `UPGRADING.md` had not been updated in #2200. so I didn't expect any breaking change from such refactoring.

I'm closing, there is nothing we can do here.

dm1try:
though, we can create a patch for compatibility with 1.6.0 and up a minor version
```ruby
Grape::Validations::Base = Grape::Validations::Validators:Base
```

ro-savage:
@dm1try - We've updated our codebase to use `Grape::Validations::Validators:Base`
But I think it make sense to release `1.6.3` that contains this - so that the latest 1.6.x isn't a breaking change from 1.6.0

dblock:
> @dm1try - We've updated our codebase to use `Grape::Validations::Validators:Base`
> But I think it make sense to release `1.6.3` that contains this - so that the latest 1.6.x isn't a breaking change from 1.6.0

This was a miss. Care to PR this fix @ro-savage ?

ro-savage:
@dblock - Unfortunately I am not a ruby developer, I actually wouldn't know where to put the alias.

Jack12816:
:( Breaking change in a patch version. This does not only affect `Grape::Validations::Base`, but also users who extended existing built-in validators or compositions of it. This is even worse for external package maintainers who rely on the constants.

dblock:
Agreed. Can someone please PR a fix? @dm1try ?

dm1try:
@dblock take a look at #2244

PR #2244, opened 2022-02-17
fix a breaking change provided in 1.6.1

dm1try:
ref #2230 #2200

### DIFF
--- a/lib/grape/validations/validators/base.rb
+++ b/lib/grape/validations/validators/base.rb
@@ -93,3 +93,10 @@ module Grape
     end
   end
 end
+
+Grape::Validations::Base = Class.new(Grape::Validations::Validators::Base) do
+  def initialize(*)
+    super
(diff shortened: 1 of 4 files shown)


dblock:
Because this was a regression let's add it to CHANGELOG?

dblock (approved):

lib/grape/validations/validators/base.rb
```diff
@@ -93,3 +93,10 @@ def fail_fast?
     end
   end
 end
+
+Grape::Validations::Base = Class.new(Grape::Validations::Validators::Base) do
+  def initialize(*)
+    super
+    warn '[DEPRECATION] `Grape::Validations::Base` is deprecated. Use `Grape::Validations::Validators::Base` instead.'
```
ericproulx, line 100: I'll add a PR for adding a warn message to MissingGroupType and UnsupportedGroupType

PR #2249, opened 2022-02-27
Split CI matrix, extract edge.

dblock:
GitHub CI doesn't have a feature to ignore certain steps in CI, so it looks like Grape build is continuously broken. This isn't the case. This PR extracts edge CI into a separate workflow that only runs on pull requests.

### DIFF
--- a/.github/workflows/edge.yml
+++ b/.github/workflows/edge.yml
@@ -1,0 +1,35 @@
+---
+name: edge
+on:
+  pull_request:
+    branches:
+      - "*"
+jobs:
+  test:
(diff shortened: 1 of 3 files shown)


dblock:
@dm1try WDYT?

dm1try:
@dblock it totally makes sense to me. let's add this ref https://github.com/actions/toolkit/issues/399 also.

and sorry, but I'm currently not in "working conditions" because of the war in Ukraine, so I'm not going to respond/interact for some time.

dm1try (approved):

PR #2265, opened 2022-06-14
Use COVERALLS_PARALLEL as recommended to combine coverage reports.

dblock:
Let's see if this does anything? https://docs.coveralls.io/parallel-build-webhook

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,6 +60,7 @@ jobs:
     env:
       BUNDLE_GEMFILE: ${{ matrix.gemfile }}
       COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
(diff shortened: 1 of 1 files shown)


duffn:
If this doesn't work, it could be worthwhile to explore using the Coveralls GitHub Action (https://github.com/marketplace/actions/coveralls-github-action) instead. Both Coveralls and the `coveralls-reborn` package itself even recommend it https://github.com/tagliala/coveralls-ruby-reborn#github-actions

dblock:
@duffn I think this doesn't work and we should get closer to https://github.com/tagliala/coveralls-ruby-reborn#github-actions (just minus Rails). The GHA app is authorized, so it will also work on pull requests (no secret key required).

duffn:
https://github.com/ruby-grape/grape/pull/2266

PR #2266, opened 2022-06-15
Use Coveralls GitHub Action

duffn:
This uses the Coveralls GitHub Action in place of `coveralls-reborn` as recommended in the gem README https://github.com/tagliala/coveralls-ruby-reborn#github-actions and as discussed here https://github.com/ruby-grape/grape/pull/2265. The parallel test steps are taken from the Action README https://github.com/coverallsapp/github-action

Examples on my fork: 
- Job for a merge showing individual Ruby version & Gemfile matrix https://coveralls.io/builds/50032540
- Master after a merge https://coveralls.io/github/duffn/grape

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,1 +1,1 @@
-service_name: github-actions
+service_name: github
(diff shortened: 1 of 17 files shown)


dblock:
Much better! Add a changelog with something like “Fixed code coverage” and we are good to go.

PR #2271, opened 2022-07-21
Numeric is not a dry-types built-in type, fix coercion/validation

Vasfed:
In 1.3+ if someone had a `Numeric` type parameter which worked in `<1.3` - any valid non-blank value will silently fail validation
```ruby
params { optional :price, type: Numeric }
```

This happens because dry-types does not contain built-in type for `DryTypes::SomeScope::Numeric` and `const_get` by default searches ancestors, thus finding `::Numeric`.
Runtime error ``undefined method `[]' for Numeric:Class`` is treated as a validation error and leads to unexpected behavior.

This PR adds mapping for Numeric that fixes it, also fallback to `const_get(.., false)` will make similar issues easier to debug.

### DIFF
--- a/lib/grape/validations/types/primitive_coercer.rb
+++ b/lib/grape/validations/types/primitive_coercer.rb
@@ -12,6 +12,7 @@ module Grape
         MAPPING = {
           Grape::API::Boolean => DryTypes::Params::Bool,
           BigDecimal => DryTypes::Params::Decimal,
+          Numeric => DryTypes::Params::Integer | DryTypes::Params::Float | DryTypes::Params::Decimal,
 
           # unfortunately, a +Params+ scope doesn't contain String
           String => DryTypes::Coercible::String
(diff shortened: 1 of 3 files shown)


dblock:
Right on. The change to not lookup inherited types in `const_get` seems to break a bunch of other stuff, though.

```
 30) Grape::Endpoint#declared when the param is missing and include_missing=false sets nested objects to be nil
      Failure/Error: MAPPING.fetch(type) { scope.const_get(type.name, false) }
      NameError:
        uninitialized constant #<Module:0x00005598ebeb08a0>::Set
```

Vasfed:
Yep, reverted the `const_get` change, it's not that important for the fix

lib/grape/validations/types/primitive_coercer.rb
```diff
@@ -12,25 +12,29 @@ class PrimitiveCoercer < DryTypeCoercer
...
+          BigDecimal => DryTypes::Strict::Decimal,
+          Numeric => DryTypes::Strict::Integer | DryTypes::Strict::Float | DryTypes::Strict::Decimal
         }.freeze
 
         def initialize(type, strict = false)
           super
 
           @type = type
 
+          # TODO: use const_get(.., false) to prevent vague errors on unknown types, but it breaks a lot of other stuff
```
dblock, line 32: Can we just revert this part of the change and open a new issue on GH please?

Vasfed: Done

dblock (approved):

dblock:
Thanks! Merged.

PR #2272, opened 2022-07-22
Check for dry-types type existence and prevent unexpected top level const lookup

Vasfed:
This is a follow-up to https://github.com/ruby-grape/grape/pull/2271 - prevent silent validation fails for types not built-in in dry-types.
Instead of validation fail - check for `[]` method presence in provided class at init time.


dblock:
Solid! fix Rubocop and it's good with me.

Vasfed:
> Solid! fix Rubocop and it's good with me.

Done

PR #2274, opened 2022-07-30
Error middleware support using rack util's symbols as status

dhruvCW:
support the ability to use the Rack::Utils's status_code symbols instead of integers.


dblock:
I don't see why not. Update CHANGELOG and we can merge.

dblock (changes requested):
Looking at this more carefully, I saw that we were documenting that you could set status with a code in https://github.com/ruby-grape/grape#http-status-code, and we have specs that show you can do this [here](https://github.com/ruby-grape/grape/blob/master/spec/grape/api_spec.rb#L3042).

```ruby
      it 'can be set with a status code symbol' do
        subject.get '/foo' do
          status :see_other
        end
        get '/foo'
        expect(last_response.status).to eq 303
      end
```

So why do we need this? Or at least should we be removing the call in https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/inside_route.rb#L196?

dhruvCW:
The reason is so the status code can be a symbol when using the error middleware to generate an error like so [error!(response, :forbidden)](https://github.com/ruby-grape/grape/blob/9bce86f554ee058d73efeb00dea5aeff136f8180/lib/grape/middleware/error.rb#L53)

dblock (approved):

dblock:
Ok, my only concern was if we call `Rack::Utils.status_code` twice all the time, but it doesn't look like it. Merged.

PR #2276, opened 2022-08-09
Fix exception super

### LINKED ISSUE #2273: Grape:Exceptions::Base missing super?
Hello,

While looking at some exceptions in our logs, I found that the attribute `message` wasn't shown when doing `to_s`. For instance,
```ruby
e = Grape::Exceptions::EmptyMessageBody.new('my message')
e.to_s =  "Grape::Exceptions::EmptyMessageBody"
```
When reading about [Exception](https://ruby-doc.org/core-2.5.1/Exception.html)
```
to_s → string
Returns exception's message (or the name of the exception if no message is set).
```

If we add `super(message)` in missing in Grape::Exceptions::Base, it works as expected

I'll open a PR

Thanks.

ericproulx:
Fix  #2273

### DIFF
--- a/lib/grape/exceptions/base.rb
+++ b/lib/grape/exceptions/base.rb
@@ -7,11 +7,10 @@ module Grape
       BASE_ATTRIBUTES_KEY = 'grape.errors.attributes'
       FALLBACK_LOCALE = :en
 
-      attr_reader :status, :message, :headers
+      attr_reader :status, :headers
 
       def initialize(status: nil, message: nil, headers: nil, **_options)
         @status  = status
-        @message = message
         @headers = headers
         super(message)
       end
--- a/lib/grape/exceptions/base.rb
+++ b/lib/grape/exceptions/base.rb
@@ -13,6 +13,7 @@ module Grape
         @status  = status
         @message = message
         @headers = headers
+        super(message)
       end
 
       def [](index)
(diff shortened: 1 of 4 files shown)


lib/grape/exceptions/base.rb
```diff
@@ -13,6 +13,7 @@ def initialize(status: nil, message: nil, headers: nil, **_options)
         @status  = status
         @message = message
         @headers = headers
+        super(message)
```
dblock, line 16: Does this duplicate storing `message`, i.e. is `super.to_s` or something of the kind provides the property and we can remove `@message` from our implementation?

ericproulx: Makes total sense

lib/grape/exceptions/base.rb
```diff
@@ -7,12 +7,12 @@ class Base < StandardError
       BASE_ATTRIBUTES_KEY = 'grape.errors.attributes'
       FALLBACK_LOCALE = :en
 
-      attr_reader :status, :message, :headers
```
dblock, line 10: This is a breaking change like so, meaning that I cannot do .message on an exception, right? Let's reimplement `def message` that calls `to_s`?

dblock (changes requested):

lib/grape/exceptions/base.rb
ericproulx: `message` exists through inheritance from StandardError

ericproulx: I've just pushed a `spec` for `message`

dblock (approved):

lib/grape/exceptions/base.rb
```diff
@@ -7,12 +7,12 @@ class Base < StandardError
...
+      attr_reader :status, :headers
 
       def initialize(status: nil, message: nil, headers: nil, **_options)
         @status  = status
-        @message = message
         @headers = headers
+        super(message)
```
kuraga, line 15: Should be two lines above :)

dblock: Does it matter? If you think it does, PR?

kuraga: #2304

Issue #2278, opened 2022-08-21
test failure with rack-test 2.0


terceiro:
If I make this change to `Gemfile`

```diff
diff --git a/Gemfile b/Gemfile
index acc721aa..f0cff687 100644
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,7 @@ group :test do
   gem 'maruku'
   gem 'mime-types'
   gem 'rack-jsonp', require: 'rack/jsonp'
-  gem 'rack-test', '~> 1.1.0'
+  gem 'rack-test', '~> 2.0'
   gem 'rspec', '~> 3.11.0'
   gem 'ruby-grape-danger', '~> 0.2.0', require: false
   gem 'simplecov', '~> 0.21.2'
```

and run the tests (after running `bundle`), I get the following test failures:

```
Failures:

  1) Grape::Endpoint#headers includes request headers
     Failure/Error:
       expect(JSON.parse(last_response.body)).to eq(
         'Host' => 'example.org',
         'Cookie' => ''
       )
     
       expected: {"Cookie"=>"", "Host"=>"example.org"}
            got: {"Cookie"=>"", "Host"=>"example.org", "Version"=>"HTTP/1.0"}
     
       (compared using ==)
     
       Diff:
       @@ -1,3 +1,4 @@
        "Cookie" => "",
        "Host" => "example.org",
       +"Version" => "HTTP/1.0",
       
     # ./spec/grape/endpoint_spec.rb:141:in `block (3 levels) in <top (required)>'

  2) Grape::Endpoint#params from body parameters returns a 400 if given an invalid multipart body
     Failure/Error: expect(last_response.body).to eq('empty message body supplied with multipart/form-data; boundary=foobar content-type')
     
       expected: "empty message body supplied with multipart/form-data; boundary=foobar content-type"
            got: "file is invalid"
     
       (compared using ==)
     # ./spec/grape/endpoint_spec.rb:435:in `block (4 levels) in <top (required)>'

Finished in 4.61 seconds (files took 1.81 seconds to load)
2152 examples, 2 failures

Failed examples:

rspec ./spec/grape/endpoint_spec.rb:139 # Grape::Endpoint#headers includes request headers
rspec ./spec/grape/endpoint_spec.rb:426 # Grape::Endpoint#params from body parameters returns a 400 if given an invalid multipart body
```

It would be nice if these tests could be made to work with rack-test 2.0.

PR #2279, opened 2022-08-22
Test with rack-test 2.0

terceiro:
Those two tests endpoint_spec are in the end testing internal details of
rack-test. Change them so that they still test what was intended, but in
a more robust way that does not break on changes on the implementation
details of rack-test.

Fixes: https://github.com/ruby-grape/grape/issues/2278

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -138,10 +138,9 @@ describe Grape::Endpoint do
 
     it 'includes request headers' do
       get '/headers'
-      expect(JSON.parse(last_response.body)).to eq(
-        'Host' => 'example.org',
-        'Cookie' => ''
-      )
+      response = JSON.parse(last_response.body)
+      expect(response['Host']).to eq('example.org')
+      expect(response['Cookie']).to eq('')
     end
 
     it 'includes additional request headers' do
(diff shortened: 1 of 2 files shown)


spec/grape/endpoint_spec.rb
```diff
@@ -432,7 +431,7 @@ def app
         end
         post '/upload', { file: '' }, 'CONTENT_TYPE' => 'multipart/form-data; boundary=foobar'
         expect(last_response.status).to eq(400)
-        expect(last_response.body).to eq('empty message body supplied with multipart/form-data; boundary=foobar content-type')
+        expect(last_response.body).not_to be_empty
```
dblock, line 434: Are we losing something in this test? Is there something we can assert that ensures that the error is empty message body?

terceiro: good question. TBH I could not figure out yet where this message in the body comes from. I thought it came from rack-test, but it doesn't. It seems that the new rack-test changes the way it composes requests with file uploads though.

dblock:
Want to finish this @terceiro ?

terceiro:
was this closed by mistake? I think we should get support for rack-test 2.0 at some point

terceiro:
no, sorry for the noise. that MR does seem to address the issues here

PR #2284, opened 2022-10-20
An unexpected backtick

zysend:
found an unexpected backtick while learning


dblock:
Weird that somehow made it into the codebase! Looks like we never hit this code in a spec? Care to add one, please?

zysend:
Weird it seems that there are double backticks.
Thanks for the reminder.

Issue #2275, opened 2022-08-03
Why given block condition not match, but declared return block inside params?


YaEvan:
I want declared not to allow unvalidated parameters if the given condition is not met

# case1

```
desc 'Create T'
params do
        optional :device_type, type: String, desc: 'device_type', values: ['type1', 'type2']
        given device_type: ->(val) { val == 'type1' } do
          requires :device_config, type: Hash do
            requires :rtsp, type: String, regexp: %r{rtsp://}
          end
        end
        given device_type: ->(val) { val == 'type2' } do
          requires :device_config, type: Hash do
            requires :number, type: Integer
          end
        end
end
put ':id' do
   puts declared(params)
   puts declared(params, include_missing: false)
end
```

params

```
{"device_config"=>{"rtsp"=>"fake_rtsp", number=>'fake_number'}}
```

result 

```
{"device_type"=>nil, "device_config"=>{"rtsp"=>"fake_rtsp"}}
{"device_type"=>nil, "device_config"=>{"rtsp"=>"fake_rtsp"}} // requires not work, but declared return
```

# case2

```
desc 'Create T'
params do
        optional :device_type, type: String, desc: 'device_type', values: ['type1', 'type2']
        given device_type: ->(val) { val == 'type1' } do
          requires :device_config, type: Hash do
            requires :rtsp, type: String, regexp: %r{rtsp://}
          end
        end
        given device_type: ->(val) { val == 'type2' } do
          requires :device_config, type: Hash do
            requires :number, type: Integer
          end
        end
end
put ':id' do
   puts declared(params)
   puts declared(params, include_missing: false)
end
```

params

```
{"device_config"=>{"rtsp"=>"fake_rtsp", number=>"fake_rtsp"}}
```

result 

```
{"device_type"=>nil, "device_config"=>{"number"=>"fake_rtsp"}}
{"device_config"=>{"number"=>"fake_rtsp"}} // requires not work, but declared return last delimit
```

dblock:
This does look like a bug where we are not evaluating `given` in `declared`? Care to maybe PR some failing tests/fix?

dblock:
Closed via https://github.com/ruby-grape/grape/pull/2285

PR #2285, opened 2022-10-25
try to fix the declared with not given

zysend:


zysend:
# try to fix #2275 


```
puts declared(params)
puts declared(params, include_missing: false)
puts declared(params, include_missing: false, include_not_dependent: false)
```
## case 1,2
### params
```
{"device_config":{"rtsp":"fake_rtsp", "number":"fake_rtsp"}}
```
### result
```
{"device_type"=>nil, "device_config"=>{"number"=>"fake_rtsp"}}
{"device_config"=>{"number"=>"fake_rtsp"}}
{}
```

## case 3
### params
```
{"device_type": "type1","device_config":{"rtsp":"fake_rtsp", "number":"fake_rtsp"}}
```
### result
```
{"device_type"=>"type1", "device_config"=>{"number"=>"fake_rtsp"}}
{"device_type"=>"type1", "device_config"=>{"number"=>"fake_rtsp"}}
{"device_type"=>"type1", "device_config"=>{"rtsp"=>"fake_rtsp"}}
```

dblock (changes requested):
This is great, albeit a bit complicated. I don't have anything better to suggest, but consider some refactoring to make it a bit cleaner/better organized/.

- I think `include_not_dependent` should be called something else, what exactly does it do? exclude anything that is not `given` (given is evaluated?) or anything that's inside any `given` block? So maybe `evaluate_given` or just `given: true/false`?
- Will nested given work? Let's add some tests/open some more bugs?
- This needs documentation, and CHANGELOG.

spec/grape/validations/params_scope_spec.rb
```diff
@@ -664,6 +664,35 @@ def initialize(value)
...
+        given a: ->(val) { val == 'y' } do
+          requires :b, type: Hash, allow_blank: false do
+            requires :e
+          end
+        end
+        given a: ->(val) { val == 'z' } do
+          requires :c
+        end
+      end
+      subject.get('/include_not_dependent') { declared(params, include_not_dependent: false).to_json }
```
dblock, line 685: Add spec with `include_not_dependent: true`.

zysend:
Thanks for your advice.

- include_not_dependent is not logical and readable，so I choose the evaluate_given.

- My previous tests covered too few cases, so I added more tests.
Then I found a bug on the declared_params_with_scope and I think again if the way is really organized.
Finally, I choose a new binding way between the attr and scope instand of the attrs and scope.
The new way is cleaner, but maybe unsafely(changes declared_params directly) and slowly(A scope may run meets_dependency? many times).

zysend:
-  Added documentation, and CHANGELOG.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 * [#2256](https://github.com/ruby-grape/grape/pull/2256): Raise `Grape::Exceptions::MultipartPartLimitError` from Rack when too many files are uploaded - [@bschmeck](https://github.com/bschmeck).
 * [#2266](https://github.com/ruby-grape/grape/pull/2266): Fix code coverage - [@duffn](https://github.com/duffn).
 * [#2284](https://github.com/ruby-grape/grape/pull/2284): Fix an unexpected backtick - [@zysend](https://github.com/zysend).
+* [#2285](https://github.com/ruby-grape/grape/pull/2285): Try to fix the declared with not given - [@zysend](https://github.com/zysend).
```
dblock, line 30: Try -> Fix ...

UPGRADING.md
```diff
@@ -16,6 +16,25 @@ See [#2227](https://github.com/ruby-grape/grape/pull/2227) for more information.
 
 Rack supports a configurable limit on the number of files created from multipart parameters (`Rack::Utils.multipart_part_limit`) and raises an error if params are received that create too many files.  If you were handling the Rack error directly, Grape now wraps that error in `Grape::Execeptions::TooManyMultipartFiles`.  Additionally, Grape will return a 413 status code if the exception goes unhandled.
 
+#### Declared params can choose not to evaluate given
+
+The [`#declared(params)`] will return parameters even if depend on another parameters not given,but it will not return these if `evaluate_given=false` was set.
```
dblock, line 21: Since the default behavior hasn't changed you don't need anything in UPGRADING. Move this into README.md and only document the new `evaluate_given` option. I can help with the English text.

lib/grape/validations/params_scope.rb
```diff
@@ -130,13 +165,14 @@ def required?
       # Adds a parameter declaration to our list of validations.
       # @param attrs [Array] (see Grape::DSL::Parameters#requires)
       def push_declared_params(attrs, **opts)
+        opts[:declared_params_scope] ||= self
```
dblock, line 168: Dup this structure, don't modify caller. 

```
opts = opts.merge(decared_params_scope: self) unless opts.key?(:declared_params_scope)
```

dblock (changes requested):
Code looks good, some minor stuff below. Let's work on the documentation.

Regarding performance: this is only a problem when calling `declared(params)`, right? So there's no regression during execution in most cases, so I am not worried. Let me know if I misunderstood.

zysend:
I changed the structure of `@declared_params`, it will be initialized when api is initialized and only used when `#declared(params)` is called.
if you do not set `decared_params_scope=false` when using `#declared(params)`, `@declared_params` will behave the same as the old data structure.

Thanks for the review, I will revise it soon.

lib/grape/validations/params_scope.rb
zysend: Yes,I should be more rigorous.

zysend:
I revised the docs and modified the default value of `evaluate_given`(`evaluate_given: false` means that it does not evaluate given and will return all declared_params).

README.md
```diff
@@ -1078,6 +1079,102 @@ curl -X POST -H "Content-Type: application/json" localhost:9292/users/signup -d
 }
 ````
 
+### Evaluate Given
+
+By default `declared(params)` will return parameters even if depend on another parameters not given. If you do not want to return these, you can use the `evaluate_given` option. By default, `evaluate_given` is set to `false`. Consider the following API:
```
dblock, line 1084: By default `declared(params)` will return parameters even if depend on another parameters not given. If you do not want to return these, you can use the `evaluate_given` option. By default, `evaluate_given` is set to `false`.

-> 

By default `declared(params)` will not evaluate `given` and return all parameters. Use `evaluate_given` to evaluate all `given` blocks and return only parameters that satisfy `given` conditions.

lib/grape/dsl/inside_route.rb
```diff
@@ -47,41 +47,47 @@ def declared_array(passed_params, options, declared_params, params_nested_path)
         end
 
         def declared_hash(passed_params, options, declared_params, params_nested_path)
-          renamed_params = route_setting(:renamed_params) || {}
+          declared_params.each_with_object(passed_params.class.new) do |declared_param_attr, memo|
+            # Check given
```
dblock, line 51: Nit: would remove this comment, it's pretty clear what this line does.

dblock (approved):
Looks great! Let's fix the English a little in README and this is good to go!

dblock:
I merged the change, thanks for contributing @zysend !

zysend:
Thanks for your advise.

Issue #2295, opened 2022-12-23
Using shared parameters with options, without passing in any options broken


magni-:
From the README:
```ruby
module SharedParams
  extend Grape::API::Helpers

  params :order do |options|
    optional :order_by, type: Symbol, values: options[:order_by], default: options[:default_order_by]
    optional :order, type: Symbol, values: %i(asc desc), default: options[:default_order]
  end
end

class API < Grape::API
  helpers SharedParams

  desc 'Get a sorted collection.'
  params do
    use :order, order_by: %i(id created_at), default_order_by: :created_at, default_order: :asc
  end

  get do
    Collection.send(params[:order], params[:order_by])
  end
end
```
This code worked in 1.6.2 and still works in 1.7.0. However, using `use :order` without passing in any options used to work in 1.6.2 and no longer works in 1.7.0: ``NoMethodError: undefined method `[]' for nil:NilClass``.

This seems to be caused by https://github.com/ruby-grape/grape/pull/2232 (fixing https://github.com/ruby-grape/grape/issues/2231). Since this wasn't mentioned in the PR or in `UPGRADING`, I'm guessing this wasn't an intentional breaking change?

The end-user workaround is to define a default value for `options`:
```ruby
params :order do |options = {}|
  optional :order_by, type: Symbol, values: options[:order_by], default: options[:default_order_by]
  optional :order, type: Symbol, values: %i(asc desc), default: options[:default_order]
end
```
But it would be nice if we didn't need to do so.

magni-:
Actually, it's much cleaner for us to switch to keyword arguments now that these are supported:
```ruby
params :order do |order_by: nil, default_order: nil, default_order_by: nil|
  optional :order_by, type: Symbol, values: options[:order_by], default: options[:default_order_by]
  optional :order, type: Symbol, values: %i(asc desc), default: options[:default_order]
end
```

Leaving the issue open in case people still think the previous behavior should be brought back?

dblock:
@magni- Any such breaking change was not intentional. We really want to get to a point where we're strictly adhering to semver. Help us by writing a spec that shows how it breaks? Let's fix it?

cc: @dm1try

dm1try:
Yeah, it's certainly not something I expected to break.
Though, the current interface feels more explicit to me based to the examples above.

dm1try:
I've opened https://github.com/ruby-grape/grape/pull/2299, just in case.

[...]

PR #2299, opened 2022-12-29
fix, do not use kwargs for empty args 

dm1try:
for backward compatibility with the previous implemention
where options param behaves as an empty hash by default

see #2295


dblock:
Thanks for the fix @dm1try !

PR #2301, opened 2023-01-01
Reviste GitHub workflows

### LINKED ISSUE #2300: rails_edge on 2.7 needs newer RubyGems
While working #2292, I found that [error](https://github.com/ruby-grape/grape/actions/runs/3817326578/jobs/6493515044) 
```
Because every version of rails depends on RubyGems >= 3.3.13
    and rails_edge.gemfile depends on rails >= 0,
    RubyGems >= 3.3.13 is required.
  So, because current RubyGems version is = 3.1.6,
    version solving has failed.
```

Easy to fix. We add `rubygems: latest` to the `ruby/setup-ruby@v1` step. See https://github.com/ruby/setup-ruby/blob/master/action.yml

ericproulx:
This PR updates current GH workflows.

- Refactor ruby-matrix
- Add Ruby 3.2
- Add rack_3_0.gemfile
- Rename some gemfiles
- Use ubuntu-latest
- Should fix #2300


.github/workflows/test.yml
```diff
@@ -24,40 +18,16 @@ jobs:
...
-          - ruby: 2.7
-            gemfile: 'gemfiles/multi_json.gemfile'
-          - ruby: 2.7
-            gemfile: 'gemfiles/multi_xml.gemfile'
-          - ruby: 2.7
-            gemfile: 'gemfiles/rails_7.gemfile'
-    runs-on: ubuntu-20.04
+        os: [ubuntu-latest]
+        ruby: ['2.6', '2.7', '3.0', '3.1', '3.2']
+        gemfile: [multi_json, multi_xml, rack_1_0, rack_2_0, rack_2_2, rails_5_2, rails_6_0, rails_6_1, rails_7_0]
```
dblock, line 23: This becomes a massive matrix and the list of `exclude` will grow.

We could make something like https://github.com/slack-ruby/slack-ruby-client/blob/master/.github/workflows/test.yml#L13 potentially to make it easier to grok, WDYT?

dblock (commented):
Wondering whether it's really useful to run all the permutations on 5 versions of Ruby. This is 62 builds.

.github/workflows/test.yml
ericproulx: I've reduced it and add some specific inclusion.

dblock (approved):

dblock:
It's a step in the right direction thanks!

PR #2310, opened 2023-03-19
Fix YARD docs markdown rendering

duffn:
This fixes the rendering of the YARD docs in markdown. RudyDoc reads `.yardopts` so this will work there too.

As one small example:

Before

Screen Shot 2023-03-18 at 6 57 36 PM

After

Screen Shot 2023-03-18 at 6 58 56 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ dist
 Gemfile.lock
 gemfiles/*.lock
 tmp
+.yardoc
 
 ## Rubinius
 .rbx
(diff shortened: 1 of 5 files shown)


dblock:
Looks like there's an unrelated rubocop failure, care to fix so we can get a green build?

duffn:
Failures are now due to 

https://github.com/rack/rack-test/blob/main/History.md#210--2023-03-14

```Digest authentication support, deprecated in 2.0.0, has been removed (Jeremy Evans #307)```

duffn:
https://github.com/ruby-grape/grape/pull/2311

PR #2317, opened 2023-04-16
Remove maruku and rubocop-ast as dependencies

### LINKED ISSUE #2316: maruku and rubocop-ast ?
We have `maruku` and `rubocop-ast` in Gemfile but none of then are actually used in the code. We could remove them safely. Also,  `rubocop-ast` is a direct dependency of `rubocop`.

ericproulx:
Fix #2316


dblock:
Thanks

PR #2323, opened 2023-05-06
fix using endless ranges for values parameter.

dhruvCW:
allows using endless range for values validator.

e.g.

```ruby
params do
  requires :minimum, type: Integer, values: 10..
  optional :maximum, type: Integer, values: ..10 
end
```


dblock:
Perfect thank you.

dblock:
I missed a failure here in CI because I was trained not to see it in all the edge failures :(

```
rspec ./spec/grape/validations/validators/values_spec.rb:390 # Grape::Validations::Validators::ValuesValidator does not allow an invalid value for a parameter using an endless range
rspec ./spec/grape/validations/validators/values_spec.rb:384 # Grape::Validations::Validators::ValuesValidator validates against values in an endless range
```

dhruvCW:
>I disabled it for ActiveSupport < 6 in https://github.com/ruby-grape/grape/pull/2325, but please do PR a fix if you know of one.

thanks for patching the tests. The issue is because of a bug in rails 5.2 which was fixed in rails 6 but not backported. 

https://github.com/rails/rails/pull/36460

given rails 5.2 has already reached end of life a while back I don't think it's worth the effort to try and work around this problem.

PR #2324, opened 2023-05-08
expose default in the description dsl.

dhruvCW:
grape-swagger 1.6.0 [exposes the ability to define the default response](https://github.com/ruby-grape/grape-swagger/pull/868) but the description dsl does not support it yet.

This PR adds support to define the `default` response using the description dsl.


dblock:
Looks good. Needs a README update, please.

PR #2326, opened 2023-05-09
Use Active support functions

ericproulx:
This PR replace some custom made functions in favour of ActiveSupport's equivalent.

- Use ActiveSupport `deep_symbolize_keys` and `deep_merge!`
- Replace `object.present? ? object : nil` by `presence`
- Replace `!include?` by `exclude?`
- Use `dup` and `duplicable?` instead of `duplicatable?` and `duplicate`


dblock:
🤞 that the ActiveSupport versions perform as well

ericproulx:
> 🤞 that the ActiveSupport versions perform as well

Should we add a minimum version for active support ? Since we're testing rails 5.2 to 7.0, tests are covering ActiveSupport 5 to 7. Thus, it doesn't test < 4. At least we know it works from 5 and beyond :)

ericproulx:
> > 🤞 that the ActiveSupport versions perform as well
> 
> Should we add a minimum version for active support ? Since we're testing rails 5.2 to 7.0, tests are covering ActiveSupport 5 to 7. Thus, it doesn't test < 5.2. At least we know it works from 5 and beyond :)

I've restrict to ActiveSupport >= 5. It doesn't work below because of [this](https://github.com/rails/rails/blob/0ecaaf76d1b79cf2717cdac754e55b4114ad6599/activesupport/lib/active_support/core_ext/object/duplicable.rb#L105-L119) and we're talking of Ruby 1.9

dblock:
LGTM want to finish this PR?

ericproulx:
I'll run a quick test suites on our codebase to see if there's anything wrong.

ericproulx:
> I'll run a quick test suites on our codebase to see if there's anything wrong.

Everything looks good. There's only the danger part that is failing.

dblock:
Not sure what the Danger problem is, but there's an extra `*` in the CHANGELOG. Maybe that's it? I can poke it at if you can't figure it out.

dblock:
I think we should either release 1.7.1 before merging this, or make this 1.8.0, wdyt?

ericproulx:
> I think we should either release 1.7.1 before merging this, or make this 1.8.0, wdyt?

Good idea, after this PR, I'll open a another that will use ActiveSupport::Deprecation instead of `warn [DEPRECATION]`. Sticking with ActiveSupport seems the best on the long run.

dblock:
All yours to release/bump!

ericproulx:
> All yours to release/bump!

Done. I hope everything is ok. I'm not releasing that often.

dblock:
Let's increment version to 1.8.0 as part of this PR? Feel free to merge after.

PR #2331, opened 2023-05-20
run_validators memory optmization

ericproulx:
This PR change the way validator are created before calling `validate`. `run_validators` has always created all the validators before calling `validate`. In a case where an exception is raised and the validator `fail_fast?` returns true, the remaining validators won't be `validate`. Thus, we can optimize the validation process by yielding validator instead of creating all of them at first. It's a small memory optimization and maybe it will speed up a little bit any endpoint that has a validator with `fail_fast`

In the same file, I also replaced some `|| []` by `&.`


dblock:
Nice

PR #2332, opened 2023-05-27
Replace Grape::Config by ActiveSupport::Configurable

ericproulx:
This PR replaces Grape::Config::Configuration by ActiveSupport::Configurable. Same behavior is expected


dblock (commented):
What do you think about inheriting/extending `Config` from `ActiveSupport::Configurable` instead and keeping `reset`? It's a common pattern used in tests (you had to copy that code too here).

ericproulx:
> What do you think about inheriting/extending `Config` from `ActiveSupport::Configurable` instead and keeping `reset`? It's a common pattern used in tests (you had to copy that code too here).

The test wasn't actually testing the config part but the `build_param_with` options. I got rid of the around and use the option properly. I know it's a common practice in test to do the `reset` but it's only for testing not production. I tend to avoid functions that are just needed when testing.

PR #2333, opened 2023-06-01
Feature: Allows procs with arity 1 to validate and use custom messages 

### LINKED ISSUE #2280: set custom message when using proc to validate value
How can I use a proc to validate and use a custom message?

This way is deprecated:
```ruby
params do
  requires :e, values: { proc: -> (v) { v.even? }, message: 'must be even' }
end
```

The new way doesn't support a custom message:
```ruby
params do
  requires :e, values: -> (v) { v.even? }
end
```

TheDevJoao:
This PR fixes the issue presented here: #2280

As this is my first contribution to the repo, I am open to any form of feedback!

PS: I also created the tests with the example shown in the issue in mind.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Features
 
-* : The `validate_param!` method to allow procs to validate and use custom messages [@thedevjoao](https://github.com/TheDevJoao).
+* [#2333] (https://github.com/ruby-grape/grape/pull/2333): The `validate_param!` now allows procs to validate and use custom messages [@thedevjoao](https://github.com/TheDevJoao).
 * [#2326](https://github.com/ruby-grape/grape/pull/2326): Use ActiveSupport extensions - [@ericproulx](https://github.com/ericproulx).
 * [#2327](https://github.com/ruby-grape/grape/pull/2327): Use ActiveSupport deprecation - [@ericproulx](https://github.com/ericproulx).
 * [#2330](https://github.com/ruby-grape/grape/pull/2330): Use ActiveSupport inflector - [@ericproulx](https://github.com/ericproulx).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
 * [#2330](https://github.com/ruby-grape/grape/pull/2330): Use ActiveSupport inflector - [@ericproulx](https://github.com/ericproulx).
 * [#2331](https://github.com/ruby-grape/grape/pull/2331): Memory optimization when running validators - [@ericproulx](https://github.com/ericproulx).
 * [#2332](https://github.com/ruby-grape/grape/pull/2332): Use ActiveSupport configurable - [@ericproulx](https://github.com/ericproulx).
-* Your contribution here.
```
dblock, line 10: Put back this line.

dblock (changes requested):
I don't see `:proc_message` anywhere in specs, how is this wired up?! I must be missing something obvious 😅

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#2333] (https://github.com/ruby-grape/grape/pull/2333): The `validate_param!` now allows procs to validate and use custom messages [@thedevjoao](https://github.com/TheDevJoao).
```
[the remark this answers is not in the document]
TheDevJoao: > I don't see `:proc_message` anywhere in specs, how is this wired up?! I must be missing something obvious sweat_smile

Hey @dblock , thanks for the review! Your comment helped out a ton, turns out what I was doing wasn't necessary. I recently refactored the code to cover what was suggested [here](https://github.com/ruby-grape/grape/issues/2280#issuecomment-1248400550) and it should make more sense as an added feature. 😃

lib/grape/validations/validators/values_validator.rb
```diff
@@ -68,6 +68,17 @@ def check_excepts(param_array)
...
         end
 
+        def validate_proc(proc, param_array)
+          case proc.arity
+          when 0
+            param_array.all? { |_param| proc.call }
+          when 1
+            param_array.all? { |param| proc.call(param) }
+          else
+            raise ArgumentError, 'proc arity must be 0 or 1'
```
dblock, line 78: Do we get some kind of useful call stack with a line number such as that the developer knows where the error comes from?

dblock (approved):
This makes a lot more sense now!

dblock:
@TheDevJoao add a test for when arity is > 1 and we can merge?

lib/grape/validations/validators/values_validator.rb
TheDevJoao: Yeah, considering the ruby and rails versions that grape supports, the stack trace will show every sequence leading to this raise statement.

Issue #1527, opened 2016-11-28
unknown validator: is_array


lumean:
Hi,
I'm not sure if the issue is in the right place here. If not, please point me to the correct repository to open the issue there  (grape-entity,  grape-swagger, grape-swagger/entity, ....)

Here is my problem - code attached  [repro_api.txt](https://github.com/ruby-grape/grape/files/616127/repro_api.txt) (use it as config.ru). 

As soon as I switch from 
```
params do
      requires :all, type: Entity::Accounts  # this will be rendered as body parameter by grape-swagger
end
```
to
```
params do
      requires :all, using: Entity::Accounts.documentation, documentation: { param_type: 'body'}
end
```
I get an error. I want to switch to 'using' because I don't want the 'all' key in the required input json but instead directly the Accounts Entity at root level.
```
c:/ruby/ruby225-p319/lib/ruby/gems/2.2.0/gems/grape-0.18.0/lib/grape/validations/params_scope.rb:347:in `validate': unknown validator: is_array (Grape::Exceptions::UnknownValidator)
        from c:/ruby/ruby225-p319/lib/ruby/gems/2.2.0/gems/grape-0.18.0/lib/grape/validations/params_scope.rb:250:in `block in validates'
```
somehow, grape doesn't recognize anymore the 'is_array' property inside the documentation hash.
In addition, even when I remove the 'is_array' property,  'param_type: body'  from the documentation hash is ignored and the swagger documentation generates parameters of type 'formData' instead of 'body'.

So could someone please point out how to use  'is_array' properly in combination with  grape-entity and grape-swagger? I was not able to find a working example.

dblock:
Do you think you can build a spec that reproduces? I think this belongs in the grape-swagger.

lumean:
could you point me to some example code or a guide how write specs for grape(-swagger)? Then I think we close this issue and I'll open a new one in grape-swagger

dblock:
Start by closing this and opening one in grape-swagger. For specs there're many examples in the spec folder, I don't have much better advice, but there're a lot of people to help if you can try to get started!

migmartri:
I am running into exactly this issue. Did you happen to open the new issue in grape-swagger?

lumean:
no not yet, however I found a workaround for my case. Instead of using the params block I now use params within the description block. E.g.
```ruby
   desc 'does something' do
     detail <<-NOTE
     Some detailed explanation
     NOTE
     params Entity::Accounts.documentation
   end
[...]

Issue #1302, opened 2016-03-04
Errors with using entity documentation


onemanstartup:
I was trying to migrate my params to using entities and I'm getting different types of errors. Is this grape-entity error or grape?
```ruby
#...
      requires :all, using: Entity::Activity.documentation.except(:id)
#...
```
```ruby
module Entity
  class Activity < Base
    expose :id, documentation: {type: 'integer', desc: 'id'}
    expose :active, documentation: {type: 'Boolean', desc: 'id'}
# ...
```
This is error happens first of all.
```
lib/grape/validations/types.rb:114:in `recognized?': undefined method `ancestors' for "string":String (NoMethodError)
```

This error happens  when I'm changing type to classes. I get it we don't have Boolean type in ruby and grape uses virtus, but with Integer and String type there is no errors.
```
`<class:Activity>': uninitialized constant Entity::Activity::Boolean (NameError)
```

dblock:
To me `Entity::Activity.documentation.except(:id)` looks suspicious, why would that even work?

onemanstartup:
@dblock I took this line from readme.
```
      requires :all, except: [:ip], using: API::Entities::Status.documentation.except(:id)
```

dblock:
Ok, I guess this needs a spec in Grape or Grape::Entity. You should try to do that next.

onemanstartup:
@dblock I should write a spec? I don't understand. Documentation is a bit misleading.
Here what I found.
This works.
```ruby
    desc 'Returns your public timeline.' do
      detail 'more details'
      params Entity::Value.documentation.except(:id, :errors)
    end
```
If i want to overwrite or make field required i could do params block below but it add new param instead of overwriting previous params.

And i can't use params block within desc block.
```ruby
    desc 'Returns your public timeline.' do
      params do
        requires :all, Entity::Value.documentation.except(:id, :errors)
      end
    end
```
This gives error
```ruby
lib/grape/util/strict_hash_configuration.rb:48:in `block (2 levels) in simple_settings_methods': wrong number of arguments (given 0, expected 1) (ArgumentError)
```

dblock:
I haven't dug through this. If the documentation says something that you're trying to do and it doesn't work, it's a bug in either the code or the doc. I'll take a look later in depth, but feel free to do what you think is right and make pull requests. I know I am not being very helpful here ;)

dkniffin:
[...]

PR #2338, opened 2023-06-15
Fix unknown validator exception when using requires/optional with Entity

mscrivo:
when that entity specifies keywords that grape interprets as custom validators, ie: `is_array`, `param_type`, etc.

There have been a couple issues opened about this in the past with no real resolution: https://github.com/ruby-grape/grape/issues/1527 and https://github.com/ruby-grape/grape/issues/1302

This PR changes it so that it skips looking for a custom validator for those special documentation keywords.

This allows you to do something like:

```rb
desc 'Create some entity',
     entity: SomeEntity,
     params: SomeEntity.documentation
params do
  requires :none,
           except: %i[field1 field2],
           using:
             SomeEntity.documentation
end
post do
 ...
end
```

and SomeEntity can specify documentation like:

```rb
      class SomeEntity < Grape::Entity
        expose :id, documentation: { type: Integer, format: 'int64', desc: 'ID' }
        expose :name, documentation: { type: String, desc: 'Name', require: true, param_type: 'body' }
        expose :array_field,
               documentation: {
                 type: String,
                 desc: 'Array Field',
                 require: true,
                 is_array: true,
                 param_type: 'body'
               }
      end
```

and it won't crash on startup and give you correct documentation and param validation to boot.

Open questions:
1. Is this an exhaustive list of keywords?
2. Is there a better way to maintain/get the list of keywords?
3. Is there a better approach altogether?


dblock:
Start by writing a test for this? Let's find a better way to implement it, too.

mscrivo:
> Let's find a better way to implement it, too

Sure, but I was hoping for some guidance there since I don't know the codebase as well as others.

> Start by writing a test for this?

Will do, but I don't think it makes sense to until we have an acceptable approach

dblock:
I tend to do TDD for things like this, but in either case I don't know what the right fix is. Hardcoding the list that Grape knows nothing about is probably not. I've rarely used this feature too, but it looks like `documentation` had to match parameter options by design.

mscrivo:
> I tend to do TDD for things like this, but in either case I don't know what the right fix is. Hardcoding the list that Grape knows nothing about is probably not. I've rarely used this feature too, but it looks like `documentation` had to match parameter options by design.

There in lies the problem, how do you write a test for this, when the issue is in the integration point between the multiple repos? Seems like the fix needs to be in grape, but proper testing of it would be done in grape-swagger-entity? 

I don't know, tbh, I'm confused why grape is even split into so many repos. Seems like it would be a much more robust offering if all combined into one gem with solid integration points. Documentation would be much better too since you wouldn't have to constantly jump between 4 different READMEs to figure out what combination of things you need. But I admit my naivety here, as I really don't know anything about the project's history and thus context for various design decisions.

dblock:
> > I tend to do TDD for things like this, but in either case I don't know what the right fix is. Hardcoding the list that Grape knows nothing about is probably not. I've rarely used this feature too, but it looks like `documentation` had to match parameter options by design.
> 
> There in lies the problem, how do you write a test for this, when the issue is in the integration point between the multiple repos? Seems like the fix needs to be in grape, but proper testing of it would be done in grape-swagger-entity?

You can add an integration test that uses a thirdparty gem here: https://github.com/ruby-grape/grape/tree/master/spec/integration

> I don't know, tbh, I'm confused why grape is even split into so many repos. Seems like it would be a much more robust offering if all combined into one gem with solid integration points. Documentation would be much better too since you wouldn't have to constantly jump between 4 different READMEs to figure out what combination of things you need. But I admit my naivety here, as I really don't know anything about the project's history and thus context for various design decisions.

Cause Grape doesn't want to be Rails! :) but open to more specific suggestions and of course contributions

mscrivo:
I've made it a tiny bit more explicit as to what's going on in this code, ie. 
* That there exists some list of reserved keywords for documentation hashes that is:
  * Not codified in any one place
  * Does not have any validators associated with it

I've modified the existing tests around `requires :all` and `requires :none` such that they fail when they reserved words are not excluded from custom validator lookups and they do indeed fail if I remove the list.

lib/grape/validations/params_scope.rb
```diff
@@ -358,8 +358,12 @@ def validates(attrs, validations)
         # type casted values
         coerce_type validations, attrs, doc, opts
 
+        # There are a number of documentation options on entities that don't have
+        # corresponding validators. Since there is nowhere that enumerates them all,
+        # we maintain a list of them here and skip looking up validators for them.
+        reserved_keywords = %i[as required param_type is_array format example]
```
dblock, line 364: Let's make this a constant `RESERVED_KEYWORDS`. Maybe we should also rename it? These are documentation options so `RESERVED_DOCUMENTATION_KEYWORDS`? WDYT?

dblock (changes requested):
Let's add a test that exercises every one of these options? Maybe something like

```
ruby
ParamsScope:: RESERVED_DOCUMENTATION_KEYWORDS.each do |keywords|
   context keyword do
       it ..
```

lib/grape/validations/params_scope.rb
mscrivo: Sounds good to me!

mscrivo:
> Let's add a test that exercises every one of these options? Maybe something like
> 
> ```ruby
> ParamsScope:: RESERVED_DOCUMENTATION_KEYWORDS.each do |keywords|
>    context keyword do
>        it ..
> ```

I see where you're coming from, but my counterpoint is that each keyword is a different type, ie `is_array` & `required` are bools, whereas `param_type` is a string with 3 options and the others are just arbitrary strings, so it'd be a bit awkward to write a generic test going through each keyword with it knowing what it's type should be without creating a map of some sort. But that seems overly complex and I don't think it really increases test coverage either, since if you have a test that covers at least one of the keywords, you've covered them all, since the functionality of skipping is the same for all.

dblock (approved):

dblock:
Thanks for helping with this @mscrivo, merged.

PR #2341, opened 2023-06-24
Stop yielding skip value

### LINKED ISSUE #2329: Not yielding skipping value ?
Hi,

[multiple_attributes_iterator](https://github.com/ericproulx/grape/blob/master/lib/grape/validations/multiple_attributes_iterator.rb) and [single_attribute_iterator](https://github.com/ericproulx/grape/blob/master/lib/grape/validations/single_attribute_iterator.rb) are yielding attribute that are eventually skip [here](https://github.com/ericproulx/grape/blob/96e2faf1cb6165eab50209fe045f15784b211d59/lib/grape/validations/validators/multiple_params_base.rb#L12) and [here](https://github.com/ericproulx/grape/blob/96e2faf1cb6165eab50209fe045f15784b211d59/lib/grape/validations/validators/base.rb#L50). Does it make sense to just not yield them and forget the `next if skip_value` ? We would have 1 iteration saved per skip

ericproulx:
This PR will stop yielding `skip` value since iterators are already ignoring `skip` values.

Resolves #2329

### DIFF
--- a/spec/grape/validations/single_attribute_iterator_spec.rb
+++ b/spec/grape/validations/single_attribute_iterator_spec.rb
@@ -45,7 +45,7 @@ describe Grape::Validations::SingleAttributeIterator do
       context 'when missing optional value' do
         let(:params) { [Grape::DSL::Parameters::EmptyOptionalValue, 10] }
 
-        it 'doest not yield skipped values' do
+        it 'does not yield skipped values' do
           expect { |b| iterator.each(&b) }.to yield_successive_args(
             [params[1], :first, false], [params[1], :second, false]
           )
--- a/spec/grape/validations/single_attribute_iterator_spec.rb
+++ b/spec/grape/validations/single_attribute_iterator_spec.rb
@@ -14,7 +14,7 @@ describe Grape::Validations::SingleAttributeIterator do
 
       it 'yields params and every single attribute from the list' do
         expect { |b| iterator.each(&b) }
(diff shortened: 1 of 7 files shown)


spec/grape/validations/single_attribute_iterator_spec.rb
```diff
@@ -35,20 +35,19 @@
...
+            [params[2], :first, false], [params[2], :second, false]
           )
         end
       end
 
       context 'when missing optional value' do
         let(:params) { [Grape::DSL::Parameters::EmptyOptionalValue, 10] }
 
-        it 'marks params with skipped values' do
+        it 'doest not yield skipped values' do
```
dblock, line 48: typo: does not

dblock (commented):
This is a breaking change, but I fail to see how it can break someone 🤔 How was this a problem for your use-case?

At least we should add a paragraph to UPGRADING.

ericproulx:
> This is a breaking change, but I fail to see how it can break someone 🤔 How was this a problem for your use-case?
> 
> At least we should add a paragraph to UPGRADING.

I'm not sure its a breaking change. There's never been a validation done against a skip value. It's was always returned by base class when validating. I figured that why yield these values when it's actually returned automatically.

dblock:
> I'm not sure its a breaking change. There's never been a validation done against a skip value. It's was always returned by base class when validating. I figured that why yield these values when it's actually returned automatically.

I think you're right. I am going to YOLO it.

PR #2342, opened 2023-06-27
Allow specifying a handler for grape_exceptions

mscrivo:
This allows you to customize the format of the error response for grape exceptions:

For example, you could do something like this:

```rb
rescue_from :grape_exceptions do |e|
  error!({ errors: [{ code: 'Error', message: e.message.squish }] }, e.status)
end
```

which would render like this:

```
{
  "errors": [
    {
      "code": "Error",
      "message": "Problem: message body does not match declared format Resolution: when specifying application/json as content-type, you must pass valid application/json in the request's 'body'"
    }
  ]
}
```

The main reason for wanting this change is to have grape-exceptions (which are very useful!) conform to the shape of other errors. I ran into this by creating an :all handler with the shape I wanted and realizing that I could not customize the shape of the grape exceptions.


lib/grape/dsl/request_response.rb
```diff
@@ -116,6 +116,7 @@ def rescue_from(*args, &block)
           elsif args.include?(:grape_exceptions)
             namespace_inheritable(:rescue_all, true)
             namespace_inheritable(:rescue_grape_exceptions, true)
+            namespace_inheritable :grape_exceptions_rescue_handler, handler
```
dblock, line 119: Let's match the format, or remove the `(` above.

dblock (changes requested):
Looks interesting!

1. Add API specs that exhibit the behavior described in the PR so we know what this is for.
2. Update README and CHANGELOG.

mscrivo:
> 1. Add API specs that exhibit the behavior described in the PR so we know what this is for.

Done 

> 2. Fix tests ;)
Done

> 3. Update README and CHANGELOG.
Done

dblock:
Rubocop is unhappy, run `rubocop -a ; rubocop --auto-gen-config` and we're good to go!

lib/grape/endpoint.rb
```diff
@@ -276,6 +276,7 @@ def run
       end
     end
 
+    # rubocop:disable Metrics/MethodLength
```
dblock, line 279: Let's not do that and use `rubocop -a ; rubocop --auto-gen-config`. The second one will fix it. We don't want to litter the code with rubocop ignores, mostly using it for auto-formatting.

mscrivo: Tried that at first, but it produced all kinds of unrelated changes. The correct command is: 
`bundle exec rubocop -a ; bundle exec rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit 5000`
and really there should be a CI job that I can look at, that tells me that, but it seems you have to manually run CI checks on each commit? If you want more contributors, you should make this easier.

dblock: I tend to ignore the unrelated changes, we just use rubocop for auto-formatting and minimal linting, so the rest is whatever.

GitHub doesn't let first time contributors auto-run CI on PRs for some security reason. I'll look into turning that off. FYI you can easily turn on GHA on your fork if you need it. Also, we do have 414 contributors, and sincerely appreciate your help!

dblock (approved):

lib/grape/endpoint.rb
mscrivo: > GitHub doesn't let first time contributors auto-run CI on PRs for some security reason. I'll look into turning that off. 

That'd help a lot I think

> FYI you can easily turn on GHA on your fork if you need it. 

🤦🏼 I should have thought of that. 

Anyhow, thanks for merging!

dblock: I switched it to the first option, but it doesn't let you disable this behavior. You can read about this more here: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

Screenshot 2023-07-04 at 12 42 55 PM

Issue #2298, opened 2022-12-27
Rack3 breaking change with query param nil


ericproulx:
While updating `rack-test` locally to the latest version, `rack` got updated also to it's latest version 3.0.3

Some tests are failing due to the following [breaking change ](https://github.com/rack/rack/issues/1696):
```
BREAKING CHANGE: Query parsing now treats parameters without = as having the empty string value instead of nil value, to conform to the URL spec.
```

What does it means ? Passing a query params like `?name` will be considered by Rack (>=3) as `{ "name" => "" }` instead `{ "name" => nil }`.

Moslty, it affects all the `optional` parameters that you can pass `param_name: nil`.

Question? Since it's a breaking change, should we update the gemspec to keep `rack` < 3. That's one breaking change, but there might be others since other specs are falling.

Tests failing:
```log
rspec ./spec/grape/validations/validators/regexp_spec.rb:114 # Grape::Validations::Validators::RegexpValidator accepts nil
rspec ./spec/grape/validations/validators/regexp_spec.rb:59 # Grape::Validations::Validators::RegexpValidator custom validation message accepts nil
rspec ./spec/grape/validations/validators/regexp_spec.rb:82 # Grape::Validations::Validators::RegexpValidator custom validation message regexp with array refuses nil items
rspec ./spec/grape/validations_spec.rb:36 # Grape::Validations params optional doesn't validate when param not present
rspec './spec/grape/validations/validators/default_spec.rb[1:12:2:1]' # Grape::Validations::Validators::DefaultValidator optional with nil as value structures types respects the default value
rspec './spec/grape/validations/validators/default_spec.rb[1:12:2:2]' # Grape::Validations::Validators::DefaultValidator optional with nil as value structures types respects the default value
rspec './spec/grape/validations/validators/default_spec.rb[1:12:1:15]' # Grape::Validations::Validators::DefaultValidator optional with nil as value primitive types respects the default value
rspec ./spec/grape/validations/validators/coerce_spec.rb:212 # Grape::Validations::Validators::CoerceValidator coerce coerces String
rspec './spec/grape/validations/validators/coerce_spec.rb[1:1:5:11:2:1]' # Grape::Validations::Validators::CoerceValidator coerce coerces nil values structures types respects the nil value
rspec './spec/grape/validations/validators/coerce_spec.rb[1:1:5:11:1:9]' # Grape::Validations::Validators::CoerceValidator coerce coerces nil values primitive types respects the nil value
[...]

PR #2346, opened 2023-08-21
Adjust test expectations to conform to rack 3

kbarrette:
As suggested in https://github.com/ruby-grape/grape/issues/2298, here's a PR that adjusts test expectations to conform to rack 3


dblock (approved):
👏

Issue #2350, opened 2023-09-15
Grape recognize_path incorrectly recognizes path


glebsonik:
### Problem
I faced an issue with path recognizing using recognize_path from Grape::API ancestor, here is an example of the API structure I have

```Ruby
class Books < Grape::API
  resource :books do
    route_param :id, type: Integer do
      # GET /books/:id
      get do
        #...
      end
    end

    resource :share do
      # POST /books/share
      post do
      # ....
      end
    end
  end
end
```

If I try to use `Books.recognize_path('/books/share')` Then it will recognize it as `GET /books/:id` request instead of the `POST /books/share`

I guess this happens because the `recognize_path` does not rely on the request method. In this case Grape recognizer treats share as :id param instead of path

dblock:
Do you think you can you turn this into a failing test?

glebsonik:
If there is a way to predefine two routes or somehow simulate this then yes. Where should I set up this test?

dblock:
You can make a pull request on top of the existing specs, so https://github.com/ruby-grape/grape/blob/d39031567b913834d9e586ac8a2733ecf3166057/spec/grape/api/recognize_path_spec.rb#L4. 

Looking at the problem though, I wonder if this is a bug at all. AFAIK routes in Grape are matched in order of declaration, not like Ruby methods where the last one wins - it's the first one that matches. Amazingly, I actually couldn't find anything on this in our README, and I think we need to state it unambiguously in it. In fact, there's https://github.com/ruby-grape/grape/issues/1858 that asks for the same. Want to take it on? I would add `recognize_path` specs for all combinations (your example, nested routes, inherited APIs, mounted APIs) and add a section to README about the order of matching.

dblock:
@glebsonik Btw, in your specific example you have a `type: Integer` in the first route, and so it does make sense that it shouldn't match in theory, so could be a bug anyway.

glebsonik:
@dblock I've done a commit but I can neither create a PR to the grape repo nor commit and push my changes. I think, I just don't have access to do this. Is there any other way I should address my PR?

glebsonik:
Okay nvm, I've just created the PR in the fork https://github.com/ruby-grape/grape/pull/2352

jcagarcia:
Hi 👋 

I was taking a look at this issue, and as the first step I can add a section in the README talking about the order of matching, as you suggested in your comment above @dblock. Agree?

[...]

PR #2352, opened 2023-09-28
[ISSUE-2350] recognize_path incorrectly incorrectly recognizes path

glebsonik:
This is a failing test implementation that reproduces an issue with `recognize_path`described in https://github.com/ruby-grape/grape/issues/2350


dblock:
Thanks for the failing test! You're one step closer to fixing it :)

jcagarcia:
Solved in #2379

Issue #2254, opened 2022-03-02
Allow lowercase HTTP/2 headers


pre:
HTTP/2 defines that HTTP headers be lowercase. 

Grape consistently normalizes HTTP headers to PascalCase. This is ok since it's consistent. 

However, there are more and more API clients which normalize HTTP headers to lowercase. One example is Axios. 

This means that the exact same HTTP header name cannot be used as-is at the Grape API and Client side, since both implement a normalization of their own.

For example Rack provides `Rack::Headers` middleware which normalizes HTTP headers to lowercase per HTTP/2.  It also seems the upcoming Rack 3 will provide lowercased HTTP headers by default.

I suggest an option to somehow enable HTTP header normalization to lowercase in Grape. 

References:
* https://github.com/rack/rack/issues/1592
* https://github.com/rack/rack/pull/1800

dblock:
See https://github.com/ruby-grape/grape/pull/2355 for a beginning of solving this.

Issue #2354, opened 2023-10-15
X-Cascade header Rack 3 issue


schinery:
The `X-Cascade` header in Rack 3 applications appears to need to be lower case, and so it currently doesn't work as expected in Grape apps using Rack 3 using `cascade: true`.

The following spec, which passes in Rack 2, fails as [the default response](https://github.com/ruby-grape/grape/blob/master/lib/grape/router.rb#L140) is returned from Grape instead.

```ruby
it "raises ActionController::RoutingError" do
  expect { get "/api/foo" }.to raise_error(ActionController::RoutingError)
end
```

Manually updating my local Grape gem to use `x-cascade` results in the spec passing again.

schinery:
Related to https://github.com/ruby-grape/grape/issues/2254

schinery:
This is what they are doing for [Sidekiq Web](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/web.rb#L37):

```ruby
    if Gem::Version.new(Rack::RELEASE) < Gem::Version.new("3")
      CONTENT_LANGUAGE = "Content-Language"
      CONTENT_SECURITY_POLICY = "Content-Security-Policy"
      LOCATION = "Location"
      X_CASCADE = "X-Cascade"
    else
      CONTENT_LANGUAGE = "content-language"
      CONTENT_SECURITY_POLICY = "content-security-policy"
      LOCATION = "location"
      X_CASCADE = "x-cascade"
    end
```

Would something similar in [Grape Headers](https://github.com/ruby-grape/grape/blob/master/lib/grape/http/headers.rb) work?

dblock:
It would probably work. Start by writing a Rails 6 and 7 spec?

schinery:
Another option could be to make the `X-Cascade` and `Content-Type` (assuming this should also be changed) to lowercase, as these should work fine in Rack < 3 apps.

> Since HTTP/2 requires it and HTTP/1 works fine with it

https://github.com/rack/rack/issues/1592#issuecomment-584459820

schinery:
@dblock I've thrown something together for both approaches, please let me know your thoughts and whether either of these is going in the right direction:

* Detect Rack version: https://github.com/ruby-grape/grape/pull/2355
* Use lowercase headers as default: https://github.com/ruby-grape/grape/pull/2356

dblock:
Thanks. Let's dig up why that was changed in Rack before deciding which way to go?

dblock:
[...]

PR #2355, opened 2023-10-16
Set response headers based on Rack version

schinery:
https://github.com/ruby-grape/grape/issues/2354


dblock:
I like this better than https://github.com/ruby-grape/grape/pull/2356. Same question as in that issue, let's dig up why this was changed.

schinery:
> One option would be to major-version increment Grape as part of this PR and do the same as in Rack

Do you need something adding to this PR for this? 

I also pushed fixing the integration spec paths and updated changelog.

dblock:
Let's see if other maintainers have an opinion. If we major increment the version here we can make the backwards incompatible change. I am 50/50 on what's best.

dblock:
@ioquatix would like your eyes and opinion here if you have a moment?

dblock:
@schinery looks like we may need to do extra work to support Rack 1.x.

schinery:
> @schinery looks like we may need to do extra work to support Rack 1.x.

Ah, seems that `Rack::RELEASE` was added after 1.x., but there is `Rack.release` in all 3 versions ([1.x.](https://github.com/rack/rack/blob/1-6-stable/lib/rack.rb#L22), [2.x.](https://github.com/rack/rack/blob/2-2-stable/lib/rack/version.rb#L26) and [3.x.](https://github.com/rack/rack/blob/3-0-stable/lib/rack/version.rb#L31)) so switched to use that for the version check.

schinery:
> @schinery Let's increment the version to 1.9 as part of this PR since at the very least it's a big change, and add a section to UPGRADING?

@dblock like this? https://github.com/ruby-grape/grape/pull/2355/commits/8c109338224a182043074f5406f7ef18b2f95631

CHANGELOG.md
```diff
@@ -1,13 +1,19 @@
...
 #### Features
 
-* [#2353](https://github.com/ruby-grape/grape/pull/2353): Added Rails 7.1 support - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 #### Fixes
 
 * Your contribution here.
+* 
+### 1.9.0 (2023/10/18)
```
dblock, line 11: We didn't release 1.9.0 yet, so remove this and say above `### 1.9.0 (Next)`.

README.md
```diff
@@ -160,9 +160,9 @@ content negotiation, versioning and much more.
 
 ## Stable Release
 
-You're reading the documentation for the next release of Grape, which should be **1.8.1**.
+You're reading the documentation for the next release of Grape, which should be **1.9.1**.
 Please read [UPGRADING](UPGRADING.md) when upgrading from a previous version.
-The current stable release is [1.8.0](https://github.com/ruby-grape/grape/blob/v1.8.0/README.md).
+The current stable release is [1.9.0](https://github.com/ruby-grape/grape/blob/v1.9.0/README.md).
```
dblock, line 165: Current stable is 1.8.0, revert this file change.

UPGRADING.md
```diff
@@ -1,6 +1,20 @@
...
+For Rack >= 3 the following response headers are returned lowercase:
+
+* `content-type`
+* `x-cascade`
+
+For Rack < 3 the following response headers are returned capitalized:
+
+* `Content-Type`
+* `X-Cascade`
+
```
dblock, line 17: Add a link for more information see this PR like below.

I think we need more text explaining what happens in what code. Give an example of an API that sets one of these headers, or explain when they are set?

UPGRADING.md
```diff
@@ -1,6 +1,20 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.9.0
+
+#### Response Headers
+
+For Rack >= 3 the following response headers are returned lowercase:
```
dblock, line 8: This should say something like "As per https://github.com/rack/rack/issues/1592 Rack 3.0 is enoforcing the HTTP/2 semantics, and thus treats all headers as lowercase. Starting with Grape 1.9.0, the following headers are now lowercase: ..."

schinery:
@dblock hopefully done all the PR feedback correctly, with the exception of "I think we need more text explaining what happens in what code. Give an example of an API that sets one of these headers, or explain when they are set?" as not 100% sure what you want for this.

Do you have an example in mind?

schinery:
Also, just going through the README, and wondering if the following needs updating:

* https://github.com/ruby-grape/grape#headers
* https://github.com/ruby-grape/grape#rails-3x

There is also https://github.com/ruby-grape/grape#header-case-handling, which this PR hasn’t altered as it has only changed the two hardcoded headers, but should there be a code change to lower case these for Rack 3 too?

dblock:
@schinery Yes to all your questions above about https://github.com/ruby-grape/grape/issues/2254 and related issues. I can merge as is, or do you think we alter all related behavior for Rack 3 (i.e. resolve #2254) in this one PR or separately?

schinery:
@dblock just trying to get a sense of the other changes needed to resolve #2254 to see if we do it in this PR, so far I have:

1: Update `#transform_header` in https://github.com/ruby-grape/grape/blob/master/lib/grape/request.rb#L49 to be Rack specific, so capitalize for Rack < 2, lowercase for Rack >= 3. Think this is fairly straightforward as long as we're ok to use the same Rack check used in the headers class?

2: Fix the specs, like https://github.com/ruby-grape/grape/blob/master/spec/grape/request_spec.rb#L121, so they are Rack version specific. Am guessing this would mean that they would go in the integration specs like we've added in this PR?

3: Update the README, such as https://github.com/ruby-grape/grape#header-case-handling, so that it has changes for both header casings.

4: Don't think we need to do anything for the `#header` method defined in https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/headers.rb#L11, as this is down to whatever people use for the header key.

Not sure what else, partly as this is my first time truly digging around the code. Anything else you think is needed?

schinery:
Does also make me wonder... would it be an idea to leave version 1 of Grape supporting Rack < 3 and create a version 2 of Grape that supports Rack >= 3. 

There would be more overhead for maintaining etc depending how long you want to maintain version 1, but the code would be cleaner without the Rack version detection code in it. 

Happy to continue with https://github.com/ruby-grape/grape/pull/2355/#issuecomment-1773821081 though, let me know your thoughts.

schinery:
@dblock more for https://github.com/ruby-grape/grape/pull/2355#issuecomment-1773821081

I threw [this PR](https://github.com/schinery/grape/pull/1) together quick just to look at what might need changing (minus the README changes), especially which specs would need updating. There were 5 specs that need the header in the right case based on what Rack version is used in the test run.

dblock:
> Does also make me wonder... would it be an idea to leave version 1 of Grape (1.x branch off of master) supporting Rack < 3 and create a version 2 of Grape (current master) that supports Rack >= 3.

I would be totally fine with incrementing major to 2.0 as part of this PR, and lock it at Rack >= 3, instead of trying to do both if you prefer?

I don't love the idea of also maintaining a 1.x. Are you interested in doing so? It's just a 1.x branch and releasing.

schinery:
>I would be totally fine with incrementing major to 2.0 as part of this PR, and lock it at Rack >= 3, instead of trying to do both if you prefer?

It just feels a bit cleaner. As you saw in https://github.com/schinery/grape/pull/1 (which again was more to highlight than a solution) some of the issues are in the testing, plus I guess you also have to have update the docs to support both Rack versions. 

> I don't love the idea of also maintaining a 1.x. Are you interested in doing so? It's just a 1.x branch and releasing.

Not really interested either, but I guess then you have something if you need to patch 1.x. in the future, and I guess then you could then just release a 1.8.1 or 1.9.0 release with https://github.com/ruby-grape/grape/pull/2360 and updated README etc, with any new features etc going in the 2.x version.

If it is decided that this is approach to take then is this what needs to happen?

* Update this PR to:
  * Bump Grape to 2.0.0
  * Lock Rack to >= 3
  * Potentially lock ActiveSupport to >= 7.1 (not sure if 100% needed but Rails 7.0 uses Rack 2.0)
  * Update the test matrix to `gemfile: [rack_3_0, rails_7_1]` only and clean out old gemfiles
  * Update the `transform_headers` to downcase only and update the specs where needed
  * Update the README, CHANGELOG and UPGRADING docs
* You'd create a 1.x branch and then we'd need a PR for that branch that:
  * Bumped Grape to 1.8.1 or 1.9.0
  * Lock Rack < 3
  * Potentially lock ActiveSupport to < 7.1
  * Update the test matrix to not include `rack_3_0` and  `rails_7_1`
   * Update the README, CHANGELOG and UPGRADING docs 

Does it feel like we need a second opinion on this too?

schinery:
One thing I was wrong about... it wasn't Rails 7.1 that bumped Rack to 3, as the deps in the Gemfile.lock are as follows:

```Gemfile
    actionpack (7.1.1)
      actionview (= 7.1.1)
      activesupport (= 7.1.1)
      nokogiri (>= 1.8.5)
      rack (>= 2.2.4)
      rack-session (>= 1.0.1)
      rack-test (>= 0.6.3)
      rails-dom-testing (~> 2.2)
      rails-html-sanitizer (~> 1.6)
```

However, I've just been playing about with locking to Rack 3 in a branch and I get this error using `rails_7_0.gemfile`. 

```
And because actionpack >= 7.0.5, < 7.1.0 depends on rack >= 2.2.4, < 3.A
  and every version of grape depends on rack >= 3.0.0,
  every version of grape is incompatible with rails >= 7.0.0, < 7.1.0.beta1.
So, because rails_7_0.gemfile depends on rails ~> 7.0.0
  and rails_7_0.gemfile depends on grape >= 0,
  version solving has failed.
```

So, if Grape was locked to Rack 3 it looks like potentially that version might only work with Rails 7.1. 

Which means I'm now flip flopping between whether Rack should be locked in Grape or not. Suspect not...

schinery:
I've updated https://github.com/schinery/grape/pull/1 with the additional changes I think are needed to support both Rack versions in Grape. I haven't addressed your additional comments in that PR, partly as my brain is tied in knots.

schinery:
Merged https://github.com/schinery/grape/pull/1 in. 

Will ponder the "a case-insensitive structure" for the request headers, would be ace if there was a way to do it but there is nothing screaming out at me at the moment.

dblock:
> Will ponder the "a case-insensitive structure" for the request headers, would be ace if there was a way to do it but there is nothing screaming out at me at the moment.

In https://github.com/schinery/grape/blob/51b081cef9d4d91cb75f80dd33222805612c7b95/lib/grape/request.rb#L38

This builds a `Hash`. You could try with `ActiveSupport::HashWithIndifferentAccess`, I don't remember if it's also case-insensitive.

UPGRADING.md
dblock: I still want an example here that shows what happens when I was doing `request.heaeders['Content-Type']` before and what I should be doing now. Should I change that code to `request.headers[Grape::Http::Headers::CONTENT_TYPE]` for example? What will the server return before/after?

lib/grape/request.rb
```diff
@@ -47,7 +47,11 @@ def build_headers
     end
 
     def transform_header(header)
-      -header[5..].split('_').each(&:capitalize!).join('-')
+      if Gem::Version.new(Rack.release) < Gem::Version.new('3')
```
dblock, line 50: This executes in a tight loop, which is bad for performance. You want this check outside. We also have it in a few places. It should look like this:

```ruby
if Grape::Rack.rack3?
  def transform_header(header)
      -header[5..].split('_').map(&:capitalize).join('-')
  end
else
  def transform_header(header)
      -header[5..].tr('_', '-').downcase
  end
end
```

spec/grape/api_spec.rb
```diff
@@ -689,7 +689,7 @@ class DummyFormatClass
         'example'
       end
       put '/example'
-      expect(last_response.headers['Content-Type']).to eql 'text/plain'
+      expect(last_response.headers[rack_versioned_headers[:content_type]]).to eql 'text/plain'
```
dblock, line 692: Can we get rid of `rack_versioned_headers` and instead do `Grape::HTTP::Headers::CONTENT_TYPE`?

spec/support/headers_helpers.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Spec
+  module Support
+    module Helpers
+      def rack_versioned_headers
```
dblock, line 6: Get rid of this class, use `Grape::HTTP::Headers::CACHE_CONTROL` and friends.

dblock (changes requested):
Getting there ...

lib/grape/request.rb
schinery: Added `.rack3?` method and done this change, although I've added it to `Grape` rather than `Grape::Rack` otherwise there was a load of Rack errors raised because we would need to change all the Rack references to `::Rack`.

Can do this though if really needs to be `Grape::Rack`

spec/support/headers_helpers.rb
schinery: The only issue with this is what to do in the specs that are testing with non `Grape::HTTP::Headers` headers, such as `Grape-Likes-Symbolic` or `X-Access-Token`, as these need to be Rack versioned in the specs. 

I've removed the headers helper though and put these inline, so they are defined in whatever specs needed them

schinery:
> > Will ponder the "a case-insensitive structure" for the request headers, would be ace if there was a way to do it but there is nothing screaming out at me at the moment.
> 
> In https://github.com/schinery/grape/blob/51b081cef9d4d91cb75f80dd33222805612c7b95/lib/grape/request.rb#L38
> 
> This builds a `Hash`. You could try with `ActiveSupport::HashWithIndifferentAccess`, I don't remember if it's also case-insensitive.

Yeah, it doesn't...

```sh
irb(main):001:0> hash = ActiveSupport::HashWithIndifferentAccess.new(a: 1)
=> {"a"=>1}
irb(main):002:0> hash["A"]
=> nil
irb(main):003:0>
```

Whether something like [this](https://stackoverflow.com/a/2030565/1517515) would work. I'm not sure what code changes would be required though.

spec/grape/api/custom_validations_spec.rb
```diff
@@ -162,13 +162,19 @@ def validate(request)
           return unless request.params.key? @attrs.first
           # check if admin flag is set to true
           return unless @option
+
           # check if user is admin or not
           # as an example get a token from request and check if it's admin or not
-          raise Grape::Exceptions::Validation.new(params: @attrs, message: 'Can not set Admin only field.') unless request.headers['X-Access-Token'] == 'admin'
+          raise Grape::Exceptions::Validation.new(params: @attrs, message: 'Can not set Admin only field.') unless request.headers[access_header] == 'admin'
```
dblock, line 168: Could use `x_access_token_header` here.

dblock (approved):
I'm going to merge and let it sit.

spec/grape/api/custom_validations_spec.rb
schinery: Yeah, tried that, but got...

```
     NameError:
       undefined local variable or method `x_access_token_header' for #<#<Class:0x0000ffff8f8f9b18>:0x0000ffff928d97e8 @attrs=[:admin_field], @option=true, @required=false, @scope=#<Grape::Validations::ParamsScope:0x0000ffff8f8f96b8 @element=nil, @element_renamed=nil, @parent=nil, @api=#<Class:0x0000ffff8f8f9898>, @optional=false, @type=Hash, @group=nil, @dependent_on=nil, @declared_params=nil, @index=nil>, @fail_fast=false, @allow_blank=nil>
```

which was annoying, as don't like the defining it twice either.

UPGRADING.md
```diff
@@ -1,6 +1,35 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 1.9.0
+
+#### Headers
+
+As per [rack/rack#1592](https://github.com/rack/rack/issues/1592) Rack 3.0 is enforcing the HTTP/2 semantics, and thus treats all headers as lowercase. Starting with Grape 1.9.0, headers will be cased based on what version of Rack you are using.
```
ioquatix, line 8: > Rack 3.0 is enforcing the HTTP/2 semantics, and thus treats all headers as lowercase. Starting with Grape 1.9.0, headers will be cased based on what version of Rack you are using.

Rack 3 is following the HTTP/2+ semantics which require header names to be lower case. To avoid compatibility issues, starting with Grape 1.9.0, headers will be cased based on what version of Rack you are using.

lib/grape.rb
```diff
@@ -42,6 +42,10 @@ def self.deprecator
     @deprecator ||= ActiveSupport::Deprecation.new('2.0', 'Grape')
   end
 
+  def self.rack3?
+    Gem::Version.new(::Rack.release) >= Gem::Version.new('3')
```
ioquatix, line 46: If you are just trying to detect whether to use lower case headers or not, I suggest a slightly different approach:

```ruby
LOWER_CASE_HEADERS = Rack::CONTENT_TYPE == "content-type"
```

You could then use that directly or in a method.

lib/grape/dsl/inside_route.rb
```diff
@@ -328,9 +328,9 @@ def sendfile(value = nil)
       def stream(value = nil)
         return if value.nil? && @stream.nil?
 
-        header 'Content-Length', nil
-        header 'Transfer-Encoding', nil
-        header 'Cache-Control', 'no-cache' # Skips ETag generation (reading the response up front)
+        header Grape::Http::Headers::CONTENT_LENGTH, nil
```
ioquatix, line 331: You can use `Rack::CONTENT_LENGTH` if that suits your use case.

ioquatix (commented):
This generally looks okay to me.

The biggest issue is, people trying to support both Rack 2 and Rack 3 will need to be careful about the assertions. Direct assertions against the hash can be problematic. It's important to have a proper normalization step in assertions IMHO, as the expectation should be "A header named X has a value Y", but the case of X can depend on the HTTP version, server, etc. In Rails, there were many instances of header key and value case which had ossified to whatever Rack (the gem) was generating, even thought that was only a subset of valid possible results.

I know this was a pretty annoying change in some regards, but overall it will simplify things and aligns the code with the HTTP specification more closely in a way which makes it easier to know that code is doing the right thing. e.g. https://github.com/rack/rack-test/issues/229

dblock:
Thanks @ioquatix ! @schinery will you extract the comments into another PR please?

schinery:
https://github.com/ruby-grape/grape/pull/2362

PR #2356, opened 2023-10-16
Lowercase content-type and x-cascade headers

schinery:
Questions:

* The files `spec/grape/entity_spec.rb` and `spec/grape/middleware/formatter_spec.rb` has specs testing for `Content-type`, should these be lower case?


dblock (changes requested):
@schinery There's likely a lot of people relying on the fact that these headers are capitalized the way they are in Grape. I am 100% sure I've written code like that. In the original issue you say "The X-Cascade header in Rack 3 applications appears to need to be lower case..." - can you please help find the source of that and why it was changed? If it's an RFC of some kind then we can convince ourselves it's the right change.

schinery:
Closing as https://github.com/ruby-grape/grape/pull/2355 seems to be the preferred approach

PR #2357, opened 2023-10-18
Updated class names in Custom Validators section of the README 

schinery:
Updated the Custom Validators section in the README as it appears that the validation files need to be called `SomethingValidator` and not just `Something`.

For example, I'd created a custom validation using the same setup as in the current README:

```ruby
class ValidUrl < Grape::Validations::Validators::Base
    #...
  end
end
```

This resulted in:

```sh
Grape::Exceptions::UnknownValidator:
  unknown validator: valid_url
```

When I renamed it to `ValidUrlValidator` then it worked as expected and the validator was used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1849,7 +1849,7 @@ end
 ### Custom Validators
 
 ```ruby
-class AlphaNumeric < Grape::Validations::Validators::Base
+class AlphaNumericValidator < Grape::Validations::Validators::Base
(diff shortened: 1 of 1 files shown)


schinery:
Closing as it seems I might be wrong on this, and the issue is where I've put the validator in the codebase 🤦🏻

PR #2360, opened 2023-10-21
Reduce gem size by removing test_files (spec)

ericproulx:
It occurred to me while installing the gem that spec files are included. It represents half the size of the gem. Nowadays, most  gem does not include them and I think Grape should do the same.


dblock (approved):
You'd think that it would know not to include `test_files`?

ericproulx:
> You'd think that it would know not to include `test_files`?

I know right! It's been deprecated for quite a long time now.

PR #2363, opened 2023-10-28
Introducing zeitwerk

ericproulx:
In the past, I tried to make some changes regarding autoload #2200, #2207, #2209. Unfortunately, #2207 and #2209 had to be reverted.

After reading this [article](https://railsatscale.com/2023-01-12-how-to-load-code-efficiently-in-ruby/), I thought it would be great to use zeitwerk instead of autoload.

### Benefits

1. manages the `autoload`
2. `eager_load` proof
3. no explicit requires except dependencies

### Changes
 - Applied the file convention structure
 - eager_load is done when loading `grape` instead of at compile!
 - `eager_load_spec` has been removed
 -  Removed the `validator` registration process (were loaded dynamically through `inherited`)
 -  Removed the cache for coercers (@collection_coercers)


dblock:
What's the downside?

ericproulx:
> What's the downside?

Follow the file structure convention if we can label that has a "downside".

UPGRADING.md
```diff
@@ -523,7 +523,7 @@ For PUT, POST, PATCH, and DELETE requests where a non-empty body and a "Content-
 
 Grape no longer uses `multi_json` or `multi_xml` by default and uses `JSON` and `ActiveSupport::XmlMini` instead. This has no visible impact on JSON processing, but the default behavior of the XML parser has changed. For example, an XML POST containing `<user>Bobby T.</user>` was parsed as `Bobby T.` with `multi_xml`, and as now parsed as `{"__content__"=>"Bobby T."}` with `XmlMini`.
 
-If you were using `MultiJson.load`, `MultiJson.dump` or `MultiXml.parse`, you can substitute those with `Grape::Json.load`, `Grape::Json.dump`, `::Grape::Xml.parse`, or directly with `JSON.load`, `JSON.dump`, `XmlMini.parse`, etc.
+If you were using `MultiJson.load`, `MultiJson.dump` or `MultiXml.parse`, you can substitute those with `Grape::Util::Json.load`, `Grape::Util::Json.dump`, `::Grape::Util::Xml.parse`, or directly with `JSON.load`, `JSON.dump`, `XmlMini.parse`, etc.
```
dblock, line 526: Are the namespace changes of `Grape::Json` to `Grape::Util::Json` breaking or does `Grape::Json` still work? 

Either way add a new section in UPGRADING that says to use `Grape::Util::Json` instead of `Grape::Json`.

dblock:
Rebase? I'll take a look.

UPGRADING.md
ericproulx: I can keep the old namespace. I'll just need to move files from the `utils` folder and remove the `utils` modules.

lib/grape.rb
```diff
@@ -27,292 +19,56 @@
...
+require 'set'
+require 'singleton'
+require 'zeitwerk'
+
+loader = Zeitwerk::Loader.for_gem
+loader.inflector.inflect(
+  'api' => 'API',
+  'dsl' => 'DSL'
+)
+railtie = "#{__dir__}/grape/railtie.rb"
```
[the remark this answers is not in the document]
ericproulx: The file is `require` at the end if Rails version >= 7.1. This is why I configured zeitwerk to not load the file

dblock (approved):
I'd like to merge this. 

1. Update the language around auto/lazy loading/reloading in README.
2. What externally visible contracts are we breaking? Should we be incrementing the version to 3.0?

ericproulx:
> I'd like to merge this.
> 
>     1. Update the language around auto/lazy loading/reloading in README.
> 
>     2. What externally visible contracts are we breaking? Should we be incrementing the version to 3.0?

There's a breaking change with custom validator. Now, they must follow the convention `Grape::Validations::Validators::[CustomName]` The registration part was done through inheritance. I'll put it back.

dblock:
> There's a breaking change with custom validator. Now, they must follow the convention `Grape::Validations::Validators::[CustomName]` The registration part was done through inheritance. I'll put it back.

Sounds good, I'll wait to merge.

ericproulx:
@dblock I've found something else related to gem `grape-swagger`. Versions below 2.0 are  ` grape ~> 1.3` and `ruby >= 2.7`. Above 2.0, its `ruby >= 3.0` and `'grape', '>= 1.7', '< 3.0'`.

Next Grape release is `2.1.0` and `Grape` is still `ruby >= 2.7.0`.  Should they follow each other ?

dblock:
> Next Grape release is `2.1.0` and `Grape` is still `ruby >= 2.7.0`. Should they follow each other related to `ruby` version ?

I don't think they have to. Generally we remove rubies going EOL or if there's "good reason". If you can spell it out in a PR we can do whatever :)

dblock:
The code looks good here. I think we still need to say something in UGPRADING/README about auto-loading? WDYT? It feels like a scary change to go "quietly" :)

dblock:
@ericproulx Want to add some words to those files and let's merge?

ericproulx:
> @ericproulx Want to add some words to those files and let's merge?

Done. I mentionned something about Moneky Patch that might result in a Zeitwerk Error.

PR #2366, opened 2023-11-02
Quality arguments to `Accept` header should default to 1.0

hiddewie:
This PR ensures that Accept headers with mime types without a given quality value, are parsed as having a quality value of `1.0`.

The current value for mime types without a quality value is 0.0 (`nil.to_f`).

From the HTTP specification (RFC 9110), https://www.rfc-editor.org/rfc/rfc9110.html#name-accept and https://www.rfc-editor.org/rfc/rfc9110.html#name-quality-values:

> The weight is normalized to a real number in the range 0 through 1, where 0.001 is the least preferred and 1 is the most preferred; a value of 0 means "not acceptable". If no "q" parameter is present, the default weight is 1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 #### Fixes
 
 * [#2364](https://github.com/ruby-grape/grape/pull/2364): Add missing requires - [@ericproulx](https://github.com/ericproulx).
+* [#2366](https://github.com/ruby-grape/grape/pull/2366): Quality arguments to `accept` header should default to 1.0 - [@hiddewie](https://github.com/hiddewie).
 * Your contribution here.
 
 ### 1.8.0 (2023/08/30)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,11 @@
 * [#2360](https://github.com/ruby-grape/grape/pull/2360): Reduce gem size by removing specs - [@ericproulx](https://github.com/ericproulx).
 * [#2361](https://github.com/ruby-grape/grape/pull/2361): Remove `Rack::Auth::Digest` - [@ninoseki](https://github.com/ninoseki).
 * Your contribution here.
-  
+
 #### Fixes
 
 * [#2364](https://github.com/ruby-grape/grape/pull/2364): Add missing requires - [@ericproulx](https://github.com/ericproulx).
-* [#2366](https://github.com/ruby-grape/grape/pull/2366): Quality arguments to `accept` header should default to 1.0 - [@hiddewie](https://github.com/hiddewie).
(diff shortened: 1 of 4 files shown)


lib/grape/middleware/formatter.rb
```diff
@@ -171,7 +171,7 @@ def mime_array
         vendor_prefix_pattern = /vnd\.[^+]+\+/
 
         accept.scan(accept_into_mime_and_quality)
-              .sort_by { |_, quality_preference| -quality_preference.to_f }
+              .sort_by { |_, quality_preference| -(quality_preference || '1.0').to_f }
```
dblock, line 174: Does it matter that it doesn't work with `quality_preference = ''`:

```ruby
3.1.2 :012 > quality_preference = ''
 => "" 
3.1.2 :013 > (quality_preference || '1.0').to_f
 => 0.0 
```

If it's really about `nil` (and not an explicitly passed in 0.0), then I would just do this to avoid parsing the 1.0 float every time.

```
quality_preference ? quality_preference.to_f : 1.0
```

dblock (changes requested):

dblock (commented):
Add a test that with `nil`, `0.0`, and `invalid`, too?

lib/grape/middleware/formatter.rb
hiddewie: Yes, sounds good!

hiddewie:
Thanks for the review. I changed the parsing to include all `q=...` values. This will try to parse every value to a floating point number.  Invalid (textual) values will be parsed as 0.0. The empty `q` value will become a default quality of 1.0 which looks like a fine behaviour for me.

dblock:
Add to CHANGELOG and this is good to go!

hiddewie:
Thanks, done!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 #### Fixes
 
 * [#2364](https://github.com/ruby-grape/grape/pull/2364): Add missing requires - [@ericproulx](https://github.com/ericproulx).
+* [#2366](https://github.com/ruby-grape/grape/pull/2366): Quality arguments to `accept` header should default to 1.0 - [@hiddewie](https://github.com/hiddewie).
```
dblock, line 14: Let's say "Default quality to 1.0 in the Accept header when omitted".

dblock (commented):
One more, should we also document this in README? We don't need to explain the entire purpose of `q`, but maybe call it out in https://github.com/ruby-grape/grape#header with an example that chooses JSON over XML? Maybe something like this?

"Grape will evaluate the relative quality preference included in Accept headers and default to a quality of 1.0 when omitted. In the following example a Grape API that supports XML and JSON in that order will return JSON."

```
curl ... -H "Accept: text/xml;q=0.8, application/json;q=0.9"
```

hiddewie:
Sounds good, done!

dblock (approved):

dblock:
Merged, nice work, thanks for hanging in here with me!

PR #2370, opened 2023-11-14
method missing to_ary

ericproulx:
While profiling #2363, I found that `to_ary` is leaking as `Allocated String Report`.
It has been fixed in by #466 but lost since.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,8 +160,9 @@ content negotiation, versioning and much more.
 
 ## Stable Release
 
-You're reading the documentation for the stable release of Grape, **2.0.0**.
+You're reading the documentation for the next release of Grape, which should be **2.0.1**.
 Please read [UPGRADING](UPGRADING.md) when upgrading from a previous version.
+The current stable release is [2.0.0](https://github.com/ruby-grape/grape/blob/v2.0.0/README.md).
 
 
 ## Project Resources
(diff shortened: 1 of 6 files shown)


dblock (changes requested):
When does this happen? Add a test or we'll break it again!

ericproulx:
> When does this happen? Add a test or we'll break it again!

Its happening on implicit conversions. Anyway, the `method_missing` was introduced in [2016](https://github.com/ruby-grape/grape/blob/v0.16.0/UPGRADING.md#upgrading-to--0160) to deprecate `route_xyz` methods. I would consider remove the entire missing and warning part. That way, we won't have to deal with the `to_ary`

dblock:
Should we add a section into UPGRADING and save someone who wasn't paying attention to the deprecation some trouble?

README.md
```diff
@@ -3506,7 +3506,7 @@ TwitterAPI::routes[0].description # => 'Includes custom settings.'
 TwitterAPI::routes[0].settings[:custom] # => { key: 'value' }
 ```
 
-Note that `Route#route_xyz` methods have been deprecated since 0.15.0.
+Note that `Route#route_xyz` methods have been deprecated since 0.15.0 and removed since 2.0.1
```
dblock, line 3509: Add missing period.

dblock (approved):
One more period and let's merge. Thanks for doing this.

Issue #2290, opened 2022-12-16
Use param foo in default for bar


jamesarosen:
What I'd like to do:
```rb
params do
  requires :foo
  optional :bar, default: -> (other_params) { "bar from foo #{other_params[:foo]} }
end
```
or perhaps
```rb
params do
  requires :foo
  optional :bar, coerce: -> (bar_value, other_params) { "bar from foo #{other_params[:foo]} }
end
```

Unfortunately, neither `default` nor `coerce` seems to accept the other params as an argument. It may be possible to get them from `self` (which is a `ParamsScope`) within that block, but it's not obvious how.

dblock:
Does `params[:foo]` work? If not this is a great feature request.

dblock:
Closed via https://github.com/ruby-grape/grape/pull/2371.

PR #2371, opened 2023-11-17
feature(#2290): Use a param value as the `default` value of other param

jcagarcia:
Hey! 👋 

I think the feature requested in #2290 is pretty interesting. So I've added this PR for allowing to use a param value as the `default` value of some other parameter.

So now, we can do:

```ruby
params do
  optional :color, type: String, default: 'blue'
  optional :primary_color, type: String, default: -> (params) { params[:color] }
end
```

I've updated the README for including this new feature. Don't know if the next version should be `2.1.0` instead of `2.0.1`. Let me know if I should change it.

Thanks!!


dblock (approved):
Very nice! Yes, let's make this 2.1.0 in this PR. Thanks.

jcagarcia:
Thanks @dblock ! I've just updated the version to `2.1.0`. Let me know if anything else is needed 👍

dblock:
Stable release mentions 2.0.1 too, needs to be 2.1.0, thanks!

https://github.com/ruby-grape/grape#stable-release

jcagarcia:
Sure! I've just commit the change for updating the stable release section :) Hope it helps!

README.md
```diff
@@ -1242,6 +1242,17 @@ params do
 end
 ```
 
+You can use the value of one parameter as the default value of some other parameter. In this
+case, if the `primary_color` parameter is not provided, it will have the same value as the `color`
+one. If both of them not provided, both of them will have `blue` value.
```
dblock, line 1247: One last thing, I promise that's it after that. We don't wrap text in markdown manually since it shows up, make this all 1-line.

jcagarcia: Hi @dblock ! Don't worry at all, I really appreciate the reviews :) 

I've just updated my addition to the README file for following 1-line format. However, I did it in that way because all the other sections of the document. 

As you can see here:

https://github.com/ruby-grape/grape/blob/master/README.md?plain=1#L153-L159

If the 1-line is the new convention you want to follow, do you think I can help on moving all the README to 1-line format?

dblock: Absolutely. Maybe we can find a linter to run in PRs too?

dblock:
Merged, thanks for hanging in here with me!

README.md
jcagarcia: Sure! Let me try to PR this :)

Issue #2195, opened 2021-10-18
Declared broken for hash params with overlapping names


johnk87:
Upgrading from 1.3.3, noticed several bugs around `declared`. I think some of them already have issues/PRs (eg. https://github.com/ruby-grape/grape/pull/2112 https://github.com/ruby-grape/grape/pull/2001) but don't see this one:

when passing `null` to hash params with unique names, the output is consistent (but still incorrect I think, but that's a different issue, see a note at the bottom):
```ruby
gem 'grape', '1.6.0'

require 'grape'

class API < Grape::API
  format :json

  params do
    optional(:aa, type: Hash)
    optional(:ab, type: Hash)
  end

  put do
    {
      params: params,
      declared: declared(params, include_missing: false)
    }
  end
end

run API

```

```shell
> curl -X PUT http://127.0.0.1:9292 -H 'Content-Type: application/json' -d '{"aa": null, "ab": null}' | jq
{
  "params": {
    "aa": null,
    "ab": null
  },
  "declared": {
    "aa": {},
    "ab": {}
  }
}
```

but when the param names overlap:
```ruby
params do
  optional(:aa, type: Hash)
  optional(:aab, type: Hash)
end
```

```shell
> curl -X PUT http://127.0.0.1:9292 -H 'Content-Type: application/json' -d '{"aa": null, "aab": null}' | jq
{
  "params": {
    "aa": null,
    "aab": null
  },
  "declared": {
    "aa": null, # expected {}
    "aab": {}
  }
}
```

Present since 1.5.0, so it seems to be introduced by https://github.com/ruby-grape/grape/pull/2103, my guess is in this line https://github.com/ruby-grape/grape/blob/43936ac719ee524fd0f8ccd830d1eb50abd721c7/lib/grape/dsl/inside_route.rb#L98

---
Other than that: 
[Upgrading to >= 1.5.0](https://github.com/ruby-grape/grape/blob/master/UPGRADING.md#upgrading-to--150) says behaviour changes only `when params are missing and include_missing=true`
[Upgrading to >= 1.3.3](https://github.com/ruby-grape/grape/blob/master/UPGRADING.md#upgrading-to--133) says that `For now on, nil values stay nil values for all types, including arrays, sets and hashes.`

so I assume changing `nil` to `{}` in the first example when `null` is passed explicitly (and `include_missing` isn't `true` as well) is an unintended behaviour as well?

[...]

PR #2372, opened 2023-11-17
fix(#2195): Fix `declared` method for hash params with overlapping names

jcagarcia:
Hi 👋 

I was taking a look to the issue #2195 .

Seems the error is because the logic inside the `handle_passed_param` method is using the `start_with?(key)` method. So, if we have two params like `:empty_hash` and `:empty_hash_two`, as both of them share part of the key, it detects that the `:empty_hash` parameter has children, and it behaves in a different way and it ends returning a `nil` value.

For solving it, I'm checking that for having children the key of the children should start with the same key name + a `[`. Maybe not the better solution for checking this 🤔 so open to discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


dblock:
This is simple and all specs pass, so let's go for it.

PR #2373, opened 2023-11-19
fix: Updating markdown files for following one-line format

jcagarcia:
From a conversation with @dblock in https://github.com/ruby-grape/grape/pull/2371#discussion_r1398210065 I've manually fixed the markdown files for following 1-line format.

I've tried to add the [rubocop-md](https://github.com/rubocop/rubocop-md) gem, so it can be detected in future PRs. However, there is not a specific rule for this purpose.

An alternative would be add this validation to your https://github.com/ruby-grape/danger project. 

What do you think? Do you think is interesting to merge this "fix"?


dblock (approved):
Merging. A danger plugin could be a good idea, but maybe a proper markdown linter w/option is better.

PR #1944, opened 2019-12-14
AttributeTranslator optimization

ericproulx:
Still memory profiling
```log
Allocated String Report
-----------------------------------
      7640  "d"
      3820  /Users/eproulx/.rbenv/versions/2.4.9/lib/ruby/gems/2.4.0/gems/grape-1.2.6/lib/grape/router/attribute_translator.rb:16
      3820  /Users/eproulx/.rbenv/versions/2.4.9/lib/ruby/gems/2.4.0/gems/grape-1.2.6/lib/grape/router/attribute_translator.rb:18

      1248  "s"
       624  /Users/eproulx/.rbenv/versions/2.4.9/lib/ruby/gems/2.4.0/gems/grape-1.2.6/lib/grape/router/attribute_translator.rb:16
       624  /Users/eproulx/.rbenv/versions/2.4.9/lib/ruby/gems/2.4.0/gems/grape-1.2.6/lib/grape/router/attribute_translator.rb:18

       936  "p"
       468  /Users/eproulx/.rbenv/versions/2.4.9/lib/ruby/gems/2.4.0/gems/grape-1.2.6/lib/grape/router/attribute_translator.rb:16
       468  /Users/eproulx/.rbenv/versions/2.4.9/lib/ruby/gems/2.4.0/gems/grape-1.2.6/lib/grape/router/attribute_translator.rb:18
```
d = [request_method](https://github.com/ruby-grape/grape/blob/1d35559084432883f335cdea4b483ddb6dd8d53a/lib/grape/router/route.rb#L73)
s = `:requirements`
p = [regexp](https://github.com/ruby-grape/grape/blob/1d35559084432883f335cdea4b483ddb6dd8d53a/lib/grape/router.rb#L54)

Instead of using `method_missing` to fetch the data, I thought it would be better to define the methods

PR #2375, opened 2023-11-20
Fix attribute translator setter

Jell:
This was simply not working before. Doesn't seem like there are any unit tests for this class, so I haven't added new ones, but might be worth considering.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 #### Fixes
 
+* [#2375](https://github.com/ruby-grape/grape/pull/2375): Fix setter methods for `Grape::Router::AttributeTranslator` - [@Jell](https://github.com/Jell).
 * [#2370](https://github.com/ruby-grape/grape/pull/2370): Remove route_xyz method_missing deprecation - [@ericproulx](https://github.com/ericproulx).
 * [#2372](https://github.com/ruby-grape/grape/pull/2372): Fix `declared` method for hash params with overlapping names - [@jcagarcia](https://github.com/jcagarcia).
 * [#2373](https://github.com/ruby-grape/grape/pull/2373): Fix markdown files for following 1-line format - [@jcagarcia](https://github.com/jcagarcia).
--- a/lib/grape/router/attribute_translator.rb
+++ b/lib/grape/router/attribute_translator.rb
@@ -38,15 +38,15 @@ module Grape
       end
 
       def method_missing(method_name, *args)
-        if setter?(method_name[-1])
-          attributes[method_name[0..]] = *args
+        if setter?(method_name)
+          attributes[method_name.to_s.chomp('=').to_sym] = args.first
         else
           attributes[method_name]
         end
       end
 
       def respond_to_missing?(method_name, _include_private = false)
-        if setter?(method_name[-1])
+        if setter?(method_name)
           true
         else
           @attributes.key?(method_name)
@@ -56,7 +56,7 @@ module Grape
       private
 
       def setter?(method_name)
-        method_name[-1] == '='
+        method_name.end_with?('=')
       end
     end
   end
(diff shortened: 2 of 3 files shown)


dblock (changes requested):
Thanks, and yes, please add a test / CHANGELOG.

dblock:
Not sure if failed tests are related, but I suspect they are.

ericproulx:
`ends_with?` on symbol comes in 2.7 but we're still >= 2.6.
https://blog.saeloun.com/2019/12/09/ruby-2-7-added-symbol-start_with-and-symbol-end_with-method
@dblock I think we should consider deprecating 2.6. After all, rubocop dropped it and rubygems-update also. Might be a good time.

dblock:
I am not against it.

ericproulx:
@Jell if you rebase your branch, all tests should pass.

Jell:
@ericproulx @dblock I've rebased, added tests and a line in the changelog

ericproulx:
@dblock I thought I messed up my [refactoring](https://github.com/ruby-grape/grape/pull/1944) in 2019 but it always been like this since #1351. What @Jell did makes total sense but it changes the setter behavior. For instance,

```ruby
describe "#header=" do
  subject { instance.header = 'new_value' }
  let(:instance) { Grape::Router::AttributeTranslator.new(header: 'value') }
  it "sets value for header" do
    expect do
      subject
      puts "to_h = #{instance.to_h}"
    end.to change(instance, :header).from('value').to('new_value')
  end
end
```

```
Grape::Router::AttributeTranslator
  #header=
to_h = {:header=>"value", "header="=>["new_value"]}
    sets value for header (FAILED - 1)

Failures:

  1) Grape::Router::AttributeTranslator#header= sets value for header
     Failure/Error:
       expect do
         subject
         puts "to_h = #{instance.to_h}"
       end.to change(instance, :header).from('value').to('new_value')
     
       expected `Grape::Router::AttributeTranslator#header` to have changed from "value" to "new_value", but did not change
     # ./spec/grape/router/attribute_translator_spec.rb:32:in `block (3 levels) in <top (required)>'

Finished in 0.0123 seconds (files took 0.28833 seconds to load)
1 example, 1 failure
```

On the other hand, I feel the setter methods aren't really used and everything is done through the initialize. For instance, removing the `setter` part in the `method_missing` and `respond_to_missing?` like this doesn't break any specs.

```ruby
def method_missing(method_name, *args)
  return attributes[method_name] unless setter?(method_name[-1])

  super
end

def respond_to_missing?(method_name, _include_private = false)
  @attributes.key?(method_name)
end
```

Maybe we should consider removing the setter part and let the initialize do its work. After all, these attributes doesn't change after the APIs are compiled.

Jell:
@ericproulx FYI the reason for me doing the PR in the first place is because I wanted to set some default headers for Swagger documentation purposes after the route was compiled, something along the lines of:

```
routes.each do |route|
  add_default_headers(route)
  add_default_response_codes(route)end

**Comment by ericproulx** [comment]
@dblock #1348 your ruby [bug](https://bugs.ruby-lang.org/issues/12251) has been fixed 3 years ago. It starts with Ruby 3.0. The issue is still present in 2.7

File: CHANGELOG.md [hunk 1]
```diff
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#2375](https://github.com/ruby-grape/grape/pull/2375): Fix setter methods for `Grape::Router::AttributeTranslator` - [@Jell](https://github.com/Jell)
```
dblock, line 5: It's still a fix even though it never worked. Move it down to fixes?

dblock (approved):

CHANGELOG.md
dblock: It's also missing a period at the end ;)

Jell: ah of course 🤦

Jell: done!

dblock:
Good work @Jell!

Issue #1922, opened 2019-10-23
helpers are not available inside the rescue_from block - Why?


aruprakshit:
I have current_user defined inside the before hook, and then we try to read it from `rescue_from` method, we get `nil`.

```ruby
before do
  token         = nil
  token         = headers.fetch("Authorization", "").split("Basic ")[1] unless report_url?
  @current_user = ::User.find_by(auth_token: token) if token.present?

  PaperTrail.request.whodunnit = @current_user.id if @current_user
end

helpers Pundit

helpers do
  attr_reader :current_user

  def report_url?
    request.path_info == "/v1/share_reports/#{request.params['token']}"
  end

  # ....
end


rescue_from Pundit::NotAuthorizedError do |e|
  Airbrake.notify(e, {
    user: {
      id: current_user.id,
      email: current_user.email
    }
  })

  error!(
    {
      messages: {
        userNotification: e.message
      }
    }, 401
  )
end
```

We get `nil` error when `current_user` being accessed inside the `rescue_from` block as `current_user.id`. How can I make available `current_user` inside the `rescue_from` block?

dblock:
I *think* I'd want this to work. Doing `@something = ` in `before` makes it available in the execution of the API, but not in `rescue_from`. Let's call it a bug or a non-feature. Maybe someone could try to fix it?

As a workaround, the rescue block is not executed in the same scope as the request, so you don't have access to attributes. You do have access to `context` (since https://github.com/ruby-grape/grape/pull/1918, in a released version you can reach into ENV), which gives you the endpoint and then gives you endpoint-based context and settings. I think you generally want to do what Warden does, which is using ENV to pass around something across various contexts of the request (before, during, after) and not `@`.

dblock:
Yep. I will leave this open to make `@variable` work. At the very least we should document what the scope of `@` is in a `rescue_from` block. 

@aruprakshit This is a good place for someone to write some tests and contribute :) Wink wink

dblock:
Thanks for the detailed explanation! I don't love copying instance variables at all because modifying them will not work across copies, so I think fixing `self` is the way to go. I understand the other concerns.

jcagarcia:
Totally agree @dblock . I've just tried to copy the instance variables from one instance to another to see if after that they were accessible. 

I've applied the `context` fix solution in #2377

PR #2377, opened 2023-11-21
fix(#1922): Allow to use instance variables defined in the endpoints inside rescue_from

jcagarcia:
Hey 👋 

I was taking a look at #1922 and after some checks I can see that we are in a different instance when we are inside the `rescue_from` method.

```ruby
require 'grape'

class MyAPI < Grape::API
  class MyError < StandardError; end

  before do
    puts "self => #{self.class}"
    @my_var = 328
  end

  rescue_from MyError do |e|
    puts "self => #{self.class}"
    puts "my var value => #{@my_var}"
    error!({ my_var: @my_var }, 401)
  end

  get '/wadus' do
    puts "my var value => #{@my_var}"
    puts "self => #{self.class}"
    raise MyError.new
  end
end
```

The current output of the previous code is:

```shell
self =>#<Class:0x0000000107c79d88>
my var value => 328
self => #<Class:0x0000000107c79d88>
self => ##<Class:0x0000000104c77950>
my var value =>
```

This means that we cannot have access to the same instance variables declared in the `before` because we are not in the same instance.

For solving it, I'm adding the scope of the endpoint when performing the handler. So now I'm performing `endpoint.instance_exec(&handler)` instead of `instance_exec(&handler)` where the `endpoint` is the instance obtained from `@env["api.endpoint"]` [here](https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/error.rb#L36-L49). 

Doing it in that way, we are executing the handler under the endpoint instance scope and we have the variables accessible. 

```shell
self => #<Class:0x000000010448a148>
my var value => 328
self => #<Class:0x000000010448a148>
self => #<Class:0x000000010448a148>
my var value => 328
```

However, there are some trade-offs that we need to fix in this issue in order to pass all the tests.

As now we are performing the handler under the `endpoint` instance, the methods called inside the `rescue_from` should be defined in the `inside_route.rb` file. I've experienced problems with two of those methods:

* The `error!` method. Is defined both times in `inside_routes.rb` and in the `error.rb`. However, the one defined in the `inside_routes.rb` is throwing an error and the one defined in the `error.rb` is just returning a rack response. 
  * **Proposed Solution**: Catch the thrown error when performing the handler and if an error appears because someone calls the `error!` method, call the `error!` method from the `error.rb` again for returning the rack response. Check [this](https://github.com/ruby-grape/grape/compare/master...jcagarcia:issue-1922?expand=1#diff-4dd2d3b9eb97044b27af6778457a4fb2d0991647867f4a6a728919038d3d06c2R134)
* The `rack_response` method is used in several tests inside the `rescue_from`. However, this method is not defined in the `inside_route.rb` file, what causes that a `NoMethodError` is raised.
  * **Proposed Solution**: Define the `rack_response` method in the `inside_route.rb` file, so now endpoints can also perform this method for building a rack response.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -11,9 +11,45 @@ Upgrading Grape
 
 #### Instance variables scope
 
-Due to the changes done in [#2377](https://github.com/ruby-grape/grape/pull/2377), the instance variables defined inside each of the endpoints (or inside a `before` validator) are now accessible inside the `rescue_from`. This means the scope of the instance variables has changed.
+Due to the changes done in [#2377](https://github.com/ruby-grape/grape/pull/2377), the instance variables defined inside each of the endpoints (or inside a `before` validator) are now accessible inside the `rescue_from`. The behavior of the instance variables was undefined until `2.1.0`.
 
 If you were using the same variable name defined inside an endpoint or `before` validator inside a `rescue_from` handler, you need to take in mind that you can start getting different values or you can be overriding values.
+
+Before:
+```ruby
+class TwitterAPI < Grape::API
+  before do
+    @var = 1
+  end
+
+  get '/' do
+    puts @var # => 1
+    raise
+  end
+
+  rescue_from :all do
+    puts @var # => nil
+  end
+end
+```
+
+After:
+```ruby
+class TwitterAPI < Grape::API
+  before do
+    @var = 1
+  end
+
+  get '/' do
+    puts @var # => 1
+    raise
+  end
+
+  rescue_from :all do
+    puts @var # => 1
+  end
+end
+```
 
 ### Upgrading to >= 2.0.0
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,13 +1,19 @@
 Upgrading Grape
 ===============
 
-### Upgrading to >= 2.0.1
+### Upgrading to >= 2.1.0
 
 #### Grape::Router::Route.route_xxx methods have been removed
 
 - `route_method` is accessible through `request_method`
 - `route_path` is accessible through `path`
 - Any other `route_xyz` are accessible through `options[xyz]`
+
+#### Instance variables scope
+
+Due to the changes done in [#2377](https://github.com/ruby-grape/grape/pull/2377), the instance variables defined inside each of the endpoints (or inside a `before` validator) are now accessible inside the `rescue_from`. This means the scope of the instance variables has changed.
+
+If you were using the same variable name defined inside an endpoint or `before` validator inside a `rescue_from` handler, you need to take in mind that you can start getting different values or you can be overriding values.
 
 ### Upgrading to >= 2.0.0
 
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -167,6 +167,23 @@ module Grape
         throw :error, message: message, status: self.status, headers: headers
       end
 
+      # Creates a Rack response based on the provided message, status, and headers.
+      # The content type in the headers is set to the default content type unless provided.
+      # The message is HTML-escaped if the content type is 'text/html'.
+      #
+      # @param message [String] The content of the response.
+      # @param status [Integer] The HTTP status code.
+      # @params headers [Hash] (optional) Headers for the response
+      #                      (default: {Rack::CONTENT_TYPE => content_type}).
+      #
+      # Returns:
+      # A Rack::Response object containing the specified message, status, and headers.
+      #
+      def rack_response(message, status, headers = { Rack::CONTENT_TYPE => content_type })
+        message = ERB::Util.html_escape(message) if headers[Rack::CONTENT_TYPE] == 'text/html'
+        Rack::Response.new([message], Rack::Utils.status_code(status), headers)
+      end
+
       # Redirect to a new url.
       #
       # @param url [String] The url to be redirect.
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -179,7 +179,7 @@ module Grape
       # Returns:
       # A Rack::Response object containing the specified message, status, and headers.
       #
-      def rack_response(message, status, headers = { Rack::CONTENT_TYPE => content_type })
+      def rack_response(message, status = 200, headers = { Rack::CONTENT_TYPE => content_type })
         message = ERB::Util.html_escape(message) if headers[Rack::CONTENT_TYPE] == 'text/html'
         Rack::Response.new([message], Rack::Utils.status_code(status), headers)
       end
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -46,7 +46,7 @@ module Grape
             rescue_handler_for_any_class(e.class) ||
             raise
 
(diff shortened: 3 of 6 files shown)


lib/grape/middleware/error.rb
```diff
@@ -119,21 +119,29 @@ def rescue_handler_for_any_class(klass)
...
 
           handler = public_method(handler)
         end
 
-        response = handler.arity.zero? ? instance_exec(&handler) : instance_exec(error, &handler)
+        response = (catch(:error) do
+          handler.arity.zero? ? endpoint.instance_exec(&handler) : endpoint.instance_exec(error, &handler)
+        end)
+
+        response = error!(response[:message], response[:status], response[:headers]) if error?(response)
```
jcagarcia, line 133: As now we are performing the handler under the endpoint instance, the methods called inside the `rescue_from` should be defined in the `inside_route.rb` file. I've experienced problems with this `error!` method:

The `error!` method. Is defined both times in `inside_routes.rb` and in the `error.rb`. However, the one defined in the `inside_routes.rb` is throwing an error and the one defined in the `error.rb` is just returning a rack response.

**Proposed Solution**: Catch the thrown error when performing the handler and if an error appears because someone calls the error! method, call the error! method from the `error.rb` again for returning the rack response.

lib/grape/dsl/inside_route.rb
```diff
@@ -167,6 +167,23 @@ def error!(message, status = nil, additional_headers = nil)
...
+      #
+      # @param message [String] The content of the response.
+      # @param status [Integer] The HTTP status code.
+      # @params headers [Hash] (optional) Headers for the response
+      #                      (default: {Rack::CONTENT_TYPE => content_type}).
+      #
+      # Returns:
+      # A Rack::Response object containing the specified message, status, and headers.
+      #
+      def rack_response(message, status = 200, headers = { Rack::CONTENT_TYPE => content_type })
```
jcagarcia, line 182: As now we are performing the handler under the endpoint instance, the methods called inside the `rescue_from` should be defined in the `inside_route.rb` file. I've experienced problems with the `rack_response` method:

The `rack_response` method is used in several tests inside the `rescue_from`. However, this method is not defined in the `inside_route.rb` file, what causes that a `NoMethodError` is raised.

**Proposed Solution**: Define the `rack_response` method in the `inside_route.rb` file, so now endpoints can also perform this method for building a rack response.

dblock (commented):
This looks very good! No other specs are failing, which means we're on the right track to defining this behavior.

1. What happens if I do `@var = 1` in the first call, does it exist in the second call to the same API? What's the scope of `@var` across multiple API methods? All these need tests.
3. Document this behavior in README.
4. The meaning of `@var` has changed in `rescue_from`, it will need an UPGRADING explanation. No specs were broken, but is it a breaking change?

jcagarcia:
Hi @dblock  👋 thanks!

For the first point:

If you do `@var = 1` in the first call, `@var` should not be available in a second call. This is because a new instance is created of each of the different endpoints. Here an example:

```ruby
class MyAPI < Grape::API
  class MyError < StandardError; end

  rescue_from MyError do |e|
    puts "rescue_from_self => #{self.class}"
    puts "rescue_from_my_var => #{@my_var}"
    error!({ my_var: @my_var }, 401)
  end

  get '/first' do
    @my_var = 1
    puts "my_var => #{@my_var}"
    puts "self => #{self.class}"
    raise MyError.new
  end

  get '/second' do
    puts "my_var => #{@my_var}"
    puts "self => #{self.class}"
    raise MyError.new
  end
end

```
 And here the output after performing `/first`, `/second` and `/first` again.

```shell
my_var => 1
self => #<Class:0x0000000105ffa518>
rescue_from_self => #<Class:0x0000000105ffa518>
rescue_from_my_var => 1
::1 - - [22/Nov/2023:16:34:44 +0100] "GET /first HTTP/1.1" 401 12 0.0334
my_var =>
self => #<Class:0x0000000105ff9578>
rescue_from_self => #<Class:0x0000000105ff9578>
rescue_from_my_var =>
::1 - - [22/Nov/2023:16:34:47 +0100] "GET /second HTTP/1.1" 401 15 0.0012
my_var => 1
self => #<Class:0x0000000105ffa518>
rescue_from_self => #<Class:0x0000000105ffa518>
rescue_from_my_var => 1
::1 - - [22/Nov/2023:16:34:51 +0100] "GET /first HTTP/1.1" 401 12 0.0014
```

As you can see, different instances for each of the endpoints, so variables are not accessible across multiple API methods.

So do you think we should add several tests for checking that this behavior does not change?

For point number 2:

Totally agree on document this behavior in the README 👍  let me add a commit to this PR

For point number 3:

Yes. Before these changes, the meaning of `@var` inside the `rescue_from` was totally different, so let me add an explanation to the `UPGRADING` document 👍 

Thank you again for the review!!

dblock:
> If you do `@var = 1` in the first call, `@var` should not be available in a second call. This is because a new instance is created for each of the different endpoints.

Great. Let's write a spec like this.

Thanks again!

jcagarcia:
Hey @dblock  👋 

I've just updated the specs for including the cases we discussed above. Also, I've updated the `UPGRADING` and `README` files for explaining the behavior of the instance variables.

Let me know if anything else is needed 😄

UPGRADING.md
```diff
@@ -1,14 +1,20 @@
...
 
 - `route_method` is accessible through `request_method`
 - `route_path` is accessible through `path`
 - Any other `route_xyz` are accessible through `options[xyz]`
 
+#### Instance variables scope
+
+Due to the changes done in [#2377](https://github.com/ruby-grape/grape/pull/2377), the instance variables defined inside each of the endpoints (or inside a `before` validator) are now accessible inside the `rescue_from`. This means the scope of the instance variables has changed.
+
+If you were using the same variable name defined inside an endpoint or `before` validator inside a `rescue_from` handler, you need to take in mind that you can start getting different values or you can be overriding values.
```
dblock, line 16: Add a short before/after example?

dblock (changes requested):
Looks good. Can I ask for a quick example in UPGRADING and also say "This behavior was undefined until 2.1.0." so that it doesn't sound like a breaking change.

UPGRADING.md
jcagarcia: Sure!

jcagarcia: Requested change added!

dblock:
Merged, thank you.

numbata:
Just FYI. Because the context for calling a rescue_from block has changed, the `error!` method that is called from inside the block has also changed along with the signature. The old one was in `Grape::Middleware::Error`:
```
def error!(message, status = options[:default_status], headers = {}, backtrace = [], original_exception = nil)
```
and now the `error!` is defined by the `Grape::DSL::InsideRoute`:
```
def error!(message, status = nil, additional_headers = nil)
```
So if for some reason the `error!` was called with the additional `backtrace` and/or `original_exception` arguments, it will now break.

dblock:
@numbata So should these method be aligned? Either way, it would be helpful to write a test that demonstrates this, and/or add it to UGPRADING.md.

numbata:
Let's align them! Will do that here: https://github.com/ruby-grape/grape/pull/2468

PR #2378, opened 2023-11-23
Do not overwrite route_param with a regular one if they share same name

arg:
https://github.com/ruby-grape/grape/pull/2326 changed the behavior of params merging. Before v1.8.0 the following route:

```ruby
resources :users do
  route_param :id do
    get do
      user = User.find(params[:id])
      { id: user.id }
    end
  end
end
```

would return `{ id: 12345 }` if called as

```
GET /users/12345?id=1
```

Now, since v1.8.0, it returns `{ id: 1 }` as `id` passed in params overwrites `route_param`.

This PR restores the previous behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@
   - [Include Parent Namespaces](#include-parent-namespaces)
   - [Include Missing](#include-missing)
   - [Evaluate Given](#evaluate-given)
+  - [Parameter Precedence](#parameter-precedence)
 - [Parameter Validation and Coercion](#parameter-validation-and-coercion)
   - [Supported Parameter Types](#supported-parameter-types)
   - [Integer/Fixnum and Coercions](#integerfixnum-and-coercions)
@@ -1197,6 +1198,35 @@ curl -X POST -H "Content-Type: application/json" localhost:9292/child -d '{"chil
   }
 }
 ````
+
+### Parameter Precedence
+
+Using `route_param` takes higher precedence over a regular parameter defined with same name:
+
+```ruby
+params do
+  requires :foo, type: String
+end
+route_param :foo do
+  get do
+    { value: params[:foo] }
+  end
+end
+```
+
+**Request**
+
+```bash
+curl -X POST -H "Content-Type: application/json" localhost:9292/bar -d '{"foo": "baz"}'
+```
+
+**Response**
+
+```json
+{
+  "value": "bar"
+}
+```
 
 ## Parameter Validation and Coercion
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -9,6 +9,59 @@ The `rack_response` method has been deprecated and the `error_response` method h
 
 See [#2414](https://github.com/ruby-grape/grape/pull/2414) for more information.
 
+#### Change in parameters precedence
+
+When using together with `Grape::Extensions::Hash::ParamBuilder`, `route_param` takes higher precedence over a regular parameter defined with same name, which now matches the default param builder behavior.
+
+This was a regression introduced by [#2326](https://github.com/ruby-grape/grape/pull/2326) in Grape v1.8.0.
+
+```ruby
+grape.configure do |config|
+  config.param_builder = Grape::Extensions::Hash::ParamBuilder
+end
+
+params do
+  requires :foo, type: String
+end
+route_param :foo do
+  get do
+    { value: params[:foo] }
+  end
+end
+```
+
+Request:
+
+```bash
+curl -X POST -H "Content-Type: application/json" localhost:9292/bar -d '{"foo": "baz"}'
+```
+
+Response prior to v1.8.0:
+
+```json
+{
+  "value": "bar"
+}
+```
+
+v1.8.0..v2.0.0:
+
+```json
+{
+  "value": "baz"
+}
+```
+
+v2.1.0+:
+
+```json
+{
+  "value": "bar"
+}
+```
+
+See [#2378](https://github.com/ruby-grape/grape/pull/2378) for details.
+
 #### Grape::Router::Route.route_xxx methods have been removed
 
 - `route_method` is accessible through `request_method`
--- a/lib/grape/extensions/hash.rb
+++ b/lib/grape/extensions/hash.rb
@@ -12,8 +12,12 @@ module Grape
 
         def build_params
           rack_params.deep_dup.tap do |params|
-            params.deep_merge!(grape_routing_args) if env.key?(Grape::Env::GRAPE_ROUTING_ARGS)
             params.deep_symbolize_keys!
+
+            if env.key?(Grape::Env::GRAPE_ROUTING_ARGS)
+              grape_routing_args.deep_symbolize_keys!
+              params.deep_merge!(grape_routing_args)
+            end
           end
         end
       end
--- a/spec/grape/extensions/param_builders/hash_spec.rb
+++ b/spec/grape/extensions/param_builders/hash_spec.rb
@@ -79,5 +79,34 @@ describe Grape::Extensions::Hash::ParamBuilder do
       expect(last_response.status).to eq(200)
       expect(last_response.body).to eq('["bar", nil]')
     end
+
+    it 'does not overwrite route_param with a regular param if they have same name' do
+      subject.namespace :route_param do
+        route_param :foo do
+          get { params.to_json }
+        end
+      end
+
+      get '/route_param/bar', foo: 'baz'
+      expect(last_response.status).to eq(200)
+      expect(last_response.body).to eq('{"foo":"bar"}')
+    end
+
+    it 'does not overwrite route_param with a defined regular param if they have same name' do
+      subject.namespace :route_param do
+        params do
+          requires :foo, type: String
+        end
+        route_param :foo do
+          get do
+            params[:foo]
+          end
+        end
+      end
+
+      get '/route_param/bar', foo: 'baz'
+      expect(last_response.status).to eq(200)
+      expect(last_response.body).to eq('bar')
+    end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 * [#2387](https://github.com/ruby-grape/grape/pull/2387): Fix rubygems version within workflows - [@ericproulx](https://github.com/ericproulx).
 * [#2405](https://github.com/ruby-grape/grape/pull/2405): Fix edge workflow - [@ericproulx](https://github.com/ericproulx).
 * [#2414](https://github.com/ruby-grape/grape/pull/2414): Fix Rack::Lint missing content-type - [@ericproulx](https://github.com/ericproulx).
+* [#2378](https://github.com/ruby-grape/grape/pull/2378): Do not overwrite `route_param` with a regular one if they share same name - [@arg](https://github.com/arg).
 * Your contribution here.
 
 ### 2.0.0 (2023/11/11)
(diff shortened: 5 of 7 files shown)


lib/grape/extensions/hash.rb
```diff
@@ -12,8 +12,8 @@ module ParamBuilder
 
         def build_params
           rack_params.deep_dup.tap do |params|
-            params.deep_merge!(grape_routing_args) if env.key?(Grape::Env::GRAPE_ROUTING_ARGS)
             params.deep_symbolize_keys!
+            params.deep_merge!(grape_routing_args.deep_symbolize_keys!) if env.key?(Grape::Env::GRAPE_ROUTING_ARGS)
```
dblock, line 16: This modifies the input `grape_routing_args.deep_symbolize_keys!` which may be an undesirable side-effect. 

Either remove `!` which would make a copy, or perform `grape_routing_args.deep_symbolize_keys!` elsewhere in all conditions ensuring that's intended?

dblock (changes requested):
Thanks! See below.

arg:
@dblock could you please take a look, I did the requested changes but PR is still in "changes requested" state for some reason.

dblock (changes requested):
Code looks good.

Looking at this deeper, it sounds like we're changing the undocumented/undefined behavior of a regular param overriding a route param before? 

1. Is that the right thing to do? Why wouldn't a regular parameter specified by the caller override a route parameter?
2. Does anything change if I declare a regular parameter over a route parameter for the `get` API in your example? Add a spec for that.
3. Document this behavior in README.
4. This is a behavior change, but not breaking since it wasn't documented. Either way add a section to UPGRADING.

dblock:
@arg want to finish this?

arg:
> @arg want to finish this?

Yes, I will. A bit busy now but will try to find some time in the next few days.

TrevorA-TrevorA:
I just encountered this issue as well.

I can see that `grape_routing_args` already has symbolized keys here, so after the merge, you have something like...
 `{ 'id'=>1, :id=>2 }`

This problem occurs because `Hash#deep_symbolize_keys!` preserves the first value in the key conflict:

```
[63] pry(main)> h = {'id'=>1, id:2}
=> {"id"=>1, :id=>2}
[64] pry(main)> h.deep_symbolize_keys
=> {:id=>2}
[65] pry(main)> h.deep_symbolize_keys!
=> {:id=>1}
```

This is because operations that transform keys in place (e.g. deep_symbolize_keys!) iterate through the keys and [refer back to the hash object](https://github.com/rails/rails/blob/v7.0.4.3/activesupport/lib/active_support/core_ext/hash/keys.rb#L133) for the value associated with the key on each iteration. So, when the string key of the first id parameter is iterated, it is converted to a symbol and its value overwrites the value of the second id parameter. When the second (already symbol) `:id` key is iterated, its associated value has already been overwritten by the value of the first id parameter.

Operations that return a new hash with transformed keys (e.g. deep_symbolize_keys) iterate through key/value pairs and [use the paired value](https://github.com/rails/rails/blob/v7.0.4.3/activesupport/lib/active_support/core_ext/hash/keys.rb#L120) rather than referring back to the hash object with each iteration, so the last value in a key conflict is preserved.

lib/grape/extensions/hash.rb
```diff
@@ -12,8 +12,8 @@ module ParamBuilder
 
         def build_params
           rack_params.deep_dup.tap do |params|
-            params.deep_merge!(grape_routing_args) if env.key?(Grape::Env::GRAPE_ROUTING_ARGS)
             params.deep_symbolize_keys!
+            params.deep_merge!(grape_routing_args.deep_symbolize_keys) if env.key?(Grape::Env::GRAPE_ROUTING_ARGS)
```
DmytroVasin, line 16: @arg 
Could you please re-check - do we need to have `deep_symbolize_keys` for `grape_routing_args`? 
Don't we always have there a `symbolized hash` because of [this](https://github.com/ruby-grape/grape/blob/v2.0.0/lib/grape/router/route.rb#L77)

If we keep it just to be sure that the response will be symbolized - please ignore my message

DmytroVasin:
@dblock
I encountered the same issue, and if you don't mind, here's what I've found:

> Why wouldn't a regular parameter specified by the caller override a route parameter?

[`Hashie::Mash`](https://github.com/ruby-grape/grape/blob/master/lib/grape/extensions/hashie/mash.rb#L18-L20), [`HashWithIndifferentAccess`](https://github.com/ruby-grape/grape/blob/master/lib/grape/extensions/active_support/hash_with_indifferent_access.rb#L19-L21) and [`Hash (prior 1.8.0)`](https://github.com/ruby-grape/grape/blob/v1.7.1/lib/grape/extensions/hash.rb#L14-L16) follow a similar logic: **`rack_params` are taken as a base and overwritten with `grape_routing_args`**  (thus, `grape_routing_args` is strongly preferred).

NOTE:
Old [`deep symbolize behavior`](https://github.com/ruby-grape/grape/blob/v1.7.1/lib/grape/extensions/deep_symbolize_hash.rb),
is functionally equivalent to [`deep_symbolize_keys`](https://apidock.com/rails/v6.1.3.1/Hash/_deep_transform_keys_in_object)

```ruby
deep_symbolize_keys_in({"id" => nil}.merge(id: '123'))
#=> {:id=>"123"}

{"id" => nil}.merge(id: '123').deep_symbolize_keys
#=> {:id=>"123"}
```

**However**
In [Use Active support functions](https://github.com/ruby-grape/grape/pull/2326) we replaced `deep_symbolize_keys_in` with `deep_symbolize_keys!` (with a bang), that has a bit [different behavior](https://apidock.com/rails/Hash/_deep_transform_keys_in_object%21)

```ruby
{"id" => nil}.merge(id: '123').deep_symbolize_keys!
#=> {:id=>nil}
```

The priority works for `Hashie::Mash` and `HashWithIndifferentAccess` perfectly, when `Hash` relies on a sequence of arguments.
I agree that it was not documented, but the idea of priority of params was broken by our changes

dblock:
@DmytroVasin thanks for the deep dive, makes sense, let's get this PR finished and make sure we have the right specs - is there anything else missing here?

arg:
> @DmytroVasin thanks for the deep dive, makes sense, let's get this PR finished and make sure we have the right specs - is there anything else missing here?

@dblock so what are the changes to make? Just the API test and README note?

dblock:
I think so, basically what I said in https://github.com/ruby-grape/grape/pull/2378#pullrequestreview-1748045405. Thanks for picking this up!

DmytroVasin:
@arg Any updates on this? 

Sorry for disturbing you, but I can proceed with completing this MR if you don't mind

arg:
@dblock @DmytroVasin sorry, have been quite busy the last couple of months. So anyways, I added test to cover scenario when both `route_param` and regular param are defined with same nave, also added `README` and `UPGRADING` entries. Please review

lib/grape/extensions/hash.rb
arg: > If we keep it just to be sure that the response will be symbolized

As far as I remember, it was added just to be sure all the parameters are always symbolized.

README.md
```diff
@@ -39,6 +39,7 @@
   - [Include Parent Namespaces](#include-parent-namespaces)
   - [Include Missing](#include-missing)
   - [Evaluate Given](#evaluate-given)
+  - [Parameters precedence](#parameters-precedence)
```
dblock, line 42: It should be "Parameter Precedence" (singular, capitalize both words)

README.md
```diff
@@ -1198,6 +1199,35 @@ curl -X POST -H "Content-Type: application/json" localhost:9292/child -d '{"chil
 }
 ````
 
+### Parameters precedence
```
dblock, line 1202: Parameter Precedence

README.md
```diff
@@ -1198,6 +1199,35 @@ curl -X POST -H "Content-Type: application/json" localhost:9292/child -d '{"chil
...
+
+`route_param` takes higher precedence that a regular parameter defined with same name:
```
dblock, line 1204: Using `route_param` ...

that -> than or over

UPGRADING.md
```diff
@@ -9,6 +9,60 @@ The `rack_response` method has been deprecated and the `error_response` method h
 
 See [#2414](https://github.com/ruby-grape/grape/pull/2414) for more information.
 
+#### Change in parameters precedence
+
+*(only applicable if Grape::Extensions::Hash::ParamBuilder is used)*
```
dblock, line 14: Say "When using together with `Grape::Extensions::Hash::ParamBuilder`, `route_param` takes ... , which now matches the default param bulider behavior."

Instead of original behavior say "This was a regression introduced by [link to PR] in Grape v1.8.0."

lib/grape/extensions/hash.rb
dblock: Let's break this up to avoid hiding a side effect:

```ruby
if env.key?(Grape::Env::GRAPE_ROUTING_ARGS)
   grape_routing_args.deep_symbolize_keys!
   params.deep_merge!(grape_routing_args)
end
```

spec/grape/extensions/param_builders/hash_spec.rb
```diff
@@ -79,5 +79,34 @@ def app
       expect(last_response.status).to eq(200)
       expect(last_response.body).to eq('["bar", nil]')
     end
+
+    it 'does not overwrite route_param with a regular param if they have same name' do
```
dblock, line 83: Do we have similar specs without `Grape::Extensions::Hash::ParamBuilder` for the default behavior? Let's add otherwise.

dblock (changes requested):
Thanks! I have some more asks, mostly in README/UPGRADING.

DmytroVasin:
@arg Any updates on this?

Sorry for annoying you :)

arg:
Hey, thanks for reminding and sorry for the delay.I’m coming back tomorrow and will try to find some time to finish the PRBest regards,AndriiOn 29 Mar 2024, at 08:06, Dmytro Vasin ***@***.***> wrote:﻿
@arg Any updates on this?
Sorry for annoying you :)

README.md
arg: Updated.

README.md
arg: Updated.

UPGRADING.md
arg: Updated.

spec/grape/extensions/param_builders/hash_spec.rb
arg: Sure, added similar tests for all param builders.

arg:
It looks like all requested changes are applied, please review

UPGRADING.md
```diff
@@ -9,6 +9,60 @@ The `rack_response` method has been deprecated and the `error_response` method h
 
 See [#2414](https://github.com/ruby-grape/grape/pull/2414) for more information.
 
+#### Change in parameters precedence
+
+When using together with `Grape::Extensions::Hash::ParamBuilder`, `route_param` takes higher precedence
+over a regular parameter defined with same name, which now matches the default param bulider behavior.
```
dblock, line 15: typo: builder

UPGRADING.md
```diff
@@ -9,6 +9,60 @@ The `rack_response` method has been deprecated and the `error_response` method h
...
+
+This was a regression introduced by https://github.com/ruby-grape/grape/pull/2326 in Grape v1.8.0.
```
dblock, line 17: Nit: link `[#2326](https://github.com/ruby-grape/grape/pull/2326)` like all other links.

dblock (approved):
Spotted one typo, otherwise good to go!

UPGRADING.md
dblock: Also remove the wrapping in markdown, https://code.dblock.org/2021/06/07/to-wrap-or-not-to-wrap-in-markdown.html

arg: Thanks for spotting the typo!

UPGRADING.md
arg: Changed.

arg:
@dblock ready to merge

dblock:
Thanks for the good work and for hanging in here with my OCD!

PR #2379, opened 2023-11-24
fix(#2350): Update `recognize_paths` taking into account the `route_param` type

jcagarcia:
Hey 👋 

This PR tries to fix the issue described in #2350

With this changes, the `recognize_paths` method takes into account the `route_param` type (if specified) for determine if a path matches with the performed request.

This change has an important implication that has been added to the `UPGRADED` file:

Before this change, if a request was performed to an endpoint that has specified the type of the `route_param`, the path was recognized. If finally the provided parameter in the request didn't match with the specified `route_param` type. A `400 Bad Request` response was returning, indicating that the provided param in the path was not valid. With these changes, as the path is directly not recognized if it does not accomplish with the specified `route_param` type, we are returning a `404 Not Found` response.

Is that consider a breaking change? It is a bug fix?

Any comment, suggestion and discussion is totally welcome 👍

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -53,7 +53,7 @@ end
 
 #### Recognizing Path
 
-Due to the changes done in [#2379](https://github.com/ruby-grape/grape/pull/2379), Grape now takes in mind the types of the configured `route_params` in order to determine the endpoint that matches with the performed request. 
+Grape now considers the types of the configured `route_params` in order to determine the endpoint that matches with the performed request.
 
 So taking into account this `Grape::API` class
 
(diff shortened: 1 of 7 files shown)


UPGRADING.md
```diff
@@ -51,6 +51,48 @@ class TwitterAPI < Grape::API
 end
 ```
 
+#### Recognizing Path
+
+Due to the changes done in [#2379](https://github.com/ruby-grape/grape/pull/2379), Grape now takes in mind the types of the configured `route_params` in order to determine the endpoint that matches with the performed request.
```
dblock, line 56: "Grape now considers the types of the configured ..."

"Due to the changes done in ..." -> move to the end of all this text and say "See [xyz] for more information."

UPGRADING.md
```diff
@@ -51,6 +51,48 @@ class TwitterAPI < Grape::API
...
+      # POST /books/share
+      post do
+      # ....
+      end
+    end
+  end
+end
+```
+
+This was the behavior before the changes:
```
dblock, line 80: Follow the convention, say "Before:" and "After:"

dblock (approved):
Good! Nits below.

UPGRADING.md
jcagarcia: Thank you so much for the suggestion! I've just applied it! 👏

UPGRADING.md
jcagarcia: Totally right. I've just applied it! 👏

Issue #2380, opened 2023-11-27
Allow to define the API version where a endpoint is available


jcagarcia:
Hey 👋 

I want to share a possible improvement for reusing the same endpoint through different api versions.

I have the v1 of my API that only exposes the `/foo` endpoint.

```ruby
class MyAPI < Grape::API
  class MyError < StandardError; end

  version 'v1', using: :path

  get '/foo' do
    { version: version }
  end
end
```

```shell
GET /v1/foo => { version: "v1" }
```

For the next version of my API, I want to add a new endpoint `/var`

```ruby
class MyAPI < Grape::API
  class MyError < StandardError; end

  version 'v1', using: :path

  get '/foo' do
    { version: version }
  end

  version 'v2', using: :path

  get '/var' do
    { version: version }
  end
end
```

The examples above results in:

```shell
GET /v1/foo => { version: "v1" }
GET /v2/var => { version: "v2" }
```

But if i try to access the `/foo` endpoint in my v2 version I'm getting a 404. 

```shell
GET /v2/foo => 404
GET /v2/var => { version: "v2" }
```

Taking a look to the code (not present in the documentation), seems I can specify multiple versions if passing an array of values

```ruby
class MyAPI < Grape::API
  class MyError < StandardError; end

  version 'v1', 'v2',  using: :path

  get '/foo' do
    { version: version }
  end

  get '/var' do
    { version: version }
  end
end
```

So now, both endpoints are available under v1 and v2 endpoint.

```shell
GET /v1/foo => { version: "v1" }
GET /v1/var => { version: "v1" }
GET /v2/foo => { version: "v2" }
GET /v2/var => { version: "v2" }
```

However, this is not desired behavior. As `/foo` should be available in v2, but `/var` should not be present in `/v1`.

My proposal is to configure each of the endpoints for being available in the different versions of the api. For example, using a `for_versions` method like:

```ruby
class MyAPI < Grape::API
  class MyError < StandardError; end

  version 'v1', 'v2', using: :path

  get '/foo', for_versions: ['v1', 'v2'] do
    { version: version }
  end

  get '/var', for_versions: ['v2'] do
    { version: version }
  end
end
```

And expecting this behavior:

```shell
GET /v1/foo => { version: "v1" }
GET /v1/var => 404
GET /v2/foo => { version: "v2" }
GET /v2/var => { version: "v2" }
```

What do you think about this approach?

dblock:
[...]

PR #2381, opened 2023-11-28
doc(#2380): Updating versioning section inside the README

jcagarcia:
Hi 👋 I made a suggestion about how to version in (what I thought it was) an easy way. However, this is totally solved and really well explained in https://code.dblock.org/2013/07/19/evolving-apis-using-grape-api-versioning.html

So I think my problem was that I was not able to find in the documentation that multiple versions for the same `Grape::API` class were allowed and a little guide about how to properly use versioning.

Do you think this can help?


dblock (approved):
Looks good and is very useful.

Issue #2376, opened 2023-11-21
Values validator fails fast when below root scope, violating documentation


eriklovmo:
In the documentation, it is currently written:

> Grape returns all validation and coercion errors found by default. To skip all subsequent validation checks when a specific param is found invalid, use `fail_fast: true`.

My interpretation of that statement is that, in order to skip subsequent validations when a specific param is found invalid, one must set `fail_fast` to true (though that interpretation might be far-reaching). However, I have found one example that behaves differently. When using the `values` validator in a nested scope, the validation is not exhaustive: 

https://github.com/ruby-grape/grape/blob/3f01d03b7a28b088a52313d8d264f29aeacd41fa/lib/grape/validations/validators/values_validator.rb#L32

The following snippets demonstrate what this entails.

**A**

```ruby
class Foo < Grape::API
  params do
    requires :bar, values: ->(v) { v.zero? || v.positive? }
    requires :baz, values: ->(v) { v.positive? }
  end

  post "/foo" do
    "bar"
  end
end
```

With example A, the validation includes the following errors when no params are provided:

`"bar is missing, bar does not have a valid value, baz is missing, baz does not have a valid value"`

**B**

```ruby
class Foo < Grape::API
  params do
    with(type: Integer) do
      requires :bar, values: ->(v) { v.zero? || v.positive? }
      requires :baz, values: ->(v) { v.positive? }
    end
  end

  post "/foo" do
    "bar"
  end
end
```

With example B, the validation includes the following errors when no params are provided:

`"bar is missing, baz is missing"`

It would be good to either fix this behavior if possible or to update the documentation to mention this exception, either in the section about the `values` validator or in the general paragraph mentioning "fast failing", which I have cited.

dblock:
Looks like a bug, appreciate a fix!

numbata:
Looks like unexpected expected behavior #671

eriklovmo:
> Looks like unexpected expected behavior #671

Ah, that's it! It seems like the code conflates groupings of params defined using `with` and groupings of params defined using `optional`, that the PR seemingly targeted. We need to differentiate between "nesting" and "grouping", since the short-circuiting behavior is only desired sometimes.

dblock:
@eriklovmo Are you trying to fix it? Write some specs at least that fail?

numbata:
@dblock an example of spec that fails:
```diff
[...]

PR #2382, opened 2023-12-05
Fix ValuesValidator for Params Wrapped in with Block in Grape

numbata:
This PR fixes a bug [#2376](https://github.com/ruby-grape/grape/issues/2376) in Grape's `ValuesValidator`, specifically when parameters are nested within a `with` block. Previously, the validator did not correctly handle the required validation for nested parameters in this context.

## Changes:
- The `required_for_root_scope?` method in `ValuesValidator` has been modified to correctly detect if a parameter is required when it's nested inside a `with` block. The fix involves checking the parent scope of the parameter until a non-lateral (non-`with` block) scope is found, and then determining if it's the root scope.
- A test case has been added to `values_spec.rb` to cover the scenario where a parameter nested within a `with` block doesn't meet the validation criteria, ensuring that the correct error response is generated.

## Impact:
This fix ensures more accurate validation of parameters nested within `with` blocks, and aligns the behavior with the expected norms of Grape's parameter validation. It's an important fix for Grape users who use nested parameter structures, improving the reliability and correctness of the validation logic.

I welcome any feedback or suggestions on this fix and look forward to your reviews!

### DIFF
--- a/lib/grape/validations/validators/values_validator.rb
+++ b/lib/grape/validations/validators/values_validator.rb
@@ -85,7 +85,12 @@ module Grape
         end
 
         def required_for_root_scope?
-          @required && @scope.root?
+          return false unless @required
+
+          scope = @scope
+          scope = scope.parent while scope.lateral?
+
+          scope.root?
         end
 
         def validation_exception(attr_name, message)
--- a/spec/grape/validations/validators/values_spec.rb
+++ b/spec/grape/validations/validators/values_spec.rb
@@ -261,6 +261,13 @@ describe Grape::Validations::Validators::ValuesValidator do
         optional :name, type: String, values: %w[a b], allow_blank: true
       end
       get '/allow_blank'
+
+      params do
+        with(type: String) do
+          requires :type, values: ValuesModel.values
+        end
+      end
+      get 'values_wrapped_by_with_block'
     end
   end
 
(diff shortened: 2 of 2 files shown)


lib/grape/validations/validators/values_validator.rb
```diff
@@ -85,7 +85,12 @@ def except_message
         end
 
         def required_for_root_scope?
-          @required && @scope.root?
+          return false unless @required
```
numbata, line 88: fail fast if the `param` is not required.

lib/grape/validations/validators/values_validator.rb
```diff
@@ -85,7 +85,12 @@ def except_message
...
+
+          scope = @scope
+          scope = scope.parent while scope.lateral?
```
numbata, line 91: I've updated the logic here to handle nested parameters within a `with` block. The method now traverses up the scope hierarchy to check if the parameter's root scope is required, ensuring correct validation even when parameters are deeply nested within a `with' block.

dblock:
Perfect, thank you. Update CHANGELOG and we're good to go!

lib/grape/validations/validators/values_validator.rb
dblock: You could extract this into `lateral_scope` or something like that, although I am not sure it's worth it

dblock:
Merged, thank you.

PR #2384, opened 2023-12-10
fix(#1975): Allow to use `before/after/rescue_from` methods in any order when using `mount`

jcagarcia:
This PR tries to "fix" the issue described in #1975 .

Right now, as we can read in our README, declarations as `before/after/rescue_from` must be placed before `mount` in a case where they should be inherited. 

However, I think this behavior arises from the existing implementation rather than being an intentional design. This occurs because when the `mount` method is invoked, a new `Endpoint` instance is generated, inheriting the previously registered `before/after/rescue_from` methods. However, if additional `before/after/rescue_from` methods are executed after the `mount` operation, the mounted endpoints remain unrefreshed.

Considering that:

1. Whenever a method is run within a `Grape::API` class, it is stored in the `@setup` variable in the sequence in which it was executed.
2. Upon completion of the execution, we will have an array containing the methods that were executed in the precise order they were called.

My proposal to address this issue is:

1. When executing any method, such as `before/after/rescue_from` (in fact, any method, not limited to these), verify whether there have been preceding `mount` method calls for this particular `Grape::API` class stored within the `@setup` variable.
2. If affirmative, re-apply the mounting of the same endpoint to update and refresh the inherited configurations.

By doing so, at the end of this process, all mounted endpoints will inherit all the executed methods within the `Grape::API` scope.

### DIFF
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -152,7 +152,7 @@ module Grape
         end
 
         # Updating all previously mounted classes in the case that new methods have been executed.
-        if method != :mount && @setup.size > 1
+        if method != :mount && @setup.any?
           previous_mount_steps = @setup.select { |step| step[:method] == :mount }
           previous_mount_steps.each do |mount_step|
             refresh_mount_step = mount_step.merge(method: :refresh_mounted_api)
--- a/lib/grape/api.rb
+++ b/lib/grape/api.rb
@@ -150,6 +150,19 @@ module Grape
         @instances.each do |instance|
           last_response = replay_step_on(instance, setup_step)
         end
+
+        # Updating all previously mounted classes in the case that new methods have been executed.
+        if method != :mount && @setup.size > 1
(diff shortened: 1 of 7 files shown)


lib/grape/api.rb
```diff
@@ -150,6 +150,19 @@ def add_setup(method, *args, &block)
         @instances.each do |instance|
           last_response = replay_step_on(instance, setup_step)
         end
+
+        # Updating all previously mounted classes in the case that new methods have been executed.
+        if method != :mount && @setup.size > 1
```
dblock, line 155: setup.any?

dblock (approved):
👏

dblock:
Given that this broke no specs (impressive) I say we merge, thank you! Want to rebase CHANGELOG and potentially edit my nit?

jcagarcia:
> Given that this broke no specs (impressive) I say we merge, thank you! Want to rebase CHANGELOG and potentially edit my nit?

I've just rebased and applied your suggestions! :) Happy to help!

PR #2389, opened 2023-12-18
Remove rack-accept dependency

ericproulx:
`rack-accept` hasn't been updated for many years and since most of Grape's needs are already in `rack`, I think  we can remove the dependency.


dblock (approved):
👏

PR #2393, opened 2023-12-28
Optimize AttributeTranslator

ericproulx:
After #2375, I've spent some time understanding the `Grape::Router::AttributeTranslator` across the gem. I knew already what it did but I didn't know where it was used exactly. This PR is an effort to optimize the usage of the AttributeTranslator but also some refactoring across the `Grape::Router` classes.

Changes:
- Introducing `Grape::Router::GreedyRoute` to encapsulate `Grape::Router::AttributeTranslator` in the greedy match process.
- `Grape::Router::AttributeTranslator` defines more methods to not rely on `method_missing` internally.
- `Grape::Router::Pattern` and `Grape::Router::Route` have been refactored.


dblock:
This is better! Thanks.

PR #2397, opened 2024-01-01
Add ruby 3.3 to edge and test flows

ericproulx:
This PR adds ruby 3.3 is our GH workflows.


spec/grape/endpoint_spec.rb
```diff
@@ -694,17 +694,6 @@ def app
...
-      it 'raises NoMethodError but stripping the internals of the Object class' do
-        subject.get('/hey') do
-          Object.new.x
-        end
-        expect do
-          get '/hey'
-        end.to raise_error(NoMethodError, /^undefined method `x' for #<Object:0x[0-9a-fA-F]+>$/)
-      end
-    end
   end
```
dblock, line 697: What's the reason for removing this? It was fixing a bug I believe with failure context.

ericproulx: Well, it works for the `endpoint` through the`method_missing` but otherwise, the `NoMethodError` is handled by the object it self. Also when you look at the code, its supposed to end with `in `/hey' endpoint` but it does not. We don't have the control over all the instances declared in the endpoint but we do at the endpoint's instance level.

dblock (approved):

PR #2399, opened 2024-01-01
Update rubocop* gems

ericproulx:
Autocorrect new offenses


dblock (commented):
For CHANGELOG I'd call the specific version of rubocop, e.g. "Update RuboCop to 1.59.0".

Also maybe we can roll out dependabot ...

PR #2414, opened 2024-01-21
Fix response headers from lint

### LINKED ISSUE #2411: Rack::Lint::Error =>  a header value must be a String or Array of Strings, but the value of 'content-type' is a NilClass
I've been testing [Rack::Lint](https://github.com/rack/rack/blob/main/lib/rack/lint.rb) locally and we have a lot of issues in our spec suite. I'll open issues for others but let's start with this one:

Running spec:  spec/grape/api_spec.rb:2341.
```
 Rack::Lint::LintError:
       a header value must be a String or Array of Strings, but the value of 'content-type' is a NilClass
```

This test is using `rack_response` directly in a rescue_from like this
```ruby
subject.rescue_from ApiSpec::APIErrors::ParentError, rescue_subclasses: true do |e|
 rack_response("rescued from #{e.class.name}", 500)
end
```

Unfortunately, `rack_response` doesn't act like `error!` since the last is passing through the formatting part (because of the `throw`) whereas the `rack_response` just html escaped if content_type is set to`text/html` (if content_type is set in the api)

Finally, `rack_response` is never mentioned in the README but using a plain `Rack::Response` is. Also, Using `rack_response` or `Rack::Response` doesn't pass through the formatting part of the `error` middleware since its considered a final response.

IMO, there are many ways to solve that but I found an interesting middleware within Rack called [Rack::ContentType] (https://github.com/rack/rack/blob/main/lib/rack/content_type.rb) that will set a default content-type header if needed. Default, is `text/html` but I think `text/plain` would be more appropriate since we do not need to escape. I think its a clean way to deal with that issue and we could remove some `default` behavior in the code.

ericproulx:
Fix #2411 

Changes:
- Use `Rack::Utils.escape_html` instead of `ERB::Util.escape_html`
- Some refactor in error middleware.
- Remove `rack_response` from `inside_route`
- Remove some irrelevant `if: ActiveSupport::VERSION::MAJOR >= ?` in specs

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -4,7 +4,10 @@ Upgrading Grape
 ### Upgrading to >= 2.1.0
 
 #### Changes in rescue_from
-`rack_response` has been deprecated and `error_response` has been removed. Use `error!` instead.
+
+The `rack_response` method has been deprecated and the `error_response` method has been removed. Use `error!` instead.
+
+See [#2414](https://github.com/ruby-grape/grape/pull/2414) for more information.
 
 #### Grape::Router::Route.route_xxx methods have been removed
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -3,6 +3,9 @@ Upgrading Grape
 
 ### Upgrading to >= 2.1.0
 
+#### Changes in rescue_from
+`rack_response` has been deprecated and `error_response` has been removed. Use `error!` instead.
+
 #### Grape::Router::Route.route_xxx methods have been removed
 
 - `route_method` is accessible through `request_method`
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -180,7 +180,7 @@ module Grape
       # A Rack::Response object containing the specified message, status, and headers.
       #
       def rack_response(message, status = 200, headers = { Rack::CONTENT_TYPE => content_type })
-        Grape.deprecator.warn('Use error! instead of rack_response')
+        Grape.deprecator.warn('The rack_response method has been deprecated, use error! instead.')
         message = Rack::Utils.escape_html(message) if headers[Rack::CONTENT_TYPE] == 'text/html'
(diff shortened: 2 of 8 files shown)


dblock (commented):
So we do have specs that use `rack_response` in `rescue_from`. 

1. Add a section to UPGRADING.md.
2. Is there a way we can re-implement it after these changes and mark it deprecated instead?

ericproulx:
> So we do have specs that use `rack_response` in `rescue_from`. Someone is undoubtedly relying on it.
> 
>     1. Add a section to UPGRADING.md.
> 
>     2. Is there a way we can re-implement it after these changes and mark it deprecated instead?

`error_response` has been removed also. It's an old deprecation from 0.12

UPGRADING.md
```diff
@@ -3,6 +3,9 @@ Upgrading Grape
 
 ### Upgrading to >= 2.1.0
 
+#### Changes in rescue_from
+`rack_response` has been deprecated and `error_response` has been removed. Use `error!` instead.
```
dblock, line 7: The `rack_response` method ... (just to make it a sentence)

Link to the PR here and a couple missing below.

```
See [#2414](https://github.com/ruby-grape/grape/pull/2414) for more information.
```

lib/grape/dsl/inside_route.rb
```diff
@@ -180,8 +180,9 @@ def error!(message, status = nil, additional_headers = nil)
       # A Rack::Response object containing the specified message, status, and headers.
       #
       def rack_response(message, status = 200, headers = { Rack::CONTENT_TYPE => content_type })
-        message = ERB::Util.html_escape(message) if headers[Rack::CONTENT_TYPE] == 'text/html'
-        Rack::Response.new([message], Rack::Utils.status_code(status), headers)
+        Grape.deprecator.warn('Use error! instead of rack_response')
```
dblock, line 183: I think this reads cleaner?

The rack_response method has been deprecated, use error! instead.

dblock (approved):
nits, fix if you agree and hit merge

PR #2419, opened 2024-03-21
Add the `contract` DSL

### LINKED ISSUE #2386: Define contract instead of params
Hey 👋 

Sometimes, we can have a huge amount of params to be received in an endpoint. To define all those parameters in the params section can make the received params difficult to read. Additionally, managing custom validations and error messages specific to these parameters becomes challenging.

Due to this, I've seen several implementations where they use [dry-validations contract](https://dry-rb.org/gems/dry-validation/1.8/) instead of the `Grape` params.

```ruby
post '/charge' do
  validation_result = API::V1::Controllers::Contracts::Payments::PaymentsCharge.new.call(parsed_body)
  params = parsed_body.deep_symbolize_keys.deep_merge(validation_result.to_h) if validation_result.success?
  ...
end
```

Would be great to provide this feature directly by `Grape` like:

```ruby
contract API::V1::Controllers::Contracts::Payments::PaymentsCharge
post '/charge' do
  ...
end
```

I can see that we are already using `dry-types` for validating `params`  (#1347 ).

dgutov:
Hi!

This seems to cover most or everything we've discussed in the issue comments (except for "rules", as expected). It turned out pretty compact.

Not many docs yet. Comments about naming or organization welcome.

I've thought about moving the new `Validator` to a separate file, which would allow testing it in isolation, but then there wouldn't be much left in `ContractScope`, and all the validators are tested integration-style anyway.

From `Validator`'s constructor arguments, we can see that this stretches the current `ValidatorFactory` protocol, but I figured that it's not worth a major change just for this.

Also: I wasn't sure what would be the correct signature for the `klass` parameter in the doc, hence this syntax. Maybe it'd be better to leave the type not annotated.

Resolves #2386


dgutov:
@dblock @dnesteryuk @zysend @Jack12816 Tagging a few of the recent contributors, as suggested.

README.md
```diff
@@ -2086,6 +2086,35 @@ params do
 end
 ```
 
+## Parameter Validation and Coercion using `dry-schema`
```
dblock, line 2089: Add to TOC.

README.md
```diff
@@ -2086,6 +2086,35 @@ params do
...
+
+As an alternative to the `params` DSL described above, you can use a
+schema or `dry-validation` contract to describe an endpoint's
```
dblock, line 2092: Don't wrap lines, https://code.dblock.org/2021/06/07/to-wrap-or-not-to-wrap-in-markdown.html

README.md
```diff
@@ -2086,6 +2086,35 @@ params do
 end
 ```
 
+## Parameter Validation and Coercion using `dry-schema`
+
+As an alternative to the `params` DSL described above, you can use a
```
dblock, line 2091: I think we want to add a new section "Parameter Validation and Coercion" then add sub-sections for `params` and `contract`. Talk about this in general in the top section, explaining why both exists and why one may be better.

README.md
```diff
@@ -2086,6 +2086,35 @@ params do
...
+    optional(:volume).maybe(:integer, lt?: 9)
+  end
+end
+```
+
+The latter will define a coercing schema (`Dry::Schema.Params`). With
+the former syntax it's up to you to decide whether the input will need
+coercing.
+
+`params` and `contract` declarations can also be used together, e.g.
```
dblock, line 2115: Say that in the intro somewhere.

lib/grape/dsl/validations.rb
```diff
@@ -38,6 +38,18 @@ def reset_validations!
...
         end
+
+        # Declare the contract to be used for the endpoint's parameters.
+        # @param klass [Class<Dry::Validation::Contract> | Class<Dry::Schema::Processor>]
+        #   The contract or schema class to be used for validation. Optional.
+        # @yield a block yielding a new instance of Dry::Schema::Params
+        #   subclass, allowing to define the schema inline. When the
+        #   +klass+ parameter is non-nil, it will be used as a parent. Optional.
+        def contract(klass = nil, &block)
+          raise ArgumentError, 'Either klass or block must be provided' unless klass || block
```
dblock, line 49: Is the idiomatic way to do this `block_given?`?

spec/grape/validations/contract_scope_spec.rb
```diff
@@ -0,0 +1,133 @@
+# frozen_string_literal: true
+
+require 'pry'
+
+describe Grape::Validations::ContractScope do
```
dblock, line 5: Does it make sense to extract a shared spec that applies to any scope and say "behaves like a scope" here?

README.md
```diff
@@ -2086,6 +2086,35 @@ params do
...
+to describe different parts of a nested namespace for an endpoint.
+
```
dblock, line 2117: You have to say to add `dry-schema` to Gemfile here, right? We should probably have an integration spec without it as well to make sure we don't accidentally rely on something from that gem that only surfaces in production.

dblock (commented):
This is pretty great! See some comments, mostly small.

dgutov:
Also added support for `declared` just now. It re-processes the set of keys, but stops from repeating the same validations, fortunately.

This seems like the price to pay for having `declared` work having combined `params` and `contract` validations work naturally.

lib/grape/dsl/validations.rb
dgutov: That was my first instinct, but `Performance/BlockGivenWithExplicitBlock` disagreed.

spec/grape/validations/contract_scope_spec.rb
dgutov: Maybe not.

`ParamsScope` tests a lot of functionality that we don't test here because it's part of the third-party package (`dry-schema`). As a bonus, we also test here being able to combine both `contract` and `params`.

README.md
dgutov: Indeed, if we don't declare `dry-schema` a hard dependency of `grape`.

> We should probably have an integration spec without it as well to make sure we don't accidentally rely on something from that gem that only surfaces in production.

How do you suggest it be written? Probably not with a separate Gemfile. Unbinding the constant `Dry` temporarily?

README.md
dgutov: Okay, if that's your preference.

README.md
dgutov: That might be one extra nesting too much. Do you think it's okay for `Built-in Validators` to be `h4` and `allow_blank` (or `With Default`, at the end) to be `h5`?

One alternative is to put the new section under `Parameter Validation and Coercion` at `h3` and describe why you might want to use it instead. It might not hurt to see some user feedback first before making it a first-class option, especially given that `dry-schema`'s development is also stalling a little these days.

README.md
dgutov: I see that now: the Appraisal config, with indeed separate Gemfiles for certain integration test files.

The instructions in the root Gemfile do not quite work, however:

```sh
appraisal install ; rubocop -a gemfiles/*.gemfile`
```

both fails at the `rails_5` gemfile (a screenful about activesupport versions), generates some `.gemfile` files that weren't there before (apparently desynchronization happened between `rack_1_0.gemfile` and `rack1.gemfile` naming), and creates gemfiles with lots of double quotes inside them. Which in turn tiggers Rubocop for each of those lines.

What do you recommend I do here?

Regarding the integration testing itself, I think we'll be more or less safe as long as at least some of those gemfiles (the ones used for existing integration tests, for example) don't include `dry-schema`.

dgutov: Okay, I fixed the above, and indeed those existing test won't help. Looking into adding a new one.

README.md
```diff
@@ -2086,6 +2087,31 @@ params do
 end
 ```
 
+### Using `dry-validation` or `dry-schema`
+
+As an alternative to the `params` DSL described above, you can use a schema or `dry-validation` contract to describe an endpoint's parameters. This can be especially useful if you use the above already in some other parts of your application. If not, you'll need to add `dry-validation` or `dry-schema` to your `Gemfile`.
+
+Then call `contract` with a contract or schema defined previously
```
dblock, line 2094: Add a period.

README.md
```diff
@@ -2086,6 +2087,31 @@ params do
...
+
+```rb
+contract CreateFoosSchema
```
dblock, line 2097: Maybe include the code for `CreateFoosSchema` or at least the class declaration.

A different example that "Foos" would be nice, something that everybody agrees is a thing.

README.md
```diff
@@ -2086,6 +2087,31 @@ params do
...
+  required(:foos).array do
+    required(:name).filled(:string)
+    optional(:volume).maybe(:integer, lt?: 9)
+  end
+end
+```
+
+The latter will define a coercing schema (`Dry::Schema.Params`). When using the former approach, it's up to you to decide whether the input will need coercing.
+
+`params` and `contract` declarations can also be used together, e.g. to describe different parts of a nested namespace for an endpoint.
```
dblock, line 2113: The `params` ...

together in the same API ...

lib/grape/dsl/validations.rb
```diff
@@ -38,6 +38,19 @@ def reset_validations!
...
+
+        # Declare the contract to be used for the endpoint's parameters.
+        # @param contract [Class<Dry::Validation::Contract> | Dry::Schema::Processor]
+        #   The contract or schema to be used for validation. Optional.
+        # @yield a block yielding a new instance of Dry::Schema::Params
+        #   subclass, allowing to define the schema inline. When the
+        #   +contract+ parameter is a schema, it will be used as a parent. Optional.
+        def contract(contract = nil, &block)
+          raise ArgumentError, 'Either contract or block must be provided' unless contract || block
+          raise ArgumentError, 'Cannot inherit from contract, only schema' if block && contract.respond_to?(:schema)
```
dblock, line 50: I see `if block_given?` instead of `if block` in other places, let's be consistent?

dblock (changes requested):
Looks like workflow is borked, https://github.com/ruby-grape/grape/actions/runs/8383399387. Plus some nitpicks. 

Looks great otherwise!

README.md
dgutov: Added, see the new push.

README.md
dgutov: Added a colon, if you don't mind.

dgutov:
> Danger still complains about the TOC, which needs to be fixed.

I've added a TOC entry yesterday, and it works: https://github.com/dgutov/grape/tree/contract_dsl_for_dry_schema?tab=readme-ov-file#using-dry-validation-or-dry-schema

Perhaps the checker is thrown off by backtick formatting in the text?

Is there a way to see its output somewhere? The workflow just says "Danger has failed this build". And when I try to run it locally, I get various errors, such as

```
fatal: ambiguous argument 'origin/master': unknown revision or path not in the working tree.
```

or

```
.../gems/danger-9.4.3/lib/danger/request_sources/github/github.rb:117:in `fetch_details': undefined method `pull_request' for nil:NilClass (NoMethodError)
```

.github/workflows/test.yml
```diff
@@ -45,6 +49,7 @@ jobs:
         bundler-cache: true
 
     - name: Run tests
+      if: ${{ matrix.integration_only == false }}
```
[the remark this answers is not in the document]
dgutov: > Does this just save time or do these tests fail?

The tests creating dry schemas or contracts naturally fail. I suppose we could skip those tests inside the specs instead, but that can swing the reliability tradeoff in the other direction (someone deletes the gem from all gemfiles, and the build continues to pass).

> I generally don't like this matrix here that checks which gemfile is used. There's possibly a simpler solution

Looks good. Let me know if you want this change in my branch.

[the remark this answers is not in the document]
dblock: Nah, but if you want to take that on in a separate PR I'd appreciate it!

dblock (commented):
To fix danger complaining unquote the text in the TOC.

```
- [Using dry-validation or dry-schema](#using-dry-validation-or-dry-schema)
```

(danger comments in the PR with the expected TOC, see [above](https://github.com/ruby-grape/grape/pull/2419#issuecomment-2013187206))

dgutov:
> To fix danger complaining unquote the text in the TOC.
> 
> ```
> - [Using dry-validation or dry-schema](#using-dry-validation-or-dry-schema)
> ```

Done.

> (danger comments in the PR with the expected TOC, see [above](https://github.com/ruby-grape/grape/pull/2419#issuecomment-2013187206))

Ok I see: it edits the original comment. Neat.

dgutov:
Finally, I've added Rubocop exceptions. Those remaining warnings didn't look reasonable to me personally.

If you agree, I can try a separate PR that disables the `RSpec/FilePath` and `RSpec/SpecFilePathFormat` at least for all specs in the integration dir, and either disables the `RSpec/MultipleExpectations` cop altogether, or at least raises the limit.

README.md
dblock: We might be overthinking it. No strong opinions.

dblock (approved):

dblock:
Merged this, good work!

PR #2420, opened 2024-03-24
Declare integration tests with less repetition

dgutov:
Something discussed in the other PR.

It's not exactly the same behavior - each "integration" runs with the specified version of Ruby only, and doesn't run the regular test suite. But this seems easier to read anyway.

WDYT?

If running Rack 2 and 3 with all Ruby versions is really needed, I suppose integrations could be defined as a separate step (with its own matrix).


dblock:
This is much better. I'll merge.

I think we should be able to collapse a little further - the two tasks are very similar (`bundle exec rake spec` vs. `bundle exec rspec spec/integration/xyz`). What is being executed can probably move to the matrix itself.

dgutov:
> What is being executed can probably move to the matrix itself.

`rake spec` is kinda special: simply replacing it with `rspec spec` won't work because the rake task also contains the exclusions (the spec/integrations directory).

The separate step is now pretty short. We could remove it with some interpolations and long `... && ... || ...` conditionals, but I don't see a way to remove the repetition of the string `spec/integration/` (either in the step, or in the matrix). And with it, we repeat the conditional twice as well.

That seems hard to avoid without using an environment binding (which requires an extra step as well), or something more drastic like pre-generating the matrix in an additional process call (https://stackoverflow.com/a/65434401/615245), which is +1 step and +1 level of indirection. 

So, this works, but it might not be an improvement: https://github.com/dgutov/grape/commit/af797e9333db8fad47571cb256e65eb47d181dc3

dblock:
> So, this works, but it might not be an improvement: [dgutov@af797e9](https://github.com/dgutov/grape/commit/af797e9333db8fad47571cb256e65eb47d181dc3)

I kinda agree. Do we need to do `rake spec` in both though?

dgutov:
> Do we need to do rake spec in both though?

No, it's `rake spec` for the full suite and `rspec spec/integration/...` for the integration tests.

dblock:
> > Do we need to do rake spec in both though?
> 
> No, it's `rake spec` for the full suite and `rspec spec/integration/...` for the integration tests.

The `rake spec` runs `rspec ...something` eventually, what else does it do?

dgutov:
> what else does it do?

It adds the exclusion of integration tests:

```rb
  spec.exclude_pattern = 'spec/integration/**/*_spec.rb'
```

dblock:
You can add `--exclude-pattern` on the command line too. 

I don't know if it's better, but maybe.

dgutov:
> You can add --exclude-pattern on the command line too.

Sure. It's just that when these two invocations are different enough (we need to either call different commands in these two cases, or add different arguments anyway), it's difficult to come up with a readable way to DRY this more.

dblock:
How do you feel about https://github.com/ruby-grape/grape/pull/2423?

PR #2421, opened 2024-03-25
Tweak Rubocop rules; add more exceptions to the mail config

dgutov:
@dblock In reply to your comment in #2419:

> Usually I rubocop for auto formatting and consistency. You can always rubocop -a ; rubocop --auto-gen-config.

What I ended up doing is calling `bundle exec rubocop --regenerate-todo`. But I think that's a problem for potential contributors - to understand which rules should be followed, and which can/should just be put into the todo.

CONTRIBUTING says this, after all:

> Ruby style is enforced with [Rubocop](https://github.com/bbatsov/rubocop), run bundle exec rubocop and fix any style issues highlighted.

So here's an update to the default config which addresses the 3 main cops which gave me trouble.

Ultimately, it seems like you might prefer to just disable all three instead (and maybe more). But this is a more conservative change.


dgutov:
BTW, if I call `rubocop --auto-gen-config` now, it suggests additionally disabling the cops

```
Lint/ConstantDefinitionInBlock
Lint/EmptyBlock
RSpec/ContextWording
RSpec/LeakyConstantDeclaration
RSpec/NamedSubject
```

and changing a few other configuration parameters.

dblock (approved):

dblock:
> BTW, if I call `rubocop --auto-gen-config` now, it suggests additionally disabling the cops
> 
> ```
> Lint/ConstantDefinitionInBlock
> Lint/EmptyBlock
> RSpec/ContextWording
> RSpec/LeakyConstantDeclaration
> RSpec/NamedSubject
> ```
> 
> and changing a few other configuration parameters.

Will take any improvement you want to make here.

dgutov:
> Will take any improvement you want to make here.

Ok, see #2422.

Issue #1165, opened 2015-09-25
Tests are not independent


samotarnik:
If I am not mistaken, the tests are not independent, i.e. their success depends on the order of execution. Which is quite a problem. Compare
```
bundle exec rspec
```
with
```
bundle exec rspec --order rand
```
and finally
```
find spec/ -name *_spec.rb | xargs -i bundle exec rspec --order rand {}
```
In the second case, I get at least two failing examples (specifically with seed 35635: `bundle exec rspec --order rand:35635`):
```
spec/grape/api/deeply_included_options_spec.rb:52 # Grape::API doesn't work for format different than specified
spec/grape/api/shared_helpers_spec.rb:31 # Grape::API::Helpers defines parameters
```
And in the third case at least two other specs break:
```
spec/grape/validations/params_scope_spec.rb:286 # Grape::Validations::ParamsScope when validations are dependent on a parameter applies the validations only if the parameter is present
spec/grape/validations/params_scope_spec.rb:313 # Grape::Validations::ParamsScope when validations are dependent on a parameter returns a sensible error message within a nested context
spec/grape/middleware/formatter_spec.rb[1:7:23] # Grape::Middleware::Formatter input parses the body from an xml #{method} and copies values into rack.request.from_hash
spec/grape/middleware/formatter_spec.rb[1:7:5] # Grape::Middleware::Formatter input parses the body from an xml #{method} and copies values into rack.request.from_hash
spec/grape/middleware/formatter_spec.rb[1:7:32] # Grape::Middleware::Formatter input parses the body from an xml #{method} and copies values into rack.request.from_hash
spec/grape/middleware/formatter_spec.rb[1:7:14] # Grape::Middleware::Formatter input parses the body from an xml #{method} and copies values into rack.request.from_hash
spec/grape/middleware/formatter_spec.rb:130 # Grape::Middleware::Formatter accept header detection parses headers with vendor and api version
spec/grape/middleware/formatter_spec.rb:113 # Grape::Middleware::Formatter accept header detection uses quality rankings to determine formats
spec/grape/middleware/formatter_spec.rb:108 # Grape::Middleware::Formatter accept header detection detects from the Accept header
spec/grape/middleware/formatter_spec.rb:120 # Grape::Middleware::Formatter accept header detection handles quality rankings mixed with nothing
spec/grape/middleware/formatter_spec.rb:84 # Grape::Middleware::Formatter detection uses the format parameter if one is provided
[...]

PR #2422, opened 2024-03-26
More rubocop

dgutov:
Two steps here.

The first tweaks or disables most of the cops that `--auto-gen-config` commended.

The second is an attempt to fix the "leaky constant" problem that two other cops aim to warn about. I figured that's useful enough to have checked, to avoid this class of test interdependence problems (related: #1165). But the solution is more codegen-ish, so -- see if you like it.


dblock:
Great work, thank you!

PR #2423, opened 2024-03-26
DRY-up test workflow.

dblock:
@dgutov do you think this is better? It gets rid of the indirections by removing the concept of "integration", and making both gemfile used and specs declarative.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,22 +24,27 @@ jobs:
       fail-fast: false
       matrix:
         ruby: ['2.7', '3.0', '3.1', '3.2', '3.3']
-        gemfile: [rack_2_0, rack_3_0, rails_6_0, rails_6_1, rails_7_0, rails_7_1]
-        integration: [false]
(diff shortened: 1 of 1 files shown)


dgutov:
It adds more repetition, but on the plus side it's more transparent and explicit. One fewer step and variable as well.

On the balance, probably an improvement. :slightly_smiling_face:

.github/workflows/test.yml
```diff
@@ -24,24 +24,29 @@ jobs:
...
       matrix:
         ruby: ['2.7', '3.0', '3.1', '3.2', '3.3']
-        gemfile: [rack_2_0, rack_3_0, rails_6_0, rails_6_1, rails_7_0, rails_7_1]
-        integration: [false]
+        gemfile: [Gemfile, gemfiles/rack_2_0.gemfile, gemfiles/rack_3_0.gemfile, gemfiles/rails_6_0.gemfile, gemfiles/rails_6_1.gemfile, gemfiles/rails_7_0.gemfile, gemfiles/rails_7_1.gemfile]
+        specs: ['spec --exclude-pattern=spec/integration/**/*_spec.rb']
         include:
           - ruby: '2.7'
-            gemfile: rack_1_0
+            gemfile: gemfiles/rack_1_0.gemfile
```
ericproulx, line 31: You'll need to remove this part since rack_1_0 is not in the codebase anymore.

ericproulx (changes requested):

.github/workflows/test.yml
dblock: Done.

ericproulx (approved):

Issue #2396, opened 2023-12-31
Use Rack's HeadersHash instead of {}


ericproulx:
According to Rack's [upgrade guide](https://github.com/rack/rack/blob/v3.0.8/UPGRADE-GUIDE.md#response-headers-must-be-lower-case), we should use `Rack::Headers` instead of a single Hash to manage response headers. Furthermore, we always used a`{}` to manage headers but we should had used `Rack::Utils::HeaderHash`.

Anyway, I think it would be better than using `Grape.lowercase_headers?`

PR #2425, opened 2024-04-06
Replace Hash with Rack::Header/Rack::Utils::HeaderHash for Grape Headers

dhruvCW:
This closes https://github.com/ruby-grape/grape/issues/2396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#2426](https://github.com/ruby-grape/grape/pull/2426): Drop support for rack 1.x series - [@ericproulx](https://github.com/ericproulx).
 * [#2427](https://github.com/ruby-grape/grape/pull/2427): Use `rack-contrib` jsonp instead of rack-jsonp - [@ericproulx](https://github.com/ericproulx).
 * [#2363](https://github.com/ruby-grape/grape/pull/2363): Replace autoload by zeitwerk - [@ericproulx](https://github.com/ericproulx).
+* [#2425](https://github.com/ruby-grape/grape/pull/2425): Replace `{}` with `Rack::Header` or `Rack::Utils::HeaderHash` - [@dhruvCW](https://github.com/dhruvCW).
 * Your contribution here.
 
 #### Fixes
--- a/lib/grape/dsl/headers.rb
+++ b/lib/grape/dsl/headers.rb
@@ -12,7 +12,7 @@ module Grape
         if key
           val ? header[key.to_s] = val : header.delete(key.to_s)
         else
-          @header ||= {}
+          @header ||= Grape::Util::Header.new
         end
       end
       alias headers header
--- a/lib/grape/endpoint.rb
+++ b/lib/grape/endpoint.rb
@@ -238,7 +238,7 @@ module Grape
 
     def run
       ActiveSupport::Notifications.instrument('endpoint_run.grape', endpoint: self, env: env) do
-        @header = {}
+        @header = Grape::Util::Header.new
         @request = Grape::Request.new(env, build_params_with: namespace_inheritable(:build_params_with))
         @params = @request.params
         @headers = @request.headers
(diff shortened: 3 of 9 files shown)


ericproulx:
I think there's this one left [here](https://github.com/ruby-grape/grape/blob/5121279107b3b905203f48a94a8a2069659ec519/lib/grape/endpoint.rb#L241)

dhruvCW:
> I think there's this one left [here](https://github.com/ruby-grape/grape/blob/5121279107b3b905203f48a94a8a2069659ec519/lib/grape/endpoint.rb#L241). Thanks for doing this :)

ah thanks 😅 missed that, should be fixed now 👍

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [#2408](https://github.com/ruby-grape/grape/pull/2408): Fix params method redefined warnings - [@ericproulx](https://github.com/ericproulx).
 * [#2410](https://github.com/ruby-grape/grape/pull/2410): Gem deprecations will raise a DeprecationWarning in specs - [@ericproulx](https://github.com/ericproulx).
 * [#2389](https://github.com/ruby-grape/grape/pull/2389): Remove rack-accept dependency - [@ericproulx](https://github.com/ericproulx).
+* [#2425](https://github.com/ruby-grape/grape/pull/2425): Replace `{}` with `Rack::Header` (for rack 3.0 and greater) or Rack::Utils::HeaderHash` (for rack less than 3.0) in Grape::DSL::Headers - [@dhruvCW](https://github.com/dhruvCW).
```
dblock, line 21: Quote these consistently, `Grape::DSL::Headers`, maybe shorten.

Replace `{}` with `Rack::Header` or `Rack::Utils::HeaderHash`.

or if you really want versions here

Replace `{}` with `Rack::Header` in Rack 3.0+ and Rack::Utils::HeaderHash` for Rack 1.x and 2.x.

lib/grape/dsl/headers.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'grape/util/header'
```
dblock, line 3: Add a blank line after.

lib/grape/endpoint.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'grape/util/header'
```
dblock, line 3: Can we consolidate this require elsewhere once?

lib/grape/util/header.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Grape
+  module Util
+    begin
+      # For `Rack::Headers` (Rack 3+):
+      require 'rack/headers'
+      Header = Rack::Headers
+    rescue LoadError
```
dblock, line 9: Don't rely on an exception at runtime, use `Gem::Version.new(Rack.release) >= Gem::Version.new('3')`.

dblock (changes requested):
Thanks! Some asks below.

CHANGELOG.md
dhruvCW: yup the one without version sounds better 😅 replaced.

lib/grape/util/header.rb
dhruvCW: awesome, replaced 👍

lib/grape/dsl/headers.rb
dhruvCW: removed (not required anymore).

lib/grape/endpoint.rb
dhruvCW: removed (not required anymore)

dblock:
Great, merging. I don't think we need anything for UPGRADING.md, but if you disagree, please PR.

dblock (approved):

PR #2430, opened 2024-04-15
Isolate extensions within specific Gemfile

ericproulx:
With the newest addition of `dry-validation` that has its own Gemfile, I thought it would be great to revisit the remaining one and apply the same pattern by keeping isolation in mind since they all come with their dependencies. IMHO, it feels better this way while keeping it simple for future extensions.

This PR, isolates the following extensions within its own Gemfile and moved the specs accordingly:

- hashie
- grape-entity
- dry-validation

`dry-validation` has been revisited to be really isolated from all the others.


dblock (approved):
👏

PR #2431, opened 2024-04-16
Replace Appraisals by eval_gemfile

ericproulx:
This PR drops Appraisals in favor of eval_gemfile. It also updates some gems like rubocop, rack-test etc ...
I don't know if we need instructions if someone wants to add a new Gemfile but looking at the current ones might give a very good idea of the how to.


.rubocop_todo.yml
```diff
@@ -246,6 +246,52 @@ RSpec/VoidExpect:
     - 'spec/grape/api_spec.rb'
     - 'spec/grape/dsl/headers_spec.rb'
 
+# Offense count: 598
+# This cop supports unsafe autocorrection (--autocorrect-all).
+# Configuration parameters: ResponseMethods.
+# ResponseMethods: response, last_response
+RSpecRails/HaveHttpStatus:
```
dblock, line 253: Just disable it?

spec/support/chunks.rb
```diff
@@ -3,7 +3,7 @@
 module Chunks
   def read_chunks(body)
     buffer = []
-    body.each { |chunk| buffer << chunk }
+    body.each { |chunk| buffer << chunk } # rubocop:disable Style/MapIntoArray
```
dblock, line 6: I would rely on `rubocop --auto-gen-config`, we disable a lot of violations that way.

dblock (approved):
I can live with this :) Some nits below, up to you.

dblock (changes requested):
Actually two tests fail for good reason! Double check that the changes actually don't change the required gems. Probably need to swap the order of things.

dblock (approved):
I spot-checked that we end up with Rack 2.0 for the rack_2.0.gemfile. I think we should add tests in the rack and rails integrations that ensure that the version matches the intended one, e.g.

```ruby
expect(Gem::Version.new(Rack.release).segments.first).to eq 2
```

Up to you if you want to do this as part of this/future PR/maybe I can get to it, too.

PR #2433, opened 2024-04-26
Added Rack version specs to ensure the correct version is loaded.

dblock:
A followup for https://github.com/ruby-grape/grape/pull/2431.


dblock:
@ericproulx take a look pls?

ericproulx (approved):

PR #2434, opened 2024-04-26
Implement nested `with` support in parameter DSL

numbata:
## Motivation
Previously, the `with`' method in the Parameter DSL did not support nested structures. This forced to redefine the entire attribute set for each nested scope, which was cumbersome and prone to errors, especially in complex parameter structures.

## Solution
This PR introduces enhanced functionality to the `with` method that allows nested scopes to inherit and extend the attributes defined in parent scopes. By merging attributes from the current group with the new attributes, it simplifies parameter declarations and ensures a clean, DRY approach to specifying shared attributes in nested parameter contexts.

### DIFF
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -264,6 +264,7 @@ module Grape
           parent: self,
           optional: optional,
           type: type || Array,
+          group: @group,
           &block
         )
       end
--- a/spec/grape/dsl/parameters_spec.rb
+++ b/spec/grape/dsl/parameters_spec.rb
@@ -35,6 +35,12 @@ module Grape
           @validates
         end
 
+        def new_scope(args, _, &block)
+          nested_scope = self.class.new
+          nested_scope.new_group_scope(args, &block)
+          nested_scope
+        end
+
         def new_group_scope(args)
           @group = args.clone.first
           yield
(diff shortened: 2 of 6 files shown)


spec/grape/dsl/parameters_spec.rb
```diff
@@ -35,6 +35,12 @@ def validates_reader
           @validates
         end
 
+        def new_scope(args, _, &block)
+          nested_scope = self.class.new
+          nested_scope.new_group_scope(args, &block)
+          nested_scope
+        end
```
numbata, line 42: Not sure about this, but looks legit for me.

dblock:
Looks good. Document it in README please? Maybe add a test with 3 levels just to be sure ;)

lib/grape/validations/params_scope.rb
```diff
@@ -264,6 +264,7 @@ def new_scope(attrs, optional = false, &block)
           parent: self,
           optional: optional,
           type: type || Array,
+          group: @group,
```
numbata, line 267: @dblock  I've added more specs and found that in case of mix of nested scopes and groups, `with` doesn't work as expected (if there is anything expected). The reason is that when a new parameter with a block is defined , a new scope is created without any link to the current group.

spec/grape/dsl/parameters_spec.rb
```diff
@@ -35,9 +35,17 @@ def validates_reader
...
+
         def new_group_scope(args)
+          prev_group = @group
           @group = args.clone.first
           yield
+          @group = prev_group
```
numbata, line 48: Since we do not really create scopes, this is my imitation of it.

lib/grape/validations/params_scope.rb
dblock: Open an issue with a failed spec?

dblock (approved):
This LGTM, add to README?

lib/grape/validations/params_scope.rb
numbata: Whether it is an issue or an undocumented feature is always hard to tell in such mature projects :)

dblock (approved):

PR #2435, opened 2024-05-04
Use Rack's Constants

ericproulx:
### Motivation

Since we [dropped](#2426) support for rack 1.X, we can replace some of our constants by Rack's constants. For instance, we had defined all supported http methods like `Grape::Http::Headers::GET`, `Grape::Http::Headers::POST` etc... but its already defined as `Rack::GET`, `Rack::POST`, etc... in Rack.

We also had redefined some Rack's constants in our code base like `Grape::Env::RACK_INPUT`, `Grape::Env::RACK_REQUEST_FORM_HASH` etc ... but these are Rack's internals and we should not rely on them and use Rack's constants `Rack::INPUT`, `Rack::REQUEST_FORM_HASH`.

At least, if Rack as major internal changes, Grape's will be in line with these changes even if it's a renaming or a drop. In the end, Grape is rack based and it all makes sense to do that since we were already using several Rack's utility functions.

Although, Grape should have a smaller memory footprint since we have less string literals.

Finally, I've changed Rack 3.0's integrations tests to make sure all our response headers are lowercased instead of looking if our internal headers are in lowercase. If makes more sense with #2425.

Thanks

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,9 +34,6 @@ jobs:
             gemfile: gemfiles/multi_xml.gemfile
             specs: 'spec/integration/multi_xml'
           - ruby: '2.7'
-            gemfile: gemfiles/rack_2_0.gemfile
-            specs: 'spec/integration/rack_2_0'
-          - ruby: '2.7'
             gemfile: gemfiles/rack_3_0.gemfile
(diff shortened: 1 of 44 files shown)


dblock (approved):
Nice.

Issue #1370, opened 2016-04-23
How to allow empty array for params?


qsona:
I wrote the following code.
```
params do
  requires :items, type: Array[JSON] do
    requires :id, type: Integer
    requires :enabled, type: Boolean
  end
end
post '/' do
# ...
```
A request with following params succeeds
`{ items: [{ id: 1, enabled: true }, { id: 2, enabled: false }] }`

But fails with following.
`{ items: [] }`

I changed the code to
```
params do
  optional :items, type: Array[JSON] do
```

But, all requests failed then.

I want to allow `{ items: [] }` params. Is there any way to do it?

Thanks.

dblock:
If you're POSTing JSON, I believe this just works. Care to share the code that does the POSTing?

I'll close this. Please use [the Grape mailing list](https://groups.google.com/forum/#!forum/ruby-grape) for questions.

liquid1982:
I have the same problem, and can't figure out what a possible solution could be. It just seems a buggy behaviour. I need to allow the user to publish an article, and optionally specify a section. I also need to accept an empty array, that will result in unpublishing the article from the specified domain.

The non-working code is the following:

```rb
desc 'Publish an article'
params do
  optional :published_in, type: Hash do
    optional :'foo.com', type: Array[String], values: ALLOWED_SECTIONS['foo'], allow_blank: true
    optional :'bar.com', type: Array[String], values: ALLOWED_SECTIONS['bar'], allow_blank: true
  end
end
post 'articles/123/actions/publish' do
  # ...
end
```

But if I omit the `values` param, things work as expected:


```rb
desc 'Publish an article'
params do
  optional :published_in, type: Hash do
    optional :'foo.com', type: Array[String], allow_blank: true
    optional :'bar.com', type: Array[String], allow_blank: true
  end
end
post 'articles/123/actions/publish' do
  # ...
end
```

I'd like to specify the allowed values in the params block. Any suggestion?

dblock:
What is the posting code? What content-type is it POSting @liquid1982 ?

liquid1982:
@dblock It is something like:

```js
    $.ajax({
      url: url,
      method: 'POST',
      contentType: 'application/json',
      data: JSON.stringify(options.data)
    }).then(() => {})
```

dblock:
You should turn it into a spec @liquid1982 in this project.

PR #2437, opened 2024-05-10
Add length validator

dhruvCW:
This validator validates that the length of the parameter (if the parameter supports `#length` is within the specified limits).

I've also added the `min_length` and `max_length` limits to the attribute documentation so that they can be used to specify the limits for an array type in swagger.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1713,15 +1713,14 @@ end
 
 #### `length`
 
-Parameters can be restricted from having a specific length or size with the `:length` option.
+Parameters of type `String` or `Array` can be restricted to have a specific length or size with the `:length` option.
 
-the validator accepts `:min` or `:max` or both options to validate that they parameter is within
-the given limits.
+The validator accepts `:min` or `:max` or both options to validate that the value of the parameter is within the given limits.
 
 ```ruby
 params do
-  requires :str, length: { min: 3 }
-  requires :list, length: { min: 3, max: 5}
+  requires :str, type: String, length: { min: 3 }
+  requires :list, type: [Integer], length: { min: 3, max: 5 }
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@
     - [values](#values)
     - [except_values](#except_values)
     - [same_as](#same_as)
+    - [length](#length)
     - [regexp](#regexp)
     - [mutually_exclusive](#mutually_exclusive)
     - [exactly_one_of](#exactly_one_of)
@@ -69,6 +70,7 @@
   - [Custom Validation messages](#custom-validation-messages)
     - [presence, allow_blank, values, regexp](#presence-allow_blank-values-regexp)
     - [same_as](#same_as-1)
+    - [length](#length-1)
     - [all_or_none_of](#all_or_none_of-1)
     - [mutually_exclusive](#mutually_exclusive-1)
     - [exactly_one_of](#exactly_one_of-1)
@@ -1709,6 +1711,20 @@ params do
 end
 ```
 
+#### `length`
+
+Parameters can be restricted from having a specific length or size with the `:length` option.
+
+the validator accepts `:min` or `:max` or both options to validate that they parameter is within
+the given limits.
+
+```ruby
+params do
+  requires :str, length: { min: 3 }
+  requires :list, length: { min: 3, max: 5}
+end
+```
+
 #### `regexp`
 
 Parameters can be restricted to match a specific regular expression with the `:regexp` option. If the value does not match the regular expression an error will be returned. Note that this is true for both `requires` and `optional` parameters.
@@ -2023,6 +2039,15 @@ end
 params do
   requires :password
   requires :password_confirmation, same_as: { value: :password, message: 'not match' }
+end
+```
+
+#### `length`
+
+```ruby
+params do
+  requires :str, length: { min: 5, message: 'str is expected to be atleast 5 characters long' }
+  requires :list, length: { min: 2, max: 3, message: 'list is expected to have between 2 and 3 elements' }
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -2045,8 +2045,8 @@ end
 
 ```ruby
 params do
-  requires :str, length: { min: 5, message: 'str is expected to be atleast 5 characters long' }
-  requires :list, length: { min: 2, max: 3, message: 'list is expected to have between 2 and 3 elements' }
+  requires :str, type: String, length: { min: 5, message: 'str is expected to be atleast 5 characters long' }
+  requires :list, type: [Integer], length: { min: 2, max: 3, message: 'list is expected to have between 2 and 3 elements' }
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -1713,7 +1713,7 @@ end
 
 #### `length`
 
-Parameters of type `String` or `Array` can be restricted to have a specific length or size with the `:length` option.
+Parameters with types that support `#length` method can be restricted to have a specific length with the `:length` option.
 
 The validator accepts `:min` or `:max` or both options to validate that the value of the parameter is within the given limits.
 
@@ -1721,6 +1721,7 @@ The validator accepts `:min` or `:max` or both options to validate that the valu
 params do
   requires :str, type: String, length: { min: 3 }
   requires :list, type: [Integer], length: { min: 3, max: 5 }
+  requires :hash, type: Hash, length: { max: 5 }
 end
 ```
 
--- a/lib/grape/locale/en.yml
+++ b/lib/grape/locale/en.yml
@@ -10,6 +10,9 @@ en:
         values: 'does not have a valid value'
         except_values: 'has a value not allowed'
         same_as: 'is not the same as %{parameter}'
+        length: 'is expected to have length within %{min} and %{max}'
+        length_min: 'is expected to have length greater than or equal to %{min}'
+        length_max: 'is expected to have length less than or equal to %{max}'
         missing_vendor_option:
           problem: 'missing :vendor option'
           summary: 'when version using header, you must specify :vendor option'
--- a/lib/grape/validations/attributes_doc.rb
+++ b/lib/grape/validations/attributes_doc.rb
@@ -28,6 +28,9 @@ module Grape
         details[:documentation] = documentation if documentation
 
         details[:default] = validations[:default] if validations.key?(:default)
+
+        details[:min_length] = validations[:length][:min] if validations.key?(:length) && validations[:length].key?(:min)
+        details[:max_length] = validations[:length][:max] if validations.key?(:length) && validations[:length].key?(:max)
       end
 
       def document(attrs)
--- a/lib/grape/validations/validators/length_validator.rb
+++ b/lib/grape/validations/validators/length_validator.rb
@@ -10,6 +10,8 @@ module Grape
 
           super
 
+          raise ArgumentError, 'min must be an integer greater than or equal to zero' if !@min.nil? && (!@min.is_a?(Integer) || @min.negative?)
+          raise ArgumentError, 'max must be an integer greater than or equal to zero' if !@max.nil? && (!@max.is_a?(Integer) || @max.negative?)
           raise ArgumentError, "min #{@min} cannot be greater than max #{@max}" if !@min.nil? && !@max.nil? && @min > @max
         end
 
@@ -17,7 +19,8 @@ module Grape
           param = params[attr_name]
           param = param.compact if param.respond_to?(:compact)
 
-          return unless param.respond_to?(:length)
+          raise ArgumentError, "parameter #{param} has an unsupported type. Only strings & arrays are supported" unless params.is_a?(String) || param.is_a?(Array)
+
           return unless (!@min.nil? && param.length < @min) || (!@max.nil? && param.length > @max)
 
           raise Grape::Exceptions::Validation.new(params: [@scope.full_name(attr_name)], message: build_message)
--- a/lib/grape/validations/validators/length_validator.rb
+++ b/lib/grape/validations/validators/length_validator.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module Grape
+  module Validations
+    module Validators
+      class LengthValidator < Base
+        def initialize(attrs, options, required, scope, **opts)
+          @min = options[:min]
+          @max = options[:max]
+
+          super
+
+          raise ArgumentError, "min #{@min} cannot be greater than max #{@max}" if !@min.nil? && !@max.nil? && @min > @max
+        end
+
+        def validate_param!(attr_name, params)
+          param = params[attr_name]
+          param = param.compact if param.respond_to?(:compact)
+
+          return unless param.respond_to?(:length)
+          return unless (!@min.nil? && param.length < @min) || (!@max.nil? && param.length > @max)
+
+          raise Grape::Exceptions::Validation.new(params: [@scope.full_name(attr_name)], message: build_message)
+        end
+
+        def build_message
+          if options_key?(:message)
+            @option[:message]
+          elsif @min && @max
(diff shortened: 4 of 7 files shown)


lib/grape/locale/en.yml
```diff
@@ -10,6 +10,9 @@ en:
         values: 'does not have a valid value'
         except_values: 'has a value not allowed'
         same_as: 'is not the same as %{parameter}'
+        length_mismatch_both: 'is expected to have a length within %{min_length} and %{max_length}'
```
dblock, line 13: potentially better wordsmithing:

is expected to be between ... and ... characters long
is expected to be at least ... characters long
is expected to be less than ... characters long

dblock, line 13: let's match the variables of the other errors

```
length
length_min
length_max
```

lib/grape/validations/attributes_doc.rb
```diff
@@ -28,6 +28,12 @@ def extract_details(validations)
         details[:documentation] = documentation if documentation
 
         details[:default] = validations[:default] if validations.key?(:default)
+
+        return unless validations.key?(:length)
```
dblock, line 32: we'll be adding other validations, so let's not prematurely return, a double `if` is fine, it will get optimized away

```
details[:min_length] = validations[:length][:min] if validations.key?(:length) && validations[:length].key?(:min)
details[:max_length] = validations[:length][:max] if validations.key?(:length) && validations[:length].key?(:max)

```

lib/grape/validations/validators/length_validator.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Grape
+  module Validations
+    module Validators
+      class LengthValidator < Base
+        def initialize(attrs, options, required, scope, **opts)
+          @min_length = options[:min]
```
dblock, line 8: Call this `@min` and `@max` since we're already inside a `LengthValidator`

lib/grape/validations/validators/length_validator.rb
```diff
@@ -0,0 +1,35 @@
...
+          @max_length = options[:max]
+          super
+        end
+
+        def validate_param!(attr_name, params)
+          param = params[attr_name]
+          return unless param.respond_to?(:length)
+          return unless (@min_length && param.length < @min_length) || (@max_length && param.length > @max_length)
```
dblock, line 16: Semantically we probably don't want to treat `@min_length` of zero like this, so use `@min_length.nil?` in the condition?

dblock (changes requested):
Looks good, some nits below, plus:

1. Add README documentation.
2. Add tests for zero.
3. Handle/add tests for min = max, min > max and min > max. Probably needs to error on some.

lib/grape/validations/validators/length_validator.rb
dhruvCW: done 👍

lib/grape/validations/validators/length_validator.rb
dhruvCW: makes sense done 👍

lib/grape/validations/attributes_doc.rb
dhruvCW: done 👍

lib/grape/locale/en.yml
dhruvCW: done 👍

dhruvCW: I hesitate to use `characters long` since it doesn't make sense for arrays.

lib/grape/validations/validators/length_validator.rb
```diff
@@ -0,0 +1,40 @@
...
+  module Validations
+    module Validators
+      class LengthValidator < Base
+        def initialize(attrs, options, required, scope, **opts)
+          @min = options[:min]
+          @max = options[:max]
+
+          super
+
+          raise ArgumentError, "min #{@min} cannot be greater than max #{@max}" if !@min.nil? && !@max.nil? && @min > @max
```
dblock, line 13: Should we handle negative values here too?

spec/grape/validations/validators/length_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+        requires :list, type: [Integer], length: { max: 3 }
+      end
+      post 'with_max_only' do
+      end
+
+      params do
+        requires :list, type: Integer, length: { max: 3 }
+      end
+      post 'type_is_not_array' do
+      end
```
dblock, line 28: This is confusing. What is the length of an integer?

dblock (changes requested):
I think this needs a bit more work around types. 

1. Length is not supported on types such as Integer, and should fail accordingly. Needs tests.
2. Semantically, `length` and `size` aren't the same thing, even though hashes for example support both length and size. Should we have interchangeable `length` and `size` validators?
3. What happens with comparable types, e.g. when I say `min: 'xyz'`?
4. Should this be a more generic validator that works on anything `Comparable`?

dhruvCW:
> 1. Length is not supported on types such as Integer, and should fail accordingly. Needs tests.

makes sense should an argument error be raised in this case 🤔 

> 2. Semantically, `length` and `size` aren't the same thing, even though hashes for example support both length and size. Should we have interchangeable `length` and `size` validators?

While I agree, I see that in ruby both are interchangable so maybe an alias for length validator 🤔 

> 3. What happens with comparable types, e.g. when I say `min: 'xyz'`?
> 4. Should this be a more generic validator that works on anything `Comparable`?

IMO that won't make sense I can restrict the `min` and `max` values to be integer numbers (and maybe also that they not negative).

lib/grape/validations/attributes_doc.rb
```diff
@@ -28,6 +28,9 @@ def extract_details(validations)
         details[:documentation] = documentation if documentation
 
         details[:default] = validations[:default] if validations.key?(:default)
+
+        details[:min_length] = validations[:length][:min] if validations.key?(:length) && validations[:length].key?(:min)
```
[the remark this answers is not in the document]
dhruvCW: this results in a rubocop error that then suggests replacing the `if` with  an `unless` like I had originally written. How do you propose we proceed 🤔 

```ruby
return unless validations.key?(:length)

details[...]
details[...]
```

README.md
```diff
@@ -1709,6 +1711,20 @@ params do
...
+
+Parameters can be restricted from having a specific length or size with the `:length` option.
+
+the validator accepts `:min` or `:max` or both options to validate that they parameter is within
+the given limits.
+
+```ruby
+params do
+  requires :str, length: { min: 3 }
+  requires :list, length: { min: 3, max: 5}
```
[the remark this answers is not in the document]
dblock: Also format (missing space before }).

dblock:
> IMO that won't make sense. I can restrict the `min` and `max` values to be integer numbers (and maybe also that they not negative).

Yes, I think that would be the correct behavior.

dblock:
> > 1. Length is not supported on types such as Integer, and should fail accordingly. Needs tests.
> 
> makes sense should an argument error be raised in this case 🤔

Yes, I think anything being sent that doesn't have `.length` should yield a clear error. 
 
> > 2. Semantically, `length` and `size` aren't the same thing, even though hashes for example support both length and size. Should we have interchangeable `length` and `size` validators?
> 
> While I agree, I see that in ruby both are interchangable so maybe an alias for length validator 🤔

I think I'd like to write `length: ... ` which causes the validator to invoke `.length`, and `size: ` which causes the validator to invoke `.size`. I would implement both validators with a base "comparable" class that we can reuse for a future (or part of this PR maybe even) "compare: { min: max: }" validator that would work for strings too?

README.md
```diff
@@ -1709,6 +1711,20 @@ params do
 end
 ```
 
+#### `length`
+
+Parameters can be restricted from having a specific length or size with the `:length` option.
```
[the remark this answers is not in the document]
dhruvCW: done restricted the validator to only string or array and mentioned that in the readme as well.

lib/grape/validations/validators/length_validator.rb
dhruvCW: done restricted allowed values to non negative integer numbers.

spec/grape/validations/validators/length_spec.rb
dhruvCW: yup now if the type is not string or array it will raise an `ArgumentError`.

dhruvCW:
> > > 1. Length is not supported on types such as Integer, and should fail accordingly. Needs tests.
> > 
> > 
> > makes sense should an argument error be raised in this case 🤔
> 
> Yes, I think anything being sent that doesn't have `.length` should yield a clear error.
> 
> > > 2. Semantically, `length` and `size` aren't the same thing, even though hashes for example support both length and size. Should we have interchangeable `length` and `size` validators?
> > 
> > 
> > While I agree, I see that in ruby both are interchangable so maybe an alias for length validator 🤔
> 
> I think I'd like to write `length: ... ` which causes the validator to invoke `.length`, and `size: ` which causes the validator to invoke `.size`. I would implement both validators with a base "comparable" class that we can reuse for a future (or part of this PR maybe even) "compare: { min: max: }" validator that would work for strings too?

I have restricted the validator to only validate `String` or `Array` types. I am happy to alias the validator to support `size: ` as well but I think for this PR I want to be restrictive and only support these two concrete types instead of any class that has a `length` or `size` method.

lib/grape/validations/attributes_doc.rb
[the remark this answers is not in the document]
dblock: We can leave it.

lib/grape/validations/validators/length_validator.rb
```diff
@@ -0,0 +1,43 @@
...
+          raise ArgumentError, 'min must be an integer greater than or equal to zero' if !@min.nil? && (!@min.is_a?(Integer) || @min.negative?)
+          raise ArgumentError, 'max must be an integer greater than or equal to zero' if !@max.nil? && (!@max.is_a?(Integer) || @max.negative?)
+          raise ArgumentError, "min #{@min} cannot be greater than max #{@max}" if !@min.nil? && !@max.nil? && @min > @max
+        end
+
+        def validate_param!(attr_name, params)
+          param = params[attr_name]
+          param = param.compact if param.respond_to?(:compact)
+
+          raise ArgumentError, "parameter #{param} has an unsupported type. Only strings & arrays are supported" unless params.is_a?(String) || param.is_a?(Array)
```
dblock, line 22: Should this be anything that `.respond_to?(:length)`? 

nit: Let's spell out `&` as `and`

dblock (commented):
We can leave the size/length for a future PR. But I am not sure restricting to Array and String is correct.

lib/grape/validations/validators/length_validator.rb
dhruvCW: Nope I decided on only supporting `String` and `Array` for now since it doesn't make sense to me to support other types e.g. `Hash`.

dblock:
I noticed the `compact`. That forces an interesting behavior of removing `nil` entries, which I don't think is right, `nil`s are very valid values too, especially in content types like JSON, so let's remove it (compact can be introduced as a future option).

dhruvCW:
> I noticed the `compact`. That forces an interesting behavior of removing `nil` entries, which I don't think is right, `nil`s are very valid values too, especially in content types like JSON, so let's remove it (compact can be introduced as a future option). Add tests with a `nil` value that would fail with a `.compact`.

the problem is when an empty array is passed as the parameter the validator receives `[nil]`. How do you propose we handle this situation 🤔

dblock:
> I have restricted the validator to only validate `String` or `Array` types. I am happy to alias the validator to support `size: ` as well but I think for this PR I want to be restrictive and only support these two concrete types instead of any class that has a `length` or `size` method.

Why? I think `Hash` is an important one to support for a length validator, as an example.

dhruvCW:
> But I am not sure restricting to Array and String is correct.

could you maybe elaborate on your concern 🤔 . 

My line of thinking is that a type with a variable "length" (number of contained elements) like `String` or `Array` should have such a validation but for a type like `Hash` which is usually well specified (known set of keys) it doesn't make sense to validate the "length" or "size" of the object.

dblock:
> > I noticed the `compact`. That forces an interesting behavior of removing `nil` entries, which I don't think is right, `nil`s are very valid values too, especially in content types like JSON, so let's remove it (compact can be introduced as a future option). Add tests with a `nil` value that would fail with a `.compact`.
> 
> the problem is when an empty array is passed as the parameter the validator receives `[nil]`. How do you propose we handle this situation 🤔

That is a side effect of rack-test or rack parser that we don't control if I am not mistaken. There's a bunch of issues filed along those lines (e.g. https://github.com/ruby-grape/grape/issues/1370). For the tests for arrays switch to sending JSON which will match exactly what you're sending server-side.

dblock:
> could you maybe elaborate on your concern 🤔 .
> 
> My line of thinking is that a type with a variable "length" (number of contained elements) like `String` or `Array` should have such a validation but for a type like `Hash` which is usually well specified (known set of keys) it doesn't make sense to validate the "length" or "size" of the object.

Take a typical rollup API that receives some stats for a given, say SKU. 

```json
{
   'sku1' : {
      total: 123.
      name: 'sku one'
   },
   'sku2' : {
      total: 234.
      name: 'sku two'
   }
}
```

I choose a hash because I don't want sku's to repeat themselves, and with the length validator I'd be able to limit how many `sku`'s one could send at any given time. 

The other concern is that if we restruct `length` on certain types it will be part of the API forever. Relaxing it would require a backwards-incompatible change. If we make it `respond_to? :length`, we would defer to the broadest range of capabilities (and allow people to shoot themselves in the foot, which is fine for an API library).  

WDYT?

dblock:
Btw, thank you for hanging in here with me! Really appreciate your time and work on this.

dhruvCW:
> The other concern is that if we restrict `length` on certain types it will be part of the API forever. Relaxing it would require a backwards-incompatible change. If we make it `respond_to? :length`, we would defer to the broadest range of capabilities (and allow people to shoot themselves in the foot, which is fine for an API library).
> 
> WDYT?

Ah makes sense ✅ 

>and allow people to shoot themselves in the foot, which is fine for an API library

😅 wasn't sure if this is the library stance or not 🙈

dhruvCW:
> > > I noticed the `compact`. That forces an interesting behavior of removing `nil` entries, which I don't think is right, `nil`s are very valid values too, especially in content types like JSON, so let's remove it (compact can be introduced as a future option). Add tests with a `nil` value that would fail with a `.compact`.
> > 
> > 
> > the problem is when an empty array is passed as the parameter the validator receives `[nil]`. How do you propose we handle this situation 🤔
> 
> Either way, if a client sends an empty array, the server is expected to receive an empty array (`[nil]` is not an empty array). I believe that is a side effect of rack-test or rack parser that we don't control for GETs or POSTs with encoded form content type or something like that. There's a bunch of issues filed along those lines (e.g. #1370). For the tests for arrays switch to sending JSON which will match exactly what you're sending server-side.

just tried to change the tests to match the type to `[JSON]` instead of `[Integer]` and now when passing an empty array it fails in the coercer (as an invalid type).

lib/grape/validations/validators/length_validator.rb
dhruvCW: removed the restriction and now allows any type that responds to `#length`.

dblock (approved):

dblock:
Merged, thank you!

Want to try to implement `size` without duplicating all the code on top of this?

dhruvCW:
> Merged, thank you!
> 
> Want to try to implement `size` without duplicating all the code on top of this?

😅 For now no. I'll try the next time I have some time to spare.

dhruvCW:
@dblock will you be releasing 2.1.0 soon ?

dblock:
@dhruvCW I'll take are of it this week

PR #2445, opened 2024-05-20
Remove builder as a dependency

### LINKED ISSUE #2443: Dropping builder as dependency ?
[builder](https://github.com/jimweirich/builder) is only used for generating xml. Should we keep it as a dependency ? We're using [to_xml](https://github.com/rails/rails/blob/v7.1.3.3/activesupport/lib/active_support/core_ext/hash/conversions.rb) and active_support takes care to require builder when needed and will tell users if its not found.

For those who doesn't format :xml, `builder` is just an overhead.

ericproulx:
Fix #2443


dblock (changes requested):
The README also needs an update around XML to say builder needs to be added in certain circumstances.

Also, no tests failed?

ericproulx:
> The README also needs an update around XML to say builder needs to be added in certain circumstances.
> 
> Also, no tests failed?

It's in the our test `gemfile`

dblock:
> > The README also needs an update around XML to say builder needs to be added in certain circumstances.
> > Also, no tests failed?
> 
> It's in the our test [Gemfile](https://github.com/ruby-grape/grape/blob/13dd99c52e69c00a2ccf491635d962c1f6726576/Gemfile#L10)

Does it make sense to remove it from there and move XML tests to integration?

PR #2454, opened 2024-06-20
support Swagger Extensions

ShallmentMo:
Swagger 2.0 use something like `x-logo` as vendor extension，there is no need to validate

### DIFF
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -11,7 +11,7 @@ module Grape
       # There are a number of documentation options on entities that don't have
       # corresponding validators. Since there is nowhere that enumerates them all,
       # we maintain a list of them here and skip looking up validators for them.
-      RESERVED_DOCUMENTATION_KEYWORDS = %i[as required param_type is_array format example].freeze
+      RESERVED_DOCUMENTATION_KEYWORDS = %i[as required param_type is_array format example x].freeze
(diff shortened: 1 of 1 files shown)


dblock:
What's the effect of this? This is a backwards-incompatible behavior so we need to think about it. Either way the change needs tests and possibly documentation, including UPGRADING.md.

ShallmentMo:
I use `grape-entity` `grape-swagger-entity`, I might do something like this  https://github.com/ruby-grape/grape-swagger-entity/blob/9f44aa2a525be36fec8a7369bf1c1fd7130d879f/spec/grape-swagger/entity/attribute_parser_spec.rb#L103

but when the document contain `x`, grape will trigger `x` validator which raise unknown validator error.

ShallmentMo:
About swagger extension, please visit https://swagger.io/docs/specification/openapi-extensions/

dblock:
If it's something generic and documented I'm open to merging a PR with documentation, tests, etc., will need to read up more on it. In the meantime, does setting `RESERVED_DOCUMENTATION_KEYWORDS` in your project work around the problem?

ShallmentMo:
It's my mistake. I can use nested documentation in entity to avoid this: 
```
class Entity
   expose :attr, documentation: { type: Hash, documentation: { x: { key: "value" } } }
```

Issue #2460, opened 2024-06-25
Error messages depends on previous requests


seriousdev-gh:
Current Behavior: Error messages contain indices of elements from the previous request.

Expected Behavior: Error messages should be idempotent; for two identical requests, I expect the same error message.

Parameters:
```ruby
params do
  requires :items, type: Array do
    requires :item, type: Hash do
      requires :name, type: String
    end
  end
end
```

Sequence of Requests:

Request 1: `{}`
Response: 'items is missing, items[item][name] is missing'

Request 2: `{"items":[{"item":{"name":"foo"}},{"item":{"name":"foo"}},{}]}`
Response: 'items[2][item] is missing, items[2][item][name] is missing'

Request 3: `{}`
Response: 'items is missing, items[2][item][name] is missing'

The response to the last request should be the same as the response to the first request.

This issue is reproducible in version 1.7.1 and in the main branch.

ericproulx:
I could reproduce in within 1.5.3. That's an old one. Thank you for pointing out and the specs

ericproulx:
@seriousdev-gh found a way to fix it

PR #2461, opened 2024-06-25
add failing spec for leaking indices between requests

seriousdev-gh:
failing spec for issue https://github.com/ruby-grape/grape/issues/2460

### DIFF
--- a/spec/grape/validations_spec.rb
+++ b/spec/grape/validations_spec.rb
@@ -352,6 +352,49 @@ describe Grape::Validations do
       end
     end
 
+    # Ensure there is no leakage of indices between requests
+    context 'required with a hash inside an array' do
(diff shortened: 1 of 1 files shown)


spec/grape/validations_spec.rb
```diff
@@ -352,6 +352,49 @@ def define_requires_none
       end
     end
 
+    # Ensure there is no leakage of indices between requests
+    context 'required with a hash inside an array' do
+      before do
+        subject.params do
```
dblock, line 358: Is `subject` being cleared between tests or are we calling `params` on the same `Grape::API`?

dblock:
Fixed in https://github.com/ruby-grape/grape/pull/2463.

Issue #2347, opened 2023-08-22
Parameter renaming doesn't work for hash inside given block.


seriousdev-gh:
Hello.
Parameter renaming doesn't work if there is a hash inside given.

Params:
```
          requires :a, type: Hash do
            optional :c, type: String
            given :c do
              requires :b, type: Hash do
                requires :input_field, as: :output_field
              end
            end
          end
```

Input:
```
{ a: { b: { input_field: 'value' }, c: 'value2' } }
```

After running `declared(params, evaluate_given: true)`:

Expected:
```
{ a: { b: { output_field: 'value' }, c: 'value2' } }
```
Actual:
Parameter input_field is not renamed to output_field.
```
{ a: { b: { input_field: 'value' }, c: 'value2' } }
```

Info: grape version 1.7.1, ruby 3.0.2

dblock:
Try turning this into a spec, let's see that it's still the case in the next 1.8?

seriousdev-gh:
Added spec. Still the case in the latest main branch.

dblock:
This was fixed in https://github.com/ruby-grape/grape/pull/2469, please check it out!

PR #2462, opened 2024-06-25
add spec for renamed parameter in given block

seriousdev-gh:
failing spec for issue https://github.com/ruby-grape/grape/issues/2347


dblock:
This was fixed in https://github.com/ruby-grape/grape/pull/2469, please check it out!

PR #2464, opened 2024-06-28
The `length` validator only takes effect for parameters with types that support `#length` method

OuYangJinTing:
Thanks to @dhruvCW for adding the length validator(#2437).

I found that in the case of optional parameters, passing in nil will lead to an exception. This behavior is not very friendly.
For nil, it should be a better choice to skip verification directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#2471](https://github.com/ruby-grape/grape/pull/2471): Fix absence of original_exception and/or backtrace even if passed in error! - [@numbata](https://github.com/numbata).
 * [#2478](https://github.com/ruby-grape/grape/pull/2478): Fix rescue_from with invalid response - [@ericproulx](https://github.com/ericproulx).
 * [#2480](https://github.com/ruby-grape/grape/pull/2480): Fix rescue_from ValidationErrors exception - [@numbata](https://github.com/numbata).
+* [#2464](https://github.com/ruby-grape/grape/pull/2464): The `length` validator only takes effect for parameters with types that support `#length` method - [@OuYangJinTing](https://github.com/OuYangJinTing).
 * Your contribution here.
 
 ### 2.1.3 (2024-07-13)
--- a/lib/grape/validations/validators/length_validator.rb
+++ b/lib/grape/validations/validators/length_validator.rb
@@ -18,7 +18,7 @@ module Grape
         def validate_param!(attr_name, params)
           param = params[attr_name]
 
-          raise ArgumentError, "parameter #{param} does not support #length" unless param.respond_to?(:length)
+          return unless param.respond_to?(:length)
 
           return unless (!@min.nil? && param.length < @min) || (!@max.nil? && param.length > @max)
 
--- a/spec/grape/validations/validators/length_spec.rb
+++ b/spec/grape/validations/validators/length_spec.rb
@@ -188,11 +188,11 @@ describe Grape::Validations::Validators::LengthValidator do
   end
 
   describe '/type_is_not_array' do
-    context 'raises an error' do
+    context 'does not raise an error' do
       it do
         expect do
           post 'type_is_not_array', list: 12
-        end.to raise_error(ArgumentError, 'parameter 12 does not support #length')
+        end.not_to raise_error
       end
     end
   end
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,14 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 2.2.0
+
+### `Length` validator
+
+After Grape 2.2.0, `length` validator will only take effect for parameters with types that support `#length` method, will not throw `ArgumentError` exception.
+
+See [#2464](https://github.com/ruby-grape/grape/pull/2464) for more information.
+
 ### Upgrading to >= 2.1.0
 
 #### Optional Builder


lib/grape/validations/validators/length_validator.rb
```diff
@@ -18,6 +18,8 @@ def initialize(attrs, options, required, scope, **opts)
         def validate_param!(attr_name, params)
           param = params[attr_name]
 
+          return if param.nil?
```
dhruvCW, line 21: shouldn't this only return if the `required` field is false. We need to store that in an instance member variable now as well.

dhruvCW:
>I found that in the case of optional parameters, passing in nil will lead to an exception. This behavior is not very friendly. For nil, it should be a better choice to skip verification directly.

Ah nice catch 👏  for some reason I thought the validator won't be called if the parameter was not present (and optional) 😅

dblock:
@OuYangJinTing care to answer the q above and update CHANGELOG, please?

lib/grape/validations/validators/length_validator.rb
OuYangJinTing: Sorry, I don't understand that return here will lead to problems such as instance variables. If possible, could you please explain it more clearly?

dhruvCW: sorry for not being clear 😅 . I was wondering if we should be skipping the validation for required parameters or not ?

After thinking about it a `nil` required parameter would fail validation anyways so this should be fine ✅

OuYangJinTing: Thank～

OuYangJinTing:
> Ah nice catch 👏 for some reason I thought the validator won't be called if the parameter was not present (and optional) 😅

If the optional parameters are only passed, the verification will also be performed.
https://github.com/ruby-grape/grape/blob/v2.1.2/lib/grape/validations/validators/base.rb#L53

OuYangJinTing:
@dblock Done ～

spec/grape/validations/validators/length_spec.rb
```diff
@@ -187,6 +193,16 @@
     end
   end
 
+  describe '/nil_param' do
+    context 'no raises errors' do
```
dblock, line 197: Nit: does not raise an error.

README.md
```diff
@@ -1713,6 +1713,7 @@ end
 #### `length`
 
 Parameters with types that support `#length` method can be restricted to have a specific length with the `:length` option.
+In addition, if the received parameter value is nil, the length validation will not be triggered. If you do not allow nil, you can use the `allow_blank: false` option.
```
dblock, line 1716: Quote both `nil`'s.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#2464](https://github.com/ruby-grape/grape/pull/2464): Length validator allow nil - [@OuYangJinTing](https://github.com/OuYangJinTing).
```
dblock, line 5: Let's say "Allow `nil` in `length` validator"?

dblock (commented):
I have some nits, but more importantly, do we have a test for "If you do not allow nil, you can use the `allow_blank: false` option." If not please add?

OuYangJinTing:
> I have some nits, but more importantly, do we have a test for "If you do not allow nil, you can use the `allow_blank: false` option." If not please add?

@dblock Thank you for your feedback. At present, all feedback has been resolved. Please take a look when you have time.

spec/grape/validations/validators/length_spec.rb
```diff
@@ -187,6 +199,34 @@
...
+  describe '/length_with_disallow_blank' do
+    context 'when length is within limits' do
+      it do
+        post '/length_with_disallow_blank', list: [1, 2]
+        expect(last_response.status).to eq(201)
+        expect(last_response.body).to eq('')
+      end
+    end
+
+    context 'When a nil value is passed' do
```
dblock, line 221: Nit: lowercase "When" please?

README.md
```diff
@@ -1713,6 +1713,7 @@ end
 #### `length`
 
 Parameters with types that support `#length` method can be restricted to have a specific length with the `:length` option.
+In addition, if the received parameter value is `nil`, the length validation will not be triggered. If you do not allow `nil`, you can use the `allow_blank: false` option.
```
dblock, line 1716: Add "want to", so "If you do not want to allow..."

dblock (commented):
Looks good, one more typo below.

README.md
dhruvCW: um this will be problematic if you want to allow zero length parameters. For example 

```ruby
requires :some_list, type: [Integer], length: { min: 0, max: 10 }, allow_blank: false
requires :some_string, type: String, length: { min: 0 }, allow_blank: false
```

would both have validation errors if we provided the following parameters `{ some_list: [], some_string: '' }`

IMO this is incorrect behaviour.

lib/grape/validations/validators/length_validator.rb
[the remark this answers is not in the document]
dblock: A `nil` value is a perfectly valid value, that's why we have `allow_blank` (to distinguish the cases when the parameter itself is required and a value for the parameter is required).

README.md
dblock: I think this is exactly the expected behavior. Why do you think this is incorrect @dhruvCW?

dhruvCW: I see like I mentioned combining the two can lead to unexpected results since allow_blank validates both nil as well as empty values. I personally then would appreciate this specific case to be documented so that users are aware about this potential pitfall.

lib/grape/validations/validators/length_validator.rb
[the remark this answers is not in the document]
dhruvCW: I see I guess I contravened required with nil values.

README.md
dblock: Yes, we should, @OuYangJinTing add specs for ^ and README pls?

lib/grape/validations/validators/length_validator.rb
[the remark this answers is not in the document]
dblock: I know it's confusing. We have gone back and forth on this 🙈

README.md
OuYangJinTing: @dblock @dhruvCW Sorry, I'm a little busy with work these two days, and I didn't give timely feedback.
Based on this consideration, I think it should be better to add an allow_nil option to the `Grape::Validations::Validators::LengthValidator`. e.g:

```ruby
module Grape
  module Validations
    module Validators
      class LengthValidator < Base
        def initialize(attrs, options, required, scope, **opts)
          @min = options[:min]
          @max = options[:max]
          @allow_nil = options.key?(:allow_nil) ? options[:allow_nil] : true # allow `allow_nil` option, default to `true`

          super

          raise ArgumentError, 'min must be an integer greater than or equal to zero' if !@min.nil? && (!@min.is_a?(Integer) || @min.negative?)
          raise ArgumentError, 'max must be an integer greater than or equal to zero' if !@max.nil? && (!@max.is_a?(Integer) || @max.negative?)
          raise ArgumentError, "min #{@min} cannot be greater than max #{@max}" if !@min.nil? && !@max.nil? && @min > @max
        end

          param = params[attr_name]

          return if param.nil? && @allow_nil # when `allow_nil` is `true`, `nil` is allowed

          raise ArgumentError, "parameter #{param} does not support #length" unless param.respond_to?(:length)

          return unless (!@min.nil? && param.length < @min) || (!@max.nil? && param.length > @max)

          raise Grape::Exceptions::Validation.new(params: [@scope.full_name(attr_name)], message: build_message)
        end
        ...
end
```

If you agree, I will be happy to change it.

dhruvCW: I agree but I think the default should be `false`. maybe `ignore_nil` would be a better name ? since we are skipping the validation for `nil` values.

OuYangJinTing: @dhruvCW Thank you for your suggestions.
1. I agree that the default value be to false.
2. `ignore_nil` is a good name, but I use the name `allow_nil`, considering that there is already `allow_blank`, and it should be better to use a similar name.

dhruvCW: >ignore_nil is a good name, but I use the name allow_nil, considering that there is already allow_blank, and it should be better to use a similar name.

makes sense to keep the names consistent 😅 👍

dblock: Yes, the other reason not to change these is to remain positive vs. negative.

dblock (changes requested):
I’ll wait to merge this till everyone above is happy, including myself :) Please say “I’m good with merging this” when you are! And thank you so much for having an interesting conversation!

OuYangJinTing:
@dhruvCW @dblock Hello.
According to the above discussion, I refactored the code once. Please help review it when you have time.

dblock:
How does this interact with `allow_blank`?

Document the new validator in https://github.com/ruby-grape/grape?tab=readme-ov-file#built-in-validators.

README.md
```diff
@@ -1714,12 +1714,13 @@ end
 
 Parameters with types that support `#length` method can be restricted to have a specific length with the `:length` option.
 
-The validator accepts `:min` or `:max` or both options to validate that the value of the parameter is within the given limits.
+The validator accepts `:min` or `:max` or both options to validate that the value of the parameter is within the given limits,
+In addition, if need to allow `nil` value, please to pass the `allow_nil: true` option.
```
dblock, line 1718: For another PR, I think we're being too polite by saying `please` in the documentation. It's just a reference, no need to be nice :)

spec/grape/validations/validators/length_spec.rb
```diff
@@ -81,6 +81,18 @@
       post 'zero_max' do
       end
 
+      params do
+        requires :list, type: [Integer], length: { min: 0, allow_nil: true }
+      end
+      post 'allow_nil_param' do
+      end
```
dblock, line 88: Add a test with an explicit `allow_nil: false`.

dblock:
I'd love some other people's opinion on whether we do want the `allow_nil` option or reuse `allow_blank` for the length validator, maybe @ericproulx?

OuYangJinTing:
> Ok, it's just an option on `length`, I understand. What happens if I `allow_blank: true` on the `length` validator? I assume it throws an error that it's not something known?

I think it's not suitable to add the `allow_blank` option here because there is already an `allow_blank` validator. 
Instead of adding an `allow_blank` option, it's better to use the `allow_blank` validator directly.

spec/grape/validations/validators/length_spec.rb
dblock: We still need one test with the default (without any spec of `allow_nil`).

lib/grape/validations/validators/length_validator.rb
```diff
@@ -18,6 +19,12 @@ def initialize(attrs, options, required, scope, **opts)
         def validate_param!(attr_name, params)
           param = params[attr_name]
 
+          if param.nil?
+            return if @allow_nil
```
dhruvCW, line 23: I was looking into the code to understand how `allow_blank` interops with other validators and saw that it's an attribute in the [base class](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/base.rb#L24).

So should we just use this option instead of introducing `allow_nil` ?

dblock: That was my thinking. 🤷‍♂️

dblock: I can help decide: does `allow_nil` actually behave like `allow_blank` in this implementation? Meaning is `nil` vs. `[]` treated the same?

dhruvCW: technically no since `allow_nil` is for only `nil` values while `allow_blank` would mean ignoring both `nil` and `empty` values. Not sure if we want to use `allow_blank` or `allow_nil`. 

I personally do prefer more specific constraints thus `allow_nil` but I am also okay if we decide to ignore `blank` values if the `allow_blank` flag is set to true.

dblock: @OuYangJinTing write specs for these scenarios and let's decide?

Again, thanks so much for being patient. API changes are one-way, we can't easily undo them, so it's worth spending the time.

dblock: @dhruvCW Introducing `allow_nil` here only will make it potentially different from other parameter types. Without looking at specs and current behavior I would say `allow_blank` for arrays/lists should behave like `blank?` and if we wanted a `allow_nil` then we should probably implement it for all other parameter types next to `allow_blank`.

ericproulx:
> I'd love some other people's opinion on whether we do want the `allow_nil` option or reuse `allow_blank` for the length validator, maybe @ericproulx?

Sorry for my late response. Quite busy. I'll take a closer look this week-end

lib/grape/validations/validators/length_validator.rb
OuYangJinTing: > @OuYangJinTing write specs for these scenarios and let's decide?
> 
> Again, thanks so much for being patient. API changes are one-way, we can't easily undo them, so it's worth spending the time.

Sorry, I'm a little confused.
Do you need to roll back to commit [171407aeef935ab918335c5ea324ab18a96a19d4](https://github.com/ruby-grape/grape/commit/171407aeef935ab918335c5ea324ab18a96a19d4) and provide a more detailed explanation?

dblock: Let's wait for @ericproulx to help us out with a strong opinion

OuYangJinTing: Hmm, I'll make the changes after everyone finished feedback.

ericproulx:
There's is still something confusing about the `allow_nil` in association with an array. For instance, if we declare the following:
```ruby
params do
  requires :list, type: [Integer], length: { min: 1, allow_nil: true }
end
```

Does it mean that `[nil]` and `nil` are expected to be valid ? Unfortunately, the `coerce_validator` would raise an error since `nil` is not an Integer.

Although, I've noticed something regarding the `allow_blank` validator. If I'm not mistaken, its evaluation depends on where its declared. For instance, the following test will ✅  since it declared before the length validator (I've added the `fail_fast` on purpose to see which error will be triggered)

```ruby
describe 'when allow_blank is declared before length validator' do
  subject { last_response }

  let(:app) do
    Class.new(Grape::API) do
      params do
        requires :list, type: [Integer], allow_blank: false, length: { max: 1 }, fail_fast: true
      end
      get do
      end
    end
  end
  
  before do
    get '/', list: nil
  end

  it { is_expected.to be_bad_request}
end
```

but this one will ❌  and raise the `ArgumentError: parameter  does not support #length`

```ruby
describe 'when allow_blank is declared after length validator' do
  subject { last_response }

  let(:app) do
    Class.new(Grape::API) do
      params do
        requires :list, type: [Integer], length: { max: 1 }, fail_fast: true, allow_blank: false
      end
      get do
      end
    end
  end
  before do
    get '/', list: nil
  end

  it { is_expected.to be_bad_request}
end
```

Based on that behavior, the `length` validator must not raise an ArgumentError when `param` doesn't `respond_to?(:length)` but `return` to let the `allow_blank` catch the case.
       
Also, regarding the `min` and `max` possible values, I feel that both need to be greater than 0 to also let `allow_blank` cover some cases. `allow_blank: false` would not allow the `nil`, empty strings, strings containing only whitespace characters and `empty?` arrays, hashes.  Though, `[nil]` would not be valid since is not `empty?`.

OuYangJinTing:
@ericproulx Thank you for your detailed insights and explanations with examples.👍

Considering the above, in order to avoid bring too much complexity.
I think the length validator should simply skip nil and explain this situation in the document.

dhruvCW:
> @ericproulx Thank you for your detailed insights and explanations with examples.👍
> 
> Considering the above, in order to avoid bring too much complexity. I think the length validator should simply skip nil and explain this situation in the document.
> 
> cc @dhruvCW @dblock

Should we simply skip all values that don't respond to `length` ? I assume the only time this is true is `nil` since the type coercer should either fail or wrap say an integer if that's passed as the parameter.

dblock:
> > @ericproulx Thank you for your detailed insights and explanations with examples.👍
> > Considering the above, in order to avoid bring too much complexity. I think the length validator should simply skip nil and explain this situation in the document.
> > cc @dhruvCW @dblock
> 
> Should we simply skip all values that don't respond to `length` ? I assume the only time this is true is `nil` since the type coercer should either fail or wrap say an integer if that's passed as the parameter.

We should see what that breaks.

I like the idea that `blank` validator works for anything that `respond_to?(:blank?)` and `nil` validator for everything that `respond_to?(:nil?)` which is I suppose everything. It's probably easier to grok it that way.

OuYangJinTing:
@dblock @dhruvCW According to the feedback, I have adjusted the PR again. Please help me check if it is correct when you are free. Thank you.

ericproulx (approved):

dhruvCW (approved):
Thanks for hanging in there 🍻  LGTM 🚀

dblock (changes requested):
Thanks!

Is this a breaking change or a bug fix? If it's a breaking change, then we need to major-bump version 😱 . If it's a bug fix, then move the CHANGELOG line to fixes.

Either way please add a section to UPGRADING.md.

OuYangJinTing:
> Thanks!
> 
> Is this a breaking change or a bug fix? If it's a breaking change, then we need to major-bump version 😱 . If it's a bug fix, then move the CHANGELOG line to fixes.
> 
> Either way please add a section to UPGRADING.md.

I think this is a bug fix, because before, if parameters with types that don't support `#length` method, it directly throws an exception. In order to avoid exceptions, developers need to correct the api parameters by themselves, and after merging this PR, this situation can be avoided, this should be a security change.

dblock:
> > Thanks!
> > Is this a breaking change or a bug fix? If it's a breaking change, then we need to major-bump version 😱 . If it's a bug fix, then move the CHANGELOG line to fixes.
> > Either way please add a section to UPGRADING.md.
> 
> I think this is a bug fix, because before, if parameters with types that don't support `#length` method, it directly throws an exception. In order to avoid exceptions, developers need to correct the api parameters by themselves, and after merging this PR, this situation can be avoided, this should be a smooth change.

@ericproulx Could use a second opinion, please?

dblock (approved):

ericproulx:
> > > Thanks!
> > > Is this a breaking change or a bug fix? If it's a breaking change, then we need to major-bump version 😱 . If it's a bug fix, then move the CHANGELOG line to fixes.
> > > Either way please add a section to UPGRADING.md.
> > 
> > 
> > I think this is a bug fix, because before, if parameters with types that don't support `#length` method, it directly throws an exception. In order to avoid exceptions, developers need to correct the api parameters by themselves, and after merging this PR, this situation can be avoided, this should be a smooth change.
> 
> @ericproulx Could use a second opinion, please?

Feels like a bug fix to me :)

dblock:
> Feels like a bug fix to me :). Nonetheless, I'm not sure about the UPGRADING comments since its a bug fix.

I think better safe than sorry. Merging.

PR #2467, opened 2024-06-29
Fix repo coverage

### LINKED ISSUE #2466: Coverage is only covering specs
We need to fix the coverage since its just [looking](https://coveralls.io/github/ruby-grape/grape?branch=master) at the spec folders.

ericproulx:
Fix #2466

Also, `lcov` format will be used only when running GH Actions.


ericproulx:
@dblock it will drop the coverage to 96% but its the true one 😅

dblock (approved):

PR #2469, opened 2024-07-04
Fixes #2347 - Correct full path building for lateral scopes

### LINKED ISSUE #2347: Parameter renaming doesn't work for hash inside given block.

numbata:
Should fix the #2347 issue.

This pull request fixes an issue where the full_path method would stop building the full path when encountering lateral scopes like `with` or `given`. The updated method now correctly handles these cases, ensuring accurate path construction

Thanks @seriousdev-gh  for the [spec](https://github.com/ruby-grape/grape/pull/2462).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * [#2467](https://github.com/ruby-grape/grape/pull/2467): Fix repo coverage - [@ericproulx](https://github.com/ericproulx).
 * [#2468](https://github.com/ruby-grape/grape/pull/2468): Align `error!` method signatures across different places - [@numbata](https://github.com/numbata).
+* [#2469](https://github.com/ruby-grape/grape/pull/2469): Fix full path building for lateral scopes - [@numbata](https://github.com/numbata).
 * Your contribution here.
 
 ### 2.1.2 (2024-06-28)
--- a/lib/grape/validations/params_scope.rb
+++ b/lib/grape/validations/params_scope.rb
@@ -190,7 +190,13 @@ module Grape
       #
       # @return [Array<Symbol>] the nesting/path of the current parameter scope
       def full_path
-        nested? ? @parent.full_path + [@element] : []
+        if nested?
+          (@parent.full_path + [@element])
+        elsif lateral?
+          @parent.full_path
+        else
+          []
+        end
       end
 
       private
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -40,7 +40,7 @@ Metrics/BlockLength:
     - spec/**/*_spec.rb
 
 Metrics/ClassLength:
-  Max: 300
+  Max: 305
 
 Metrics/CyclomaticComplexity:
   Max: 15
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 * [#2467](https://github.com/ruby-grape/grape/pull/2467): Fix repo coverage - [@ericproulx](https://github.com/ericproulx).
 * [#2468](https://github.com/ruby-grape/grape/pull/2468): Align `error!` method signatures across different places - [@numbata](https://github.com/numbata).
+* [#2469](https://github.com/ruby-grape/grape/pull/2469): Fixes #2347 - Correct full path building for lateral scopes - [@numbata](https://github.com/numbata).
```
dblock, line 11: Let’s say “Fix” to match other lines.

lib/grape/validations/params_scope.rb
```diff
@@ -190,7 +190,10 @@ def push_declared_params(attrs, **opts)
       #
       # @return [Array<Symbol>] the nesting/path of the current parameter scope
       def full_path
-        nested? ? @parent.full_path + [@element] : []
+        return (@parent.full_path + [@element]) if nested?
```
dblock, line 193: I dislike the double return, can we do 

```
if
…
elsif 
…
else
…
```

then we can omit the returns altogether

dblock (changes requested):
Thanks for fixing this! Forgive me for nitpicking.

lib/grape/validations/params_scope.rb
numbata: Actually, it was my first implementation (with "if-else"), but then rubocop shamed me for huge amount of lines of class.

CHANGELOG.md
numbata: Done! ([diff](https://github.com/ruby-grape/grape/compare/58d54c34e2c33473063eca40cb9bbd3ebb0b4472..cc54313ef17063d19453b5c8ee052536b83f116c#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL11))

lib/grape/validations/params_scope.rb
numbata: fixed [here](https://github.com/ruby-grape/grape/compare/58d54c34e2c33473063eca40cb9bbd3ebb0b4472..cc54313ef17063d19453b5c8ee052536b83f116c#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL11)

dblock: I usually just do `rubocop -a ; rubocop --auto-gen-config`.

dblock (approved):

PR #2471, opened 2024-07-15
Fixes #2470 Expose original_exception and/or backtrace if present

### LINKED ISSUE #2470: original_exception and backtrace are not present even if passed in error!
Since #2468, `error!` signature has been aligned across different places. Unfortunately, even if you pass a `backtrace` and/or `original_exception` arguments, these latter won't be part of the returned message even if `backtrace: true` and/or `original_exception: true` are set at the `rescue_from` level.

Here are the specs

```ruby
 # json
context 'rescue_from include backtrace and original exception' do
  subject { JSON.parse(last_response.body) }

  let(:app) do
    Class.new(Grape::API) do
      format :json

      rescue_from :all, backtrace: true, original_exception: true do |e|
        error!(e.message, 500, nil, e.backtrace, e)
      end

      params do
        requires :param, type: Integer
      end

      get '/' do
        'it works'
      end
    end
  end

  before do
    get '/'
  end

  it { is_expected.to include('error' => a_kind_of(String), 'backtrace' => a_kind_of(Array), 'original_exception' => a_kind_of(String)) }
end

#xml
context 'rescue_from include backtrace and original exception' do
  subject { Grape::Xml.parse(last_response.body)['error'] }

  let(:app) do
    Class.new(Grape::API) do
      format :xml

      rescue_from :all, backtrace: true, original_exception: true do |e|
        error!(e.message, 500, nil, e.backtrace, e)
      end

      params do
        requires :param, type: Integer
      end

      get '/' do
        'it works'
      end
    end
  end

  before do
    get '/'
  end

  it { is_expected.to have_key('backtrace') & have_key('original-exception') }
end

#txt
context 'rescue_from include backtrace and original exception' do
  subject { last_response.body }

  let(:app) do
    Class.new(Grape::API) do
      format :txt

      rescue_from :all, backtrace: true, original_exception: true do |e|
        error!(e.message, 500, nil, e.backtrace, e)
      end

      params do
        requires :param, type: Integer
      end

      get '/' do
        'it works'
      end
    end
  end

  before do
    get '/'
  end

  it { is_expected.to include('backtrace', 'original exception') }
end
```

Although if we specify `backtrace: true` and/or `original_exception: true` and pass the original exception as message like
```ruby
rescue_from :all, original_exception: true, backtrace: true do |e|
  error!(e, 500, nil, e.backtrace, e)
end
```

It will break with an exception when format is json
```ruby
# exception
# NoMethodError:
#    undefined method `merge' for an instance of Grape::Exceptions::ValidationErrors
#  ./lib/grape/error_formatter/json.rb:13:in `call'
#  ./lib/grape/middleware/error.rb:49:in `format_message'

 context 'rescue_from include backtrace and original exception' do
  subject { JSON.parse(last_response.body) }

  let(:app) do
    Class.new(Grape::API) do
      format :json

      rescue_from :all, original_exception: true, backtrace: true do |e|
        error!(e, 500, nil, e.backtrace, e)
      end

      params do
        requires :param, type: Integer
      end

      get '/' do
        'it works'
      end
    end
  end

  before do
    get '/'
  end

  it { is_expected.to include('error' => a_kind_of(String), 'backtrace' => a_kind_of(Array), 'original_exception' => a_kind_of(String)) }
end
```

numbata:
Fix the case where backtrace and/or original_exception arguments are passed, but then do not become part of the returned message, even if `backtrace: true` and/or `original_exception: true` are set at the `rescue_from` level.

Thanks to @ericproulx for pointing out the problem and provided specs.

### DIFF
--- a/lib/grape/dsl/inside_route.rb
+++ b/lib/grape/dsl/inside_route.rb
@@ -170,7 +170,12 @@ module Grape
       def error!(message, status = nil, additional_headers = nil, backtrace = nil, original_exception = nil)
         status = self.status(status || namespace_inheritable(:default_error_status))
         headers = additional_headers.present? ? header.merge(additional_headers) : header
-        throw :error, message: message, status: status, headers: headers, backtrace: backtrace, original_exception: original_exception
+        throw :error,
+              message: message,
+              status: status,
+              headers: headers,
+              backtrace: backtrace,
+              original_exception: original_exception
       end
 
       # Creates a Rack response based on the provided message, status, and headers.
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -120,9 +120,9 @@ module Grape
           handler.arity.zero? ? endpoint.instance_exec(&handler) : endpoint.instance_exec(error, &handler)
         end
 
-        response = error!(response[:message], response[:status], response[:headers]) if error?(response)
-
-        if response.is_a?(Rack::Response)
+        if error?(response)
+          error_response(response)
+        elsif response.is_a?(Rack::Response)
           response
         else
           run_rescue_handler(:default_rescue_handler, Grape::Exceptions::InvalidResponse.new, endpoint)
@@ -137,7 +137,9 @@ module Grape
       end
 
       def error?(response)
-        response.is_a?(Hash) && response[:message] && response[:status] && response[:headers]
+        return false unless response.is_a?(Hash)
+
+        response.key?(:message) && response.key?(:status) && response.key?(:headers)
       end
     end
   end
--- a/spec/grape/api_spec.rb
+++ b/spec/grape/api_spec.rb
@@ -2599,6 +2599,150 @@ describe Grape::API do
       it_behaves_like 'a json format api', :failure
       it_behaves_like 'a json format api', { error: :failure }
     end
+
+    context 'when rescue_from enables backtrace without original exception' do
+      let(:app) do
+        response_type = response_format
+
+        Class.new(Grape::API) do
+          format response_type
+
+          rescue_from :all, backtrace: true, original_exception: false do |e|
+            error!('raining dogs and cats!', 418, {}, e.backtrace, e)
+          end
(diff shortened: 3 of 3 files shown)


ericproulx:
@dblock anything to add ?

numbata:
> The only think that bugs me is the error? function. 

Totally agree. And we can do something around that (in separate PR ofc), but I have fear that will be breaking changes.

ericproulx:
> > The only think that bugs me is the error? function.
> 
> Totally agree. And we can do something around that (in separate PR ofc), but I have fear that will be breaking changes.

If its only for internal/private usage, we should be fine with the breaking changes :)

ericproulx:
Thanks @numbata

numbata:
> If its only for internal/private usage, we should be fine with the breaking changes :)

I just allow that some of the grape users have something like:
```ruby
rescue_from do |e|
  throw :error, message: "foo", status: 200, headers: {}
end
```
or directly:
```ruby
get "/foo" do
  throw :error, message: "oopsie", status: 200, headers: {}
end
```

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 #### Fixes
 
-* Your contribution here.
```
dblock, line 9: Don't remove this line next time ;)

dblock:
> @dblock anything to add ?

Ya'll are fast than me! Thanks.

CHANGELOG.md
numbata: 🙈

numbata: Let's say that I believe that there will only be new features in the next release and no fixes 🤞.

ericproulx: Sorry about that @dblock

numbata: It looks like a new "check" can be added to the "danger-changelog" gem.

dblock: For sure, we take PRs :)

PR #2478, opened 2024-07-23
Fix rescue_from with invalid response

### LINKED ISSUE #2477: Returning an invalid response in a `rescue_from` block causes “NoMethodError: undefined method 'default_rescue_handler'”
If the following code path is executed, an exception “NoMethodError: undefined method 'default_rescue_handler'” is raised.

https://github.com/ruby-grape/grape/blob/56719693012269b0132d09e5e8fd41a906e38a6d/lib/grape/middleware/error.rb#L128

The reason is that the following code does not take private methods into account:

https://github.com/ruby-grape/grape/blob/56719693012269b0132d09e5e8fd41a906e38a6d/lib/grape/middleware/error.rb#L114-L116

The reason for the introduction of the error is that the code referred to above was forgotten to be changed in c6ad84a4b793a21ec8189fd5121749fb20e81f6f, in which `default_rescue_handler` was made private and other code was changed from using `:default_rescue_handler` to using `method(:default_rescue_handler)`. The fix for the problem is to apply that change to the code referred to above as well.

Another problem is that the code path is seemingly untested. While preparing a pull request containing a test and a fix, I was unsure where to put the test, as there are many places where `rescue_from` is tested.

ericproulx:
Fix #2477

cc @manueljacob


ericproulx:
I took the opportunity to fix the last RSpec/AnyInstance cop.

spec/grape/api_spec.rb
```diff
@@ -2196,8 +2196,6 @@ def foo
     it 'uses default_rescue_handler to handle invalid response from rescue_from' do
```
manueljacob, line 2196: I suggest to change the name of the example to reflect that it doesn’t anymore test the implementation detail that it uses `default_rescue_handler`.

manueljacob (commented):
Otherwise LGTM.

spec/grape/api_spec.rb
ericproulx: I made a change

dblock:
@manueljacob you good with this?

spec/grape/api_spec.rb
```diff
@@ -2193,14 +2193,14 @@ def foo
...
 
-    it 'uses default_rescue_handler to handle invalid response from rescue_from' do
-      subject.rescue_from(:all) { 'error' }
-      subject.get('/') { raise }
-
-      expect_any_instance_of(Grape::Middleware::Error).to receive(:default_rescue_handler).and_call_original
-      get '/'
-      expect(last_response).to be_server_error
-      expect(last_response.body).to eql 'Invalid response'
+    context 'when block returns an invalid response' do
```
manueljacob, line 2196:
```suggestion
    context 'when rescue_from block returns an invalid response' do
```

manueljacob (commented):
I have a minor suggestion, otherwise LGTM!

manueljacob (approved):

numbata:
@ericproulx great job! 👍

PR #2484, opened 2024-07-26
Revisit versioner middlewares

ericproulx:
This PR is a refactor of all Grape::Middleware::Versioner

First, they all use the `default_options` like the other middlewares. Since default `version_options` is including another Hash, I had to replace `merge` by `deep_merge` in Grape::Middleware::Base.

Second, the `using` strategy is now [dynamically](https://github.com/ruby-grape/grape/blob/c2f3578b52871eb78d87eaeb9255a8b782d90a03/lib/grape/middleware/versioner.rb#L19) solved like validators. It's all based on the name of the class.

Third, I've added a `versioner_helpers` that all Grape::Middleware::Versioner include. At least, all the `options[...]` are exposed through an helper function which reduces the `[]` calls.


lib/grape/middleware/versioner.rb
```diff
@@ -13,21 +13,12 @@ module Middleware
...
-          Header
-        when :param
-          Param
-        when :accept_version_header
-          AcceptVersionHeader
-        else
-          raise Grape::Exceptions::InvalidVersionerOption.new(strategy)
-        end
+        Grape::Middleware::Versioner.const_get(:"#{strategy.to_s.classify}")
+      rescue NameError
```
dblock, line 20: I don't like that we rely on an exception here. It also allows you to register a strategy without anyone noticing, which ends up being used as an API.

lib/grape/middleware/versioner/accept_version_header.rb
```diff
@@ -17,45 +17,22 @@ module Versioner
...
-          return if potential_version.empty?
+        def before
+          potential_version = env[Grape::Http::Headers::HTTP_ACCEPT_VERSION]&.strip
+          throw_error_not_acceptable('Accept-Version header must be set.') if strict? && potential_version.blank?
 
-          # If the requested version is not supported:
-          throw :error, status: 406, headers: error_headers, message: 'The requested version is not supported.' unless versions.any? { |v| v.to_s == potential_version }
+          return if potential_version.blank?
 
+          throw_error_not_acceptable('The requested version is not supported.') unless potential_version_match?(potential_version)
```
dblock, line 28: Probably `not_acceptable!` makes more sense (needs a bang and is similar to other throws).

lib/grape/middleware/versioner/param.rb
```diff
@@ -19,31 +19,15 @@ module Versioner
...
+        include VersionerHelpers
 
         def before
-          potential_version = Rack::Utils.parse_nested_query(env[Rack::QUERY_STRING])[paramkey]
-          return if potential_version.nil?
+          potential_version = Rack::Utils.parse_nested_query(env[Rack::QUERY_STRING])[parameter_key]
+          return if potential_version.blank?
 
-          throw :error, status: 404, message: '404 API Version Not Found', headers: { Grape::Http::Headers::X_CASCADE => 'pass' } if options[:versions] && !options[:versions].find { |v| v.to_s == potential_version }
+          throw_api_version_not_found unless potential_version_match?(potential_version)
```
dblock, line 28: version_not_found!

dblock (commented):
Comments below. 

Beef up the description of the PR a bit to explain why this is better. Why is it? :)

lib/grape/middleware/versioner.rb
ericproulx: I do understand your point about registering without anyone noticing but on the other hand we're already embracing the pattern through [coercing](https://github.com/ruby-grape/grape/blob/2b8567a1ca49b3d0236fa82deb2edef34c8839cf/lib/grape/validations/types/primitive_coercer.rb#L35) and [validators](https://github.com/ruby-grape/grape/blob/2b8567a1ca49b3d0236fa82deb2edef34c8839cf/lib/grape/validations.rb#L25). Now, everyone can build their own `versioner` and they just need to follow the naming convention like the validators.

ericproulx: In the future we could imagine having a separate gem just for versioning. It's a bit like how [Faraday](https://github.com/lostisland/faraday/blob/main/UPGRADING.md#adapters-have-moved) evolved through time by moving out the adapters. I know versioning ain't that big compared to adapters but its the idea

dblock: Ok, I buy the argument of we're already doing it.

dblock (commented):
The entire spec/grape/util/accept_header_handler_spec.rb is gone. Where ele do we cover the accept version negotiation and x-cascade handling for that one in tests?

ericproulx:
> The entire spec/grape/util/accept_header_handler_spec.rb is gone. Where ele do we cover the accept version negotiation and x-cascade handling for that one in tests?

There's already the accept_version_header_spec that covers all cases. Coverage is 100%

PR #2485, opened 2024-07-27
Add `is:` parameter to the `length`  validator

Dakad:
This PR is to complement the `length` validator introduced in #2437.
It adds an `exact:`  param to verify that the given param is exactly equal to a limit.


lib/grape/validations/attributes_doc.rb
```diff
@@ -29,6 +29,7 @@ def extract_details(validations)
 
         details[:default] = validations[:default] if validations.key?(:default)
 
+        details[:exact_length] = validations[:length][:exact] if validations.key?(:length) && validations[:length].key?(:exact)
```
Dakad, line 32: Will undo this change bc this details key is not part of the swagger specs. 
REF: https://swagger.io/docs/specification/data-models/data-types#string

dblock:
I like it. I think we may want to call this `equals` or even `value` to match other value parameters, wdyt?

Dakad:
> I like it. I think we may want to call this `equals` or even `value` to match other value parameters, wdyt?

For me, `equals` is better than `value`. What about `equal` without **s**? 
Either way, I'm not too concerned about the naming :smile:

dblock:
Is `equal` like `min` and `max`? I think `value` is more like those, but I am flexible :) What does @dhruvCW who introduced this parameter, or @ericproulx think?

Stepping back, I think `length` should simply accept a value, so 

```
 requires :code, type: String, length: 2
```

Dakad:
> Is `equal` like `min` and `max`? I think `value` is more like those, but I am flexible :)

Not sure I understand the 1st question.

If I can add my two cents, I would prefer to split into distinct length validators.
In my company, before the length validator was introduced, we currently have 3 distinct validators for the length, `exact_length: `, `min_length: ` and `max_length:` . They have `_length` suffix because we don't want to confuse the with max or min parameter value and to have use validator to such as `max_element:`, ....

dblock:
I think it's pretty easy to grok `length: { min: X, max: Y }` and `length: 2`, would love to hear other (strong) opinions.

Note that we did add `length` in #2437, and it was released, so switching to 3 validators would be a breaking change, and we should be convinced that it's somehow a lot better if we want to deprecate one and introduce the other.

dblock:
I'd be good with `is`, but Grape also has `value` elsewhere (not in a validator), so that's why I suggested that.

> Suggestion: having `in` would be great too to define a range

For large ranges doesn't that explode them causing potentially other problems? But otherwise we could support it along with `min` and `max`.

I think @Dakad is looking for some consensus. What's your strong opinion @ericproulx?

ericproulx:
> I'd be good with `is`, but Grape also has `value` elsewhere (not in a validator), so that's why I suggested that.
> 
> > Suggestion: having `in` would be great too to define a range
> 
> For large ranges doesn't that explode them causing potentially other problems? But otherwise we could support it along with `min` and `max`.
> 
> I think @Dakad is looking for some consensus. What's your strong opinion @ericproulx?

`value` is used is `except_validator` and wording makes sense since the name of the validator is `except_value`. For length, I would stick to `length { is: 1 }`.

About the range, Ruby will manage its within a `Range` object. I'm not worried about large ranges since endless ranges are managed since Ruby 2.6.

Dakad:
Change done!

PR #2496, opened 2024-09-08
Reduce object allocation while compiling

ericproulx:
This PR reduces object allocation when compiling an api.

The function `prepare_path` was calling `inheritable_setting.to_hash` 2 times to just keep `namespace_stackable` and `namespace_inheritable`.

### Current allocation / bytes
```
Total allocated: 62277446 bytes (334535 objects)
Total retained:  5954533 bytes (15516 objects)
```

### Current allocation / types
```
allocated objects by class
    218676  Array
     89913  Hash
```

### This PR allocation/bytes
```
Total allocated: 51007518 bytes (220812 objects)
Total retained:  5954325 bytes (15514 objects)
```

### This PR allocation/types
```
allocated objects by class
    128512  Array
     66354  Hash
```

This PR also changes the following
 - use `include` instead of `.send(:include)`
 - create a helpers module only if a helper is defined otherwise it's nil
 - Remove `|| {}` in favor of handling `nil` cases in some options passed to middlewares.


numbata:
Fantastic job 👍

PR #2500, opened 2024-09-25
Remove file method

ericproulx:
This PR removes the deprecated `file` method.


dblock:
Which version did we deprecate this in? If it's < 2.x then let's delete. Otherwise I think we should have a policy of deprecating in one major and deleting in the next major.

Issue #2171, opened 2021-03-30
env['api.format'] does not override formatter set by content-type 


waynerobinson:
In reference to #1989 (sorry, I have commented in there before I realised the issue was closed).

I'm having exactly the same issue.

If this is used:

```ruby
env['api.format'] = :binary # or :txt
content_type 'application/json' # or header 'Content-Type', 'application/json'
{key: 123}.to_json
```

Then it still calls the JSON formatter. If the `content_type` method is left off, it doesn't.

Here is an example using grape-on-rack:

https://github.com/waynerobinson/grape-on-rack

Please take a look at the extra methods added to `Acme::Ping`.

The only one that doesn't call the JSON formatter is the one where the content type is set to `text/plain`.

waynerobinson:
It looks like the value of the content type overrides the format. See https://github.com/ruby-grape/grape/blob/02d7113d09eb9fcb4264c841d1fdd305e3e8adb5/lib/grape/middleware/formatter.rb#L55

I'm not sure what the safest way is to reverse this?

waynerobinson:
If this test shows the intended behaviour, I don't think we'll be able to override the output by just changing the order the `api_format` is looked up in.

https://github.com/ruby-grape/grape/blob/7741f022b5928f0ddcee27de08555d95ea4556f8/spec/grape/api_spec.rb#L3732-L3739

If we did, it would break the behaviour where there is an explicit format set, but the implementation expects to be able to override it by changing the content type.

For example:

```ruby
format :json

def action
  content_type "text/plain"
  "plain text"
end
```

would still output as JSON if we reversed the lookup to `api_format = env[Grape::Env::API_FORMAT] || mime_types[headers[Grape::Http::Headers::CONTENT_TYPE]]` because it would always pull out `:json` first.

I don't think there's any non-breaking way to override the behaviour of one of the content-type associated formatters at the moment on a per-request basis.

iabdulin:
here's my use case:
I build JSON in the database and want to return it as is, no additional formatting is needed

waynerobinson:
Yeah. Our use case is similar. We have cached data that's already JSON that we'd like to just return vs having to parse and re-encode. 

I'll take a look at starting a PR tomorrow with some options. 👍

trialize:
There is an issue with order of conditions in **Grape::Middleware::Formatter.fetch_formatter**

Original code:
`      
      def fetch_formatter(headers, options)
[...]

PR #2505, opened 2024-10-12
Prioritize env API_FORMAT over Rack::CONTENT_TYPE header

trialize:
As per https://github.com/ruby-grape/grape/issues/2171

# Fix the Order of Conditions in `fetch_formatter` to Correctly Use Custom Formatters

**Issue:**

In `Grape::Middleware::Formatter.fetch_formatter`, the current order of conditions causes custom formatters specified in `env[Grape::Env::API_FORMAT]` to be ignored in favor of the default formatter associated with the `Content-Type` header.

**Original Code:**

```ruby
def fetch_formatter(headers, options)
  api_format = mime_types[headers[Grape::Http::Headers::CONTENT_TYPE]] || env[Grape::Env::API_FORMAT]
  Grape::Formatter.formatter_for(api_format, **options)
end
```

**Problem:**

This code prioritizes the `Content-Type` header over the `env[Grape::Env::API_FORMAT]`. As a result, even if a custom formatter is explicitly set in the environment, it may be overridden by the default formatter determined by the `Content-Type`.

**Solution:**

Swap the order of conditions so that `env[Grape::Env::API_FORMAT]` takes precedence:

```ruby
def fetch_formatter(headers, options)
  api_format = env[Grape::Env::API_FORMAT] || mime_types[headers[Grape::Http::Headers::CONTENT_TYPE]]
  Grape::Formatter.formatter_for(api_format, **options)
end
```

**Rationale:**

By giving priority to `env[Grape::Env::API_FORMAT]`, we ensure that custom formatters are correctly recognized and applied, allowing for more flexible and accurate response formatting.

**Example:**

```ruby
# Define a custom formatter
content_type :raw_json, 'application/json'
formatter :raw_json, ->(object, _env) { object }

# Use the custom formatter within an endpoint
get do
  env['api.format'] = :raw_json
  body <<-HERE
  {
    "glossary": {
      "title": "example glossary",
      "GlossDiv": {
        "title": "S",
        "GlossList": {
          "GlossEntry": {
            "ID": "SGML",
            "SortAs": "SGML",
            "GlossTerm": "Standard Generalized Markup Language",
            "Acronym": "SGML",
            "Abbrev": "ISO 8879:1986",
            "GlossDef": {
              "para": "A meta-markup language, used to create markup languages such as DocBook.",
              "GlossSeeAlso": ["GML", "XML"]
            },
            "GlossSee": "markup"
          }
        }
      }
    }
  }
  HERE
end
```

**Before the Fix:**

The custom formatter `:raw_json` is not applied. Instead, Grape uses its default JSON formatter.

**After the Fix:**

The custom formatter `:raw_json` is correctly applied, and the response body is returned exactly as intended.

### DIFF
--- a/lib/grape/middleware/formatter.rb
+++ b/lib/grape/middleware/formatter.rb
@@ -53,7 +53,7 @@ module Grape
       end
 
       def fetch_formatter(headers, options)
-        api_format = mime_types[headers[Rack::CONTENT_TYPE]] || env[Grape::Env::API_FORMAT]
+        api_format = env[Grape::Env::API_FORMAT] || mime_types[headers[Rack::CONTENT_TYPE]]
         Grape::Formatter.formatter_for(api_format, options[:formatters])
       end
 


dblock (changes requested):
Looks reasonable. 

This needs a test, a CHANGELOG entry, a paragraph in UPGRADING, and some documentation about order of formatters in README, please.

This is a breaking change, but also a bug. If no existing tests fail we are good with the above.

trialize:
That's interesting. Given the test (bellow), it should overwrite formatter based on `content_type` within endpoint itself.

```ruby
      it 'can be overwritten with an explicit content type' do
        subject.get '/meaning_of_life_with_content_type' do
          content_type 'text/plain'
          { meaning_of_life: 42 }.to_s
        end
        get '/meaning_of_life_with_content_type'
        expect(last_response.body).to eq({ meaning_of_life: 42 }.to_s)
      end
```

Which kind of aligns with following sections in documentation
-  https://github.com/ruby-grape/grape/tree/master?tab=readme-ov-file#content-type
-  https://github.com/ruby-grape/grape/tree/master?tab=readme-ov-file#sending-raw-or-no-data

Meaning, it's expected that

```ruby
class API < Grape::API
  get '/home_timeline_js' do
    content_type 'application/javascript'
    "var statuses = ...;"
  end
end
```

and

```ruby
class API < Grape::API
  get '/file' do
    content_type 'application/octet-stream'
    File.binread 'file.bin'
  end
end
```

Have to skip any formatters (if not defined for those content_types).

trialize:
Closed due to lack of time to address the case according to CONTRIBUTING requirements.

Following code

```ruby
get do
  content_type 'application/json'
  env['api.format'] = :binary
  '{ "meaning_of_life": 42 }'
end
```

Still would apply json formatter and produce `"{ \"meaning_of_life\": 42 }"` which is a bug.

As a workaround, for those having the need, following could be used on separate endpoints 

```ruby
module API
  class myAPI < Grape::API
    resource :raw_json do

    content_type :raw_json, 'application/json'
    formatter :raw_json, ->(object, _env) { object }

    get do
      content_type 'application/json'
      '{ "myRawJson": 123 }'
    end

  end
end
```

dblock:
Hopefully someone can finish this, please!

PR #2506, opened 2024-10-20
Fix fetch_formatter api_format

### LINKED ISSUE #2171: env['api.format'] does not override formatter set by content-type

ericproulx:
Fix #2171

This PR also introduce the function `api_format` which can be used in route. Instead of `env['api.format']`, we can now user `api_format :format`. Seems better than exposing internal `api.format`.

I've also took the opportunity to set `HTTP_VERSION` in `Grape::Http::Headers` and use it. Finally, instead of using `[] ||` pattern for a Hash, I've changed it to `.fetch() {}` which seems a bit faster.

```
irb(main):002* Benchmark.ips do |x|
irb(main):003*   x.report("fetch") { {}.fetch(:non_existance, true) }
irb(main):004*    x.report("[] ||") { {}[:non_existance] || true }
irb(main):005> end; nil
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]
               fetch     1.511M i/100ms
               [] ||     1.446M i/100ms
               fetch     15.149M (± 3.1%) i/s   (66.01 ns/i) -     77.071M in   5.093208s
               [] ||     14.418M (± 1.1%) i/s   (69.36 ns/i) -     72.310M in   5.015870s
=> nil
```


UPGRADING.md
```diff
@@ -3,6 +3,12 @@ Upgrading Grape
 
 ### Upgrading to >= 2.3.0
 
+### `content_type` vs `api.format` inside API
+
+Before 2.3.0, `content_type` would have priority over `env['api.format']` when set in an API. The priority has been flipped and `env['api.format']` will be checked first.
+In addition, the function `api_format` has been added. Instead of setting `env['api.format']` directly, you can call `api_format`.
+See [#2506] (https://github.com/ruby-grape/grape/pull/2506) for more information.
```
dblock, line 10:
```suggestion
See [#2506](https://github.com/ruby-grape/grape/pull/2506) for more information.
```

UPGRADING.md
```diff
@@ -3,6 +3,12 @@ Upgrading Grape
 
 ### Upgrading to >= 2.3.0
 
+### `content_type` vs `api.format` inside API
+
+Before 2.3.0, `content_type` would have priority over `env['api.format']` when set in an API. The priority has been flipped and `env['api.format']` will be checked first.
```
dblock, line 8:
```suggestion
Before 2.3.0, `content_type` had priority over `env['api.format']` when set in an API, which was incorrect. The priority has been flipped and `env['api.format']` will be checked first.
```

dblock (changes requested):
Some text nits. Thanks.

PR #2507, opened 2024-10-23
Fix Set type with values

nikolai-b:
Not sure if this wants a change log as it feels more like a bug fix.


dblock (changes requested):
Thanks for fixing this. 

1. Yes this needs a CHANGELOG entry since it's a bug fix, thanks for fixing it! 
2. Should the check be doing an `is_a?(Enumerable)`?

dblock (approved):

PR #2511, opened 2024-11-04
Fix Grape allowing invalid headers to be set

### LINKED ISSUE #2334: Grape allows invalid headers to be set.
Follow up from <https://github.com/socketry/protocol-rack/issues/2#issuecomment-1577206428>.

The following implementation allows non-string header values.

https://github.com/ruby-grape/grape/blob/d1dfdccca242e00d4d88064a013f92e89a76dbd2/lib/grape/dsl/headers.rb#L13

Technically, all headers should be strings, according to the rack spec.

```ruby
header "x-foo", 123
```

I'm not sure if we should change this. I see several options:

- Emit a warning if it's not a string.
- Convert values to a string.
- Do nothing.
- Later on, map `@header` key values to string values, (or array of string values, allowed by Rack 3+).

SlakrHakr:
Fixes #2334

Ensure all header values are strings according to the Rack spec.

* Convert header values to strings using `to_s` in the `header` method in `lib/grape/dsl/headers.rb`.
* Emit a warning if the header value is not a string in the `header` method in `lib/grape/dsl/headers.rb`.
* Add tests in `spec/grape/dsl/headers_spec.rb` to verify that non-string header values are converted to strings and warnings are emitted.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/ruby-grape/grape/issues/2334?shareId=639f2aab-a49d-4921-b2f6-a83f870c960d).


dblock (changes requested):
Thanks.

This needs some docs (README, CHANGELOG, possibly UPGRADING depending on the answers to below).

1. What is the current behavior of `header 'integer key', 123`? I am afraid we may be breaking some existing behavior.
2. I don't think we need a warning if we're going to automatically coerce every header value `.to_s`, we should probably either loudly fail or document it as such.

SlakrHakr:
added documentation to README, CHANGELOG

based on my answer above, is documentation in UPGRADING needed?

SlakrHakr:
fixed TOC and added documentation to UPGRADING doc

ericproulx:
> > added documentation to README, CHANGELOG
> > based on my answer above, is documentation in UPGRADING needed?
> 
> The fact that a warning is emitted now will surprise users, and we are trying them to change the code to do an explicit `.to_s`, so I think an entry in UPGRADING that explains this is in order, please. It doesn't hurt anyway.
> 
> @SlakrHakr the TOC needs a fix, see danger complaining.
> 
> @ericproulx WDYT about this change?

It makes sense but maybe we should emit a Deprecation instead of a warn. RSpec can be configured with `raise_errors_for_deprecations!` and that would probably help while upgrading instead of relying on a `warn` that emits on `stderr`.

ericproulx:
This [suggestion](https://github.com/ruby-grape/grape/issues/2334#issuecomment-1581599323) could be a nice feature since its hard to comply 100% to the rack-spec. I think we could add something like `lint_api!` similar to `do_not_route_options!`. People would have the choice to add or not a condition to enforce it since it would be disabled by default.

dblock:
I think I like a deprecation more as well. WDYT @SlakrHakr ?

ericproulx:
@SlakrHakr do you need some help ? I might make a release this weekend and your PR could be a part of it. Thanks :)

ericproulx:
I don't think this is still relevant since we're using Rack Headers and it will convert keys as expected by the spec.

PR #2512, opened 2024-11-11
Optimize hash alloc

ericproulx:
This PR is all about hash allocation.

### Double splat operator
Many functions had  a `**options` or `**_options` where a simple `options` or `options = {}` can do the work.

Double splat is useful when keywords are passed explicitly or when you have keyword arguments. Something like

```ruby
def my_function(**options)
  puts options
end

my_function(first_param: 'a', second_param: 'b')
# options = { first_param: 'a', second_param: 'b' }
```
```ruby
def my_function(first_param:, **options)
  puts options
end

my_function(first_param: 'a', second_param: 'b')
# options = { second_param: 'b' }
```

Most of our internal functions are just passing a plain Hash that had to be `**` because of the declared **options. The validator factory is a great example
```ruby
module Grape
  module Validations
    class ValidatorFactory
      def self.create_validator(**options)
        options[:validator_class].new(options[:attributes],
                                      options[:options],
                                      options[:required],
                                      options[:params_scope],
                                      **options[:opts])
      end
    end
  end
end
```
the `create_validator` function is called like this
```ruby
 yield Grape::Validations::ValidatorFactory.create_validator(**saved_validation)
```
and `saved_validation` is a plain Hash

### Namespace

I found a `@namespace_description` in `lib/grape/dsl/routing.rb#namespace` that wasn't use at all so I removed it. It was creating some hashes.

### Exceptions

- `Grape::Exceptions::Base` had a `**_options`. I had to update `Grape::Exceptions::Validation` declaration and `super` usage to fit the expected parameters.

### Validators
`Grape:: Validations::Validators::Base` had a `*opts`. It's been replaced by `opts` since its a [static hash](https://github.com/ruby-grape/grape/blob/0477baf66feb02833c0462f5f0dd4da4036703aa/lib/grape/validations/params_scope.rb#L520)

### Routing HEAD
I've optimized the `mount_in` function and `add_head_not_allowed_methods_and_options_methods`.

`HEAD` and `GET` are [identical](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.2) except that `HEAD` does not return content. Instead of creating a new Route for the `HEAD`, I've just `dup` the `GET` and replace the `@request_method` to HEAD

### Benchmark

Before

```
Total allocated: 51146165 bytes (221499 objects)
Total retained:  5971658 bytes (15564 objects)

allocated memory by class
  27633240  Hash  <=====
   9604928  String
   6485704  Array
   4663405  Regexp
   2274584  ActiveSupport::OrderedOptions
    178224  Grape::Router::Route
    166896  Grape::Router::Pattern
     72720  Grape::Path
     30432  MatchData
     29400  Grape::Router::GreedyRoute
      1704  Class
      1224  Enumerator
       560  Proc
       488  Module
       448  Rack::Utils::HeaderHash


retained memory by class
   2274584  ActiveSupport::OrderedOptions
   1811845  Regexp
    933629  String
    305928  Array
    267800  Hash  <=====
    178224  Grape::Router::Route
    166896  Grape::Router::Pattern
     29400  Grape::Router::GreedyRoute
      1240  Class
       488  Module


```

```
Total allocated: 29384381 bytes (176924 objects)
Total retained:  5177218 bytes (12359 objects)

allocated memory by class
   9604928  String
   6699632  Hash  <=====
   6157528  Array
   4663405  Regexp
   1810584  ActiveSupport::OrderedOptions
    178224  Grape::Router::Route
    130896  Grape::Router::Pattern
     72720  Grape::Path
     30432  MatchData
     29400  Grape::Router::GreedyRoute


retained memory by class
   1811845  Regexp
   1810584  ActiveSupport::OrderedOptions
    933629  String
    178224  Grape::Router::Route
    144320  Hash  <=====
    134968  Array
    130896  Grape::Router::Pattern
     29400  Grape::Router::GreedyRoute
      1240  Class
       488  Module
       400  Proc
       280  Grape::Middleware::Stack::Middleware
       256  <<Unknown>>
       224  Rack::Utils::HeaderHash
       168  ActiveSupport::HashWithIndifferentAccess
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [#2500](https://github.com/ruby-grape/grape/pull/2500): Remove deprecated `file` method - [@ericproulx](https://github.com/ericproulx).
 * [#2501](https://github.com/ruby-grape/grape/pull/2501): Remove deprecated `except` and `proc` options in values validator - [@ericproulx](https://github.com/ericproulx).
 * [#2502](https://github.com/ruby-grape/grape/pull/2502): Remove deprecation `options` in `desc` - [@ericproulx](https://github.com/ericproulx).
+* [#2512](https://github.com/ruby-grape/grape/pull/2512): Optimize hash alloc - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 #### Fixes
--- a/benchmark/compile_many_routes.rb
+++ b/benchmark/compile_many_routes.rb
@@ -3,23 +3,20 @@
 $LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', 'lib'))
 require 'grape'
 require 'benchmark/ips'
-require 'memory_profiler'
 
 class API < Grape::API
   prefix :api
   version 'v1', using: :path
 
-  namespace :test do
-    2000.times do |index|
-      namespace index do
-        get do
-          'hello'
-        end
-      end
+  2000.times do |index|
+    get "/test#{index}/" do
+      'hello'
     end
   end
 end
 
-MemoryProfiler.report(allow_files: 'grape') do
-  API.compile!
-end.pretty_print(to_file: 'without_double_splat.txt')
+Benchmark.ips do |ips|
+  ips.report('Compiling 2000 routes') do
+    API.compile!
+  end
+end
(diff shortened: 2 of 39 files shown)


dblock (approved):
👏

PR #2513, opened 2024-11-27
Optimize Grape::Path -> Reduce Hash Allocation

ericproulx:
This PR refactors the usage of Grape::Path and it comes with a lower Hash allocation. Most of the changes are in `Grape::Path` and the `to_routes` function in `Grape::Endpoint`

This is the result taken from the `benchmark/compile_many_routes.rb`
```ruby
require 'memory_profiler'

class API < Grape::API
  prefix :api
  version 'v1', using: :path

  2000.times do |index|
    get "/test#{index}/" do
      'hello'
    end
  end
end

MemoryProfiler.report(allow_files: 'grape') do
  API.compile!
end.pretty_print
```

Screenshot 2024-11-26 at 20 34 00
Screenshot 2024-11-26 at 20 33 34


dblock:
Alright thanks!

PR #2516, opened 2024-12-20
Dynamic registration

ericproulx:
This PR brings a common way to load parsers, formatters, validators and middleware versioners. With this change, it facilitates the customization through inheritance.

## Parsers, formatters and error formatters

Parsers, formatters and error formatters had static constant named `DEFAULTS` that was acting as a lookup to translate a format (:json, :xml, etc ...) to an actual class. Instead of relying on this, it will now use a dynamic registry `Grape::Util::Registry` that will be warmed up through inheritance like the validators (See `base` classes) Thus, `zeitwerk`, [eager loads](https://github.com/ericproulx/grape/blob/30b3a4377172814ca0bf0c72473f1df1bf7abcfb/lib/grape.rb#L74) classes, so the registries are filled up when Grape is required.

## Versioners

I applied the same pattern for middleware versioners since its translating a symbol to an actual class.

## Validators

For the validators, I've just replace the actual registry by `Grape::Util::Registry` and I removed the `const_get` part since all validators are eager loaded.

### Contract scope
I moved the contract scope validator to the `validators` folders and rename it to `ContractScopeValidator` to have a better short name than `validator`.

## Grape::Util::Registry
#2437 introduced a `length` validator and users that already have one with the same name might not be aware that its been overridden. Giving the situation, the registry will emit a warning if an entry has changed. This is true for all classes that are using the registry.

## DryTypeCoercer
I've refactored the part where its was calling `const_get` to use a static array.


dblock (commented):
Should we mention that `Grape::XYZ::DEFAULTS` no longer exists in UPGRADING?

ericproulx:
> Should we mention that `Grape::XYZ::DEFAULTS` no longer exists in UPGRADING?

Not sure its relevant since its been added 5 month ago. [PR](https://github.com/ruby-grape/grape/pull/2475)
I'm bringing back the registrable but on a new form.

dblock (approved):

PR #2524, opened 2025-01-19
Validators bad encoding

### LINKED ISSUE #2523: ArgumentError invalid byte sequence in UTF-8
Image

```ruby
# lib/grape/validations/validators/allow_blank_validator.rb
# frozen_string_literal: true

module Grape
  module Validations
    module Validators
      class AllowBlankValidator < Base
          return if (options_key?(:value) ? @option[:value] : @option) || !params.is_a?(Hash)

          value = params[attr_name]
          value = value.strip if value.respond_to?(:strip)

          return if value == false || value.present?

          raise Grape::Exceptions::Validation.new(params: [@scope.full_name(attr_name)], message: message(:blank))
        end
      end
    end
  end
end
```

ericproulx:
Fix #2523

It fixes the `allow_blank`, `regexp` and `values` validator.
It also covers the `Accept-Version` http header.


dblock (approved):

dblock:
LGTM, rebase.

Issue #2527, opened 2025-01-24
Was it intentional to hide the http response status from the error formatter?


drewnichols:
I'm curious if this was discussed before and done intentionally. 

For some reason, the `status` value is not passed into the `format_message` method here in `error!`:

https://github.com/ruby-grape/grape/blob/5ce44def9ce7026ad140a31a146820d081ab6798/lib/grape/middleware/error.rb#L132-L137

Passing it would make including it in the response body JSON easier. This is the way it's included in the [JSON API error response bodies](https://jsonapi.org/examples/#error-objects):

```yaml
# HTTP/1.1 422 Unprocessable Entity
# Content-Type: application/vnd.api+json

{
  "errors": [
    {
      "status": "422",
      "source": { "pointer": "/data/attributes/firstName" },
      "title":  "Invalid Attribute",
      "detail": "First name must contain at least two characters."
    }
  ]
}
```

dblock:
I don't think this was intentional at all, please feel free to PR a change.

drewnichols:
I will make it so

ericproulx:
Ain't the most intuitive way but you can access the status from the `env` variable which is Rack's `env`.

You can access endpoint's status code like this `env[Grape::Env::API_ENDPOINT].status`

```ruby
describe 'error formatter env' do
  subject { last_response.body }

  let(:app) do
    Class.new(described_class) do
      format :json

      error_formatter :json, ->(message, _backtrace, _options, env, _original_exception) {
        "#{message}:#{env[Grape::Env::API_ENDPOINT].status}"
      }

      rescue_from :all do
        error!("I'm a teapot!", 418)
      end

      get { raise StandardError }
    end
  end

  before do
    get '/'
  end

  it { is_expected.to eq("I'm a teapot!:418") }
end
```

@dblock To be more intuitive, I would suggest to revisit the signature to also include `headers`. It would be a breaking change obviously but for the better good I think.

drewnichols:
The `env[Grape::Env::API_ENDPOINT].status` seems to be set in many cases but not all. I found that it works when I have custom `rescue_from` blocks like the following. 

```ruby
  rescue_from NotAuthenticatedError do |e|
    error!(e.message, 401, {}, e.backtrace, e)
  end
```

Ironically for grape errors rescued with the following 
```ruby
    rescue_from :grape_exceptions
```

we get 
```
Grape::Env::API_ENDPOINT].status == 201
```

[...]

PR #2528, opened 2025-01-28
Passing status to the ErrorFormatter

drewnichols:
The `status` paramter is not passed into the `format_message` method here in `error!`:

https://github.com/ruby-grape/grape/blob/5ce44def9ce7026ad140a31a146820d081ab6798/lib/grape/middleware/error.rb#L132-L137

Passing it allow it to be included in the response body JSON. This is the way it's included in the [JSON API error response bodies](https://jsonapi.org/examples/#error-objects):

```yaml
# HTTP/1.1 422 Unprocessable Entity
# Content-Type: application/vnd.api+json

{
  "errors": [
    {
      "status": "422",
      "source": { "pointer": "/data/attributes/firstName" },
      "title":  "Invalid Attribute",
      "detail": "First name must contain at least two characters."
    }
  ]
}
```


lib/grape/error_formatter/base.rb
```diff
@@ -4,7 +4,7 @@ module Grape
   module ErrorFormatter
     class Base
       class << self
-        def call(message, backtrace, options = {}, env = nil, original_exception = nil)
+        def call(message, backtrace, options = {}, env = nil, original_exception = nil, _status = nil) # rubocop:disable Metrics/ParameterLists
```
dblock, line 7: Use `rubocop -a ; rubocop --auto-gen-config` instead of an inline `#rubocop:disable`, we have a lot of violations.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 * [#2514](https://github.com/ruby-grape/grape/pull/2514): Add rails 8.0 to CI - [@ericproulx](https://github.com/ericproulx).
 * [#2516](https://github.com/ruby-grape/grape/pull/2516): Dynamic registration for parsers, formatters, versioners - [@ericproulx](https://github.com/ericproulx).
 * [#2518](https://github.com/ruby-grape/grape/pull/2518): Add ruby 3.4 to CI - [@ericproulx](https://github.com/ericproulx).
+* [#2528](https://github.com/ruby-grape/grape/pull/2528): Passed status to ErrorFormatter - [@drewnichols](https://github.com/drewnichols).
```
dblock, line 14:
```suggestion
* [#2528](https://github.com/ruby-grape/grape/pull/2528): Pass `status` into `ErrorFormatter` - [@drewnichols](https://github.com/drewnichols).
```

dblock (changes requested):
Looks good, likely needs documentation in https://github.com/ruby-grape/grape?tab=readme-ov-file#exception-handling and see if we can add lower level tests that a custom formatter for this.

lib/grape/error_formatter/base.rb
drewnichols: done

README.md
```diff
@@ -2772,8 +2772,8 @@ Custom error formatters for existing and additional types can be defined with a
 
 ```ruby
-  error_formatter :txt, ->(message, backtrace, options, env, original_exception) {
-    "error: #{message} from #{backtrace}"
+  error_formatter :txt, ->(message, backtrace, options, env, original_exception, status) {
```
dblock, line 2775: Is this a breaking change? Meaning if I had the old code and upgrade Grape, will this fail with an invalid number of parameters?

dblock (commented):
Question below. 

If the answer is yes, we need a backwards compatible implementation, maybe passing `status` within `options`?

README.md
drewnichols: Best I could tell this calls `Grape::Middleware::Error.format_message` where I defaulted status to `nil`. If someone created a custom Middleware there could be a problem. Fairpoint though.. I can check this with my own app and also look into making some more specs to verify.

ericproulx: Ain't the most intuitive way but you can access the status from the `env` variable which is Rack's `env`.

You can access endpoint's status code like this `env[Grape::Env::API_ENDPOINT].status`

```ruby
describe 'error formatter env' do
  subject { last_response.body }

  let(:app) do
    Class.new(described_class) do
      format :json

      error_formatter :json, ->(message, _backtrace, _options, env, _original_exception) {
        "#{message}:#{env[Grape::Env::API_ENDPOINT].status}"
      }

      rescue_from :all do
        error!("I'm a teapot!", 418)
      end

      get { raise StandardError }
    end
  end

  before do
    get '/'
  end

  it { is_expected.to eq("I'm a teapot!:418") }
end
```

dblock: Since we document custom middleware this would be a breaking change :(

dblock (changes requested):
Let's discuss @ericproulx's comment on this interface in https://github.com/ruby-grape/grape/issues/2527#issuecomment-2622768862. If we are going to introduce a breaking change, I think we want something future-proof. 

The other option is to add `status` to the local scope, and not change this interface, which seems desirable.

dblock:
Thanks for hanging in here with us @drewnichols! I still wish we had a local scope for `request` and/or `respose` here.

drewnichols:
No worries, as this would be a breaking change it makes sense to do it in a major version update and put more thought into it to allow for futures extensibility.

PR #2529, opened 2025-01-31
Change Grape::API's `@setup` var to an Array (from a Set)

Haerezis:
A Set lead Grape to ignore identical DSL calls when mounting an API, as they were not recorded when the API was first loaded (because identical item are ignored when added to a Set).

https://github.com/ruby-grape/grape/issues/2173


dblock:
I like it, @ericproulx any comments/objections?

spec/grape/api/setup_set_vs_array_spec.rb
stephannv, line 1: I'm not sure if creating a test with Set here makes sense. It could test only the expected behavior, and not a code that will be removed from the codebase.

Haerezis: I kinda agree but that was what was asked so..

dblock: @Haerezis I think we misunderstood each other :) 

@stephannv is right, the only test we need is one that would fail without the change in lib/grape/api.rb. So you can remove `SetupSetAPI` and `SetupArrayAPI`, reduce the test to a minimum that fails without the change to `= []`.

dblock (changes requested):
@stephannv is right here, my bad. Thanks for you help!

spec/grape/api/setup_set_vs_array_spec.rb
Haerezis: Ha, sorry, we must have effectively misunderstood each other.

So in short it's not even a test changing anything about `@setup`, it's just a test checking that mounted routes have the correct number of settings and that none are missing. And this test would fail with the current Impl using a Set, but because I would also change the current impl to use an Array, it would NOT fail.

Sorry I was confused with the wording of `a test that fails without the change to = []`. I understood it as "show us a comparison test because we don't accept the change without the comparison".

At least the current spec have shown beyond the shadow of a doubt that using a Set introduce a bug.

I'll come back quickly with what you asked.

dblock: Exactly, English is my third language ;) Thank you!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * [#2514](https://github.com/ruby-grape/grape/pull/2514): Add rails 8.0 to CI - [@ericproulx](https://github.com/ericproulx).
 * [#2516](https://github.com/ruby-grape/grape/pull/2516): Dynamic registration for parsers, formatters, versioners - [@ericproulx](https://github.com/ericproulx).
 * [#2518](https://github.com/ruby-grape/grape/pull/2518): Add ruby 3.4 to CI - [@ericproulx](https://github.com/ericproulx).
+* [#2529](https://github.com/ruby-grape/grape/pull/2529): Change Grape::API's `@setup` var to an array (from a set) - [@Haerezis](https://github.com/Haerezis).
```
dblock, line 14: This is actually a bug fix, should go below under fixes with something like "Fix `route_setting` used with multiple `mount`".

dblock (changes requested):
Looks good, let's put the CHANGELOG in the right place?

spec/grape/api_remount_spec.rb
```diff
@@ -505,5 +505,65 @@ def printed_response
...
+          }
+        )
+      end
+
+      context 'when mounting it' do
+        before do
+          root_api.mount a_remounted_api
+        end
+
+        it 'still has all the settings for both routes, even after mount' do
```
dblock, line 549: redundant

```suggestion
        it 'still has all the settings for both routes' do
```

spec/grape/api_remount_spec.rb
```diff
@@ -505,5 +505,65 @@ def printed_response
         end
       end
     end
+
+    context 'with settings on defined routes' do
```
dblock, line 509:
```suggestion
    context 'with route settings' do
```

CHANGELOG.md
Haerezis: I changed the changelog and put the change in fixes section with a title I thought fit better.

dblock (approved):

dblock:
Merci @Haerezis!

Haerezis:
my pleasure. Sorry for the misunderstanding :) Thanks for the quick replies !

PR #2530, opened 2025-02-01
Force endpoint status in error_response and error!

ericproulx:
Following this [comment](https://github.com/ruby-grape/grape/issues/2527#issuecomment-2625351197), this PR fixes the issue with the endpoint's status when not invoking `error!` through a block or a helper.


dblock (approved):
👏

PR #2535, opened 2025-02-16
Delegates calls to inner object is some classes by extends Forwardable

ericproulx:
This PR will delegate some functions to some inner objects. Here are the details:

#### api.rb
Add delegation to `base_instance` and `instance_for_rack`
Use `delegate_missing_to` instead of method_missing functions

#### endpoint.rb
Delegate `params` and `headers` to `request`
Remove `LazyObject` for `build_headers` since its not forced 

#### request.rb
Use `each_header` from Rack::Request instead of `env.each_pair`
Remove `.to_s` since Ruby 2.7 added `start_with?` to symbols

#### stack.rb
Delegates some functions


dblock:
LGTM, YOLO

PR #2536, opened 2025-02-16
Update normalize_path like Rails

ericproulx:
This PR apply the same strategy as [Rails](https://github.com/rails/rails/pull/54517/commits/8d1b9fd45a8ea2a3422ca15bc6dcbcb11631ab91) when normalizing path.


lib/grape/router.rb
```diff
@@ -5,11 +5,22 @@ class Router
     attr_reader :map, :compiled
 
     def self.normalize_path(path)
+      return +'/' unless path
```
dblock, line 8: What is `+'/'` syntax?

dblock (changes requested):
In what situations does this change behavior? At the very least I feel like we need to bring in some tests from Rails for `normalize_path` if we're going to be doing it here as well.

lib/grape/router.rb
ericproulx: It's a shortcut for `dup`

dblock: TIL

dblock (approved):

PR #2537, opened 2025-02-16
Use ActiveSupport `try` pattern

ericproulx:
This PR introduces `try` in the codebase. We had several cases of if `obj.respond_to?(<name>)` && obj.<name>` and I think using `try` is more readable.


dblock (approved):
I like it a lot!

PR #2538, opened 2025-02-17
Handle json array

### LINKED ISSUE #2526: Parameter check error
I define the request parameter rule to be
```ruby
 params do
  requires :workflow_nodes, type: Array do
    requires :id, type: String
    requires :title, type: String
    optional :is_start, type: Boolean
    requires :steps, type: Array do
      requires :id, type: String
      optional :type, type: String, values: %w[send_messsge assign_user assign_team tag_conversation snooze close add_commit]
      given type: ->(val) { val == 'send_messsge' } do
        requires :message, type: Hash do
          requires :content, type: String
        end
      end
      given type: ->(val) { val == 'assign_user' } do
        requires :assign_user_id, type: String
      end
      given type: ->(val) { val == 'assign_team' } do
        requires :assign_team_id, type: String
      end
      given type: ->(val) { val == 'tag_conversation' } do
        requires :tags, type: Array[String]
      end
      given type: ->(val) { val == 'add_commit' } do
        requires :comment_content, type: String
      end
     end
  end
end

```
The content of my request is
```json
{
 "workflow_nodes": [
  {
    "id": "eqibmvEzPo8hQOSt",
    "title": "Node 1",
    "is_start": true,
    "steps": [
      {
        "id": "DvdSZaIm1hEd5XO5",
        "type": "send_messsge",
        "message": {
          "content": "打击好",
          "menus": []
        }
      },
      {
        "id": "VY6MIwycBw0b51Ib",
        "type": "add_commit",
        "comment_content": "初来乍到"
      }
    ]
  }
 ]
}

```
However, the request returns an inspection error message  `workflow_nodes[0][steps][0][message][content] is missing`

mohammednasser-32:
Fix #2526 

There is an issue in handling json arrays,
The problem is [here](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/params_scope.rb#L97) we mark a Scope with type Array as meeting dependency if at least 1 element meets the dependency,
And therefore in the next iteration (json items), `@dependant_on` is nil, `@parent` meets dependency, and so we validate all json items, even the ones that should be skipped since they are not meeting the dependency

So I added `@params_meeting_dependency` param (name might be changed) which filters only the json items that should be validated so we can skip the rest


dblock:
@ericproulx take a look?

dblock:
@mohammednasser-32 Thank you!

Run `rubocop -a ; rubocop --auto-gen-config` to make RuboCop happy in the meantime.

mohammednasser-32:
@dblock Fixed :pray:

ericproulx:
> @ericproulx take a look?

I'll take a look

ericproulx:
@mohammednasser-32 there's still an issue with the returned message
```ruby
it 'fails' do
  params = {
    array: [
      {
        a: 'a',
        json: { b: 'b' }
      },
      {
        a: 'a'
      },
      {
        a: 'a',
        json: { b: 'b' }
      }
    ]
  }
  post '/nested_array', params.to_json, 'CONTENT_TYPE' => 'application/json'
  expect(last_response.body).to eq('array[1][json] is missing, array[1][json][b] is missing')
end
```

^ this test will fail.

mohammednasser-32:
@ericproulx ah I see, the `@scope.index` is not being reset for nested json arrays, honestly with the way nested json arrays are currently handled I am not sure how to do it...maybe it requires more refactoring :/

ericproulx:
> @ericproulx ah I see, the `@scope.index` is not being reset for nested json arrays, honestly with the way nested json arrays are currently handled I am not sure how to do it...maybe it requires more refactoring :/

Yeah I know, It's a tough one.

ericproulx:
@mohammednasser-32 I'm not sure your solution works. I've simplified the issue regarding params with a `given` and without.

Both endpoints `array_without_given` and `array_with_given` are expecting an array of hashes `{ a: Integer, b: Integer }`. The `array_with_given` is just forcing the `requires :b, type: Integer` only if `a` is `odd?`

Sending the following payload `array: [ { a: 1, b: 2 }, { a: 3 }, { a: 5 } ]` where all `a` are `odd?` to both endpoints should return a body containing `array[1][b] is missing, array[2][b] is missing`. Unfortunately, only the `array_without_given` works.

Here's are the specs

```ruby
context 'array without given' do
  before do
    subject.params do
      requires :array, type: Array do
        requires :a, type: Integer
        requires :b, type: Integer
      end
    end

    subject.post '/array_without_given'
  end

  it 'fails' do
    params = {
      array: [
        {
          a: 1,
          b: 2
        },
        {
          a: 3
        },
        {
          a: 5
        }
      ]
    }
    post '/array_without_given', params.to_json, 'CONTENT_TYPE' => 'application/json'
    expect(last_response.body).to eq('array[1][b] is missing, array[2][b] is missing')
  end
end

context 'array with given' do
  before do
    subject.params do
      requires :array, type: Array do
        requires :a, type: Integer
        given a: ->(v) { v.odd? } do
          requires :b, type: Integer
        end
      end
    end

    subject.post '/array_with_given'
  end

  it 'fails' do
    params = {
      array: [
        {
          a: 1,
          b: 2
        },
        {
          a: 3
        },
        {
          a: 5
        }
      ]
    }
    post '/array_with_given', params.to_json, 'CONTENT_TYPE' => 'application/json'
    expect(last_response.body).to eq('array[1][b] is missing, array[2][b] is missing')
  end
end
```

mohammednasser-32:
@ericproulx Indeed the solution returns the correct status code but not the proper indexing, alright feel free to close the PR then..thanks for checking!

ericproulx:
> @ericproulx Indeed the solution returns the correct status code but not the proper indexing, alright feel free to close the PR then..thanks for checking!

I think I found a solution.

ericproulx:
@mohammednasser-32 could you add the tests from this [comment](https://github.com/ruby-grape/grape/pull/2538#issuecomment-2676186892) in your branch and simply switch the `@index || index` in the `full_name` function [here](https://github.com/ruby-grape/grape/blob/6e6958f35e6f502cd50e8590aeb501f4b78ae1fd/lib/grape/validations/params_scope.rb#L130). We should read `"#{@parent.full_name(@element)}#{brackets(index || @index)}#{brackets(name)}"`. It felt really strange to not prioritized the `index` param.

mohammednasser-32:
@ericproulx Yes that was it! working perfectly now..thank you! pushed [here](https://github.com/ruby-grape/grape/pull/2538/commits/3e2f133a592d4e9ff05b35d0e55cbc1320caf564)

lib/grape/validations/params_scope.rb
```diff
@@ -94,7 +95,11 @@ def should_validate?(parameters)
       def meets_dependency?(params, request_params)
         return true unless @dependent_on
         return false if @parent.present? && !@parent.meets_dependency?(@parent.params(request_params), request_params)
-        return params.any? { |param| meets_dependency?(param, request_params) } if params.is_a?(Array)
+
+        if params.is_a?(Array)
+          @params_meeting_dependency = params.filter { |param| meets_dependency?(param, request_params) }
```
dblock, line 100: This is memoized, shouldn't it be

```suggestion
          @params_meeting_dependency ||= params.filter { |param| meets_dependency?(param, request_params) }
```

dblock: I think so. Let’s get rid of the unneeded memorization?

mohammednasser-32: but I think it is already not memorized, you mean this line right?
```ruby
@params_meeting_dependency = params.filter { |param| meets_dependency?(param, request_params) }
```
I am already using `=` here, so which memorization you mean?

dblock: Nevermind, this is fine. I see we need  `@params_meeting_dependency` later. I dislike that we have a `@` in here but it can be called multiple times.

dblock:
@ericproulx good with you? merge?

ericproulx:
@mohammednasser-32 could you add the test in from that [comment](https://github.com/ruby-grape/grape/issues/2526#issuecomment-2613984105). It all started from that and I just want to make sure it passes

mohammednasser-32:
@ericproulx done, it needed one extra small fix to handle json array within json array

ericproulx:
@mohammednasser-32 just missing a changelog entry and we're good to go.

mohammednasser-32:
@ericproulx done :pray:

ericproulx:
@dblock sounds good to me.

PR #2540, opened 2025-02-25
Deprecates Grape's Extensions for ParamsBuilder in favor of build_with

ericproulx:
This PR refactors `params_builder` by using the same technique as `validators`, `parsers`, `error formatters` ... etc which is registration with a symbolized short name. IMO, it makes sense to follow the same pattern like other components that we have and the `registration` part ease the customization.

There are 3 ways to configure the `params_builder`:
- Globally by calling`Grape.config.params_builder`
- At the API level by including params_builder extensions concern
- using `build_with` inside `params` scope

This PR is an hybrid meaning it does not remove the old way, its just deprecating it. Here are the deprecations:
- Passing a class to `build_with` or `Grape.config.param_builder` 
- Including Grape's extensions like `Grape::Extensions::Hashie::Mash::ParamBuilder` 

See [UPGRADING.md](https://github.com/ruby-grape/grape/blob/625d752a2d63345198fb4020b27d793713c6448d/UPGRADING.md)

This PR also adds `build_with` a the API level since including the extensions concerns are deprecated.


README.md
```diff
@@ -719,10 +719,13 @@ For example, for the `param_builder`, the following code could run in an initial
 
 ```ruby
 Grape.configure do |config|
-  config.param_builder = Grape::Extensions::Hashie::Mash::ParamBuilder
+  config.param_builder = :hashie_mash
 end
 ```
 
+Available parameter builders are `:hash`, `:hash_with_indifferent_access`, and `:hashie_mash`.
+See [params_builder](lib/grape/params_builder)
```
dblock, line 727:
```suggestion
See [params_builder](lib/grape/params_builder).
```

README.md
```diff
@@ -803,16 +806,15 @@ The class can also be overridden on individual parameter blocks using `build_wit
...
 end
 ```
 
-Or globally with the [Configuration](#configuration) `Grape.configure.param_builder`.
-
 In the example above, `params["color"]` will return `nil` since `params` is a plain `Hash`.
 
-Available parameter builders are `Grape::Extensions::Hash::ParamBuilder`, `Grape::Extensions::ActiveSupport::HashWithIndifferentAccess::ParamBuilder` and `Grape::Extensions::Hashie::Mash::ParamBuilder`.
+Available parameter builders are `:hash`, `:hash_with_indifferent_access`, and `:hashie_mash`.
+See [params_builder](lib/grape/params_builder)
```
dblock, line 817:
```suggestion
See [params_builder](lib/grape/params_builder).
```

UPGRADING.md
```diff
@@ -1,6 +1,11 @@
 Upgrading Grape
 ===============
 
+#### Params Builder
+
+- Passing a class to `build_with` or `Grape.config.param_builder` has been deprecated in favor of a symbolized short_name. See `SHORTNAME_LOOKUP` in [params_builder](lib/grape/params_builder.rb)
```
dblock, line 6:
```suggestion
- Passing a class to `build_with` or `Grape.config.param_builder` has been deprecated in favor of a symbolized short_name. See `SHORTNAME_LOOKUP` in [params_builder](lib/grape/params_builder.rb).
```

lib/grape/locale/en.yml
```diff
@@ -57,3 +58,4 @@ en:
           problem: 'invalid version header'
           resolution: '%{message}'
         invalid_response: 'Invalid response'
+        query_parsing: "query params are not parsable."
```
dblock, line 61:
```suggestion
        query_parsing: 'query params are not parsable'
```

dblock (approved):
Does this prevent one from defining their own `ParamBuilder`? I am not sure anybody would have, but it's still a public interface. 🤔

ericproulx:
> Does this prevent one from defining their own `ParamBuilder`? I am not sure anybody would have, but it's still a public interface. 🤔

No, anyone can define their own. I've just reused the same technique as parsers, error_formatters, etc ...

dblock (approved):

PR #2548, opened 2025-03-23
Formatting from header acts like Versioning from header

ericproulx:
This PR changes the how the content_type is negotiated in `Grape::Middleware::Formatter` when looking in the ACCEPT header. Since #2389, `Grape::Middleware::Versioner::Header` is using `Rack::Utils.best_q_match` to find it but not `Grape::Middleware::Formatter`. It seems totally legit that both should act the same way.

This PR also optimize the negotiation regarding the extension found in a path lie `/api/v2/test.json`. Originally, it was using `split` without limit the number of occurrences that could be found and since its always looking at the end of string , it felts natural to just find the first dot starting from the end of the string.

Finally, `Grape::Middleware::Formatter` won't approximate the content-type. This test was expecting an approximation when using a vendor/type form
```ruby
it 'is set to closest generic for custom vendored/versioned without registered type' do
   _, headers, = subject.call(Rack::PATH_INFO => '/info', Grape::Http::Headers::HTTP_ACCEPT => 'application/vnd.test+json')
   expect(headers[Rack::CONTENT_TYPE]).to eq('application/json')
end
```
IMO, it didn't make sense since user should use versioning through `Grape::Middleware::Versioner` in their API. It was made for this.


dblock:
Since there's definite change in behavior (even if that behavior was not desired, invalid, etc.) this needs a clear UPGRADING.md. Go through the tests being removed and try to use those as examples?

ericproulx:
@dblock done

UPGRADING.md
```diff
@@ -6,6 +6,25 @@ Upgrading Grape
 - Passing a class to `build_with` or `Grape.config.param_builder` has been deprecated in favor of a symbolized short_name. See `SHORTNAME_LOOKUP` in [params_builder](lib/grape/params_builder.rb).
 - Including Grape's extensions like `Grape::Extensions::Hashie::Mash::ParamBuilder` has been deprecated in favor of using `build_with` at the route level.
 
+#### Accept Header Negotiation Harmonized
+
+[Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept) header is now fully interpreted through `Rack::Utils.best_q_match` which is following [RFC2616 14.1](https://datatracker.ietf.org/doc/html/rfc2616#section-14.1).
+Since [Grape 2.1.0](https://github.com/ruby-grape/grape/blob/master/CHANGELOG.md#210-20240615), the [header versionning strategy](https://github.com/ruby-grape/grape?tab=readme-ov-file#header) is using embracing it
```
dblock, line 12:
```suggestion
Since [Grape 2.1.0](https://github.com/ruby-grape/grape/blob/master/CHANGELOG.md#210-20240615), the [header versioning strategy](https://github.com/ruby-grape/grape?tab=readme-ov-file#header) was adhering to it, but `Grape::Middleware::Formatter` never did.
```

Remove CR/LFs, this is markdown, no need to wrap.

UPGRADING.md
```diff
@@ -6,6 +6,25 @@ Upgrading Grape
...
+but `Grape::Middleware::Formatter` never did.
+
+Your API might act differently since it will strictly follow the [RFC2616 14.1](https://datatracker.ietf.org/doc/html/rfc2616#section-14.1) when interpreting the `Accept` header.
+Here are the differences:
+
+###### Invalid or missing quality ranking
+The following will be interpreted that `application/json` is the preferred media type:
```
dblock, line 19:
```suggestion
The following used to yield ... and now will yield `application/json` as the preferred media type:
```

Explain what it was before.

UPGRADING.md
```diff
@@ -6,6 +6,25 @@ Upgrading Grape
...
+- `application/json;q=invalid,application/xml;q=0.5`
+- `application/json,application/xml;q=1.0`
+
+###### Closest generic for custom vendored/versioned
+In a case where a custom vendored/versioned like `application/vnd.test+json` is provided and 
+your API is accepting `application/json`, Grape will not consider the last and default to `text/plain`.
```
dblock, line 25:
```suggestion
your API is accepting `application/json`, Grape will no longer consider the last and default to `text/plain`.
```

Ad a before/after example.

PR #2549, opened 2025-03-23
Delegate cookies management to Grape::Request

ericproulx:
This PR delegates `cookies` to Grape::Request like the actual `params` and `headers` are. Minor tweaks have been made in `Grape::Cookies`:

- Rack's cookies will be read only if the API calls cookies
- Response cookies management is now done with a Set instead of Hash containing `true`
- Uses Forwardable module

### DIFF
--- a/lib/grape/cookies.rb
+++ b/lib/grape/cookies.rb
@@ -12,12 +12,13 @@ module Grape
 
     def_delegators :cookies, :[], :each
 
-    def initialize(lazy_rack_cookies)
-      @lazy_cookies = lazy_rack_cookies
+    def initialize(rack_cookies)
+      @cookies = rack_cookies
+      @send_cookies = nil
     end
 
-    def each_response_cookies
-      return unless defined?(@send_cookies)
+    def response_cookies
+      return unless @send_cookies
 
       send_cookies.each do |name|
         yield name, cookies[name]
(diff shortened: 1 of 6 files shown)


lib/grape/cookies.rb
```diff
@@ -2,41 +2,51 @@
...
+    def delete(name, **opts)
+      self.[]=(name, opts.merge(DELETED_COOKIES_ATTRS))
     end
 
-    def each(&block)
-      @cookies.each(&block)
+    private
+
+    def cookies
+      return @cookies if defined?(@cookies)
```
dblock, line 40: The lazy cookies thing is an interesting way to solve this. I am not opposed to it, but before we merge, isn't it easier to set `@cookies = cookies` in `initialize` and `call` here? Avoids having two variables (less state), is easier to follow IMO.

```ruby
@cookies = @cookies.call.with_indifferent_access if @cookies.respond_to?(:call)
```

lib/grape/cookies.rb
```diff
@@ -2,41 +2,51 @@
...
+
+    def initialize(lazy_rack_cookies)
+      @lazy_cookies = lazy_rack_cookies
     end
 
-    def write(header)
-      @cookies.select { |key, _value| @send_cookies[key] == true }.each do |name, value|
-        cookie_value = value.is_a?(Hash) ? value : { value: value }
-        Rack::Utils.set_cookie_header! header, name, cookie_value
+    def each_response_cookies
```
dblock, line 19: I would name this `response_cookies` because that's what an enumerator would be called if we wanted to add one.

dblock (commented):
Took me a moment to grok through this otherwise simple change, see if you want any of my suggestions?

ericproulx:
> Took me a moment to grok through this otherwise simple change, see if you want any of my suggestions?

Great suggestion! I did some rework

dblock:
> > Took me a moment to grok through this otherwise simple change, see if you want any of my suggestions?
> 
> Great suggestion! I did some rework

Much better!

PR #2550, opened 2025-03-23
Drop active support 6.0

ericproulx:
This PR set minimal active support to 6.1. We do not test 6.0 in the CI and I think its fair to drop it.


dblock:
Does 6.0 not work or is it EOL?

ericproulx:
> Does 6.0 not work or is it EOL?

https://endoflife.date/rails
3 years ago

PR #2553, opened 2025-04-05
Less query params parsing

### LINKED ISSUE #2488: Parsing query params may lead to unhandled Rack exception.
These [exceptions](https://github.com/rack/rack/blob/df241355a7f122dc22437398267c1d5f0b27e1ad/lib/rack/query_parser.rb#L13-L26) are not handled when parsing query params.

We need to rescue them when building [params](https://github.com/ruby-grape/grape/blob/12dc739e6aa8be56e4b5adb39a9ffd9df03eec72/lib/grape/request.rb#L15) and whenever we are calling `Rack::Utils.parse_nested_query`

https://github.com/ruby-grape/grape/blob/12dc739e6aa8be56e4b5adb39a9ffd9df03eec72/lib/grape/middleware/versioner/param.rb#L25
https://github.com/ruby-grape/grape/blob/12dc739e6aa8be56e4b5adb39a9ffd9df03eec72/lib/grape/middleware/formatter.rb#L143

This test will raise an `Rack::QueryParser::ParamsTooDeepError`. Nonetheless, having a `rescue_from :all` would handle it but I don't think its user's responsibility.

```ruby
context 'query params ParamsTooDeepError' do
  subject { last_response }

  let(:params) do
    "foo#{"[a]" * Rack::Utils.param_depth_limit}=bar"
  end

  let(:api) do
    Class.new(described_class) do
      get { 'yes!' }
    end
  end

  let(:app) { api }

  before { get "/?#{params}" }

  it { is_expected.to be_successful }
end
```

Here are other cases from Rack's test suite
https://github.com/rack/rack/blob/df241355a7f122dc22437398267c1d5f0b27e1ad/test/spec_utils.rb#L230-L239

In the end, we should return a 400 according to [Rack](https://github.com/rack/rack/blob/df241355a7f122dc22437398267c1d5f0b27e1ad/lib/rack/query_parser.rb#L68-L72)

ericproulx:
This PRs reduces potential query params parsing. Instead of calling `Rack::Utils.parse_nested_query(env[Rack::QUERY_STRING])` directly, Grape will call Rack's function [GET](https://github.com/rack/rack/blob/4f15e7b814922af79605be4b02c5b7c3044ba206/lib/rack/request.rb#L491) that does the same work but keeps the parsed version in `RACK_REQUEST_QUERY_HASH`. This way, parsing will be done at most once.

Also, it fixes #2488 and it comes with a tiny refactor in `Grape::Middleware::Formatter` including comments in the code. Finally, this PR adds a new function named `rack_request` that's create a instance of `Rack::Request` from `env` in `Grape::Middleware::Base`. `Grape::Middleware::Formatter` had already the equivalent but named `request`. This helps for calling `GET` in middlewares and I think user might benefit from it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [#2546](https://github.com/ruby-grape/grape/pull/2546): Fix middleware with keywords - [@ericproulx](https://github.com/ericproulx).
 * [#2547](https://github.com/ruby-grape/grape/pull/2547): Remove jsonapi related code - [@ericproulx](https://github.com/ericproulx).
 * [#2548](https://github.com/ruby-grape/grape/pull/2548): Formatting from header acts like versioning from header - [@ericproulx](https://github.com/ericproulx).
+* [#2553](https://github.com/ruby-grape/grape/pull/2553): Less query params parsing - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 2.3.0 (2025-02-08)
--- a/lib/grape/exceptions/conflicting_types.rb
+++ b/lib/grape/exceptions/conflicting_types.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Grape
+  module Exceptions
+    class ConflictingTypes < Base
+      def initialize
+        super(message: compose_message(:conflicting_types), status: 400)
+      end
+    end
+  end
+end
--- a/lib/grape/exceptions/invalid_parameters.rb
+++ b/lib/grape/exceptions/invalid_parameters.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Grape
+  module Exceptions
+    class InvalidParameters < Base
+      def initialize
+        super(message: compose_message(:invalid_parameters), status: 400)
+      end
+    end
+  end
+end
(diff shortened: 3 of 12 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [#2546](https://github.com/ruby-grape/grape/pull/2546): Fix middleware with keywords - [@ericproulx](https://github.com/ericproulx).
 * [#2547](https://github.com/ruby-grape/grape/pull/2547): Remove jsonapi related code - [@ericproulx](https://github.com/ericproulx).
 * [#2548](https://github.com/ruby-grape/grape/pull/2548): Formatting from header acts like versioning from header - [@ericproulx](https://github.com/ericproulx).
+* [#2553](https://github.com/ruby-grape/grape/pull/2553): Less query params parsing - [@ericproulx](https://github.com/ericproulx).
```
dblock, line 21: Update with the fix you're making instead, maybe make this one "Improve performance of query params parsing"

dblock (changes requested):
Nicely done. Only nits from me. Consider adding the removal of CONSTANTs into UPGRADING.

UPGRADING.md
```diff
@@ -1,6 +1,11 @@
 Upgrading Grape
 ===============
 
+### Grape::Middleware::Base
```
dblock, line 4: For later, we should have an `Upgrading >= X.Y.Z` here whatever the next version is.

dblock (approved):

dblock:
Coveralls is down for maintenance, let's wait for it to come back.

PR #2554, opened 2025-04-05
Remove Grape::Http::Headers

### LINKED ISSUE #2334: Grape allows invalid headers to be set.

ericproulx:
This PR removes `Grape::Http::Headers`. Since most constants `HTTP_` were just literal string and Grape's uses `#frozen_string_literal: true`, it made sense to just use the literal string wherever needed. Here's a list of notable changes:

- `Grape::Http::Headers::SUPPORTED_METHODS` has been moved to `Grape` module.
- `Grape::Http::Headers::HTTP_HEADERS` has been moved to `Grape::Request` and its now called `KNOWN_HEADERS`. The last has been revisited to reflect Rack 3's [KNOWN_HEADERS](https://github.com/rack/rack/blob/4f15e7b814922af79605be4b02c5b7c3044ba206/lib/rack/headers.rb#L9). Its not an exact copy. Grape's version has more.
- `Grape::Util::Lazy::Object` no longer exists. Solely used by `Grape::Http::Headers::HTTP_HEADERS`
- `Grape::Http::Headers.find_supported_method` no long exists. It wasn't used.
- Removes `to_s` when calling `headers` since its not just a plain `{}` but a `Grape::Util::Headers`. Fix #2334 
- Closes #2511


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * [#2540](https://github.com/ruby-grape/grape/pull/2540): Introduce Params builder with symbolized short name - [@ericproulx](https://github.com/ericproulx).
 * [#2550](https://github.com/ruby-grape/grape/pull/2550): Drop ActiveSupport 6.0 - [@ericproulx](https://github.com/ericproulx).
 * [#2549](https://github.com/ruby-grape/grape/pull/2549): Delegate cookies management to `Grape::Request` - [@ericproulx](https://github.com/ericproulx).
+* [#2554](https://github.com/ruby-grape/grape/pull/2554): Remove Grape::Http::Headers - [@ericproulx](https://github.com/ericproulx).
```
dblock, line 12: I think you have a fix here for https://github.com/ruby-grape/grape/issues/2334, add?

dblock, line 12:
```suggestion
* [#2554](https://github.com/ruby-grape/grape/pull/2554): Remove `Grape::Http::Headers` - [@ericproulx](https://github.com/ericproulx).
```

dblock (changes requested):
I think this needs an UPGRADING on the invalid header that can be set and the removal of some constants.

CHANGELOG.md
ericproulx: Well its a fix .. but not really. We've been using `Grape::Util::Header` for a while and since then, Grape is compliant with Rack's spec about headers.

UPGRADING.md
```diff
@@ -1,6 +1,18 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 2.4.0
```
ericproulx, line 4: @dblock Fix from your [comment](https://github.com/ruby-grape/grape/pull/2553#discussion_r2030178520)

dblock (approved):

dblock:
Rebase? Feel free to merge yourself.

PR #2556, opened 2025-04-10
Remove unused `Request::DEFAULT_PARAMS_BUILDER` constant

eriklovmo:
The constant was introduced in 45abe0dd9f18fcc0e1993b1d5a2988135cb216e1 but is never referenced and hasn't been since its inception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#2550](https://github.com/ruby-grape/grape/pull/2550): Drop ActiveSupport 6.0 - [@ericproulx](https://github.com/ericproulx).
 * [#2549](https://github.com/ruby-grape/grape/pull/2549): Delegate cookies management to `Grape::Request` - [@ericproulx](https://github.com/ericproulx).
 * [#2554](https://github.com/ruby-grape/grape/pull/2554): Remove `Grape::Http::Headers` and `Grape::Util::Lazy::Object` - [@ericproulx](https://github.com/ericproulx).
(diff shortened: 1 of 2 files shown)


ericproulx:
> The constant was introduced in [45abe0d](https://github.com/ruby-grape/grape/commit/45abe0dd9f18fcc0e1993b1d5a2988135cb216e1) but is never referenced and hasn't been since its inception.

So true, I was supposed to use it but I reused the `Grape.config.param_builder`. Thanks for pointing out

ericproulx:
> Do we want a changelog entry for this?

Please, even though its seems insignificant.

dblock (approved):

dblock:
This was shipped in a previous version? I think we want a CHANGELOG and UPGRADING that mentions the removal of the constant since it is public and could have been used.

eriklovmo:
>>  Do we want a changelog entry for this?
>
> Please, even though its seems insignificant.

@ericproulx 

> This was shipped in a previous version? I think we want a CHANGELOG and UPGRADING that mentions the removal of the constant since it is public and could have been used.

@dblock 

The constant we're removing here was introduced after version 2.3.0, so it hasn’t been officially released yet. For that reason, I’m thinking it might be confusing to mention it in the `CHANGELOG` and `UPGRADING` docs—what do you think?

dblock:
> The constant we're removing here was introduced after version 2.3.0, so it hasn’t been officially released yet. For that reason, I’m thinking we shouldn't mention it in the `UPGRADING` doc. @dblock

Yep, but also NBD.

PR #2557, opened 2025-04-12
New feature : linting

ericproulx:
This PR adds a linting feature that can activated through `Grape.config.lint = true` or `lint!` in a `Grape::Api`.
Suggested in this [comment](https://github.com/ruby-grape/grape/issues/2334#issuecomment-2795024096).

By design, Grape's CI has `Grape.config.lint = true`, so all specs calling an endpoint are linted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@
     - [Header](#header)
     - [Accept-Version Header](#accept-version-header)
     - [Param](#param)
+- [Linting](#linting)
+  - [Bug in Rack::ETag under Rack 3.X](#bug-in-racketag-under-rack-3x)
 - [Describing Methods](#describing-methods)
 - [Configuration](#configuration)
 - [Parameters](#parameters)
@@ -649,6 +651,25 @@ version 'v1', using: :param, parameter: 'v'
 
     curl http://localhost:9292/statuses/public_timeline?v=v1
 
+
+## Linting
+
+You can check if your API is in conformance with [Rack's specification](https://github.com/rack/rack/blob/main/SPEC.rdoc) by calling `lint!` at the API level or through [configuration](#configuration):
+```ruby
+# API Level
+class Api < Grape::API
+  lint!
+end
+# OR through configuration
+Grape.configure do |config|
+  config.lint = true
+end
+# OR
+Grape.config.lint = true
+```
+
+### Bug in Rack::ETag under Rack 3.X
+If you're using Rack 3.X and the `Rack::Etag` middleware (used by [Rails](https://guides.rubyonrails.org/rails_on_rack.html#inspecting-middleware-stack)), a [bug](https://github.com/rack/rack/pull/2324) related to linting has been fixed in [3.1.13](https://github.com/rack/rack/blob/v3.1.13/CHANGELOG.md#3113---2025-04-13) and [3.0.15](https://github.com/rack/rack/blob/v3.1.13/CHANGELOG.md#3015---2025-04-13) respectively.
 
 ## Describing Methods
 
--- a/README.md
+++ b/README.md
@@ -654,17 +654,19 @@ version 'v1', using: :param, parameter: 'v'
 
 ## Linting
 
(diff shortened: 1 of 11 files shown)


dblock:
Should we call it `lint!` since it's `Rack::Lint`?

But thinking about it, maybe we need something more generic that allows one to inject middleware here, including potentially the ability to remove `Rack::Head`?

ericproulx:
> Should we call it `lint!` since it's `Rack::Lint`?
> 
> But thinking about it, maybe we need something more generic that allows one to inject middleware here, including potentially the ability to remove `Rack::Head`?

The middleware stack lacks a `delete` function. I think we should do something like [ActionDispatch::MiddlewareStack](https://api.rubyonrails.org/classes/ActionDispatch/MiddlewareStack.html)

ericproulx:
> Please document this in the README if you want it merged.
> 
> That said, also consider not merging this and doing a configurable middleware implementation instead.

I like the `lint!` DSL since we can configure our entire CI to use it by default. The configurable middleware seems a good idea but I don't know how we would enforce this in the CI without adding it in every call in every specs. It's already possible to do add `Rack::Lint` so to by calling:
```ruby
insert_before Grape::Middleware::Error, Rack::Lint
```

I think the `lint!` really ease the whole thing.

README.md
```diff
@@ -650,6 +652,25 @@ version 'v1', using: :param, parameter: 'v'
     curl http://localhost:9292/statuses/public_timeline?v=v1
 
 
+## Linting
+
+You can check if your API is in conformance with [Rack's specification](https://github.com/rack/rack/blob/main/SPEC.rdoc) by calling `lint!` at the API level or through [configuration](#configuration):
```
dblock, line 657:
```suggestion
You can check whether your API is in conformance with the [Rack specification](https://github.com/rack/rack/blob/main/SPEC.rdoc) by calling `lint!` at the API level or through [configuration](#configuration).
```

Let's just break up these to make it look like the rest of the doc? People tend to copy paste things and will copy paste those together if we leave them together. And we can remove the comments because it's obvious that the last two are "through configuration".

```
class Api < Grape::API
  lint!
end
```

```
Grape.configure do |config|
  config.lint = true
end
```

```
Grape.config.lint = true
```

spec/spec_helper.rb
```diff
@@ -13,6 +13,7 @@
   end
 end
 
+Grape.config.lint = true # lint all apis by default
```
dblock, line 16: Hmmmm.... I am a bit concerned that we switch a default value in Grape in tests, what are the advantages?

dblock (changes requested):
I don't think we should be switching a default value in Grape globally in tests, that negates the default. Any strong feelings?

spec/spec_helper.rb
ericproulx: I think our CI should always check if Grape is in conformance with Rack's specification. Based on this PR and several CI runs, we can say with a high level of confidence that Grape's is in conformance with Rack 2 and 3 specifications. At least, we do our best by linting our CI constantly and I'm confident that future changes will not break the latter.

Finally, providing the `lint!`feature to everyone without Grape being conformant is not suitable as it should only catch wrongful code from the users.

ericproulx: Fun part, while doing it, I found an issue in [rack](https://github.com/rack/rack/issues/2323)

dblock: Alright, I reluctantly accept this argument.

spec/grape/api_spec.rb
```diff
@@ -4738,4 +4738,26 @@ def uniqe_id_route
...
+        get '/' do
+          status 42
+        end
+      end
+    end
+
+    around do |example|
+      Grape.config.lint = false
+      example.run
+      Grape.config.lint = true
```
dblock, line 4755: I think this needs an `ensure` because `example.run` can raise. NBD.

dblock (approved):

spec/grape/api_spec.rb
ericproulx: Makes total sense. Thank you for pointing that out

PR #2561, opened 2025-05-01
Revisit Grape's middlewares default options

ericproulx:
See copilot's [comment](https://github.com/ruby-grape/grape/pull/2561#pullrequestreview-2809654658) but it should:

- reduce `Hash` allocations since `default_options` have been replaced by frozen constants
- reduce a little bit `default_options` size since `nil` values have been removed in favor of keeping actual default.
- add clarity about the default options per middleware


lib/grape/middleware/error.rb
```diff
@@ -3,30 +3,22 @@
...
+        rescue_options: {
+          backtrace: false,
+          original_exception: false
+        }.freeze
+      }.freeze
+
+      def initialize(app, **options)
         super
-        self.class.include(@options[:helpers]) if @options[:helpers]
+        self.class.include(options.delete(:helpers)) if options[:helpers]
```
dblock, line 21: This modifies caller's `Hash` or does it not?

dblock (changes requested):
The caller's hash should be left alone even if internal.

lib/grape/middleware/error.rb
ericproulx: Its does. I will fix it

ericproulx:
@dblock is changing `Grape::Middleware::Base` signature like I did considered has a breaking change ?

dblock:
> @dblock is changing `Grape::Middleware::Base` signature like I did considered has a breaking change ?

You know, I think it may be, since we do mention `Base` in at least one example in README.md. But I don't think we mention the initializer. WDYT?

dblock (approved):

PR #2568, opened 2025-06-06
Off the back of #2464, we might also want to remove the string exampl…

pieterocp:
Off the back of #2464, we might also want to remove the string example from the readme too. 

Since these dont work at all, just error and mislead.

Let me know if I'm just a big noob and have misconfigured my project or something.


pieterocp:
> It should work with a String since the type responds to `length`. Do you have an example ?

Does work well, messed up and was using an older version in my project, thought that a more complex scenario would also error but was user error on my side.

Will close this, but a String example inside the spec might help peanut brains like myself out?

e.g.

```ruby
# frozen_string_literal: true

describe Grape::Validations::Validators::LengthValidator do
  let_it_be(:app) do
    Class.new(Grape::API) do
      # ...

      params do
        requires :list, type: String, length: { min: 1, max: 3 }
      end
      post 'string_example' do
      end
      
      # ...
      
  describe '/string_example' do
    context 'when length is between limits' do
      it do
        post '/string_example', list: 'ab'
        expect(last_response.body).to eq('')
      end
    end
    context 'when length is greater than limit' do
      it do
        post '/string_example', list: 'too long'
        expect(last_response.body).to eq('list is expected to have length within 1 and 3')
      end
    end
  end
  
  # ...
end
```

dblock:
@pieterocp Do PR it!

PR #2574, opened 2025-06-22
Less active support concern

ericproulx:


spec/integration/dry_validation/dry_validation_spec.rb
```diff
@@ -6,47 +6,21 @@
...
+            namespace_stackable_hash[key] << value
+          else
+            namespace_stackable_hash[key]
+          end
+        end
 
-      it 'evaluates block' do
-        expect { subject.params { raise 'foo' } }.to raise_error RuntimeError, 'foo'
+        def self.namespace_stackable_hash
+          @namespace_stackable_hash ||= Hash.new do |hash, key|
```
dblock, line 20: This causes a side effect, doesn't it? Clear it?

ericproulx: I've done the same pattern elsewhere. Since some module are dependent, I need to recreate the functions in the anonymous class.
This won't leak

ericproulx: Since some modules are dependent, i'll check if I can combine them.

PR #2577, opened 2025-07-02
Simplify endpoint block execution and deprecate `return` call in it

ericproulx:
### Summary

This PR deprecates the use of the `return` value from `Grape::Endpoint#call`.

### Motivation

- I don't know why we want to use `return` in an endpoint. It feels more like an habit than a need. 

### Changes

- Marked the return value of `Grape::Endpoint#call` as deprecated in the documentation and code comments.
- Added deprecation warnings where appropriate.
- Updated relevant specs and documentation.

### Impact

- No breaking changes for typical API usage.
- Users who were relying on the return value of `Grape::Endpoint#call` should update their code to avoid this pattern.

### Checklist

- [x] Deprecation warning added
- [x] Documentation updated
- [x] Specs updated


dblock:
Feel free to merge with the UPGRADING update @ericproulx.

jarthod:
Hello, I just noticed this deprecation warning while testing the edge version, and I'm wondering what is the recommended alternative for people who did use `return` ^^ Here is my endpoint which is using to return a different format (and bypasse the rest of the method) in some specific case: 
```ruby
    delete ':id' do
      if recipient = Notifier.recipients(current_user).find {|r| r.key == params[:id] }
        if recipient.type == :sms
          current_user.phone_numbers -= [recipient.identifier]
        elsif recipient.type == :webhook
          webhook = current_user.webhooks.where(id: recipient.identifier).first
          return {'deleted' => webhook.destroy}
        elsif recipient.type == :slack_compatible
          current_user.slack_compatible -= [recipient.identifier]
        elsif recipient.type == :msteams
          current_user.msteams -= [recipient.identifier]
        end
        save_model current_user
        {'deleted' => current_user.previous_changes.any?}
      else
        error! "Recipient not found: #{params[:id]}", 404, {"Cache-Control" => "public, max-age=60", "Vary" => "X-Api-Key"}
      end
    end
```

Here the `return` was quite helpful to me to avoid writing another level of conditional, I'm wondering if there's another good option to return early (similar to error! but for a success) ? Thanks !

dblock:
@ericproulx wdyt? Using `return` inside an endpoint is probably more common than we thought? Would there be a way to reimplement support for it "cleanly"?

ericproulx:
You could extract the `case`  in a method inside your current_user model. Easily testable without Grape. Here's how to woul d do it : 
```ruby
# in your `current_user` model

class RecipientNotFound < StandardError; end

def delete_recipient(recipient_id)
  recipient = Notifier.recipients(self).find { |r| r.key == recipient_id }
  raise RecipientNotFound, "Recipient not found: #{recipient_id}" unless recipient

  deleted = false
  case recipient.type
  when :sms
    self.phone_numbers -= [recipient.identifier]
  when :webhook
    webhook = webhooks.where(id: recipient.identifier).first
    deleted = webhook.destroy
  when :slack_compatible
    self.slack_compatible -= [recipient.identifier]
  when :msteams
    self.msteams -= [recipient.identifier]
  end
 
  save # assuming save_model is doing that
  deleted || previous_changes.any?
end

# Grape API

rescue_from CurrentUserModel::RecipientNotFound do |e| # 
  error! e.message, 404, { "Cache-Control" => "public, max-age=60", "Vary" => "X-Api-Key" }
end

delete ':id' do
  { 'deleted' => current_user.delete_recipient(params[:id]) }
end
```
@jarthod wdyt ?

a5-stable:
@ericproulx

Thank you for your continued work on Grape.

I’m also seeing the deprecation warnings related to using return in endpoints.
There are a few cases in our codebase where we intentionally rely on `return`.

I understand that the logic can be refactored, or `next` can technically
be used as a simple alternative by https://github.com/ruby-grape/grape/pull/2621.

It’s true that `next` does not raise a LocalJumpError, but this behavior depends
on Grape's internal implementation details, specifically the shift from using
UnboundMethod to a block-based execution model. That makes a bit harder to reason about at a glance.
I would appreciate your thoughts on this. Thank you in advance!

ericproulx:
Hi @a5-stable, thanks for reaching out about this subject. Quite frankly, having `return` is `blocks` or `lambdas` is not a common idioms in ruby and since almost everything is a block, I would rather stick with them. Not to say that some magic was done behind the scenes so that people could use `return` and I wasn't a fan. I like simple things. That being said, would you mind sharing your cases in a issue so that we could have a proper discussion about it? Maybe Grape's is missing a feature that is forcing you or your team to use `return`. I would gladly discuss about it. Thanks.

jarthod:
Thanks @ericproulx !
I made peace with using `next` in my small case but I appreaciate the consideration and I also believe like @a5-stable that it makes more sense with `return`.

As a quick reminder, if you still want to remove the unbound method thing, I believe this could also be cleanly implemented using throw / catch, like sinatra does for early response return in it's block syntax DSL:
https://avdi.codes/throw-catch-raise-rescue-im-so-confused/
https://patshaughnessy.net/2012/3/7/learning-from-the-masters-sinatra-internals

In which case it would need to be an explicit method of yours instead of return (like `response`, `render` or something like that). But that would be a more acceptable workaround I believe as it's more readable than "next" and would support the same use-cases as return (basically a drop-in replacement). @a5-stable if you have any opinion on this let us know.

PR #2586, opened 2025-07-27
Just keep `helpers` method in the public scope.

ericproulx:
### Description

This PR changes the visibility of the helper methods in `Grape::DSL::Helpers` from `protected` to `private`. This ensures that these internal methods are not accessible as part of the public or protected API, improving encapsulation and reducing the risk of accidental usage outside their intended scope.

#### Change Summary
- Changed `protected` to `private` for internal methods in `lib/grape/dsl/helpers.rb`.

#### Rationale
Marking these methods as `private` clarifies their intended usage and helps prevent misuse by consumers of the DSL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


dblock (approved):
Do we need to say it in UPGRADING? Probably not necessary.

Issue #2024, opened 2020-03-18
Returning an empty list on DELETE


kubaracek:
When presenting an empty list [] DELETE requests results with `204 No Content`

I'm trying to present an empty list to the DELETE request but grape always results with 204 and empty body. So far I've tried `present [], with: SomeEntity` and `body []`. Both return the same response.
It's a JSON api

LeFnord:
you are trying to respond something, so you have to set the status to 200

dblock:
See https://github.com/ruby-grape/grape#http-status-code.

I'll close this. Please use [the Grape mailing list](https://groups.google.com/forum/#!forum/ruby-grape) for questions.

PR #2587, opened 2025-07-30
the delete request with an empty hash/array should respond with a 200

Elvinra:
On the version 1.6 of grape the behavior was to send a 204 no content with the body when an empty array was supplied

So after the migration to the latest version, the delete request now send a 204 no content, but the body is empty.
The only way to have the body is to force an "status 200".

So the correct status code should be 200 when an empty Hash/Array is supplied to the body.

"present?" return false on a empty array/hash :/


ericproulx:
Hi, thank you for opening a PR. By design, `[]` is considered has no content so that's why its returning 204. You can set `      status 200` if you want.

See this [issue](https://github.com/ruby-grape/grape/issues/2024)
Thanks

PR #2611, opened 2025-10-07
chore: reduce objection allocations from requires and optional DSL

nvh0412:
Current approach: Always clones attrs and opts (in optional), then mutates and clones again
Optimized approach: remove redundant dup by writing our own extract_options! without mutating the input

### DIFF
--- a/lib/grape/dsl/parameters.rb
+++ b/lib/grape/dsl/parameters.rb
@@ -124,17 +124,17 @@ module Grape
       #       end
       #     end
       def requires(*attrs, &block)
-        orig_attrs = attrs.clone
-
-        opts = attrs.extract_options!.clone
+        # Extract options without mutating the original attrs array
+        attrs_clean, opts = extract_options_non_mutating(attrs)
+        opts = opts.clone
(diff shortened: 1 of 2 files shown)


ericproulx:
> Current approach: Always clones attrs and opts (in optional), then mutates and clones again Optimized approach: remove redundant dup by writing our own extract_options! without mutating the input

Love the idea and I'm wondering if it would be possible to just add and extra parameter to the `requires` and `optional` methods.

Something like:
```ruby
def requires(*attrs, **opts, &block)
 ...
end
```

If you open a console and add
```ruby
def test(*attrs, **opts, &block)
  puts "attrs = #{attrs}"
  puts "opts = #{opts}"
end
```
and type
```ruby
test :a, :b, c: :d
```

`opts` should be equal to `c: :d`

nvh0412:
I’ve thought about it, but the thing is we previously passed the original attributes (including the opts hash) into `new_scope` when there was a block. I haven't dug into `new_scope` yet, but by separating the hash and array parameters, we’d need to merge them back before passing them to new_scope to maintain backward compatibility.

Issue #2576, opened 2025-06-30
[Bug] Mount point has effect on mounted API class


bogdan:
If I mount `UsersAPI` under `/v1` namespace, this namespace gets propagated into the mounted class.
This causes a problem if same class is mounted more than once.

Reproduce:

``` ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'grape'
  gem 'rack'
  gem 'minitest'
  gem 'rack-test'
end

require 'minitest/autorun'
require 'rack/test'
require 'grape'

class GrapeAPIBugTest < Minitest::Test
  include Rack::Test::Methods

  class UsersAPI < Grape::API
    format :json

    get "/users" do
      present :users, []
    end
  end

  class RootAPI < Grape::API
    format :json

    namespace "/v1" do
      mount UsersAPI
    end
  end

  def test_nested_route_via_root_api
    env = Rack::MockRequest.env_for("/v1/users")
    status, _, body = RootAPI.call(env)

    assert_equal '{"users":[]}', body.join
    assert_equal 200, status
  end

  def test_direct_users_api
    env = Rack::MockRequest.env_for("/users")
    status, _, body = UsersAPI.call(env)

    assert_equal '{"users":[]}', body.join
    assert_equal 200, status
  end
end
```

ericproulx:
I've been testing earlier versions (too 1.5.3) and they all failed the test. I don't know how's your setup but doing the following works
```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'grape'
  gem 'rack'
  gem 'minitest'
  gem 'rack-test'
end

require 'minitest/autorun'
require 'rack/test'
require 'grape'

class GrapeAPIBugTest < Minitest::Test
  include Rack::Test::Methods

  UsersAPI = Class.new(Grape::API) do
    format :json

    get "/users" do
      present :users, []
    end
  end

  RootAPI = Class.new(Grape::API) do
    format :json

    namespace "/v1" do
      mount UsersAPI
    end

    namespace "/v2" do
      mount UsersAPI
    end

    mount UsersAPI
  end

  def test_v1_users_via_api
    env = Rack::MockRequest.env_for("/v1/users", method: Rack::GET)
    status, _, body = RootAPI.call(env)

    assert_equal '{"users":[]}', body.join
    assert_equal 200, status
  end

[...]

PR #2612, opened 2025-10-21
Avoid multiple mount pollution

alexanderadam:
Soooo, I might have missed something here but the way I see this is, that PR #1893 introduced the behaviour seen in #2576.

And the problem to be tackled was different than what was implemented.
According to the [Rack Spec](https://github.com/rack/rack/blob/main/SPEC.rdoc), a Rack app only needs to be

> a Ruby object that responds to `call`. It takes exactly one argument, the `environment` (representing an HTTP request) and returns a non-frozen `Array` of exactly three elements: the `status`, the `headers`, and the `body` (representing an HTTP response).

Therefore having `Grape::API.call` delegating to [`base_instance.call`](https://github.com/ruby-grape/grape/blob/2cbeeac50648deb11df13f726ec62f483151f75e/lib/grape/api/instance.rb#L73) should be enough.

Howeveeeeeer, [when using `Rack::Test` with def app returning a `Grape::API` class directly (e.g., `API::V1::Ping`)](https://github.com/ruby-grape/grape/pull/1893/files#diff-ba279e71539c2483403d189cb8cb691c35921fc9dff02031d6cb3bf85965a408R42-R55), tests would naturally fail with `404`.

So to me this looks like the fix with d29eb793098a871aec02cd4903af0bf5f76ddd49 was a bit weird and introduced that bug in #2576. :thinking: 

But I could also be totally wrong with _my_ understanding here.
And I'm therefore open to any corrections and learnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#2605](https://github.com/ruby-grape/grape/pull/2605): Add Rack 3.2 support with new gemfile and CI integration - [@ericproulx](https://github.com/ericproulx).
 * [#2607](https://github.com/ruby-grape/grape/pull/2607): Remove namespace_stackable and namespace_inheritable from public API - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
+* [#2612](https://github.com/ruby-grape/grape/pull/2612): Avoid multiple mount pollution - [@alexanderadam](https://github.com/alexanderadam).
 
 #### Fixes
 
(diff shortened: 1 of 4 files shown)


dblock:
@alexanderadam Thanks! Fix CI/CHANGELOG and I'll take a close look - @ericproulx wdyt?

ericproulx:
@alexanderadam could you mount your api multiple times and try with a [configuration](https://github.com/ruby-grape/grape?tab=readme-ov-file#mount-configuration).

PR #2613, opened 2025-10-21
sendfile test fix

alexanderadam:
I have no idea whether this was somehow intentionally done like this or whether that's actually something that intentionally failing or maybe something else.

However, I just thought that I could do this quickly and check what you're thinking instead of a long discussion. :wink: 


dblock:
This definitely passed at some point, we should understand why the behavior changed. Are we getting a newer version of Rack now that has a code change? I'd like to avoid masking a real change accidentally - would you mind bisecting this down to a specific change in rack before we merge the fix?

PR #2614, opened 2025-10-21
update reload documentation

alexanderadam:
refs #358

**PS:** I'm looking for a new adventure in case anybody is looking to hire or work with a Ruby/Rails/Crystal dev

PR #2615, opened 2025-10-21
Remove TOC Danger check

alexanderadam:
As discussed [here](https://github.com/ruby-grape/grape/pull/2614#issuecomment-3429212172) and as announced [here](https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/).


dblock:
This might be more work :( Are we running https://github.com/ruby-grape/danger/blob/master/Dangerfile#L29?

alexanderadam:
Aaaah, I disabled it AND removed the TOC at the same time.
I think the Danger config is taken from the main branch (otherwise somebody could temper it).

Thus I need to do two PRs, I think.

PR #2621, opened 2025-11-01
Update UPGRADING notes regarding `return` usage and simplify endpoint execution

ericproulx:
# Update UPGRADING notes regarding `return` usage and simplify endpoint execution

## Summary

This PR simplifies the endpoint execution code and improves the deprecation messaging for using `return` in endpoint blocks. It removes an unnecessary lambda wrapper and clarifies in documentation that developers should use `next` instead of `return`.

### Code Simplification

- **Simplified `Grape::Endpoint#initialize`**: Removed the unnecessary `@block` lambda wrapper that was created around the source block. The endpoint now directly stores the source block in `@source`.

- **Simplified `Grape::Endpoint#execute`**: Moved the execution logic directly into the `execute` method instead of wrapping it in a lambda. The method now:
  - Directly calls `instance_exec(&@source)` when a source block exists
  - Maintains the same instrumentation with `ActiveSupport::Notifications`
  - Preserves the deprecation warning behavior for `LocalJumpError` (when `return` is used)

### Documentation Updates

- **Updated `UPGRADING.md`**: Added clarification that developers should use `next` instead of `return` when exiting early from endpoint blocks.

### Test Updates

- **Updated deprecation warning test**: Modified the test expectation to match the updated warning message that now includes "Use `next` instead" guidance.

## Benefits

1. **Cleaner code**: Removes an unnecessary abstraction layer (the lambda wrapper)
2. **Better developer experience**: The deprecation warning now provides actionable guidance on what to use instead
3. **Consistency**: The code structure is more straightforward and easier to understand

## Related Issues/PRs

Related to [#2577](https://github.com/ruby-grape/grape/pull/2577) where `return` in endpoint execution was initially deprecated.

## Testing

- Existing tests pass
- The deprecation warning test has been updated to verify the new warning message
- Behavior remains backward compatible (using `return` still works but shows a deprecation warning)

### DIFF
--- a/spec/grape/endpoint_spec.rb
+++ b/spec/grape/endpoint_spec.rb
@@ -772,7 +772,7 @@ describe Grape::Endpoint do
   end
 
   context 'when `return`' do
-    it 'calls deprecatator' do
+    it 'calls deprecator' do
       subject.get('/home') do
         return 'Hello'
       end


dblock (approved):
My usual nitpick, sorry! Fix and merge.

dblock (approved):

PR #2637, opened 2025-12-12
Refactor: Extract declared method into separate DSL module

ericproulx:
# Refactor: Extract `declared` method into separate DSL module

## Summary

This PR extracts the `declared` method and its related logic from `Grape::DSL::InsideRoute::PostBeforeFilter` into a new dedicated `Grape::DSL::Declared` module. This refactoring simplifies the codebase by replacing the **dynamic** module extension pattern (using `extend`) with a **static** include-based approach.

### New Module (`lib/grape/dsl/declared.rb`)

- **Created `Grape::DSL::Declared` module**: Extracts all `declared` method logic, including:
  - The `declared` public method
  - Private helper methods (`declared_array`, `declared_hash`, `declared_hash_attr`, `handle_passed_param`, `optioned_param_key`, `optioned_declared_params`)
  - The `MethodNotYetAvailable` exception class

### Simplified `InsideRoute` (`lib/grape/dsl/inside_route.rb`)

- **Removed `PostBeforeFilter` module**: All logic moved to `Declared` module
- **Removed `post_filter_methods` class method**: No longer needed with the new pattern
- **Replaced dynamic extension with static include**: Now uses `include Declared` instead of dynamically extending modules via `extend post_extension` after filter execution

### Endpoint Changes (`lib/grape/endpoint.rb`)

- **Added `@before_filter_passed` instance variable**: Tracks whether before filters have completed
- **Added `before_filter_passed` attr_reader**: Allows the `Declared` module to check filter state
- **Set flag after before filters**: `@before_filter_passed = true` is set after `run_filters befores, :before`
- **Removed dynamic module extension**: No longer calls `extend post_extension` after running filters

### Test Updates (`spec/grape/dsl/inside_route_spec.rb`)

- **Updated test setup**: Added dummy class to test `Declared` module isolation

## Benefits

1. **Improved code organization**: `declared` method logic is now in its own dedicated module
2. **Simpler pattern**: Replaces **dynamic** module extension (runtime `extend`) with **static** module inclusion (compile-time `include`)
3. **Better maintainability**: Clearer separation of concerns makes the code easier to understand and modify
4. **Reduced complexity**: Eliminates the `post_filter_methods` pattern in favor of a straightforward flag check

## Technical Details

The `declared` method now checks the `before_filter_passed` flag directly instead of relying on dynamic module extension timing. Previously, the code used `extend PostBeforeFilter` dynamically after filter execution to make methods available. This has been replaced with a **static** `include Declared` that is always present, but the methods check the `before_filter_passed` flag to enforce the same behavior (preventing `declared` from being called before parameter validation). This provides the same functionality but with a simpler, more explicit, and more performant mechanism.

## Testing

- [x] All existing tests pass
- [x] No breaking changes to public API
- [x] `declared` method behavior remains unchanged

## Notes

This is a pure refactoring with no functional changes. The `declared` method's behavior and API remain exactly the same, but the internal implementation is cleaner and more maintainable. The key improvement is replacing the dynamic `extend` pattern with a static `include` pattern.


dblock (approved):
...

PR #2681, opened 2026-04-17
Extract Grape::Endpoint.before_each into Grape::Testing

ericproulx:
## Summary

- Moves `before_each`, `reset_before_each`, and `run_before_each` out of `Grape::Endpoint` into a new opt-in `Grape::Testing` module
- Production requests no longer pay any cost from this testing-only feature — `run_before_each` is only injected into `Endpoint#run` via `prepend` when `require 'grape/testing'` is called
- `before_each` now always requires a block (raises `ArgumentError` otherwise); `reset_before_each` replaces the old `before_each(nil)` pattern
- Removes `lib/grape/endpoint.rb` from `Style/OptionalBooleanParameter` rubocop todo

## Test plan

- [ ] `bundle exec rspec spec/grape/testing_spec.rb` — new dedicated spec for `Grape::Testing`
- [ ] `bundle exec rspec spec/grape/endpoint_spec.rb` — verify no regression on endpoint behaviour

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -147,7 +147,6 @@ Style/CombinableLoops:
 Style/OptionalBooleanParameter:
   Exclude:
     - 'lib/grape/dsl/parameters.rb'
-    - 'lib/grape/endpoint.rb'
     - 'lib/grape/serve_stream/sendfile_response.rb'
     - 'lib/grape/validations/types/array_coercer.rb'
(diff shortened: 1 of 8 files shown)


dblock:
If this is not needed in the actual library, it should be something shared in spec/support, and not be released at all?

ericproulx:
> If this is not needed in the actual library, it should be something shared in spec/support, and not be released at all?

It's actually a helper described in the readme so that users can mock easily. It's not just internal stuff.

PR #2693, opened 2026-04-22
Introduce Grape::Exceptions::ErrorResponse value object

ericproulx:
## Summary

`Middleware::Error#error_response` previously handled three input shapes polymorphically — Hash from `throw :error, {...}`, `Grape::Exceptions::Base` subclass, or arbitrary user-handler return.

This PR introduces `Grape::Exceptions::ErrorResponse`, a small value object with `status`, `message`, `headers`, `backtrace`, `original_exception` readers and a `coerce(input)` classifier that normalizes Hash / Exception / ErrorResponse inputs at the boundary. `error_response` is rewritten to call `coerce` once at entry and read typed attributes — the polymorphism puzzle is gone.

- Migrated 7 internal `throw :error, hash` sites to throw `ErrorResponse` instances (`Endpoint#error!`, formatter middleware, error middleware, both versioner middlewares).
- User-defined `throw :error, hash` calls continue to work via `coerce`.
- Returning or throwing a `{ message:, status:, headers: }` Hash from a `rescue_from` handler now emits a deprecation warning via `Grape.deprecator`. Behavior preserved; will be removed in a future release.

## Contract change

The only break is that the payload thrown via `throw :error, ...` is now an `ErrorResponse` value object instead of a `Hash`. Code that does `catch(:error)` and reads `payload[:status]` must switch to `payload.status`. UPGRADING entry added.

## Test plan
- [x] `bundle exec rspec` — green (added specs for `ErrorResponse` and the new deprecation).
- [x] `bundle exec rubocop` on touched files — clean.
- [x] CI green.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -29,6 +29,12 @@ end
 
 Reading `options[...]` is unchanged.
 
+#### Throw `:error` payloads are now `Grape::Exceptions::ErrorResponse`
+
+The payload thrown via `throw :error, ...` is now a `Grape::Exceptions::ErrorResponse` value object instead of a `Hash`. If you `catch(:error)` and inspect the payload, switch from `payload[:status]` to `payload.status` (or `payload[:message]` to `payload.message`, etc.). User-defined `throw :error, hash` calls continue to work — `Middleware::Error#error_response` coerces Hashes, exceptions, and `ErrorResponse` instances at the boundary.
+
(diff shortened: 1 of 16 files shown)


UPGRADING.md
```diff
@@ -3,6 +3,22 @@ Upgrading Grape
...
+`Grape::Exceptions::Base#[]` (which delegated to `__send__` so callers could read `exception[:status]`, `exception[:message]`, etc.) has been removed. Use the named accessors instead:
+
+```ruby
+# Before
+exception[:status]
+exception[:message]
+
+# After
+exception.status
+exception.message
```
dblock: Why not implement `Grape::ErrorResponse[]` to avoid having to rewrite the calling code?

dblock (approved):
-_-

PR #2696, opened 2026-05-01
Reduce per-request allocations on the request hot path

ericproulx:
## Summary

A bundle of small, behavior-preserving cleanups that reduce per-request allocations and method-dispatch chains on the request hot path. None change semantics. The unifying pattern: stable values (filter buckets, middleware options, versioner config, etc.) are computed once at compile/init time and read via `attr_reader`, instead of being re-resolved through nested hash lookups on every request.

## Changes

- **`Endpoint#run`** — precompute filter buckets (`befores`, `before_validations`, `after_validations`, `afters`, `finallies`) and `:build_params_with` once in `compile!`; replace the dynamic `define_method` accessor block with `attr_reader`; remove a dead `header 'Allow', header['Allow']` self-assignment.
- **`Versioner::Base`** — replace the two `define_method`-per-key blocks with `attr_reader` + ivars precomputed in `initialize`. Per-request `pattern`/`prefix`/`vendor`/`strict`/etc. lookups now hit attr loads. `DEFAULT_OPTIONS`, `attr_reader` list, and ivar assignments alphabetized; redundant `version_options` reader dropped.
- **`Middleware::Error` / `Middleware::Formatter`** — same `options[...]` → `attr_reader` migration. All option keys live in `DEFAULT_OPTIONS` (alphabetized), assigned to ivars in `initialize`, read via attr at request time. Drops the dead `rescue_subclasses` middleware key (the per-handler `:rescue_options` hash carries it, not the middleware options).
- **`Middleware::Base`** — freeze `@options` after `merge_default_options` since they're immutable once initialized. One spec adjusted to construct a fresh middleware rather than mutate `subject.options[...]`.
- **`Request#make_params`** — skip the `except(:version, :route_info)` hash allocation in the common case where routing args contain only the known keys.
- **`Validators::Base#validate!`** — lazy-allocate `array_errors`. Saves one Array allocation per validator per request when validation succeeds.
- **`InheritableValues#[]`** — drop the per-call `@inherited_values.merge(@new_values)` and use `Hash#fetch` with a fallback block. **1.55x faster lookup, zero allocation.** Mostly helps boot/compile time (where ~30+ reads per endpoint hit this).
- **`Grape::API::Instance`** — rename `ROOT_PREFIX_VERSIONING_KEY` → `ROOT_PREFIX_VERSIONING_KEYS` (holds 3 symbols); pair with `zip` instead of `each_with_index` + manual indexing.
- **`Validations::Types::CustomTypeCoercer#infer_coercion_method`** — guard-clause refactor of nested if/else.
- **`ErrorFormatter::Base.call`** — defer `merge_backtrace` / `merge_original_exception` lookups so the two `options.dig(:rescue_options, ...)` calls only run when `wrap_message` returned a Hash.

## Throughput impact

`benchmark/version_throughput/` (added in this PR) — single-threaded `Benchmark.ips` against `BenchAPI.call(env)` hitting `/api/v1/hello` with a small JSON response. Ruby 4.0.3, arm64-darwin25:

| Variant | No-YJIT (i/s) | μs/req | YJIT (i/s) | μs/req |
|---|---:|---:|---:|---:|
| Grape 3.2.1 | 42,106 | 23.75 | 79,452 | 12.59 |
| master (pre-PR, `d234bfa5`) | 52,164 | 19.17 | 102,619 | 9.74 |
| **this PR** (`392bbd79`) | **57,677** | **17.34** | **105,588** | **9.47** |

Deltas:

| Comparison | No-YJIT | YJIT |
|---|---:|---:|
| this PR vs master (pre-PR) | **+10.6%** | **+2.9%** |
| this PR vs 3.2.1 | **+37.0%** | **+32.9%** |

YJIT delta vs master is small because the JIT was already specializing most of the hot-path indirection; the no-YJIT delta is where this PR shows clearly. ±5-8% noise floor — single-run numbers, not 3-run averages.

Reproduce locally: `GRAPE_VERSIONS="3.2.1,master" ruby benchmark/version_throughput/run.rb`. See `benchmark/version_throughput/README.md` for methodology.

## Test plan
- [x] `bundle exec rspec` — full suite green, no behavior changes
- [x] `bundle exec rubocop` on touched files — clean
- [x] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [#2692](https://github.com/ruby-grape/grape/pull/2692): Replace per-request `Proc` allocation in `Router#transaction` with a `halt?` helper - [@ericproulx](https://github.com/ericproulx).
 * [#2694](https://github.com/ruby-grape/grape/pull/2694): Split `Versioner::Base#available_media_types` into an `attr_reader` plus `build_available_media_types` - [@ericproulx](https://github.com/ericproulx).
 * [#2695](https://github.com/ruby-grape/grape/pull/2695): Lift trailing `if/else` into guard clauses - [@ericproulx](https://github.com/ericproulx).
+* [#2696](https://github.com/ruby-grape/grape/pull/2696): Reduce per-request allocations on the request hot path; migrate middleware options to `attr_reader` and freeze `@options` post-init - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 #### Fixes
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -3,6 +3,32 @@ Upgrading Grape
 
 ### Upgrading to >= 3.3
 
+#### `Grape::Middleware::Base#options` is now frozen
+
+`@options` is frozen at the end of `Grape::Middleware::Base#initialize` (after `merge_default_options`). The hash is initialized once and treated as immutable for the lifetime of the middleware. Custom middleware that mutates `options[...]` at runtime will now raise `FrozenError`.
+
+If your custom middleware was patching its own options on the fly:
+
+```ruby
+# Before
+class MyMiddleware < Grape::Middleware::Base
+  def before
+    options[:flag] = compute_flag
+    # ...
+  end
+end
+
+# After — store mutable runtime state on a dedicated ivar
+class MyMiddleware < Grape::Middleware::Base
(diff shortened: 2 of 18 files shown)


dblock:
You're collecting quite a few valuable PRs for performance, time to publish an agent skill for optimizing ruby code based on it!

PR #2702, opened 2026-05-08
Add oneof: option for polymorphic Hash parameters (#2385)

### LINKED ISSUE #2385: Param with multiple acceptable Hash Types
How would I define that a given param can have multiple possible hash values?

As a very simplified example, say that _either_ of these hashes are valid params:
```json
"value": {
    "time_unit": "hour",
    "rate": 100.0
}
```

```json
"value": {
    "fixed_price": 100.0,
}
```
but that value can take only one or the other of those forms. It's also possible I could add more acceptable formats in the future, but that there will always be a finite number of acceptable formats to the "value" key.

I attempted something like this but it does not seem to be possible:
```ruby
params do
  requires :value, desc: "TODO", types: [
    Hash {
      requires :fixed_price, type: Float
    },
    Hash {
      requires :time_unit, type: String
      requires :rate, type: Float
    },
  ]
end
```

I can sort of achieve what I want right now by delcaring its type as hash, declaring all of the possible fields as optional, and setting some complicated rules with `mutually_exclusive` and `given`, but this is messy and it is difficult to parse at a glance what the expected possible schemas are.

Is there a better way to achieve this?

ericproulx:
Resolves #2385. Adds an `oneof:` option for `requires`/`optional` so a Hash parameter can accept one of several variant schemas without the `optional + mutually_exclusive + given` workaround the issue describes.

## API

```ruby
params do
  requires :value, type: Hash, oneof: [
    proc { requires :fixed_price, type: Float },
    proc do
      requires :time_unit, type: String
      requires :rate, type: Float
    end
  ]
end
post '/pricing' do
  params[:value]
end
```

Both `{"value":{"fixed_price":100.0}}` and `{"value":{"time_unit":"hour","rate":50.0}}` succeed. A request that doesn't match any variant returns `400` with `value does not match any of the allowed schemas`.

Each variant is a `Proc` that uses the normal Grape `params` DSL — so `requires`, `optional`, `regexp:`, `values:`, `allow_blank:`, nested `Hash`/`Array`, `Grape::API::Boolean`, custom types, etc. all work inside variants without re-implementation. Coercion through the winning variant is applied to the parent params (e.g. `"150.5"` → `150.5`).

Variants are tried in declaration order; the first variant whose validators raise no exceptions wins (deterministic).

## Why a separate `oneof:` and not `types: [hash_schema {...}]`

I considered overloading `types:` per @dblock's note in the issue thread, but kept `oneof:` distinct because:

- `types:` already drives `MultipleTypeCoercer` for primitive variants (`types: [Integer, String]`).
- Mixing primitives and hash schemas in one array — `types: [Integer, hash_schema { ... }]` — produces a non-obvious dispatch.
- `oneof:` aligns with JSON Schema vocabulary and makes the user's intent explicit.

If the maintainers prefer the `types:` overload, the validator and collector machinery in this PR are reusable — only `infer_coercion`/`process_oneof!` would need to change.

## Implementation

| File | What |
|---|---|
| `lib/grape/validations/oneof_collector.rb` | Minimal `@api`-compatible shim. Exposes `OneofCollector.collect(block)` which evaluates the variant block in a fresh `ParamsScope` and returns the captured validators. |
| `lib/grape/validations/validators/oneof_validator.rb` | `OneofValidator < Base`. At request time deep-dups the candidate hash, runs each variant's validators against the dup, takes the first variant that produces no exceptions, and replaces `params[attr_name]` with the (possibly coerced) result. Raises `Validation` with i18n key `oneof` if no variant matches. |
| `lib/grape/validations/params_scope.rb` | `validates` calls `process_oneof!` when `:oneof` is in the validations hash; the helper validates shape (non-empty Array of Procs, `type: Hash`) and replaces each Proc with its `OneofCollector.collect`-built validator list before the standard `validate` loop instantiates `OneofValidator`. |
| `lib/grape/locale/en.yml` | New `oneof: 'does not match any of the allowed schemas'` key. |
| `README.md` | New §"Multiple Hash Schemas with `oneof`" between "Multiple Allowed Types" and "Validation of Nested Parameters". |
| `CHANGELOG.md` | Features entry referencing #2385. |

## Definition-time guards

- `oneof:` requires `type: Hash` (raises `ArgumentError`)
- Variants array must be non-empty (raises `ArgumentError`)
- Every entry must be a `Proc` (raises `ArgumentError`)

## Tests

19 new examples in `spec/grape/validations/validators/oneof_validator_spec.rb`:

- Definition-time errors (4 cases)
- Two-flat-variants matching/coercion/rejection (6 cases)
- `optional :value` with missing/invalid (2 cases)
- Full DSL inside variants — `values:`, `regexp:` (4 cases)
- Deeply nested `Hash` inside variants (3 cases)

## Test plan

- [x] `bundle exec rspec spec/grape/validations/validators/oneof_validator_spec.rb` — 19 examples, 0 failures
- [x] `bundle exec rspec` — full suite green (2,273 examples, up from 2,254)
- [x] `bundle exec rubocop lib/ spec/` — clean

## Credits

Same feature requested as in #2661 by @jcagarcia and @mia-n. This PR re-implements from scratch using the existing `ParamsScope` infrastructure (instead of a parallel mini-DSL) so all Grape validators work inside variants by design.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * [#2706](https://github.com/ruby-grape/grape/pull/2706): Refactor `ParamsScope#validates` and `ParamsDocumentation` around a frozen `Grape::Validations::ValidationsSpec` value object; the validations hash supplied by the DSL is no longer mutated and the helper chain becomes pure - [@ericproulx](https://github.com/ericproulx).
 * [#2707](https://github.com/ruby-grape/grape/pull/2707): Tighten six guard conditions in `lib/` via De Morgan and `blank?`/`present?`/`include?` rewrites; no behaviour change - [@ericproulx](https://github.com/ericproulx).
 * [#2709](https://github.com/ruby-grape/grape/pull/2709): Lift trailing `if/else` into guard clauses; tighten `Util::Lazy::ValueEnumerable` - [@ericproulx](https://github.com/ericproulx).
+* [#2702](https://github.com/ruby-grape/grape/pull/2702): Add `oneof:` option for `requires`/`optional` to accept a Hash parameter matching one of several variant schemas (resolves [#2385](https://github.com/ruby-grape/grape/issues/2385)) - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 #### Fixes
--- a/README.md
+++ b/README.md
@@ -1335,6 +1335,56 @@ end
 client.get('/', status_codes: %w(1 two)) # => [1, "two"]
 ```
 
+### Multiple Hash Schemas with `oneof`
+
+A Hash parameter that may take one of several different shapes can be declared with the `oneof:` option. Each variant is a `Proc` that uses the normal `params` DSL — so the full validator surface (`requires`, `optional`, `regexp:`, `values:`, `allow_blank:`, nested `Hash`/`Array`, etc.) is available inside.
+
+```ruby
+params do
+  requires :value, type: Hash, oneof: [
+    proc { requires :fixed_price, type: Float },
+    proc do
+      requires :time_unit, type: String
+      requires :rate, type: Float
+    end
+  ]
+end
+post '/pricing' do
+  params[:value]
+end
+```
+
+Both of these requests succeed:
+
+```bash
+curl -d '{"value":{"fixed_price":100.0}}' -H 'Content-Type: application/json' /pricing
+curl -d '{"value":{"time_unit":"hour","rate":50.0}}' -H 'Content-Type: application/json' /pricing
+```
+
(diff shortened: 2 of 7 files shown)


dblock:
I like `oneof` with the use of `proc` very much, nice solution! Merge at will after we hear from @jcagarcia and @mia-n?

dblock (approved):

dblock:
Haven't heard from @jcagarcia or @mia-n - should we merge this @ericproulx? I'm good with it.

ericproulx:
> Haven't heard from @jcagarcia or @mia-n - should we merge this @ericproulx? I'm good with it.

I'm good with it.

dblock:
Squash it.

PR #2711, opened 2026-05-13
Update version number in UPGRADING.md

ysksn:
Update example version to 1.3.3


dblock:
Thanks for fixing this one!

PR #2714, opened 2026-05-14
Drop Grape::Middleware::Globals

ericproulx:
## Summary

`Grape::Middleware::Globals` and the three env constants it sets (`Grape::Env::GRAPE_REQUEST`, `Grape::Env::GRAPE_REQUEST_HEADERS`, `Grape::Env::GRAPE_REQUEST_PARAMS`) have been dead code for years. Removing the class, its spec, and the constants.

### Why

- **Never mounted in `lib/`.** No `use Grape::Middleware::Globals` anywhere in production code. `Grape::Endpoint#build_stack` does not include it.
- **No internal consumers of its env keys.** Nothing in `lib/` reads `env['grape.request']`, `env['grape.request.headers']`, or `env['grape.request.params']`. `Grape::Endpoint` builds its `Grape::Request` directly (`endpoint.rb:151`), bypassing the middleware entirely.
- **No documentation references.** README and UPGRADING never mention the class.
- **Origin is incidental.** Introduced 2013 in commit `9987090b` ("Add required option for oauth2 middleware") — bundled into an unrelated oauth2 change.

### Breaking change

Anyone who mounted `Grape::Middleware::Globals` in their own custom Rack stack (to populate `env['grape.request']` for downstream middleware) will lose it. UPGRADING includes a self-contained replacement snippet they can drop into their app.

The original implementation is preserved in git history at [`6b4111b3:lib/grape/middleware/globals.rb`](https://github.com/ruby-grape/grape/blob/6b4111b3/lib/grape/middleware/globals.rb).

## Test plan

- [x] `bundle exec rspec` — 2288 examples, 0 failures (5 fewer than before; those were the deleted `globals_spec.rb`)


UPGRADING.md
```diff
@@ -3,6 +3,28 @@ Upgrading Grape
...
+    request = Grape::Request.new(env)
+    env['grape.request'] = request
+    env['grape.request.headers'] = request.headers
+    env['grape.request.params'] = request.params if env['rack.input']
+    @app.call(env)
+  end
+end
+```
+
+The original implementation is preserved in git history — `git show 6b4111b3:lib/grape/middleware/globals.rb`.
```
dblock: This doesn't work for me ... maybe link to a revision on GitHub?

PR #2723, opened 2026-05-15
Deprecate positional options Hash for `desc`; add grape-swagger integration spec

ericproulx:
#### Summary

`Grape::DSL::Desc#desc` now accepts its options as **keyword arguments**. Passing a positional options Hash still works but emits a deprecation warning via `Grape.deprecator`, so downstream callers keep functioning through the deprecation cycle.

```ruby
# preferred
desc 'Create a user', detail: 'creates a user', tags: %w[users]

# still works, now deprecated
desc 'Create a user', { detail: 'creates a user', tags: %w[users] }
```

#### grape-swagger integration coverage

Adds `spec/integration/grape_swagger/grape_swagger_spec.rb`, a `gemfiles/grape_swagger.gemfile`, and a dedicated CI matrix entry (mirroring the existing `grape_entity` / `dry_validation` integration jobs).

The spec stands up a real API with `add_swagger_documentation` and asserts the generated OpenAPI 2.0 document (paths, descriptions carried from `desc`, `success:` responses, declared params, per-resource `/swagger_doc/:name`).

It also documents why the shim matters: grape-swagger 2.1.4 still calls `desc 'text', options_hash` internally (`lib/grape-swagger/doc_methods.rb`), so without the compatibility path the latest grape-swagger would break outright under the keyword-only signature. An upstream issue has been opened on `ruby-grape/grape-swagger` to migrate to keyword arguments.

#### Tests

- New grape-swagger integration spec: green under `gemfiles/grape_swagger.gemfile`
- `spec/grape/dsl/desc_spec.rb`: covers the keyword form and the deprecated positional form (warns + still works when silenced)
- All `desc`-exercising specs pass under the default gemfile

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,11 +33,14 @@ jobs:
           - gemfiles/rails_7_2.gemfile
           - gemfiles/rails_8_0.gemfile
           - gemfiles/rails_8_1.gemfile
-        specs: ['spec --exclude-pattern=spec/integration/{grape_entity,hashie,dry_validation,multi_*}/*_spec.rb']
+        specs: ['spec --exclude-pattern=spec/integration/{grape_entity,grape_swagger,hashie,dry_validation,multi_*}/*_spec.rb']
         include:
           - ruby: '4.0'
             gemfile: gemfiles/grape_entity.gemfile
             specs: 'spec/integration/grape_entity'
+          - ruby: '4.0'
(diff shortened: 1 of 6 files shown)


ericproulx:
Upstream grape-swagger issue for the internal positional-Hash `desc` usage: ruby-grape/grape-swagger#977

dblock (approved):

PR #2732, opened 2026-05-20
fix: remove redundant code patterns

dashitongzhi:
## Summary
Remove redundant code patterns in Ruby code.

## Changes
- lib/grape/router.rb: Remove redundant .dup call
- lib/grape/dsl/parameters.rb: Remove redundant attrs check

## Motivation
Clean up unnecessary operations for better performance.

### DIFF
--- a/lib/grape/dsl/parameters.rb
+++ b/lib/grape/dsl/parameters.rb
@@ -141,7 +141,7 @@ module Grape
         opts = @group.deep_merge(opts) if @group
 
         # check type for optional parameter group
-        if attrs && block
+        if block
           raise Grape::Exceptions::MissingGroupType if type.nil?
           raise Grape::Exceptions::UnsupportedGroupType unless Grape::Validations::Types.group?(type)
         end
--- a/lib/grape/router.rb
+++ b/lib/grape/router.rb
@@ -135,7 +135,7 @@ module Grape
     end
 
     def default_response
-      [404, DEFAULT_RESPONSE_HEADERS.dup, DEFAULT_RESPONSE_BODY.dup]
+      [404, DEFAULT_RESPONSE_HEADERS.dup, DEFAULT_RESPONSE_BODY]
     end
 
     def match?(input, method)


ericproulx (commented):
Thanks for the cleanup! Both changes are interesting; one of them deserves a deeper look before merging.

## `attrs && block` → `block`

The simplification is technically correct — `*attrs` is always an Array (truthy), so `attrs &&` was always equivalent to `block`. But the fact that it was *written* `attrs && block` is a signal the original author intended a different guard. Compare with the sibling code in `Grape::Validations::ParamsScope#new_scope` (`params_scope.rb:270`):

```ruby
if element && !optional
  raise Grape::Exceptions::MissingGroupType if type.nil?
  raise Grape::Exceptions::UnsupportedGroupType unless …
end
```

`element` there is `attrs.first`. The likely-intended shape in `optional` was `attrs.first && block` (or `attrs.any? && block`) — i.e., skip the type-check only when there are no attrs. As-written, the check fires for the typo-shaped `optional do … end` call (anonymous block, no attrs), and raises `MissingGroupType` — which is the *wrong* diagnosis (the user's mistake isn't missing a type, it's calling `optional` without any attribute names).

Concretely, both `requires` and `optional` are nonsensical without attrs — the DSL is declaring named parameters. An empty `*attrs` list means "declare nothing," which there's no legitimate reason to call. The right fix is to make that precondition explicit at the top of both methods, *then* the `attrs &&` simplification follows for free:

```ruby
def requires(*attrs, **opts, &block)
  raise ArgumentError, 'requires expects at least one attribute name' if attrs.empty?
  …
end

def optional(*attrs, **opts, &block)
  raise ArgumentError, 'optional expects at least one attribute name' if attrs.empty?
  …
  if block   # attrs.any? is now a precondition; redundant `attrs &&` drops cleanly
    raise Grape::Exceptions::MissingGroupType if type.nil?
    raise Grape::Exceptions::UnsupportedGroupType unless Grape::Validations::Types.group?(type)
  end
  …
end
```

That swap also addresses a subtler issue: without the empty-attrs guard, `optional do … end` traces through `new_scope(nil, …)` and ends up creating an anonymous lateral optional scope (`@parent && !@element`) whose declared params get pushed at the parent's level. That's not what the DSL was designed to express, but `new_scope`'s `element && !optional` guard skips the type-check for that path, so the scope construction proceeds. The current `attrs && block` accidentally blocks this via `MissingGroupType`. The clean fix replaces that accidental block with an explicit one.

If you'd like to take this on:

1. Add the explicit `attrs.empty?` guard to both `requires` and `optional`.
2. Then keep your `attrs && block` → `block` simplification — it's unambiguously correct after step 1.
3. Add specs in `spec/grape/validations/params_scope_spec.rb` locking down `requires` / `optional` raising `ArgumentError` when called without any attribute names.

That turns this from a behavior-preserving cleanup into a behavior-locking-down one, which is more valuable. If you'd rather not expand the scope, an alternative is to just add a spec asserting that `optional do … end` raises `MissingGroupType` *intentionally* — but the diagnostic is misleading enough that I'd lean toward step 1.

## `DEFAULT_RESPONSE_BODY.dup`

`DEFAULT_RESPONSE_BODY = ['404 Not Found'].freeze` is consumed read-only by Rack (each is idempotent; close is checked via `respond_to?` and a frozen Array doesn't respond, so it's skipped). The Rack 3 SPEC doesn't require per-response body identity. So your change is spec-compliant.

It does reduce defensive headroom: the `.dup` made the body mutable so middleware that appends to the body Array (logging shims, ETag debuggers) wouldn't hit a `FrozenError`. Grape's own test suite doesn't exercise such middleware, but they exist in the wild. Worth a CHANGELOG note (perf-flavored) and probably also keeping in mind that the `HEADERS.dup` next to it isn't strictly necessary either — both were copy-paste defensiveness. If you want to take the body change as-is I'd suggest a one-liner in the perf section of CHANGELOG documenting the reasoning.

## Misc

- Title is `fix:` but neither change is a bug fix — `chore:` or `perf:` fits better.
- No CHANGELOG entry — every recent PR has one.

Happy to take this over and land it if you'd rather not iterate. Thanks for the contribution!

dblock:
@ericproulx this has been a spam account with some random AI-generated PRs adding Python-based CI, watch out

PR #2733, opened 2026-05-20
Drop dead reverse_merge require; call HWIA constructor directly

ericproulx:
## Summary

- Removes the now-dead `active_support/core_ext/hash/reverse_merge` require (no remaining call sites in `lib/`); the two `reverse_merge` usages become explicit `{ defaults }.merge!(other)` literals.
- Swaps `.with_indifferent_access` call sites for `ActiveSupport::HashWithIndifferentAccess.new(...)` and adds an explicit `require 'active_support/hash_with_indifferent_access'`, so the class dependency is documented at the top of `lib/grape.rb` rather than implicit via the core_ext. The `core_ext/hash/indifferent_access` require is intentionally kept — `HWIA#convert_value` invokes `Hash#nested_under_indifferent_access` internally when wrapping nested hash values.

No behavior change; same semantics in all touched paths.

## Test plan

- [x] `bundle exec rspec spec/grape/params_builder spec/grape/middleware spec/grape/util/registry_spec.rb spec/grape/endpoint_spec.rb spec/grape/request_spec.rb` (488 examples, 0 failures)

### DIFF
--- a/lib/grape/middleware/error.rb
+++ b/lib/grape/middleware/error.rb
@@ -195,7 +195,7 @@ module Grape
 
       def error!(message, status = default_status, headers = {}, backtrace = [], original_exception = nil)
         env[Grape::Env::API_ENDPOINT].status(status) # not error! inside route
-        merged_headers = headers.reverse_merge(Rack::CONTENT_TYPE => content_type)
+        merged_headers = { Rack::CONTENT_TYPE => content_type }.merge!(headers)
         error = Grape::Exceptions::ErrorResponse.new(
           status:, message:, headers: merged_headers, backtrace:, original_exception:
         )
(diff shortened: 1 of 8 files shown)


lib/grape/middleware/error.rb
```diff
@@ -195,7 +195,7 @@ def framework_default(endpoint)
 
       def error!(message, status = default_status, headers = {}, backtrace = [], original_exception = nil)
         env[Grape::Env::API_ENDPOINT].status(status) # not error! inside route
-        merged_headers = headers.reverse_merge(Rack::CONTENT_TYPE => content_type)
+        merged_headers = { Rack::CONTENT_TYPE => content_type }.merge!(headers)
```
dblock, line 198: I was wondering how much faster `merge!` is vs. `merge` here.

```
● Benchmark merge vs merge! on a fresh hash literal (shell)
  │ ruby -r benchmark -e '
  │ require "benchmark/ips" rescue nil
  │ 
  │ headers = { "X-Custom" => "value", "X-Other" => "thing" }
  │ content_type = "application/json"…
  └ 19 lines...

● Yes, merge! is ~19% faster in this pattern.
```

PR #2738, opened 2026-05-23
Skip redundant GRAPE_ROUTING_ARGS rewrite when route has no params to merge

ericproulx:
## Summary

- `Router#process_route` reads `env[GRAPE_ROUTING_ARGS]`, computes `route_params`, then writes the merged Hash back. When `env[GRAPE_ROUTING_ARGS]` was already set and `route_params` was blank, the ternary returned the existing Hash unchanged and we wrote the same value back to env — a no-op assignment on every cascade match through a route with no URL placeholders.
- Use `||=` for the default seed and only re-assign env when there are new params to actually merge.

## Test plan

- [x] `bundle exec rspec` — 2313 examples, 0 failures
- [x] `bundle exec rubocop lib/grape/router.rb` — clean
- [ ] CI green


ericproulx:
Closing — the perf win (one Hash slot rewrite on cascade-matched no-param routes) is too small to justify the readability regression in either form (the `args = env[...] ||= default` chord or the duplicated-branch restructure). Leaving master's three-line ternary as-is.

PR #2747, opened 2026-05-25
Drop Enumerable from Grape::Exceptions::ValidationErrors

ericproulx:
## Summary
- `Grape::Exceptions::ValidationErrors` no longer mixes in `Enumerable` and no longer defines a public `#each`. The Enumerable surface was undocumented (the README only covers `#errors`, `#full_messages`, `#message`, and `rescue_from`), untested, and had exactly one consumer: `#full_messages` calling `map` on `self`.
- `#full_messages` now walks `#errors` directly via `flat_map`, documenting the data shape (`{ params => [validations] }`) at the call site.
- `UPGRADING.md` includes a before/after migration snippet for any `rescue_from` blocks that were iterating the exception instance.

## Test plan
- [x] `bundle exec rspec` — 2313 examples, 0 failures
- [x] No references to `each`/`map`/`to_a` on `ValidationErrors` instances remain in `lib/` or `spec/` (verified by grep)


dblock:
Why drop the enumerator?

ericproulx:
> Why drop the enumerator?

https://blog.arkency.com/2014/01/ruby-to-enum-for-enumerator/
I don't think we really need it but I could add back the `each`

PR #2755, opened 2026-06-01
Inline mustermann-grape into Grape::Router::MustermannPattern

ericproulx:
## Why

`mustermann-grape` is a ~40-line `Mustermann::AST::Pattern` subclass that defines Grape's path syntax (`:param`, `*splat`, `{name}` / `{+splat}`, `( )` optionals, `|`, and an Integer digit-only constraint). In practice it exists **only** for Grape — and it has already grown Grape-specific: the Integer constraint reaches into Grape's own `params` option.

```ruby
param_type = pattern&.options&.dig(:params, param_name, :type)
constraint = param_type == 'Integer' ? /\d/ : '[^/?#.]'
```

Keeping it as a separate gem means a two-step release for any syntax change (cut a `mustermann-grape` release, then bump Grape's `~> 1.1.0` pin), plus its own CI and versioning — a lot of ceremony for 40 lines only Grape uses. Grape also always instantiated the class **directly** (`Mustermann::Grape.new`), never through the `type: :grape` registry.

## What

- Move the grammar into Grape as **`Grape::Router::MustermannPattern`** (MIT attribution to the original authors — namusyaka, Konstantin Haase, Daniel Doubrovkine — preserved), and instantiate it directly from `Grape::Router::Pattern`.
- Depend on **`mustermann`** directly instead of `mustermann-grape`.
- Drop the `register :grape` call — Grape never used the registry, and not squatting a global Mustermann type keeps the inlined grammar a private Grape detail.

This is **not** a deep dependency reduction: the grammar extends `Mustermann::AST::Pattern`, so the `mustermann` core dependency stays. The graph just collapses from `grape → mustermann-grape → mustermann` to `grape → mustermann`, and Grape owns the 40 lines it was already coupled to. The win is maintenance: pattern-syntax changes become a single PR instead of a cross-gem release dance.

## Compatibility

One behavioral change: the pattern class is no longer registered as `type: :grape`. Apps that called `Mustermann.new(_, type: :grape)` and relied on Grape pulling in `mustermann-grape` transitively should add `gem 'mustermann-grape'` to their Gemfile directly. Documented in `UPGRADING.md`.

## Validation

- `bundle install` now resolves **without `mustermann-grape`** (only `mustermann 4.0.0`) — Grape is self-sufficient.
- Full suite **2,316 examples, 0 failures**; RuboCop clean.
- Integer-constraint behavior verified: `/api/:id` matches `42`, rejects `foo`.


dblock:
Makes sense, let's just check-in with @namusyaka on whether there was a reason to have a separate gem for this?

dblock (approved):

PR #2758, opened 2026-06-02
Fix: VariantCollectionCoercer#to_s returns "Array[Type, ...]" notation

bogdan:
## Problem

`type: [Integer, String]` and `types: [Integer, String]` have different semantics:

- `types: [Integer, String]` — the parameter itself can be Integer **or** String (variant types)
- `type: [Integer, String]` / `type: Array[Integer, String]` — the parameter is always an Array whose **members** can be Integer or String (variant-member-type collection)

However, `VariantCollectionCoercer` had no custom `to_s`, so `route.params[:type]` for the `type:` form serialised to a garbage object string (`"#<Grape::Validations::Types::VariantCollectionCoercer:0x...>"`). Documentation tools such as grape-swagger and grape-oas could not distinguish it from the plain `"[Integer, String]"` produced by `types:`, and could not generate a correct schema for either case.

## Fix

Add `VariantCollectionCoercer#to_s` returning the Grape DSL notation:

- `type: [Integer, String]` → `"Array[Integer, String]"`
- `type: Set[Integer, String]` → `"Set[Integer, String]"`
- `types: [Integer, String]` → `"[Integer, String]"` (unchanged)

This matches the notation users write in the DSL and gives documentation tools an unambiguous signal to emit `{type: array, items: {oneOf: [...]}}` rather than a plain `oneOf`.

## Compatibility

- `declared_params_handler.rb` reads `route.params[:type]` but only matches `'Array'`, `'Hash'`, and the single-type `[Type]` pattern — `"Array[Integer, String]"` does not collide with any of those branches.
- For the Set form: `"Set[Integer, String]"` **does** match line 111's `start_with?('Set')` check, so a `type: Set[Integer, String]` param now gets `Set.new` as its declared default. This is correct and consistent with how `type: Set` and `type: Set[String]` already behave; a spec has been added to pin it down.
- grape-swagger's type-parsing regexes (`/\[(.*)\\]$/`, `/\A\[.*\]\z/`) are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@
 * [#2678](https://github.com/ruby-grape/grape/pull/2678): Update rubocop to 1.86.0 and autocorrect offenses - [@ericproulx](https://github.com/ericproulx).
 * [#2682](https://github.com/ruby-grape/grape/pull/2682): Fix `Style/OptionalBooleanParameter` offenses - [@ericproulx](https://github.com/ericproulx).
 * [#2699](https://github.com/ruby-grape/grape/pull/2699): Fix `Grape::Validations::Types::CustomTypeCoercer` dropping symbolized hash keys for `Array`/`Set` types; refactor the class for readability - [@ericproulx](https://github.com/ericproulx).
+* [#2758](https://github.com/ruby-grape/grape/pull/2758): Fix `VariantCollectionCoercer#to_s` to return `"Array[Type, ...]"` notation instead of a raw object string - [@bogdan](https://github.com/bogdan).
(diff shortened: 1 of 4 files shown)


ericproulx (changes requested):
Nice fix. Two asks: (1) add a spec for the Set form (type: Set[Integer, String] → "Set[Integer, String]") — it's an untested branch. (2) The compatibility note isn't quite right for Set: "Set[Integer, String]" does match declared_params_handler.rb:111's start_with?('Set'), so a type: Set[…] param now declares Set.new instead of falling through. That looks like the correct/consistent behaviour, but please call it out and add a declared-params test to pin it down.

ericproulx (approved):

PR #2759, opened 2026-06-10
Use create_additions: false in Grape::Json.load

dblock:
## What

When `MultiJson` is not available, `Grape::Json` falls back to stdlib `::JSON`. `Grape::Parser::Json` calls `Grape::Json.load` to parse incoming request bodies.

`JSON.load` (unlike `JSON.parse`) honours the `json_class` magic key: it looks up the named constant and calls `.json_create` on it, passing all decoded data. This allows a remote caller to instantiate arbitrary Ruby objects that are already loaded in the process.

This PR wraps the stdlib fallback in a thin module and passes `create_additions: false` to `JSON.load`, which disables that dispatch. The `MultiJson` path is untouched.

## Why this is not a CVE

- **Zero default attack surface.** No classes expose `json_create` with a bare `require 'json'`. The dangerous `json/add/core` (which does add `json_create` to `Regexp`, `Exception`, `Symbol`, etc.) is not loaded by Grape, Rack, or Rails.
- **No known RCE gadget chain.** The stdlib classes that do define `json_create` do not lead to code execution. This is fundamentally different from the Rails YAML/Psych CVEs (e.g. CVE-2013-0156) where RCE was achievable via common library classes.
- **The reporter agreed.** The original security report explicitly stated they would claim it is not worthy of a CVE assignment.
- **Upstream is already fixing it.** Since `json` gem 2.8.0, `JSON.load` emits a deprecation warning when `create_additions` is not explicitly set, and `json` 3.0 will default it to `false`.

The real concern is the information-disclosure angle: `JSON.load` calls `const_get` on the class name, so an attacker can probe what constants are loaded in the process by observing whether they get a 500 vs. a normal response. That alone is enough reason to apply this fix.

## Changes

- `lib/grape/json.rb`: wrap the stdlib JSON fallback in a `Grape::Json` module whose `load` delegates to `JSON.load(..., create_additions: false)`
- `spec/grape/parser/json_spec.rb`: regression spec verifying `json_class` payloads are not deserialized

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@
 * [#2703](https://github.com/ruby-grape/grape/pull/2703): Catch exceptions raised inside `rescue_from` blocks; new `rescue_from :internal_grape_exceptions` opt-in for unrecognised internal errors (resolves [#2482](https://github.com/ruby-grape/grape/issues/2482)) - [@ericproulx](https://github.com/ericproulx).
 * [#2706](https://github.com/ruby-grape/grape/pull/2706): Fix `optional :foo, message: 'oops'` raising `UnknownValidator` - [@ericproulx](https://github.com/ericproulx).
 * [#2751](https://github.com/ruby-grape/grape/pull/2751): Fix structured error messages leaking the raw i18n key for an undefined optional step such as `summary` (closes #2748) - [@ericproulx](https://github.com/ericproulx).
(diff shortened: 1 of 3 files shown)


ericproulx (approved):

ericproulx:
I was wondering why not just use `parse` method ?

dblock:
> I was wondering why not just use `parse` method ?

Take a look at https://github.com/ruby-grape/grape/pull/2761, I think it's weird that we'd overload `load` with `parse`.

PR #2760, opened 2026-06-10
Improve developer experience

olivier-thatch:
This PR sets up the repo to provide an improved developer experience:

1. Include a basic `.editorconfig`, because [EditorConfig](https://editorconfig.org/) is awesome.

2. `.ruby-version` and `Gemfile.lock` are now committed
    * In the past, common practice was to not include these files for gems/libraries, but the current best practice is to commit these to provide a unified developer experience (everyone is running the same set of dependencies) and help with reproducibility.
    * No Ruby version manager is mandated. `.ruby-version` can be read by rbenv, rvm, mise, etc.
    * `.ruby-version` is set to 3.3.11, the latest in the 3.3 branch which is the oldest non-EOL branch. This helps prevent contributors from using modern Ruby features not available in the oldest supported version.
    * `Gemfile` now includes `ruby file: '.ruby-version'` to keep everything in sync
    * `Gemfile.lock` is generated with `--normalize-platforms` and `--add-checksums` to ensure it works on all platforms and that dependencies are pinned with their checksums (pretty important in this age of constant supply chain attacks)

3. VSCode (and VSCode-compatible editors like Cursor etc.) settings:
    * recommended extensions:
        * most importantly Shopify's excellent [Ruby LSP](https://marketplace.visualstudio.com/items?itemName=Shopify.ruby-lsp)
        * [EditorConfig](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)
        * [GitLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens)
        * [Open in GitHub](https://marketplace.visualstudio.com/items?itemName=sysoev.vscode-open-in-github)
    * configure VSCode to format on save. The Ruby LSP will automatically apply Rubocop fixes based on the project's Rubocop configure

4. Ran `rubocop --auto-gen-config` to keep `.rubocop_todo.yml` up-to-date.


.gitignore
```diff
@@ -26,13 +26,11 @@ coverage
 doc
 pkg
 .rvmrc
-.ruby-version
 .ruby-gemset
 .rspec_status
 .bundle
 .byebug_history
 dist
-Gemfile.lock
```
dblock, line 35: This is a gem, so we don't commit a lock file.

dblock (commented):
I don't kind things like EditorConfig or .vscode, will leave open for a bit for others to chime in. Generally in the copilot/claude world we care less and less.  I do recommend splitting these into separate PRs so we can have separate CHANGELOG entries.

PR #2761, opened 2026-06-10
Replace JSON.load with JSON.parse in Grape::Json stdlib fallback

dblock:
Follow-up to #2759.

`JSON.parse` is the semantically correct API for parsing untrusted input. `JSON.load` is documented for internal serialization of trusted data (analogous to `Marshal.load`). The previous fix passed `create_additions: false` to neuter `JSON.load` — but that just makes it behave like `JSON.parse`. May as well use the right call directly.

To me it looks strange that we'd overload `load` with `parse`, so I think we don't merge this.


ericproulx:
I think [this](https://github.com/ruby-grape/grape/pull/2762) is better

PR #2765, opened 2026-06-14
Detect the MultiXML constant to avoid the multi_xml deprecation

ericproulx:
## Background

multi_xml **0.9.0** renamed its constant from `MultiXml` to `MultiXML` (same modernization as multi_json 1.21.0, by the same author). `MultiXml` is now a deprecated alias, removed in v1.0, that emits a one-time warning on constant access or method call:

```
The MultiXml constant is deprecated and will be removed in v1.0. Use MultiXML instead.
```

`lib/grape/xml.rb` detected the backend via `::MultiXml` and `Grape::Xml.parse` resolved to `MultiXml.parse`, so users on multi_xml >= 0.9 would hit the deprecation.

## What this PR does

Detect `::MultiXML` first, then fall back to the legacy `::MultiXml` constant, then `ActiveSupport::XmlMini`:

| Backend | `Grape::Xml` |
| --- | --- |
| multi_xml >= 0.9 | `::MultiXML` |
| multi_xml < 0.9 | `::MultiXml` |
| no multi_xml | `::ActiveSupport::XmlMini` |

**Unlike the multi_json counterpart (#2764), no facade is needed.** multi_xml's only method-level deprecation is `load` → `parse`, and Grape calls only `parse`, which is canonical (not renamed) on every backend — so a direct constant alias suffices.

### CI covers both multi_xml lines

`gemfiles/multi_xml.gemfile` is pinned to `>= 0.9` and a new `gemfiles/multi_xml_0_8.gemfile` pins `< 0.9`; both run the `spec/integration/multi_xml` suite, exercising the `MultiXML` and legacy-`MultiXml` paths. The integration spec now drives a real Grape API that parses an XML request body through `Grape::Xml.parse` (rather than a static constant-equality check), so a regression onto the deprecated constant surfaces via the suite's deprecation handler.

## Testing

- `BUNDLE_GEMFILE=gemfiles/multi_xml.gemfile bundle exec rspec spec/integration/multi_xml` (>= 0.9, resolves to 0.9.1): 1 example, 0 failures
- `BUNDLE_GEMFILE=gemfiles/multi_xml_0_8.gemfile bundle exec rspec spec/integration/multi_xml` (< 0.9, resolves to 0.8.1): 1 example, 0 failures
- `bundle exec rspec` (no multi_xml, `XmlMini` fallback): 2322 examples, 0 failures
- Verified the integration spec fails against the pre-fix `lib/grape/xml.rb` under multi_xml 0.9.1 (the deprecation surfaces), confirming it guards the regression.
- RuboCop clean.

This is the multi_xml counterpart to #2764 (multi_json).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,6 +53,9 @@ jobs:
           - ruby: '4.0'
             gemfile: gemfiles/multi_xml.gemfile
             specs: 'spec/integration/multi_xml'
+          - ruby: '4.0'
+            gemfile: gemfiles/multi_xml_0_8.gemfile
+            specs: 'spec/integration/multi_xml'
     runs-on: ubuntu-latest
     env:
       BUNDLE_GEMFILE: ${{ github.workspace }}/${{ matrix.gemfile }}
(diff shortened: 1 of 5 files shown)


ericproulx:
@dblock I'm gonna cut a version after this one.

PR #2769, opened 2026-06-21
Always format JSON via Grape::Json.dump

### LINKED ISSUE #2486: Formatting with multi_json ?
Am I wrong to say that [Grape::Formatter::JSON](https://github.com/ruby-grape/grape/blob/2b8567a1ca49b3d0236fa82deb2edef34c8839cf/lib/grape/formatter/json.rb#L8) will never call `Grape::Json::Dump` since every object responds to `:to_json` ?

ericproulx:
Closes #2486.

## Summary

`Grape::Formatter::Json` and `Grape::Formatter::SerializableHash` short-circuited to `object.to_json` whenever the object responded to it. Since virtually every object responds to `to_json`, `Grape::Json.dump` was effectively never reached during formatting — so a configured `multi_json` back-end (e.g. `oj`) was silently bypassed, inconsistent with `Grape::ErrorFormatter::Json` and `Grape::Parser::Json`, which already go through `Grape::Json`.

This removes the `to_json` short-circuit from both formatters so JSON is always serialized through `Grape::Json.dump`, honoring the configured back-end consistently.

## Changes

- `lib/grape/formatter/json.rb`: drop `return object.to_json if object.respond_to?(:to_json)`.
- `lib/grape/formatter/serializable_hash.rb`: drop the same `to_json` fallback.
- `README.md`: update the `:json` formatter description.
- `UPGRADING.md`: document the behavior change (breaking).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#2769](https://github.com/ruby-grape/grape/pull/2769): Always format JSON via `Grape::Json.dump` so a configured `multi_json` back-end is honored - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 3.3.0 (2026-06-20)
--- a/README.md
+++ b/README.md
@@ -3160,7 +3160,7 @@ end
 
 Built-in formatters are the following.
 
-* `:json`: use object's `to_json` when available, otherwise call `MultiJson.dump`
+* `:json`: serializes the object via Grape's JSON back-end (`JSON` by default, or `multi_json` when available)
 * `:xml`: use object's `to_xml` when available, usually via `MultiXml`
 * `:txt`: use object's `to_txt` when available, otherwise `to_s`
 * `:serializable_hash`: use object's `serializable_hash` when available, otherwise fallback to `:json`
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,14 @@
 Upgrading Grape
 ===============
 
+### Upgrading to >= 4.0
+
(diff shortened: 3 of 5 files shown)


dblock:
Test failures look legit.

ericproulx:
Closing this — I'm keeping the `to_json` short-circuit.

The premise was that `object.to_json` and `Grape::Json.dump(object)` are equivalent and the short-circuit only had the side effect of bypassing the configured `multi_json` back-end. That's not the case: they produce **different output**, so this isn't a no-op consistency fix — it's a silent serialization change.

`to_json` routes through each object's own JSON representation (ActiveSupport's `as_json`), while `Grape::Json.dump` (→ `JSON.generate`) only handles JSON primitives natively and falls back to `to_s` for everything else. Removing the short-circuit would change the output for `Time`, `Date`, `BigDecimal`, symbols, and any object defining `as_json`/`to_json` — and can even flip the string encoding.

Concretely, for a `Time`:

```ruby
t = Time.at(0).getlocal("+00:00")

t.to_json            # => "1970-01-01T00:00:00.000+00:00"  (ISO 8601, honors as_json)
Grape::Json.dump(t)  # => "1970-01-01 00:00:00 +0000"      (Time#to_s fallback)
```

That's a behavior regression for anyone relying on `as_json`/ISO 8601 output (which is the common case with ActiveSupport loaded), and a far wider contract break than "honor the configured back-end." The consistency gain with `ErrorFormatter::Json` / `Parser::Json` doesn't justify regressing established serialization semantics, so I'd rather leave the formatters as-is.

#2486 should be revisited with an approach that preserves `as_json` semantics if we still want the back-end to apply in the formatter path.

PR #2777, opened 2026-07-04
Add mounted_app reader and app: keyword to Grape::Endpoint

ericproulx:
> Stacked on #2776 → #2775 → #2774 — review/merge those first. The base auto-retargets as the stack lands.

### Summary

`app` (the Rack app or Grape API mounted at an endpoint) was only reachable by digging into the public options bag as `endpoint.options[:app]`. This is **step 1** of moving it to a first-class accessor — additive and non-breaking.

### Changes

- `Grape::Endpoint#initialize` accepts `app:` as an explicit keyword (was pulled from `**options`).
- New public `Grape::Endpoint#mounted_app` reader → `config.app` (the mounted app; `nil` for a plain block endpoint).
- Grape's own mount-refresh dedup (`refresh_already_mounted`) now reads `endpoint.mounted_app` instead of `endpoint.options[:app]`.
- Added coverage for `#mounted_app` and for the `refresh_already_mounted` dedup path, which previously had **none**.

### Non-breaking

`endpoint.options[:app]` is still populated exactly as before (same object; key absent for plain endpoints), so grape-swagger and any external readers keep working. No deprecation warning yet.

### Why `mounted_app`, not `app`

`@app` already means the *runnable* Rack app (`config.app || build_stack`, invoked by `call`). A public `app` returning the raw `config.app` would be `nil` for ordinary block endpoints even though they're callable — confusing. `mounted_app` says exactly what it is.

### Follow-ups (separate PRs)

- Migrate grape-swagger to `endpoint.mounted_app` / `endpoint.endpoints`.
- Drop `:app` from the options Hash in a major, with an UPGRADING note.


dblock:
Document it?

PR #2778, opened 2026-07-04
Rename Grape::Endpoint's for: keyword to api: and expose #api

ericproulx:
> Stacked on #2777 → #2776 → #2775 → #2774 — review/merge those first. The base auto-retargets as the stack lands.

### Summary

`for` identified the API an endpoint belongs to, but it's a reserved word — it can only reach `Grape::Endpoint::Options` through `**rest`, and its value can't be referenced as a local (`x = for` is a syntax error). This renames it to an explicit `api:` keyword and drops the `**rest` workaround.

### Changes

- `Grape::Endpoint#initialize` takes an explicit `api:` keyword (was `for:` via `**options`/`**rest`); the routing DSL passes `api: self`.
- `Grape::Endpoint::Options` member `:for` → `:api`, now a normal keyword (no `**rest`).
- New public `Grape::Endpoint#api` reader, and `#logger`, exposed via `Forwardable` delegation to the config value object.
- `endpoint.options[:for]` removed from the public options Hash; the one in-repo reader (`named_api_spec`) migrated to `endpoint.api`.

### Why `api` stays in `config`

Unlike the earlier `:app`/`:path`/`:method` moves, this one **can't** leave `config`. The endpoint reaches the mounting API's `configuration` through it — `config.api.configuration.evaluate` — and that configuration is set per-mount (`mount ..., with: {...}`), reassigned during setup, and read **lazily at request time**. Capturing it at construction would go stale, so the live API reference is required.

### Upgrading

Two breaking surfaces, documented in `UPGRADING.md`: `Grape::Endpoint.new`'s `for:` keyword → `api:`, and `endpoint.options[:for]` → `endpoint.api`.


dblock:
Document it along with `mounted_app`?

PR #2783, opened 2026-07-05
Read a route's version, anchor and requirements from its pattern

ericproulx:
Rebased onto master now that #2782 is merged.

`version`, `anchor` and `requirements` are the attributes that shape the matcher — they're passed to `Grape::Router::Pattern` to build the mustermann pattern. But they were *also* passed to `Grape::Router::Route` and stored as ivars, solely to back the `route.version`/`route.anchor`/`route.requirements` readers. That's a redundant second copy.

### Change

- `Grape::Router::Pattern` now stores and exposes `anchor`, `version`, `requirements` (`attr_reader`). Its `initialize` signature is unchanged.
- `Grape::Router::BaseRoute` delegates those three readers to `@pattern`, alongside the existing `def_delegators :@pattern, :path, :origin`. They're dropped from its `attr_reader`, `initialize` keywords and ivars.
- `Endpoint#to_routes` no longer passes them to `Route.new`.
- Metadata that does **not** shape matching — `namespace`, `prefix`, `settings` — stays as `Route` keyword arguments.

The line is principled: **the pattern owns the match parameters; the route owns the metadata.**

Also collapses `extract_capture` to a guard plus a single merge. With the `format` keyword gone (#2782) it just adds an optional `version` constraint to `requirements`, so the intermediate accumulator hash was noise — this is the simplification that didn't make it into #2782 before it merged.

### Notes

- grape-swagger already passes `anchor`/`version`/`requirements` to `Pattern.new` and passes **none** of them to `Route.new`, so its routes' readers become correct instead of `nil`.
- `GreedyRoute` borrows another route's pattern, so its `version`/`anchor`/`requirements` now reflect that pattern rather than `nil`. Nothing reads them (the OPTIONS/405 suite is green), so it's invisible — noted for completeness.


dblock (approved):
Merge on clean.

PR #2786, opened 2026-07-06
Make requirements and anchor first-class endpoint inputs

ericproulx:
Stacked on #2785.

Follows the same direction as #2785 (first-class `params`): `route`'s two documented, Grape-consumed options — `requirements:` and `anchor:` — become explicit keyword arguments instead of opaque keys read out of the `route_options` bag. `**route_options` still catches any custom options (e.g. grape-swagger metadata passed inline).

### What changed

- **`route(methods, paths, requirements: nil, anchor: true, **route_options, &)`** — the two documented options are now named parameters; everything else flows through `**route_options` as before.
- **`Grape::Endpoint::Options`** gains `:requirements` and `:anchor` members, and **`Grape::Endpoint.new`** matching `requirements:` / `anchor:` keywords.
- **`to_routes`** reads `config.requirements` / `config.anchor` instead of `route_options[:requirements]` / `route_options.fetch(:anchor, true)`.
- **`mount`** passes `anchor: false` as a keyword rather than `route_options: { anchor: false }`.
- The **verb methods** (`get`/`post`/…) are left as pure `**options` forwarders — `route` is the single source of truth for these options and their defaults.

### Behavior

- `get(':id', requirements: { id: /[0-9]+/ })` still enforces the constraint; `get('/(*:path)', anchor: false)` and mounted apps still match as catch-alls.
- Like `params` in #2785, `requirements`/`anchor` no longer appear in `route.options` (`route.options[:anchor]` → `nil`); the effective value comes from the `route.requirements` / `route.anchor` readers. This refines the note added in #2779 (they previously still appeared in the bag). Nothing in Grape or grape-swagger read them that way.

UPGRADING updated (the #2779 caveat corrected + a parallel note added). Full suite green (2353 examples, 0 failures); RuboCop clean.


ericproulx:
@dblock this is a stack PR.

PR #2787, opened 2026-07-06
Stop populating the write-only namespace description in desc

ericproulx:
`desc` wrote its settings into two places — the route scope (`route_setting(:description)`) and the namespace scope (`namespace_setting(:description)`) — but the namespace copy was pure dead state:

- Grape's runtime reads only `route[:description]` (in `route()`, and only for the `:params` key).
- The namespace scope isn't wired for inheritance — `InheritableSetting#inherit_from` sets `inherited_values` for `namespace_inheritable`/`namespace_stackable`/`namespace_reverse_stackable` and merges `route`, but never touches `@namespace`. So namespace-scope values are only cloned by `point_in_time_copy` and never consulted.
- grape-swagger doesn't read it either (checked against the local checkout).

Confirmed by experiment: dropping the write breaks only the `desc_spec.rb` assertions that checked the write itself — the rest of the suite is unaffected.

### Change

`desc` now writes only the route scope. `namespace_setting(:description)` returns `nil`; a route's description remains available via `route_setting(:description)` and the `route.description` reader. UPGRADING note added.

Full suite green (2351 examples, 0 failures); RuboCop clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [#2782](https://github.com/ruby-grape/grape/pull/2782): Remove the dead `format` keyword from `Grape::Router::Pattern` - [@ericproulx](https://github.com/ericproulx).
 * [#2783](https://github.com/ruby-grape/grape/pull/2783): Read a route's `version`, `anchor` and `requirements` from its pattern instead of storing them again on the route - [@ericproulx](https://github.com/ericproulx).
 * [#2784](https://github.com/ruby-grape/grape/pull/2784): Move HEAD route creation into `Grape::Router::Route#to_head` - [@ericproulx](https://github.com/ericproulx).
+* [#2787](https://github.com/ruby-grape/grape/pull/2787): Stop populating the write-only `namespace_setting(:description)` in `desc` - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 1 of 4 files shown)


dblock:
Can a namespace have its own/different description?

ericproulx:
No — and it couldn't before this PR either. `desc` has only ever described the *next endpoint*: the runtime consumes only `route[:description]`, and the namespace-scope copy removed here was write-only (the `@namespace` scope isn't wired into `inherit_from`, and neither Grape nor grape-swagger ever read it).

What `desc` before a `namespace` actually does is unchanged by this PR — it attaches to the first route defined inside (verified against this branch):

```ruby
desc 'namespace-level description'
namespace :users do
  get :first   # description: "namespace-level description"

  desc 'route-level description'
  get :second  # "route-level description"

  get :third   # nil
end
```

So routes inside a namespace can each have their own description, exactly as before. For metadata on the namespace itself, the supported channel is options on the `namespace` call — `namespace :users, desc: 'User operations'` — which land in `Grape::Namespace#options` and are what grape-swagger walks via `namespace_stackable[:namespace]`.

(Side note: an unconsumed `desc` also leaks into the first route of the *next sibling* namespace — a pre-existing quirk, untouched here; happy to open a follow-up issue.)

dblock:
> An unconsumed `desc` also leaks into the first route of the _next sibling_ namespace — a pre-existing quirk, untouched here; happy to open a follow-up issue.

Yes, this looks unexpected. I think `desc` on a namespace should cause an error.

PR #2788, opened 2026-07-08
Compare the base API, not to_s, when refreshing a mounted app

ericproulx:
### Problem

`refresh_already_mounted` deduplicates endpoints for the "same" mounted app with a string compare:

```ruby
endpoints.delete_if { |endpoint| endpoint.mounted_app.to_s == app.to_s }
```

This only works by accident. `mount` turns every mounted Grape API into a throwaway `mount_instance`, and `mount_instance` builds a **fresh `Class.new(@base_parent)` on every call** (`lib/grape/api.rb`), so two mounts of the same API are *distinct objects* — `app == mounted_app` is always `false`. What makes the `.to_s` compare land is that `Grape::API::Instance` delegates `to_s` to its `@base` (the original API):

```ruby
def_delegators :@base, :to_s   # lib/grape/api/instance.rb
```

So `mounted_app.to_s == app.to_s` is really asking *"do these two throwaway instances share the same base API?"* — through a **string proxy**. That's brittle: it's an identity check dressed up as a string compare, and two distinct APIs that happen to share a name (same `to_s`) falsely dedup.

### Fix

Compare the base API by identity — the real signal the string proxy stands in for. `@base` was writer-only (`base=`), so this exposes a reader and compares it:

```ruby
def same_mounted_app?(mounted, app)
  return mounted.base.equal?(app.base) if mounted.respond_to?(:base) && app.respond_to?(:base)

  mounted.equal?(app)
end
```

Plain Rack apps have no base and are mounted as-is (same object across mounts), so they fall back to object identity.

### Tests

- Existing `refresh_already_mounted` specs still pass (same API → same base → deduped; no flag → duplicated).
- Added a spec that forces a `#to_s` collision between two distinct APIs and asserts the refresh keys off base identity, not the string — it **fails on the old `to_s` compare** (removes the wrong endpoint) and passes with this change.
- Full suite: 2352 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/grape/dsl/routing.rb
+++ b/lib/grape/dsl/routing.rb
@@ -152,7 +152,7 @@ module Grape
           refresh_already_mounted = opts.any? ? opts.first[:refresh_already_mounted] : false
           if refresh_already_mounted && !endpoints.empty?
             endpoints.delete_if do |endpoint|
-              endpoint.mounted_app.to_s == app.to_s
+              same_mounted_app?(endpoint.mounted_app, app)
             end
           end
 
@@ -292,6 +292,18 @@ module Grape
       def refresh_mounted_api(mounts, *opts)
         opts << { refresh_already_mounted: true }
         mount(mounts, *opts)
+      end
+
+      # Two mounts refer to the same app when they share the same base Grape
+      # API. +mount+ turns every mounted Grape API into a throwaway
+      # +mount_instance+ (a fresh +Class.new+ per mount), so object identity
+      # never holds across mounts; comparing the base is the real signal.
+      # Plain Rack apps have no base and are mounted as-is, so they fall back
+      # to object identity.
+      def same_mounted_app?(mounted, app)
+        return mounted.base.equal?(app.base) if mounted.respond_to?(:base) && app.respond_to?(:base)
+
+        mounted.equal?(app)
       end
 
       # Execute first the provided block, then each of the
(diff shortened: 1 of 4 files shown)


lib/grape/dsl/routing.rb
```diff
@@ -274,6 +274,18 @@ def refresh_mounted_api(mounts, *opts)
         mount(mounts, *opts)
       end
 
+      # Two mounts refer to the same app when they share the same base Grape
+      # API. +mount+ turns every mounted Grape API into a throwaway
+      # +mount_instance+ (a fresh +Class.new+ per mount), so object identity
+      # never holds across mounts; comparing the base is the real signal.
+      # Plain Rack apps have no base and are mounted as-is, so they fall back
+      # to object identity.
+      def same_mounted_app?(mounted, app)
```
dblock, line 303: Do we have `app` in scope here? In which case it could be called `mounted_in?(mounted_app)`?

dblock (approved):
Don't love the naming/implementation, but conceptually LGTM.

lib/grape/dsl/routing.rb
ericproulx: `app` is in scope at the call site — the `delete_if` block closes over it — but not inside `same_mounted_app?`. That's a plain private method, so it can't reach the caller's local, which is why it takes both `mounted` and `app`. Renaming to a one-arg `mounted_in?(mounted_app)` wouldn't compile as-is; getting that shape would mean inlining the comparison back into the block rather than renaming the method.

I'd lean toward keeping the extracted method — the name plus the doc comment earn their keep. And `same_mounted_app?` fits better than `mounted_in?` since the check is symmetric: it compares `mounted.base` to `app.base` (falling back to object identity for plain Rack apps, which have no base), not a containment relationship.

Happy to inline it if you'd prefer that instead.

dblock: All good, thanks for the detailed explanation.

PR #2790, opened 2026-07-10
Don't mutate router maps at request time

ericproulx:
### Problem

`Grape::Router` builds its lookup tables as auto-vivifying hashes:

```ruby
@map           = Hash.new { |hash, key| hash[key] = [] }
@optimized_map = Hash.new { |hash, key| hash[key] = // }
```

`compile!` only populates keys for HTTP methods that have routes, but at request time `#match?` and `#rotation` index the maps by the *request's* method. Any method with no routes triggers the default block and **inserts a key into the shared hashes at request time**. `@router.freeze` doesn't help — `freeze` is shallow, so the inner hashes stay mutable.

Demonstration on a GET-only API (starts with `["GET", "HEAD"]`):

```
POST /hello      -> 405, maps now ["GET","HEAD","POST","*"]
PROPFIND /hello  -> maps now [..., "PROPFIND"]
200 concurrent requests, each a distinct custom method -> maps grow to 203 keys
```

Impact:
- **JRuby / TruffleRuby (no GVL):** concurrent `Hash#[]=` on a plain hash is a data race → lost updates / corrupted state.
- **All runtimes:** unbounded growth — every distinct method string becomes a permanent key, so a client able to send arbitrary methods grows the maps without limit (memory-exhaustion vector).
- **MRI:** currently safe only because the GVL serializes each insert; still shared mutable state on the hot path.

### Fix

- Use plain hashes (no auto-vivifying default).
- Create the routes array explicitly on `append` (`(@map[method] ||= []) << route`).
- `freeze` both maps at the end of `compile!` so any future request-time write raises loudly instead of silently mutating.
- Read missing keys without inserting: `#match?` via safe navigation, `#rotation` via `&.each`.

Requests with an unrouted method still resolve to 404/405 exactly as before.

### Tests

New `spec/grape/router_spec.rb` regressions assert the maps are frozen after compilation, that routing methods with no routes does not mutate them, and that such requests still return the default 404 without error. Full suite: 2354 examples, 0 failures. RuboCop clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 #### Fixes
 
 * [#2767](https://github.com/ruby-grape/grape/pull/2767): Update rubocop to 1.88.0 and rubocop-rspec to 3.10.2 - [@ericproulx](https://github.com/ericproulx).
+* [#2790](https://github.com/ruby-grape/grape/pull/2790): Stop mutating `Grape::Router`'s internal maps at request time for HTTP methods with no routes (data race on JRuby/TruffleRuby and unbounded growth) - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 3.3.2 (2026-07-05)
--- a/lib/grape/router.rb
+++ b/lib/grape/router.rb
@@ -13,8 +13,11 @@ module Grape
     def initialize
       @neutral_map = []
       @neutral_regexes = []
-      @map = Hash.new { |hash, key| hash[key] = [] }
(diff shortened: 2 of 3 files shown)


dblock (approved):
TIL. This is pretty damn interesting. How did you find this?

ericproulx:
> TIL. This is pretty damn interesting. How did you find this?

Fable 5 did. I was surprised too.

dblock:
@ericproulx release a patch soon?

PR #2798, opened 2026-07-18
Internalize callback filters behind Grape::Util::InheritableSetting accessors

ericproulx:
Continues the `InheritableSetting` cleanup from #2795/#2796: the filter blocks registered by the callbacks DSL (`before`, `before_validation`, `after_validation`, `after`, `finally`) are now recorded and read through intention-revealing accessors on `Grape::Util::InheritableSetting`, so no caller indexes into `namespace_stackable` for these keys anymore.

### New accessors on `InheritableSetting`

Following the `rescue_handlers` / `add_rescue_handlers` precedent:

| Accessor | Replaces |
|---|---|
| `callbacks` | `namespace_stackable[:befores]`, `[:before_validations]`, `[:after_validations]`, `[:afters]`, `[:finallies]` |
| `add_callback(name, block)` | `namespace_stackable[<pluralized key>] = block` |

Notes:

* `callbacks` returns a Hash keyed by the DSL method names (`callbacks[:before]`, `callbacks[:finally]`, …), matching how users write the DSL and the `type` labels already used by `run_filters` / instrumentation payloads. The singular→pluralized mapping now lives in one place: `InheritableSetting::CALLBACK_STORE_KEYS`.
* `add_callback` uses `fetch`, so a typo'd callback name raises `KeyError` at definition time; `Endpoint#compile!` reads with `callbacks.fetch(...)`, so key drift fails loudly at boot instead of leaving a `nil` ivar that would only crash at request time inside `run_filters`.
* `DSL::Callbacks` no longer knows anything about storage — it collapses to five method names delegating to `add_callback`.
* Storage under the pluralized `namespace_stackable` keys is unchanged, so `to_hash` output and external readers (e.g. grape-swagger) see no difference; the pluralized keys should now be considered internal.
* The DSL spec now asserts through the public `callbacks` accessor and gains coverage for the previously untested `.finally`.


dblock:
rebase

PR #2799, opened 2026-07-18
Internalize rescue configuration behind Grape::Util::InheritableSetting accessors

ericproulx:
Continues the `InheritableSetting` cleanup from #2795/#2796/#2797/#2798: the remaining rescue state written by `rescue_from` — the stacked `RescueOptions` and the meta-selector handlers/flags — is now recorded and read through intention-revealing accessors on `Grape::Util::InheritableSetting`, completing the rescue-configuration encapsulation started with `rescue_handlers` / `add_rescue_handlers`.

### New accessors on `InheritableSetting`

| Accessor | Replaces |
|---|---|
| `rescue_options` / `add_rescue_options(options)` | `namespace_stackable[:rescue_options]` |
| `add_all_rescue_handler(handler)` | `namespace_inheritable[:rescue_all] = true` + `[:all_rescue_handler] = handler` |
| `add_grape_exceptions_rescue_handler(handler)` | the three `namespace_inheritable` writes for `:grape_exceptions` |
| `add_internal_grape_exceptions_rescue_handler(handler)` | `namespace_inheritable[:internal_grape_exceptions_rescue_handler] = handler` |
| `rescue_all?`, `rescue_grape_exceptions?`, `all_rescue_handler`, `grape_exceptions_rescue_handler`, `internal_grape_exceptions_rescue_handler` | the corresponding `namespace_inheritable` reads |

Notes:

* The `rescue_options` reader absorbs the nearest-scope-wins convention: it returns the latest registered `Grape::DSL::RescueOptions` (or `nil`), hiding the raw stack and the `&.last` previously done at the read site in `Endpoint#error_middleware_options`. Since `StackableValues#[]` returns a frozen empty array for never-written keys, plain `.last` is already nil-safe.
* Each meta-selector writer records its handler *and* flips the associated flags, so the `rescue_all`/`rescue_grape_exceptions` coupling lives in one place instead of being spelled out in the DSL.
* `rescue_from` simplifies: after the mixed-arguments guard, a present meta selector implies it was the only argument, so the method now cases on `meta_selector` directly and the `namespace_inheritable` / `arg` prelude is gone.
* `rescue_all?` / `rescue_grape_exceptions?` return `false` rather than `nil` when never set. `Grape::Middleware::Error` only ever uses them in boolean context, so behavior is unchanged.
* The keys' storage is unchanged, so `to_hash` output and any external readers see the same values; the raw keys should now be considered internal.
* `DSL::RequestResponse` specs now assert through the accessors, and `rescue_from :internal_grape_exceptions` gains its first direct spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * [#2786](https://github.com/ruby-grape/grape/pull/2786): Make route `requirements` and `anchor` explicit keyword arguments and first-class endpoint inputs instead of opaque `route_options` keys - [@ericproulx](https://github.com/ericproulx).
 * [#2788](https://github.com/ruby-grape/grape/pull/2788): Compare the base API instead of `to_s` when refreshing a mounted app - [@ericproulx](https://github.com/ericproulx).
 * [#2796](https://github.com/ruby-grape/grape/pull/2796): Remove `Grape::Util::ReverseStackableValues`, storing rescue handlers in plain per-scope hashes merged over `InheritableSetting#parent` - [@ericproulx](https://github.com/ericproulx).
+* [#2799](https://github.com/ruby-grape/grape/pull/2799): Internalize the remaining rescue configuration (`rescue_options`, meta-selector handlers and flags) behind `Grape::Util::InheritableSetting` accessors - [@ericproulx](https://github.com/ericproulx).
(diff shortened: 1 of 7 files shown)


dblock:
rebase

PR #2813, opened 2026-07-23
Fix type: Array + values: [...] coercing to the guessed element type instead of Array

### LINKED ISSUE #2812: Possible regression from 3.2.1 - type: Array + values: [...] produces an extra "is invalid" error in 3.3.0+
While bumping an API from Grape 3.2.1 → 3.3.3, a number of request specs whose params used `type: Array, values: [...]` for filtering (e.g. `skills:`, `job_types:`) started failing with an additional "is invalid" error being returned as well as the initial value is invalid error.

This can be resolved by declaring the param with an explicit element type instead of just `Array`, e.g. `type: [String]` instead of `type: Array`, but I wanted to check if this was an intentional change or a regression from 3.2.1?

### Reproduction

```ruby
# Gemfile: gem "grape", "3.3.3" (also reproduces on 3.3.0, 3.3.1, 3.3.2)
require 'grape'
require 'rack/test'

class API < Grape::API
  format :json
  params do
    optional :skills, type: Array, values: %w[a b c]
  end
  get '/test' do
    { ok: true }
  end
end

include Rack::Test::Methods
def app; API; end

get '/test', skills: ['foo']
puts last_response.body
```

**Grape 3.2.1:**
```json
{"error":"skills does not have a valid value"}
```

**Grape 3.3.0 – 3.3.3:**
```json
{"error":"skills is invalid, skills does not have a valid value"}
```

The extra `"skills is invalid"` appears to come from `CoerceValidator` and should it as `["foo"]` is a perfectly valid `Array`, it just contains a value outside the allowed set, which is exactly what `ValuesValidator` already reports.

### Versions

- Grape: 3.3.0+
- Ruby 4.0.6

schinery:
`guess_coerce_type`'s result was overwriting `@coerce_type` in place, so `type: Array` + `values: [...]` param coerced at runtime using the guessed element type instead of `Array` — rejecting valid array input and duplicating the `values` error on invalid input. 

The guessed type is now only used locally for the existing declaration-time `validate_value_coercion` check, never for `coerce_options`.

Fixes #2812.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 #### Fixes
 
 * [#2767](https://github.com/ruby-grape/grape/pull/2767): Update rubocop to 1.88.0 and rubocop-rspec to 3.10.2 - [@ericproulx](https://github.com/ericproulx).
+* [#2812](https://github.com/ruby-grape/grape/pull/2813): Fix `type: Array` combined with `values: [...]` list coercing to the guessed element type instead of `Array` - [@schinery](https://github.com/schinery).
 * Your contribution here.
 
 ### 3.3.3 (2026-07-17)
--- a/lib/grape/validations/validations_spec.rb
+++ b/lib/grape/validations/validations_spec.rb
@@ -51,6 +51,7 @@ module Grape
         @allow_blank = resolve_value(raw[:allow_blank])
         @fail_fast = raw[:fail_fast] || false
 
+        @guessed_coerce_type = guess_coerce_type(@coerce_type, @values, @except_values)
         @shared_opts = { allow_blank: @allow_blank, fail_fast: @fail_fast }.freeze
         @validator_entries = build_validator_entries(raw)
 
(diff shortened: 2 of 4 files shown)


ericproulx:
👋  @schinery I think the local `guessed_coerce_type` will always have the same result overtime since its value is based on previously initialized ivar and not any runtime values. Maybe we could add a `@guessed_coerce_type` in the initialize so we would have the original @coerce_type and the "guessed" one. Wdyt ?

schinery:
@ericproulx yeah good point. I've added `@guessed_coerce_type` into the initialize so it's computed once alongside `@coerce_type` rather than recomputed as a local inside `validate!` as well as an additional spec  in https://github.com/ruby-grape/grape/pull/2813/commits/888cab888e3dbc9efa521958f6eb4f6a1238d570

Lemme know what you think.

ericproulx:
> @ericproulx yeah good point. I've added `@guessed_coerce_type` into the initialize so it's computed once alongside `@coerce_type` rather than recomputed as a local inside `validate!` as well as an additional spec in [888cab8](https://github.com/ruby-grape/grape/commit/888cab888e3dbc9efa521958f6eb4f6a1238d570)
> 
> Lemme know what you think.

Amazing!

ericproulx:
@schinery I'll make a release this weekend.

PR #2815, opened 2026-07-25
Fix line-anchored regexes matching per line instead of whole string

ericproulx:
## Summary

Two whole-string checks used `^`/`$`, which in Ruby anchor to **lines**, not string boundaries. Both misfired on multi-line input.

### `Grape::Validations::Types::Json#parse` — user-visible bug

The "allow nulls and blank strings" guard was:

```ruby
return if input.nil? || input.match?(/^\s*$/)
```

`/^\s*$/` matches any string *containing* a blank (or whitespace-only) line, so perfectly valid JSON payloads were treated as blank and silently coerced to `nil`:

```ruby
"[1,2]\n\n".match?(/^\s*$/)                 # => true (trailing blank line)
"{\"a\": 1,\n\n \"b\": 2}".match?(/^\s*$/)  # => true (pretty-printed with an empty line)
```

For a `type: JSON` param this means the payload vanishes — the param behaves as if it were never sent. Fixed by replacing the regex with `input.blank?` (ActiveSupport is already a dependency and `blank?` is used elsewhere in the validators), which anchors correctly, covers the `nil` check, and drops the regex entirely.

### `Grape::Router::Pattern::Path#valid_part?` — latent

```ruby
part&.match?(/^\S/)   # meant: starts with a non-whitespace character
```

A part like `"\n/path"` passed because `^\S` matches at the start of its second line. Path parts virtually never contain newlines, so this is latent rather than user-visible, but it is the identical anchoring mistake — fixed with `\A\S`.

For reference, `Grape::Util::MediaType::VENDOR_VERSION_HEADER_REGEX` is the in-repo precedent doing this correctly with `\A...\z`.

## Tests

- New `spec/grape/validations/types/json_spec.rb` unit-covering `Json.parse`/`JsonArray.parse`: nil/blank inputs still return `nil`; payloads with an interior blank line or a trailing blank line now parse.
- Integration case in `coerce_validator_spec.rb` (`first-class JSON` context): a `type: JSON` param whose payload contains a blank line reaches the endpoint parsed.
- Suffix case in `router/pattern/path_spec.rb`: a part starting with whitespace no longer counts as a valid part.

Full suite: 2376 examples, 0 failures. RuboCop clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 #### Fixes
 
 * [#2767](https://github.com/ruby-grape/grape/pull/2767): Update rubocop to 1.88.0 and rubocop-rspec to 3.10.2 - [@ericproulx](https://github.com/ericproulx).
+* [#2815](https://github.com/ruby-grape/grape/pull/2815): Fix line-anchored regexes: `type: JSON` payloads containing a blank line were silently coerced to `nil` - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 3.3.4 (2026-07-25)
--- a/lib/grape/router/pattern/path.rb
+++ b/lib/grape/router/pattern/path.rb
@@ -61,7 +61,7 @@ module Grape
         end
 
         def valid_part?(part)
-          part&.match?(/^\S/) && not_slash?(part)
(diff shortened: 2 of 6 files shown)


dblock:
Interesting!

PR #2822, opened 2026-07-26
Add a numericality validator

ericproulx:
## Summary

Grape currently has no built-in validator for bounding or shape-checking a numeric parameter beyond `values:` (a closed inclusive range or an explicit list/predicate). Common cases like "must be a positive integer", "0 to 100 with an exclusive upper bound", or "must be a perfect score of exactly 5" require either abusing `values:` with a lambda (weak error messages, awkward for open-ended bounds) or writing a full `contract_scope` schema for a single field.

This PR adds a `numericality` validator, following `ActiveModel::Validations::NumericalityValidator`'s option names for familiarity to anyone coming from Rails:

```ruby
params do
  requires :quantity, type: Integer, numericality: { greater_than: 0 }
  requires :discount, type: Float, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 100 }
  requires :rating,   type: Integer, numericality: { equal_to: 5 }
  requires :page,     type: Integer, numericality: { greater_than: 0, only_integer: true }
end
```

Supported options (any combination): `greater_than`, `greater_than_or_equal_to`, `less_than`, `less_than_or_equal_to`, `equal_to`, `other_than`, `only_integer`, `odd`, `even`. Each has its own i18n message key (`numericality_greater_than`, etc.), mirroring how `length` already has separate `length_min`/`length_max`/`length_is` keys. Per-field `message:` overrides work the same way as every other validator.

## Design notes

- **Non-Numeric values are skipped, not errored.** Numeric-ness is `type:`/coercion's job; this validator only owns bounds/parity, so it composes with `coerce` rather than duplicating its work. A param with no `type:`/coerce still round-trips through unchanged if it isn't already a `Numeric`.
- **Typed arrays are checked element-wise.** `requires :numbers, type: [Integer], numericality: { greater_than: 0 }` validates every element (via `Array.wrap`), matching how `values` already treats arrays. `length` is the one validator that intentionally owns the array's own size, so no overlap there.
- **`only_integer` accepts whole-numbered `Float`/`BigDecimal`**, not just literal `Integer` — `val == val.to_i` — so `type: Float, numericality: { only_integer: true }` accepts `5.0` and rejects `5.5`.
- **Obviously contradictory combinations raise `ArgumentError` at definition time**, not at request time: `greater_than` + `greater_than_or_equal_to`, `less_than` + `less_than_or_equal_to`, `odd` + `even`, `equal_to` + any other comparison option, and a reversed lower/upper bound (e.g. `greater_than: 10, less_than: 5`). This mirrors `LengthValidator`'s existing `min > max` guard.

No existing validator's behavior changes; this is purely additive (new validator class, new locale keys, new README section). No `UPGRADING.md` entry needed.

## Test plan

- [x] New `spec/grape/validations/validators/numericality_validator_spec.rb` (34 examples): every option's pass/fail case, the range combination, the `only_integer` Float case, odd/even, array-element-wise checking, non-Numeric skip behavior, `allow_blank` interplay, custom `message:` override, and every `ArgumentError` guard.
- [x] Full RSpec suite passes locally (2401 examples, 0 failures).
- [x] RuboCop clean across the repo.
- [ ] CI green.


dblock:
Supporting open ranges would be cleaner because we already support closed ranges, no?

```ruby
params do
  requires :quantity, type: Integer, values: 1.. # numericality: { greater_than: 0 }
  requires :discount, type: Float, values: 0..100 # numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 100 }
  requires :rating,   type: Integer, values: 5 # numericality: { equal_to: 5 }
  requires :page,     type: Integer, values: 1.. # this one has to be integer because of type ... numericality: { greater_than: 0, only_integer: true }
end

ericproulx:
Yeah, makes sense. I'll drop it.

dblock:
> Yeah, makes sense. I'll drop it.

~Will you add support for open ranges? It's a useful feature.~

Copilot is working ...

dblock:
https://github.com/ruby-grape/grape/pull/2831

PR #2826, opened 2026-07-26
Set api.version when the version is the only path segment

ericproulx:
## Problem

`Grape::Middleware::Versioner::Path` extracts the version from the first path segment — but only when a **second** slash follows it:

```ruby
slash_position = path_info.index('/', 1) # omit the first one
return unless slash_position
```

For the root route of a path-versioned API there is no second segment, so the middleware returned early and never set `env['api.version']`:

```ruby
class API < Grape::API
  version 'v1', using: :path
  get '/' do
    { version_seen: version }
  end
end
```

`GET /v1` → `200` with `{"version_seen":null}`. The route matches (the version lives in the compiled route pattern), but everything that reads `env['api.version']` sees nothing: the `version` helper, `Grape::DSL::Entity#entity_representation_for` (entities lose their `version:` embed, breaking conditional exposures), and `ErrorFormatter::Base#present`. The same happens with a format suffix (`GET /v1.json`) and behind a prefix (`GET /api/v1`).

Only the versioner middlewares write `api.version`, so nothing downstream recovers it from the route's version capture either. The logic is identical at least as far back as v2.4.0 — long-standing, not a regression.

## Fix

When the (prefix-stripped) path is a single segment, record the version on an **exact match** against the declared versions, trimming a trailing `.format` suffix right-to-left so dotted versions still work:

* `/v1` → `v1`
* `/v1.json` → `v1` (suffix trimmed)
* `/v1.2.json` with `version 'v1.2'` → `v1.2` (rightmost dot trimmed first)
* `/awesome` → no version, **no 404** — the segment is left for the router to resolve

The single-segment branch deliberately never raises `version_not_found!`: with no following segment there is nothing to disambiguate a version from a plain path, so an unknown segment falls through to the router's 404 exactly as before. Multi-segment behavior (`/v3/foo` → 404 with `X-Cascade: pass` for an undeclared version) is unchanged.

## Tests

* Middleware-level cases for all of the above in `versioner/path_spec.rb`.
* A `sets the API version for the root route` example added to the shared versioning examples, so the root route is now covered end-to-end for **all four** strategies (`:path` — previously failing — plus `:param`, `:header`, `:accept_version_header`).

### DIFF
--- a/spec/grape/middleware/versioner/path_spec.rb
+++ b/spec/grape/middleware/versioner/path_spec.rb
@@ -59,4 +59,33 @@ describe Grape::Middleware::Versioner::Path do
       expect(subject.call(Rack::PATH_INFO => '/mounted/v1/foo').last).to eq('v1')
     end
   end
+
+  # Regression: the version was only extracted when a second slash followed it,
+  # so the root route of a path-versioned API (`GET /v1`) ran without
+  # env['api.version'] being set.
+  context 'when the version is the only path segment' do
+    let(:options) { { versions: %w[v1] } }
+
+    it 'sets the version' do
+      expect(subject.call(Rack::PATH_INFO => '/v1').last).to eq('v1')
+    end
+
+    it 'sets the version when a format suffix follows' do
+      expect(subject.call(Rack::PATH_INFO => '/v1.json').last).to eq('v1')
+    end
+
+    it 'sets a dotted version when a format suffix follows' do
+      options[:versions] = %w[v1.2]
+      expect(subject.call(Rack::PATH_INFO => '/v1.2.json').last).to eq('v1.2')
+    end
+
+    it 'sets the version behind a prefix' do
+      options[:prefix] = '/api'
+      expect(subject.call(Rack::PATH_INFO => '/api/v1').last).to eq('v1')
+    end
+
+    it 'leaves an unknown segment for the router instead of failing with 404' do
+      expect(subject.call(Rack::PATH_INFO => '/awesome').last).to be_nil
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


spec/grape/middleware/versioner/path_spec.rb
```diff
@@ -59,4 +59,33 @@
...
+      expect(subject.call(Rack::PATH_INFO => '/v1').last).to eq('v1')
+    end
+
+    it 'sets the version when a format suffix follows' do
+      expect(subject.call(Rack::PATH_INFO => '/v1.json').last).to eq('v1')
+    end
+
+    it 'sets a dotted version when a format suffix follows' do
+      options[:versions] = %w[v1.2]
+      expect(subject.call(Rack::PATH_INFO => '/v1.2.json').last).to eq('v1.2')
```
dblock: I don't understand this one, is it supposed to do v1.2 at the root for a json document?

ericproulx: Good question! The test case was a bit unclear. I've refactored it to use nested RSpec contexts to make the distinction explicit:

```ruby
context 'when the version is the only path segment' do
  let(:options) { { versions: %w[v1] } }
  # ...simple cases with v1
  
  context 'with a dotted version' do
    let(:options) { { versions: %w[v1.2] } }
    
    it 'sets the version when a format suffix follows' do
      expect(subject.call(Rack::PATH_INFO => '/v1.2.json').last).to eq('v1.2')
    end
  end
end
```

To answer your question directly: **Yes**, when an API declares `version 'v1.2', using: :path`, a client requesting the root with a format suffix (`GET /v1.2.json`) should have:
- `env['api.version']` set to `v1.2`
- The `.json` suffix parsed separately as the format

This is needed because dotted versions (semantic versioning) are common, and clients need to specify both version and format. The middleware extracts the version by matching against declared versions and stripping the format suffix from the right.

PR #2841, opened 2026-08-01
Give point-in-time copies their own stackable and rescue-handler stores

ericproulx:
## Summary

`InheritableSetting#point_in_time_copy` copied a scope's stackable store **and its rescue-handler maps** shallowly, so the nested Arrays and Hashes stayed shared with the source. A registration made *after* an endpoint was defined therefore still reached that endpoint — but only when the key already held a registration at the time the copy was taken, since otherwise the writer allocated a fresh store on the source alone.

The outcome depended on something the API never expressed:

```ruby
use Middleware1
get('/x') { }
use Middleware2          # applied to GET /x

get('/x') { }
use Middleware2          # did NOT apply to GET /x
```

```ruby
rescue_from(ArgumentError) { }
get('/x') { raise Boom }
rescue_from(Boom) { }    # rescued GET /x

get('/x') { raise Boom }
rescue_from(Boom) { }    # did NOT rescue GET /x
```

| | before | after |
| --- | --- | --- |
| `use MW1` · route · `use MW2` | MW1, **MW2** | MW1 |
| route · `use MW2` | — | — |
| `rescue_from A` · route · `rescue_from B` | **B rescues the route** | raises |
| route · `rescue_from B` | raises | raises |

Two APIs stating the same thing, behaving differently. Not a deliberate "late registration" feature either — helpers defined after an endpoint already did *not* leak, because they resolve down a different path.

## Approach

Dup the nested stores as well as the Hash holding them. A copy is a point in time: what the source registers afterwards must not reach it.

Inheritance is untouched — it resolves by walking `#parent`, so a scope still sees values an enclosing scope gains later. That is what the existing `decouples namespace stackable values` spec actually exercised: its value lives on the *parent*, so nothing was ever shared and it passed with or without this change. The new specs register on the scope itself before taking the copy, which is the shape that leaked.

## Backward compatibility

UPGRADING entry added. An API that declares `use` / `helpers` / `before` / `rescue_from` below its routes and relies on it applying to them will now see it silently not run. That arrangement only ever worked when an earlier registration for the same key happened to seed the store, so it was never dependable; the fix is to move the registration above the routes it should cover, which is where the documentation has always placed it.

## Perf

Definition-time only. 300 endpoints under scopes carrying 8 middleware, 4 helpers and 6 filters:

| | allocations | boot |
| --- | --- | --- |
| before | 463,062 | 0.127s |
| after | 464,862 (+0.4%) | 0.128s |

## Test plan

- [x] 5 new unit examples in `inheritable_setting_spec.rb` (stackable and rescue-handler leakage in both directions, sibling independence, base-only handlers) and 4 integration examples in `api_spec.rb` covering `use` and `rescue_from` in both the seeded and unseeded arrangements; verified the behavioural ones fail without the `lib/` change.
- [x] Full RSpec suite passes locally (2552 examples, 0 failures).
- [x] RuboCop clean.
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@
 * [#2842](https://github.com/ruby-grape/grape/pull/2842): Warn at definition time when a `rescue_from` class is already covered by one registered earlier in the same scope, since the later handler never runs - [@ericproulx](https://github.com/ericproulx).
 * [#2853](https://github.com/ruby-grape/grape/pull/2853): Restore, behind a deprecation warning, the trailing positional options Hash of `requires`, `optional` and `use`, which #2618 turned into a parameter name - [@ericproulx](https://github.com/ericproulx).
 * [#2856](https://github.com/ruby-grape/grape/pull/2856): Update simplecov - [@ericproulx](https://github.com/ericproulx).
+* [#2841](https://github.com/ruby-grape/grape/pull/2841): Stop `use`, `helpers`, `rescue_from` and other registrations declared below a route from reaching it when an earlier registration had seeded the same key (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 3.3.5 (2026-07-30)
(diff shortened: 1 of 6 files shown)


dblock (approved):
Let's update README with this?

PR #2842, opened 2026-08-01
Warn when a rescue_from handler can never run

ericproulx:
## Summary

`Grape::Middleware::Error` resolves a registered handler with `#find`, so within a scope the **first matching class wins**. A handler registered for a class an earlier one already covers is therefore dead code, and silently so:

```ruby
rescue_from StandardError do ... end   # wins
rescue_from ArgumentError do ... end   # never runs
```

Swapping the two lines is all it takes, but nothing said so — and the neighbouring `rescue_from :all` behaves the *other* way round, since it lives in `all_rescue_handler` and is consulted only after the registered handlers:

| declaration | raising `ArgumentError` |
| --- | --- |
| `rescue_from :all` then `rescue_from ArgumentError` | `ArgumentError` handler ✅ |
| `rescue_from StandardError` then `rescue_from ArgumentError` | **`StandardError` handler** |

Two documented ways of saying "catch everything", ordering differently against a later specific handler — while the README describes `rescue_from :all` as rescuing "all `StandardError` exceptions".

## Approach

Warn at definition time rather than reorder. Which handler should win is the author's call, and `:all` already offers "broad first, specific still wins" to anyone who wants it. No behaviour changes.

```
Grape: rescue_from ArgumentError will never run — StandardError was registered earlier
in the same scope and is matched first. Register the more specific class before the broader one.
```

Deliberately quiet in the cases that are not mistakes:

| case | warns? |
| --- | --- |
| broader class registered first | ✅ |
| same class registered twice (first handler is kept) | ✅ (own message) |
| narrower class registered first | — |
| unrelated classes | — |
| `rescue_from A, B` sharing one handler | — |
| same handler via `with:` on both | — |
| `rescue_subclasses: false` pairs | — |
| inner scope shadowing an enclosing one | — |
| `rescue_from :all` then a specific class | — |

Across scopes the nearest registration deliberately wins, so an inner `rescue_from StandardError` shadowing an outer `rescue_from ArgumentError` is documented behaviour, not a mistake. Exact-match handlers are consulted before the subclass-matching ones and never match a descendant, so they cannot shadow each other either.

The check lives in `Grape::Util::ShadowedRescueHandlers`, called from `#add_rescue_handlers` where a scope's own registrations are known — kept out of `InheritableSetting` itself, which is at its `Metrics/ClassLength` limit and is a settings store rather than a place for DSL diagnostics.

README gains the ordering rule, which was only implied.

## Test plan

- [x] 7 new examples in `inheritable_setting_spec.rb` covering both warning messages and the five quiet cases; verified the two warning ones fail without the `lib/` change.
- [x] Full RSpec suite passes locally (2550 examples, 0 failures) — no existing spec registers a shadowed handler, so no new noise.
- [x] RuboCop clean.
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 * [#2827](https://github.com/ruby-grape/grape/pull/2827): Make the `cascade` DSL getter return the configured value (`cascade false` read back as `true`) - [@ericproulx](https://github.com/ericproulx).
 * [#2829](https://github.com/ruby-grape/grape/pull/2829): Fix a cascading route handing over only to the last route registered for the path, making a middle version (3+ mounted versions with a catch-all) answer 406 - [@ericproulx](https://github.com/ericproulx).
 * [#2826](https://github.com/ruby-grape/grape/pull/2826): Fix `api.version` not being set for the root route of a path-versioned API (`GET /v1`) - [@ericproulx](https://github.com/ericproulx).
+* [#2842](https://github.com/ruby-grape/grape/pull/2842): Warn at definition time when a `rescue_from` class is already covered by one registered earlier in the same scope, since the later handler never runs - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 3.3.5 (2026-07-30)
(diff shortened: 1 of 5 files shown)


dblock (approved):
I like it.

PR #2845, opened 2026-08-01
Render a redirect message as the plain text it claims to be

ericproulx:
## Summary

`#redirect` announces its message as `text/plain` — and has since it was introduced in 2015, in a commit literally titled *"Redirect as plain text with optional message override"* — but it only ever set the header. The body was still handed to the API's own formatter:

```ruby
class API < Grape::API
  format :json
  get('/r') { redirect '/there' }
end
```

```
HTTP/1.1 302 Found
Location: /there
Content-Type: text/plain

"This resource has been moved temporarily to /there."
```

Quotes included. That is neither valid plain text nor what a client reading the content type would expect, and the header and body contradict each other.

| API format | before | after |
| --- | --- | --- |
| `format :json` | `302`, `text/plain` + `"…moved temporarily to /there."` (quoted) | `302`, `text/plain` + `…moved temporarily to /there.` |
| `format :txt` | correct already | unchanged |
| `format :xml` | **`500`, `application/xml`, no `Location` header** | `302`, `Location: /there`, `text/plain` + the message |

The `:xml` row is worse than it first looked: the XML formatter cannot serialize a String, so it raised and the error path took over. `redirect` on an XML API has never redirected at all — it answered `500` with `<error><message>cannot convert String to xml</message></error>` and no `Location`. There is a spec pinning the `302` now.

The existing `#redirect` specs missed it because they run on the default `:txt` format, where the formatter is a no-op.

## Approach

Set `api.format` alongside the header for the generated message — the same lever an endpoint already has via `#api_format` (spec'd in `api_spec.rb` as *'can be overwritten with an explicit api_format'*) — so the message is rendered by the txt formatter whatever the API declares. `:txt` is always resolvable: `Grape::Formatter.formatter_for` falls back to the built-in registry, which is not narrowed by the API's `format`.

It is per-request rack env, so other routes on the same API are unaffected — there is a spec pinning that.

## Backward compatibility

**UPGRADING entry added** — @dblock is right that this is a breaking change, and the original description of it here was wrong.

The body of a redirect changes on any API whose format is not `:txt`. Concretely, `JSON.parse(response.body)` on a redirect succeeded before and now raises. On a JSON API the `Location` header, the status and the `Content-Type` are unchanged, so a client that follows the redirect is unaffected — but code that reads the body is.

On a format that could not serialize the message at all, such as `:xml`, the status, `Content-Type` and `Location` all change — from a `500` with no `Location` to the `302` that was always intended.

Longstanding rather than a regression: present since v0.14.0 (`d1bba79d`, 2015) and identical in 3.3.4.

## Scoped to the generated message

The first version of this PR set `api_format :txt` unconditionally, which also caught a body the caller passed. That was worse than the bug on a JSON API:

```ruby
redirect '/there', body: { message: 'moved' }
# before:  {"message":"moved"}
# with the unconditional version:  {message: "moved"}   # Hash#to_s — neither JSON nor useful text
```

So the format is now only forced for the message Grape generates, which is the one body known to be plain text:

```ruby
api_format :txt unless body
```

A caller-supplied body keeps the API's format and is untouched by this PR. The trade-off is that `redirect url, body: 'go away'` on a JSON API still returns `"go away"` quoted under `text/plain` — the same contradiction, for a body Grape cannot assume anything about. Narrowing it further (say, rendering any `String` body as text) is a judgement call I left out; happy to widen it if you'd prefer.

## Test plan

- [x] 5 new examples in `endpoint_spec.rb` under `#redirect`: plain-text body on a JSON API, a structured and a string caller-supplied body both keeping the API's format, a guard that other routes keep it too, and a `302` + `Location` on an XML API. Verified the XML one fails without the `lib/` change (`expected: 302`).
- [x] Full RSpec suite passes locally (2596 examples, 0 failures).
- [x] RuboCop clean.
- [ ] CI green.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -38,6 +38,41 @@ end
 
 Nothing changes for the ordinary arrangement — registrations declared before a route, or inherited from an enclosing namespace or a mounting API, still apply exactly as before, including values an enclosing scope gains after the nested scope was created.
 
+#### `redirect` renders its default message as plain text
+
+The `redirect` API set the content type to `text/plain`, but delegated body rendering to the formatter. With an API defaulting to JSON, a `redirect` would render the body as JSON with a `text/plain` content-type header.
+
+```ruby
+class API < Grape::API
+  format :json
+  get('/r') { redirect '/there' }
+end
+```
+
+```
+Content-Type: text/plain
+
+"This resource has been moved temporarily to /there."
+```
+
+Grape now renders the message it generates with the txt formatter, so the body is the plain sentence the content type claims:
+
+```
+Content-Type: text/plain
+
+This resource has been moved temporarily to /there.
+```
+
+**What can break.** Code that parses a redirect body — `JSON.parse(response.body)` on a redirect succeeded before and now raises — or a test asserting on the encoded form. On a JSON API the `Location` header, the status and the `Content-Type` are unchanged, so a client that follows the redirect is unaffected.
+
+On an API whose formatter cannot serialize a String, such as `format :xml`, `redirect` did not work at all: the formatter raised, and the response was a `500` carrying an error document and no `Location` header. Those APIs now get the `302` and the `Location` they always should have.
+
+This applies only to the message Grape generates. A body you pass yourself is still rendered by the API's formatter, unchanged:
+
+```ruby
+redirect '/there', body: { message: 'moved' }   # still {"message":"moved"} on a JSON API
+```
+
 #### Path params are tagged UTF-8 instead of ASCII-8BIT
 
 Params captured from the request path — `route_param`, `:id`-style segments, splats — now come back tagged `UTF-8`. They used to carry the `ASCII-8BIT` encoding of Rack's `PATH_INFO`, because Mustermann decodes the path against that raw string and nothing re-tagged the result. Query and body params were already `UTF-8`, since Rack tags those itself.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@
 * [#2841](https://github.com/ruby-grape/grape/pull/2841): Stop `use`, `helpers`, `rescue_from` and other registrations declared below a route from reaching it when an earlier registration had seeded the same key (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
 * [#2846](https://github.com/ruby-grape/grape/pull/2846): Keep a `:version` path capture in `params` when the API declares no version, instead of always dropping it as Grape's own - [@ericproulx](https://github.com/ericproulx).
 * [#2839](https://github.com/ruby-grape/grape/pull/2839): Tag path params as UTF-8 instead of leaving them ASCII-8BIT, so they compare equal to the non-ASCII literals an API declares (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
+* [#2845](https://github.com/ruby-grape/grape/pull/2845): Render `redirect`'s default message as the plain text its content type announces, instead of letting the API's formatter re-encode it (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 3.3.5 (2026-07-30)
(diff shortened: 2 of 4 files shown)


dblock (approved):
This is a backwards incompatible change, right? Needs UPGRADING?

dblock (approved):
This is a backwards incompatible change, right? Needs UPGRADING?

ericproulx:
You're right on both counts — thanks, my "no UPGRADING entry" claim was wrong.

The body of a redirect changes on any API whose format isn't `:txt`. `JSON.parse(response.body)` on a redirect succeeded before and now raises. Added an UPGRADING entry saying so, and corrected the description above.

Checking the blast radius also turned up something the first version got wrong. Setting `api_format :txt` unconditionally caught a body the caller passed, which on a JSON API was worse than the bug:

```ruby
redirect '/there', body: { message: 'moved' }
# before: {"message":"moved"}
# then:   {message: "moved"}    # Hash#to_s — neither JSON nor useful text
```

So it's now scoped to the message Grape generates, which is the one body known to be plain text:

```ruby
api_format :txt unless body
```

The trade-off is that `redirect url, body: 'go away'` on a JSON API still comes back as `"go away"` quoted under `text/plain`. I left that alone rather than guessing at a caller's intent, but rendering any `String` body as text would be a one-line change if you'd rather have it.

UPGRADING.md
```diff
@@ -37,6 +37,38 @@ end
 ```
 
 Nothing changes for the ordinary arrangement — registrations declared before a route, or inherited from an enclosing namespace or a mounting API, still apply exactly as before, including values an enclosing scope gains after the nested scope was created.
+#### `redirect` renders its default message as plain text
```
dblock, line 41: Add a line above.

UPGRADING.md
```diff
@@ -37,6 +37,38 @@ end
...
+
+`redirect` has announced its body as `text/plain` since it was introduced in 0.14.0, but it only ever set the header — the message itself was still handed to the API's own formatter. On an API declaring a format other than `:txt`, the sentence came back encoded by that formatter under a `text/plain` content type:
```
dblock: I think this can be explained simpler.

The `redirect` API set the content-type to `text/plain`, but delegated body rendering to the formatter. With an API defaulting to JSON, a `redirect` would render the body as JSON with a `text/plain` content-type header.

Example: ...

dblock (approved):

UPGRADING.md
ericproulx: Done — blank line added. The heading was butting straight against the previous section.

UPGRADING.md
ericproulx: Agreed, and taken almost verbatim — the version-history framing was the convoluted part. It now opens with:

> The `redirect` API set the content type to `text/plain`, but delegated body rendering to the formatter. With an API defaulting to JSON, a `redirect` would render the body as JSON with a `text/plain` content-type header.

followed by the JSON API example and its before/after body. Kept the two short notes after it: what can break (parsing a redirect body), and that a body you pass yourself still goes through the API formatter.

ericproulx:
Went back over the description claim by claim and found one I had wrong — the `format :xml` row.

I had it as "quoted/encoded by the XML formatter". It is worse than that: the XML formatter cannot serialize a String, so it raises and the error path takes over. On master a redirect from an XML API answers

```
500 Internal Server Error
Content-Type: application/xml

<?xml version="1.0" encoding="UTF-8"?>
<error>
  <message>cannot convert String to xml</message>
</error>
```

with no `Location` header at all — `redirect` has never redirected on an XML API. With this PR it is a `302` with `Location: /there` and the plain message. Added a spec pinning that, and confirmed it fails without the change (`expected: 302`).

That also made "the `Location` header, the status and the `Content-Type` are all unchanged" wrong as an absolute — it holds for JSON, not for a format that could not serialize the message. Corrected in both the description and the UPGRADING entry, which now notes the XML case in one sentence.

Also swapped the `api_spec.rb:4494` reference for the spec name, since the line number drifts on every rebase; it had already gone stale.

PR #2847, opened 2026-08-01
Match Accept media types case-insensitively

ericproulx:
## Summary

Media types are case-insensitive (RFC 9110 §8.3.1), but the registered ones are spelled in lower case and matched literally, so a differently-cased `Accept` header found nothing:

| request | before | after |
| --- | --- | --- |
| `Accept: TEXT/PLAIN` against a `text/plain` content type | served **`application/json`** (the default format) | `text/plain` |
| `Accept: APPLICATION/VND.TWITTER-V1+JSON` on a header-versioned API | `api.version` **nil** | `"v1"` |
| `Accept: application/VND.twitter-v1+json` | `api.version` **nil** | `"v1"` |

Neither failed loudly. Content negotiation fell through to the default format, and header versioning behaved as though no version had been asked for — so the request was served by whichever version matched first, and the client quietly got something other than what it asked for.

## Approach

Three sites decided this, all comparing against the lower-case registered types:

- `Middleware::Formatter#format_from_header` — the `Rack::Utils.best_q_match` lookup
- `Util::MediaType.best_quality_media_type` — the same lookup on the versioning path
- `Util::MediaType.parse` / `.match?` — the vendor pattern

Down-case the incoming media type at each. The vendor pattern itself stays lower-case: that is the case a vendor and version are declared in via the DSL, and therefore the case they are compared in.

Grape already treats media types this way when deciding whether to escape an error body — `Middleware::Error#html_content_type?`, added by #2789, compares with `casecmp?`.

## Backward compatibility

No UPGRADING entry. A differently-cased `Accept` previously matched nothing, so nothing could depend on the old outcome beyond receiving the default format. Lower-case headers — effectively all real traffic — are unaffected.

Longstanding rather than a regression: identical in 3.3.4.

## Test plan

- [x] 2 new examples in `media_type_spec.rb` (vendor/version/format parsed from an upper-case header, and type/subtype down-cased) and 2 in `api_spec.rb` (content negotiation and vendor version resolution end to end); verified all four fail without the `lib/` change.
- [x] Full RSpec suite passes locally (2547 examples, 0 failures).
- [x] RuboCop clean.
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 * [#2827](https://github.com/ruby-grape/grape/pull/2827): Make the `cascade` DSL getter return the configured value (`cascade false` read back as `true`) - [@ericproulx](https://github.com/ericproulx).
 * [#2829](https://github.com/ruby-grape/grape/pull/2829): Fix a cascading route handing over only to the last route registered for the path, making a middle version (3+ mounted versions with a catch-all) answer 406 - [@ericproulx](https://github.com/ericproulx).
 * [#2826](https://github.com/ruby-grape/grape/pull/2826): Fix `api.version` not being set for the root route of a path-versioned API (`GET /v1`) - [@ericproulx](https://github.com/ericproulx).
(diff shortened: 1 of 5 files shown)


dblock (approved):
Add to UPGRADING?

PR #2850, opened 2026-08-01
Adversarial sweep: eleven longstanding fixes

ericproulx:
## Summary

The eleven changes from the 2026-08 adversarial sweep, gathered into one branch — [#2838](https://github.com/ruby-grape/grape/pull/2838) through [#2849](https://github.com/ruby-grape/grape/pull/2849) except [#2848](https://github.com/ruby-grape/grape/pull/2848), one commit each, cherry-picked in order and applying cleanly.

Every one is **longstanding, not a regression**: each was reproduced on released 3.3.4 before being fixed. Each individual PR carries the full reasoning, benchmarks and rejected-alternatives; this one exists to review or land them together.

| | fix | area |
| --- | --- | --- |
| [#2838](https://github.com/ruby-grape/grape/pull/2838) | Params nested deeper than an array scope declares passed validation, then crashed the endpoint with `TypeError`. Also gives `type: Array[JSON]` its element indices back | validation |
| [#2839](https://github.com/ruby-grape/grape/pull/2839) | Path params arrived ASCII-8BIT while query params were UTF-8, so `values: ['café']` matched `?id=café` but not `/café` | encoding |
| [#2840](https://github.com/ruby-grape/grape/pull/2840) | An exception while rendering an error body escaped the whole middleware stack; `rescue_from :all` did not help | error handling |
| [#2841](https://github.com/ruby-grape/grape/pull/2841) | `use` / `helpers` / `rescue_from` declared *below* a route still applied to it — but only when an earlier registration had seeded the store | settings |
| [#2842](https://github.com/ruby-grape/grape/pull/2842) | Warn when a `rescue_from` class is already covered by one registered earlier in the same scope, making it dead code | error handling |
| [#2843](https://github.com/ruby-grape/grape/pull/2843) | `rescue_from :grape_exceptions` was inert against a catch-all written as a class, so validation failures still answered 500 | error handling |
| [#2844](https://github.com/ruby-grape/grape/pull/2844) | `represent` was silently skipped for any model answering `#first` or `#klass` — a `Struct`, or anything `Enumerable` | entities |
| [#2845](https://github.com/ruby-grape/grape/pull/2845) | `redirect` announced `text/plain` and then let the API's formatter re-encode the body | responses |
| [#2846](https://github.com/ruby-grape/grape/pull/2846) | `route_param :version` on an API declaring no version reached the endpoint as `nil` | routing |
| [#2847](https://github.com/ruby-grape/grape/pull/2847) | `Accept` media types were matched case-sensitively, so `TEXT/PLAIN` silently got the default format | content negotiation |
| [#2849](https://github.com/ruby-grape/grape/pull/2849) | Remove `http_digest`, which has had no strategy behind it since 2.0.0 and raised on the first request rather than at definition | auth |

## Contract changes

Five carry UPGRADING entries, plus a rewrite of the existing `auth` keyword-arguments section:

- path params tagged UTF-8 (#2839)
- a failed error rendering answers 500 rather than escaping (#2840)
- registrations below a route no longer reach it (#2841)
- `:grape_exceptions` outranks a catch-all class handler (#2843)
- `http_digest` removed (#2849)

#2844, #2845, #2846, #2847 and #2848 need none: each only makes something work that previously could not, on inputs that had no working behaviour to depend on.

## Supersedes

Closes #2838
Closes #2839
Closes #2840
Closes #2841
Closes #2842
Closes #2843
Closes #2844
Closes #2845
Closes #2846
Closes #2847
Closes #2849

Each of those carries the full reasoning for its fix — reproduction, benchmarks, rejected alternatives — and the CHANGELOG entries here keep their numbers, so the discussion stays reachable.

## On the shape of this branch

Eleven commits rather than one, deliberately — squashing eleven unrelated fixes would throw away the per-fix rationale, and each message carries the reproduction and the reasoning. Every commit is `cherry-pick -x`'d, so it references the commit it came from.

The CHANGELOG entries keep their original PR numbers, which point at the individual discussions.

## Perf

Suite time is flat. Interleaved A/B, four runs each, against `master`:

| | mean | examples |
| --- | --- | --- |
| master | 2.19s | 2543 |
| this branch | 2.24s | 2598 |

The +2.3% matches the +2.2% examples this branch adds — per-example cost is unchanged.

[#2848](https://github.com/ruby-grape/grape/pull/2848) was dropped from this branch and closed. It made route declaration invalidate the compiled router, which only mattered when mutating an API that had already been compiled — declaring routes after `mount`, after `helpers`, or after reading `.routes` all work on master today. That leaves runtime mutation of a live API, which is neither documented nor thread-safe, and the change cost O(n²) in exactly that case.

## Test plan

- [x] Full RSpec suite passes on the combined branch (2594 examples, 0 failures).
- [x] RuboCop clean across all 338 files.
- [x] All eleven cherry-picks applied without conflict, including the CHANGELOG and UPGRADING additions.
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@
 * [#2827](https://github.com/ruby-grape/grape/pull/2827): Make the `cascade` DSL getter return the configured value (`cascade false` read back as `true`) - [@ericproulx](https://github.com/ericproulx).
 * [#2829](https://github.com/ruby-grape/grape/pull/2829): Fix a cascading route handing over only to the last route registered for the path, making a middle version (3+ mounted versions with a catch-all) answer 406 - [@ericproulx](https://github.com/ericproulx).
 * [#2826](https://github.com/ruby-grape/grape/pull/2826): Fix `api.version` not being set for the root route of a path-versioned API (`GET /v1`) - [@ericproulx](https://github.com/ericproulx).
+* [#2834](https://github.com/ruby-grape/grape/pull/2834): Restore the #2824 fix for cascaded routes leaking `route_info` and path captures, silently reverted by #2829 - [@ericproulx](https://github.com/ericproulx).
+* [#2838](https://github.com/ruby-grape/grape/pull/2838): Reject request params nested in more arrays than the block declares, instead of silently unwrapping them and passing validation, and report `type: Array[JSON]` errors against the element that failed - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 3.3.5 (2026-07-30)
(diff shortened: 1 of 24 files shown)


dblock:
LGTM 🤷‍♂️

PR #2852, opened 2026-08-05
Preserve positional options for named params

### LINKED ISSUE #2851: Replacing extract_options! with *args, **options in #2618 breaks callers that pass options as a positional Hash
### Description

#2618 replaced several `args.extract_options!` call sites with an explicit `**options` (or `**kwargs`) parameter, e.g. `def use(*names, **options)` in `lib/grape/dsl/parameters.rb`. This changes observable behavior for any caller that passes the trailing options as a literal `Hash` instead of using the `**` splat (or bare `key: value` keywords), because Ruby 3+ no longer auto-converts a trailing `Hash` argument into keyword arguments unless it is explicitly double-splatted at the call site.

`extract_options!` and `*args, **options` are not equivalent:

```ruby
def debug1(*args)
  options = args.count > 1 ? args.extract_options! : {}
  puts args.inspect
  puts options.inspect
end

def debug2(*args, **options)
  puts args.inspect
  puts options.inspect
end

debug1(:key, :object, { k: :v })
# args: [:key, :object]
# options: {:k=>:v}

debug2(:key, :object, { k: :v })
# args: [:key, :object, {:k=>:v}]
# options: {}
```

With `extract_options!`, the trailing hash is pulled out of `args` and treated as options regardless of how it was passed. With `*args, **options`, a hash literal (or a hash stored in a variable) that isn't double-splatted stays a positional argument and lands in `args`, while `options` is silently empty.

This affects any existing caller who builds an options hash and passes it positionally, e.g.:

```ruby
params do
  use :pagination, { some: :option }
end
```

instead of:

```ruby
params do
  use :pagination, **{ some: :option }
end
```

The former used to work with `extract_options!` and silently stops working after #2618, since `:pagination` and `{ some: :option }` both end up in `names` and `options` stays `{}`.

### Steps to reproduce

```ruby
require 'grape'

class API < Grape::API
  params do
    use :pagination, { max_per_page: 100 }
  end
  get '/items' do
    params
  end
end
```

### Expected behavior

The options hash passed positionally is extracted and forwarded to the named param block, same as before #2618.

### Actual behavior

The positional hash is treated as another name in `*names` instead of being merged into `**options`, so `Params :{max_per_page: 100} not found!` is raised (or, more generally, the options are silently dropped).

cyphercodes:
## Summary

`Grape::DSL::Parameters#use` still accepts keyword options, but Ruby 3 no longer turns a trailing positional Hash into keywords. After #2618, calls like `use :pagination, { max_per_page: 100 }` treated the Hash as another named params key instead of as options.

This restores the old `extract_options!` behavior for the multi-argument case by extracting a trailing positional Hash when no keyword options were supplied, while preserving single-argument named param lookup.

Fixes #2851.

## Test plan

- [x] Added a regression example for `use :params_group, options`; verified it fails before the fix.
- [x] `bundle exec rspec spec/grape/dsl/parameters_spec.rb` (Ruby 3.4 Docker): 24 examples, 0 failures.
- [x] `bundle exec rubocop lib/grape/dsl/parameters.rb spec/grape/dsl/parameters_spec.rb`: 2 files inspected, no offenses detected.
- [x] `bundle exec rake` (Ruby 3.4 Docker): RuboCop 338 files inspected, no offenses detected; RSpec 2551 examples, 0 failures.
- [ ] CI green.

🤖 AI-assisted with Hermes Agent; changes were reviewed and verified locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@
 * [#2829](https://github.com/ruby-grape/grape/pull/2829): Fix a cascading route handing over only to the last route registered for the path, making a middle version (3+ mounted versions with a catch-all) answer 406 - [@ericproulx](https://github.com/ericproulx).
 * [#2826](https://github.com/ruby-grape/grape/pull/2826): Fix `api.version` not being set for the root route of a path-versioned API (`GET /v1`) - [@ericproulx](https://github.com/ericproulx).
 * [#2842](https://github.com/ruby-grape/grape/pull/2842): Warn at definition time when a `rescue_from` class is already covered by one registered earlier in the same scope, since the later handler never runs - [@ericproulx](https://github.com/ericproulx).
+* [#2852](https://github.com/ruby-grape/grape/pull/2852): Preserve positional options hashes passed to named params - [@cyphercodes](https://github.com/cyphercodes).
 * Your contribution here.
 
 ### 3.3.5 (2026-07-30)
--- a/lib/grape/dsl/parameters.rb
+++ b/lib/grape/dsl/parameters.rb
@@ -54,6 +54,8 @@ module Grape
       #       end
       #     end
       def use(*names, **options)
+        options = names.pop if options.empty? && names.size > 1 && names.last.is_a?(Hash)
+
         named_params = @api.inheritable_setting.named_params || {}
         names.each do |name|
           params_block = named_params.fetch(name) do
--- a/spec/grape/dsl/parameters_spec.rb
+++ b/spec/grape/dsl/parameters_spec.rb
@@ -63,6 +63,13 @@ describe Grape::DSL::Parameters do
       subject.use :params_group, **options
     end
 
+    it 'extracts a positional options hash from named params' do
+      subject.api = Class.new { include Grape::DSL::Settings }.new
+      subject.api.inheritable_setting.add_named_params(named_params)
+      expect(subject).to receive(:instance_exec).with(options).and_yield
+      subject.use :params_group, options
+    end
+
     it 'raises error when non-existent named param is called' do
       subject.api = Class.new { include Grape::DSL::Settings }.new
       expect { subject.use :params_group }.to raise_error('Params :params_group not found!')


OuYangJinTing:
Thanks for the quick turnaround on this @cyphercodes!

While the `use` fix looks correct for the case described in #2851, the underlying problem from #2618 (`extract_options!` semantics not being equivalent to a literal `*args, **options` / `*names, **options` signature) is broader than just `Grape::DSL::Parameters#use`. The following methods introduced/changed by #2618 have the same pattern and are, as far as I can tell, still affected on `master`:

- `Grape::DSL::Entity#present`
- `Grape::DSL::Parameters#requires`
- `Grape::DSL::Parameters#optional`
- `Grape::DSL::RequestResponse#rescue_from`
- `Grape::DSL::Routing#version`
- `Grape::API.override_all_methods!`
- `Grape::API.replay_step_on`
- `Grape::API.skip_immediate_run?`

So this PR fixes one instance of the symptom, but the current fix does not fully resolve the problem — the same class of bug is reachable through several other entry points.

To recap the case already covered by #2851/this PR:

### Steps to reproduce

```ruby
require 'grape'

class API < Grape::API
  params do
    use :pagination, { max_per_page: 100 }
  end
  get '/items' do
    params
  end
end
```

### Expected behavior

The options hash passed positionally is extracted and forwarded to the named param block, same as before #2618.

### Actual behavior

The positional hash is treated as another name in `*names` instead of being merged into `**options`, so `Params :{max_per_page: 100} not found!` is raised (or, more generally, the options are silently dropped).

Here's a case this PR does **not** fix, on `present`:

### Steps to reproduce

```ruby
require 'grape'

class API < Grape::API
  get 'version' do
    present version: '1.0.0'
  end
end
```

### Expected behavior

The response body should be `{ version: '1.0.0' }`.

### Actual behavior

The response body is `null`.

`present(*args, root: nil, with: nil, **options)` binds the bare `version: '1.0.0'` keyword to `**options` instead of leaving it as the positional Hash object to represent (which is what happened before #2618, when `present(*args)` had no explicit keyword parameters and Ruby collapsed the trailing keywords into a Hash appended to `args`). Since `args` ends up empty, `object` is `nil`, and the response body becomes `null`.

Given how many call sites share this pattern, it might be worth addressing this class of bug more systematically (e.g. reverting to `extract_options!`-style handling, or auditing/fixing each affected method individually) rather than patching them one at a time as they're reported.

dblock:
> Given how many call sites share this pattern, it might be worth addressing this class of bug more systematically (e.g. reverting to `extract_options!`-style handling, or auditing/fixing each affected method individually) rather than patching them one at a time as they're reported.

@ericproulx

ericproulx:
@cyphercodes thank you for creating a PR. This a special case where the method definition contains a splat and a double splat operator. In the case of `use`, defining `max_per_page` without the brackets should solve the issue like the following example:

```ruby
require 'grape'

class API < Grape::API
  params do
    use :pagination, max_per_page: 100
  end
  get '/items' do
    params
  end
end
```

I will add a `deprecation` when the last argument of the array is a `hash` so that users can fix the issue properly.

For the `present` method, since `version` is your body and not an option, you should add the `brackets` so that ruby does pass it into `options` but `args`. The following example should work:

```ruby
require 'grape'

class API < Grape::API
  get 'version' do
    present { version: '1.0.0' }
  end
end
```

PR #2860, opened 2026-08-27
Read Route#default from the description, not from Hash#default

ericproulx:
A route's `desc` keys are delegated to its options Hash, an `ActiveSupport::OrderedOptions`, whose `method_missing` answers an unknown name with that key's value. `default` is the one name in `Grape::Util::ApiDescription::DSL_METHODS` that is *not* unknown to it — it is `Hash#default`, the Hash's own default value — so delegating it reported `nil` for every route, whatever the description documented.

```ruby
route = Grape::Router::BaseRoute.new(pattern, { default: { code: 400 } })
route.default          # => nil                (Hash#default)
route.options[:default] # => { code: 400 }
```

`delegate_missing_to :@options` wouldn't have caught it either: `default` isn't missing on the Hash, so the fallback lands in the same place.

`default` is the only collision — all 17 delegated names were checked against `ActiveSupport::OrderedOptions.method_defined?` / `private_method_defined?`. It's dropped from the delegated list and read explicitly, alongside the `success` and `failure` readers that exist for the same reason. Subtracting it from the list rather than letting an explicit `def default` shadow the generated delegator keeps the suite's `--warnings` run free of a method-redefined warning.


ericproulx:
Closing in favour of a fresh branch; the original commit and PR body carried a link that should not have been on a public repo.

PR #2861, opened 2026-08-27
Add `default_response`, the name grape-swagger reads, and fix `Route#default`

ericproulx:
Two things, and the second is the one that matters to users.

### Adds `default_response`

grape-swagger asks a route for its default response as `route.default_response` ([`GrapeSwagger::Endpoint#default_code_from_route`](https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/endpoint.rb)). Grape never defined that name — the `desc` DSL wrote the key as `default` — so the call fell through `BaseRoute`'s options bag and returned `nil`. **A default response declared exactly the way grape's README documents it never reached the generated OpenAPI document:**

```ruby
desc('Delete a widget') { default code: 'default', message: 'boom' }
# route.options.keys     => [:description, :default]
# route.default_response => nil      # what grape-swagger asks for
```

`default_response` becomes the name on both sides — the `desc` DSL and the route reader — with `default` kept as a deprecated alias:

| surface | new | old (warns) |
| --- | --- | --- |
| `desc` block | `default_response code: 400` | `default code: 400` |
| `desc` keyword | `desc 'x', default_response: {…}` | `desc 'x', default: {…}` |
| route reader | `route.default_response` | `route.default` |

The deprecation shim lives at the write boundary: `desc` warns on a `default:` keyword and stores it as `:default_response`, so only one key ever reaches a route and every reader stays a plain key read. Unlike `success`/`failure`, which support two spellings permanently, this is one name replacing another.

### Fixes `default`

The `default` reader grape did define was broken anyway, which is why nothing could have been documented through it either. Every `desc` key is delegated to an `ActiveSupport::OrderedOptions`, whose `method_missing` answers an unknown name with that key's value — but `default` is not unknown to it. It is `Hash#default`, the Hash's own default value:

```ruby
route = Grape::Router::BaseRoute.new(pattern, { default: { code: 400 } })
route.default           # => nil               (Hash#default)
route.options[:default] # => { code: 400 }
```

`delegate_missing_to :@options` wouldn't have caught it either: `default` isn't missing on the Hash, so the fallback lands in the same place. `default` is the only such collision — all 17 delegated names were checked against `ActiveSupport::OrderedOptions.method_defined?` / `private_method_defined?` — and it is now the one reader written out by hand rather than generated as a delegator. Every other `desc` name, `default_response` included, is delegated as before.

### Notes

- Includes #2862, merged into this branch.
- The `desc 'x', default:` keyword deprecation is the one piece beyond a rename of the method: without it that spelling would migrate with no notice. Easy to drop if you'd rather deprecate only the method calls.
- Replaces #2860.

- [x] Full RSpec suite (2624), plus the grape_swagger (2631) and grape_entity (2647) gemfiles.
- [x] New integration example asserts a `default_response` reaches the generated OpenAPI document as the `default` response — it did not before.
- [x] RuboCop clean; no method-redefined warnings under the suite's `--warnings`.
- [ ] CI green.


lib/grape/dsl/desc.rb
```diff
@@ -56,6 +59,12 @@ def desc(description, *legacy_options, **options, &config_block)
           options = legacy_options.first.merge(options)
         end
 
+        if options.key?(:default)
+          Grape.deprecator.warn('The `default` option of `desc` is deprecated. Use `default_response` instead.')
+          legacy_default = options.delete(:default)
```
dblock: Does this modify caller's `options`, dup first.

dblock (commented):
Needs UPGRADING anyway.

lib/grape/dsl/desc.rb
ericproulx: It doesn't — `**options` collects into a fresh Hash even when the caller splats one in (`desc 'x', **h`), and the positional path reassigns `options` to `merge`'s result, so neither ever aliases the caller's Hash:

```ruby
h = { default: { code: 400 } }
desc('x', **h)   # h unchanged
desc('x', h)     # h unchanged; options.equal?(h) => false
```

But it reads like it might, so rather than a `dup` that implies a hazard that isn't there, I rebuilt it instead of mutating in place:

```ruby
options = { default_response: options[:default] }.merge(options.except(:default))
```

An explicit `default_response:` still wins, since the merged Hash keeps its key. Pinned with specs for both the keyword and the positional form asserting the caller's Hash is unchanged.

ericproulx:
Added an UPGRADING entry — `#### The desc default key is renamed to default_response`, at the top of the 4.0.0 section. It covers the deprecated and preferred forms side by side for all three surfaces (`desc` block, `desc` keyword, route reader), and the one thing that actually moves: a `default:` keyword is now stored under `:default_response`, so code reading `route.options[:default]` directly has to follow the key. CHANGELOG line carries `(see UPGRADING)`.

The `options` question is answered inline — no caller mutation, but rebuilt so it doesn't read like there might be.

Pushed as 4dca0da3. Full suite 2624, grape_swagger gemfile 2631, RuboCop clean.

PR #2865, opened 2026-08-29
Pass `required` as an argument instead of writing a presence key

ericproulx:
## The problem

`requires` announced that a parameter is required by writing into the options Hash it had just been handed:

```ruby
opts[:presence] = { value: true, message: opts[:message] }
```

That puts an internal signal in the same namespace as an API's own options, and the two share a key. Which one wins depends on which method you called, and neither outcome is documented:

```ruby
requires :a, presence: false   # ignored   — :a is still required (requires overwrites it)
optional :a, presence: true    # honoured  — :a is actually required (nothing overwrites it)
```

So `presence:` is silently discarded where someone might plausibly mean it, and silently effective as an undocumented way to turn `optional` into `requires`.

## The fix

The flag travels as an argument — `required: true` from `requires`, `false` from `optional` — through `validate_attributes` to `validates`, which builds the presence entry itself. The DSL stops writing into the caller's Hash.

```ruby
validate_attributes(attrs, opts, required: true, &block)
```

`required?` already existed on `ValidationsSpec`; it just derived from the smuggled key. This makes the existing concept explicit rather than inventing one.

## A second bug, fixed on the way

The presence entry was assembled from `opts[:message]` *before* the enclosing `with` attributes were merged in, so a group-level `message:` never reached the presence validator:

```ruby
params do
  with(message: 'custom missing') { requires :a }
end

# before => {"error":"a is missing"}
# now    => {"error":"a custom missing"}
```

Same shape as the pre-merge read fixed in #2864, in a different key.

## Nothing changes behaviourally

A `presence:` supplied by an API still decides the outcome exactly as it used to — deprecated rather than dropped, so this is a full deprecation cycle rather than a break:

```
requires :a, presence: false   => 400  (unchanged)  + deprecation warning
optional :a, presence: true    => 400  (unchanged)  + deprecation warning
optional :a                    => 200  (unchanged)
```

The removal is one `.except(:presence)` line in a future major.

## A note on where the flag lives

I first put `required:` on `ValidationsSpec.from`, which broke 19 specs: `from(type: Integer)` is a brace-less Hash, and Ruby 3 passes that as keywords, so the positional argument got nothing. The same trap is why `validates(attrs, validator => {...})` needed explicit braces in this diff. A method whose Hash argument is normally written brace-less is a poor place for a new keyword, so the flag lives on `validates` and `ValidationsSpec` is untouched.

## Test plan

- [x] Full RSpec suite: 2616 examples, 0 failures.
- [x] Specs added: both spellings warn; `optional :a, presence: true` still answers 400 with deprecations silenced (pinning that the cycle preserves behaviour); `optional` signals `required: false` (nothing covered that before).
- [x] RuboCop clean over `lib` and `spec`.
- [x] UPGRADING entries for the deprecation and for the `with(message:)` change.
- [ ] CI green.

Independent of #2864 — the two apply cleanly on top of each other in either order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@
 * [#2859](https://github.com/ruby-grape/grape/pull/2859): Answer `Grape::Router::Route#success` and `#failure` from the description whichever way it was written - [@ericproulx](https://github.com/ericproulx).
 * [#2861](https://github.com/ruby-grape/grape/pull/2861): Read `Grape::Router::Route#default` from the description instead of from the options Hash's own default value, which was always `nil` - [@ericproulx](https://github.com/ericproulx).
 * [#2864](https://github.com/ruby-grape/grape/pull/2864): Apply a group's type check to `optional` the same way as `requires`, so a type supplied by an enclosing `with` is no longer invisible to it (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
+* [#2865](https://github.com/ruby-grape/grape/pull/2865): Pass `required` from `requires`/`optional` as an argument instead of writing a `presence` key into the caller's options, deprecating a user-supplied `presence` and letting a `with` block's `message` reach the presence validator (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 3.3.5 (2026-07-30)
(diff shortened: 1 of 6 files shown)


ericproulx:
Done through #2867

PR #2869, opened 2026-08-29
Add Grape::PrecompiledJson for bodies that are already JSON

ericproulx:
## Summary

The JSON formatters encode whatever the endpoint returned, so a String that already holds JSON is encoded a second time — `'{"a":1}'` comes back as `"\"{\\\"a\\\":1}\""`. This comes up whenever the JSON was rendered somewhere else: a blob cached in Redis, a `json_agg` column read straight from Postgres, output from another serializer.

`Grape::PrecompiledJson` marks one body as already rendered:

```ruby
class API < Grape::API
  format :json

  get '/cached' do
    body Grape::PrecompiledJson.new(Rails.cache.read('payload')) # => '{"id":1}', served as-is
  end
end
```

`Formatter::Json` and `Formatter::SerializableHash` return it verbatim; everything unwrapped encodes exactly as before.

An `Array` is joined without its members being parsed, so a cached collection is spliced together rather than round-tripped. `Array#join` calls `to_s`, so members may themselves be wrappers.

```ruby
Grape::PrecompiledJson.new(['{"id":1}', '{"id":2}']).to_s # => '[{"id":1},{"id":2}]'
```

## Why a wrapper rather than a passthrough formatter

The opt-in is per response rather than per API on purpose. A formatter that passed every String through cannot tell pre-rendered JSON from an endpoint that returns `'hello'`, and would answer with a bare `hello` — invalid JSON, silently. No type check catches that, since a String is precisely what such a formatter is looking for. A marker never has to guess.

The `Array` join is load-bearing rather than a convenience. Handing an encoder an array of wrappers renders them as ordinary objects:

```ruby
[w, w].to_json                     # => [{"value":"..."},{"value":"..."}]
ActiveSupport::JSON.encode(a: w)   # => {"a":{"value":"{\"a\":1}"}}
```

So the join has to happen in the wrapper, and for the same reason a wrapper is only valid as the whole body — nesting one inside a Hash is documented as unsupported, because no encoder knows to unwrap it.

A value that is neither a String nor an Array raises `Grape::Exceptions::InvalidFormatter`, which `Middleware::Formatter#build_formatted_response` already rescues into a 500, rather than letting a body Rack cannot serve reach the SPEC check. Errors are untouched — they are rendered by the error formatter, so `error!` still returns properly encoded JSON.

## Naming

`Grape::PrecompiledJson` rather than `Grape::Json::Precompiled`: `Grape::Json` is stdlib `::JSON` itself unless multi_json is loaded, so the nested name would define `JSON::Precompiled` in the stdlib namespace, and only in some bundles.

## Prior art

GitLab has carried this for years as `Gitlab::Json::Precompiled` plus a custom `Gitlab::Json::GrapeFormatter` installed with `formatter :json, ...` in `lib/api/api.rb`. Honouring the wrapper in the built-in formatters means no formatter declaration is needed.

## Backward compatibility

Purely additive — a new guard for a new class in front of the existing formatter bodies. Nothing that does not use the wrapper behaves differently, so no UPGRADING entry.

## Test plan

- [x] Full RSpec suite passes locally (2629 examples, 0 failures).
- [x] RuboCop clean (340 files, no offenses).
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
 * [#2849](https://github.com/ruby-grape/grape/pull/2849): Remove `http_digest`, which has had no strategy behind it since 2.0.0 and raised on the first request rather than when the API was defined (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
 * [#2862](https://github.com/ruby-grape/grape/pull/2862): Rename the `desc` `default` key to `default_response`, the name grape-swagger reads, deprecating `default` (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
 * [#2866](https://github.com/ruby-grape/grape/pull/2866): Take `as` as a keyword argument in `requires` and `optional` instead of carrying it in the options Hash, where it was dispatched as a no-op `AsValidator` - [@ericproulx](https://github.com/ericproulx).
+* [#2869](https://github.com/ruby-grape/grape/pull/2869): Add `Grape::PrecompiledJson`, a body wrapper the JSON formatters serve verbatim instead of encoding a second time - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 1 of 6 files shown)


dblock (approved):
I like it.

PR #2871, opened 2026-08-29
Keep `refresh_already_mounted` out of `mount`'s public signature

ericproulx:
## The option

`mount` accepts an undocumented `refresh_already_mounted` option that replaces any endpoint already mounted for the same base API instead of adding a second one.

**The behaviour is essential.** `Grape::API.refresh_mount_step` replays every earlier mount whenever a class-level method runs after one, so the mounted API picks up the new setting. Replaying a mount naively appends a second endpoint, so without the replacement a single `format :json` after a mount does this:

```
with the replacement:     endpoints=1  routes=["/thing/x(.json)"]
without it:               endpoints=3  routes=["/thing/x(.:format)", "/thing/x(.:format)",
                                               "/thing/x(.json)",    "/thing/x(.json)"]
```

**The option is not.** Its only legitimate caller is `refresh_mounted_api`, which is private, is not even reachable from an API class (`NoMethodError` — `refresh_mount_step` reaches it on the instance via `__send__`), and always passes `true`. So a constant belonging to a private method was threaded out through a public one, where any user could reach it and no documentation mentioned it.

## The change

The replacement moves into `refresh_mounted_api`. That works because a mounted Grape API's throwaway instance and the class that was written share a base — the same signal `same_mounted_app?` already keys on:

```
stored mounted_app.base  → #<Class:0x…ecf8>
original app.base        → #<Class:0x…ecf8>   equal? true
```

Passing `refresh_already_mounted` to `mount` still works and now warns. There is no replacement to point users at, because an API has no reason to ask for it — mounting the same app twice on purpose is still two endpoints.

## Two things that fall out

Once an internal flag is no longer threaded through the signature, two oddities in it stop paying for themselves.

**`respond_to?(:each_pair)` → `is_a?(Hash)`.** A `Struct` and an `OpenStruct` answer `each_pair` without being able to express an app => path mapping — their keys are member names. So a Rack app that happened to be one was read as a mapping:

```ruby
RackStruct = Struct.new(:name) do
  def call(env) = [200, { 'content-type' => 'text/plain' }, ['from struct']]
end

mount RackStruct.new('x')

# before => routes ["/x/?*path(.:format)"]        the field value became the path
# after  => 200 "from struct", routes ["/?*path(.:format)"]
```

Every Hash-like people actually pass — `HashWithIndifferentAccess`, `Hashie::Mash`, `ActiveSupport::OrderedOptions` — subclasses `Hash`. The `respond_to?` dates to `b5df2f84`, the original Rack-mounting commit. The normalisation is now shared with the endpoint replacement instead of written twice.

**`*opts` → `opts = {}`.** The splat could only ever hold zero or one element, so every read spelled `opts.any? ? opts.first[:key] : default`. An optional positional Hash accepts exactly the same call forms:

| call | `*opts` | `opts = {}` |
| --- | --- | --- |
| `mount App` | `mounts=App opts=[]` | `mounts=App opts={}` |
| `mount App => path` | `mounts={…} opts=[]` | `mounts={…} opts={}` |
| `mount({…}, with: …)` | `opts=[{with: …}]` | `opts={with: …}` |
| `mount({…}, {with: …})` | `opts=[{with: …}]` | `opts={with: …}` |

It stays **positional** deliberately. `mount` must declare no keyword parameters: Ruby routes the brace-less Hash of `mount App => '/path'` to keywords when any are declared, and the mount itself goes missing (`wrong number of arguments (given 0, expected 1)`). That also rules out `with:` as a keyword, and rules out `**mounts`, which additionally breaks `mount App` and `mount({App => path}, with: …)`.

## Specs

The two existing specs asserted the flag rather than the feature — both called `mount(…, refresh_already_mounted: true)` directly, which no longer exercises the real path. Rewritten to go through it: mount, then a class-level method, which is what triggers `refresh_mount_step`. Plus two for the deprecation itself.

One detail worth noting: the deprecation fired twice per mount at first, because `mount` recurses for a Grape API and the recursion re-entered with the same options. The flag is dropped before recursing.

## Test plan

- [x] Full RSpec suite: 2615 examples, 0 failures.
- [x] RuboCop clean over `lib` and `spec`.
- [x] All four `mount` call forms verified end to end, including that `with:` still reaches `configuration` in both the keyword and explicitly-braced spellings.
- [x] UPGRADING entries for the deprecation and for the `Struct`/`OpenStruct` change.
- [ ] CI green.

Independent of #2870.

## Also here: the CHANGELOG line #2867 never got (`e2daea11`)

Unrelated to the rest of this PR, riding along because it is one line. #2867 was merged as a squash of a branch that also carried #2865's commit, so its merge shipped that PR's `presence` deprecation as well as its own refactor. The CHANGELOG ended up crediting #2865, which is still open, and saying nothing about #2867, which is what actually landed. The #2865 line stays — it describes behaviour that is in master and links to a real PR — and this adds the missing one for #2867.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -69,6 +69,30 @@ end
 ```
 
 An API that sets `message:` on a `with` block and relies on the generic wording for missing parameters will see its own message instead.
+#### `refresh_already_mounted` is no longer a `mount` option
+
+`mount` accepted an undocumented `refresh_already_mounted` option that replaced any endpoint already mounted for the same base API instead of adding a second one. It exists for Grape's own re-mounting machinery: when a class-level method runs after a `mount`, `Grape::API.refresh_mount_step` replays that mount, and replaying it naively would duplicate the endpoint.
+
+That decision belongs to the replay, not to the call, so the endpoint replacement moved into the private `refresh_mounted_api` which is the only caller that ever wanted it. Nothing changes for the re-mounting itself.
+
+Passing the option to `mount` still works and now warns; it will be ignored in a future release. There is no replacement, because an API has no reason to ask for it — mounting the same app twice on purpose is still two endpoints.
+
+#### A `mount` target is recognised as a mapping only when it is a `Hash`
+
+`mount` decided between "a Hash of app => path" and "a bare app mounted at /" with `mounts.respond_to?(:each_pair)`. A `Struct` and an `OpenStruct` answer that too, and neither can express an app => path mapping — their keys are member names — so a Rack app that happened to be one was read as a mapping and silently mounted its own field value as a path:
+
+```ruby
+RackStruct = Struct.new(:name) do
+  def call(env) = [200, {}, ['ok']]
+end
+
+mount RackStruct.new('x')
+
+# before: mounted at "/x", because :name => 'x' was read as app => path
+# now:    mounted at "/", as the Rack app it is
+```
+
+The test is now `mounts.is_a?(Hash)`. Every Hash-like people pass — `HashWithIndifferentAccess`, `Hashie::Mash`, `ActiveSupport::OrderedOptions` — subclasses `Hash` and is unaffected.
 
 #### `use`, `helpers`, `rescue_from` and other registrations no longer reach routes defined above them
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -69,6 +69,7 @@ end
 ```
 
 An API that sets `message:` on a `with` block and relies on the generic wording for missing parameters will see its own message instead.
+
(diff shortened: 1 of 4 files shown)


UPGRADING.md
```diff
@@ -69,6 +69,30 @@ end
 ```
 
 An API that sets `message:` on a `with` block and relies on the generic wording for missing parameters will see its own message instead.
+#### `refresh_already_mounted` is no longer a `mount` option
```
dblock, line 73: Add a line above.

dblock (approved):

PR #2887, opened 2026-09-02
Add support for the HTTP QUERY method (RFC 10008)

ericproulx:
[RFC 10008](https://www.rfc-editor.org/info/rfc10008/) defines `QUERY`: a **safe and idempotent** method whose *content* carries the query, for when a query is too voluminous, too structured, or too sensitive to encode into a request URI. It sits between `GET` (which cannot carry a body) and `POST` (which is neither safe nor idempotent, so it cannot be retried or cached).

Grape had no support for it. `query` was not a verb, and a route declared as `route :query` was answered with 405 while still being listed in `Allow`.

### Changes

* **`Grape::HTTP_SUPPORTED_METHODS` gains `QUERY`**, which mints the `query` DSL verb, routes the method, and lists it in the resource's `Allow` header. Rack has no constant for the method yet, so `Grape::QUERY` holds the literal.
* **`Formatter::BODY_CARRYING_METHODS` gains `QUERY`**, so the content is parsed into `params` the way a `POST` body is. Without this the request that *is* a query arrived with `params` empty — the urlencoded case worked only by accident, since Rack's own `form_data?` keys off media type rather than method.
* **A `QUERY` sending content without a `Content-Type` is rejected with 400** rather than read as the API's default format ([Sections 2 and 2.1](https://www.rfc-editor.org/rfc/rfc10008#section-2)). Every other method keeps the fallback.

```ruby
params do
  requires :q, type: String
  optional :limit, type: Integer
end
query '/search' do
  Status.search(params[:q], limit: params[:limit])
end
```

```
curl -X QUERY 'http://localhost:9292/search' -H Content-Type:application/json -d '{"q": "grape", "limit": 10}'
```

### Already correct, now reachable

`default_status` answers 200 rather than 201 for `QUERY`, matching Section 2. The unsupported-media-type (415), unparseable-content (400) and unacceptable-`Accept` (406) paths all behave as Section 2.1 requires — they simply were never reached, because the body was never read.

### Not included

The `Accept-Query` response header ([Section 3](https://www.rfc-editor.org/rfc/rfc10008#section-3)) is optional in the spec ("can be used"), and its Structured Fields serialization plus the question of which responses should carry it are worth their own change.

### Compatibility

Additive. `Allow` is built from the routes an API actually declares, so nothing changes for an API that defines no `QUERY` route. A user-defined `def self.query` on an API subclass still wins over the DSL verb, since the DSL is reached through `method_missing`. grape-swagger reads `route.request_method` from declared routes and never the supported-method list, so its output is unchanged unless a `query` route is declared.

Adding `QUERY` to the existing `%w[POST PATCH PUT DELETE]` loop in the formatter spec runs the whole body-parsing suite against it; the API-level specs cover the verb end to end, and the missing/unsupported content-type rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@
 * [#2882](https://github.com/ruby-grape/grape/pull/2882): Strip mount path and prefix in `Grape::Middleware::Versioner::Path` with `each` instead of `Enumerable#reduce` - [@ericproulx](https://github.com/ericproulx).
 * [#2881](https://github.com/ruby-grape/grape/pull/2881): Decide whether a request carries a body from the env, so `Grape::Middleware::Formatter` no longer builds a `Rack::Request` for GET, HEAD and OPTIONS - [@ericproulx](https://github.com/ericproulx).
 * [#2883](https://github.com/ruby-grape/grape/pull/2883): Resolve a format's content type through a plain Hash carrying both spellings instead of a `HashWithIndifferentAccess` that converts the key on every read - [@ericproulx](https://github.com/ericproulx).
+* [#2887](https://github.com/ruby-grape/grape/pull/2887): Add support for the HTTP `QUERY` method (RFC 10008), including the `query` DSL verb and parsing the request content into `params` - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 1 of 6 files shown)


dblock:
I'm excited to beat Rack and Rails on QUERY support 🎈

PR #2896, opened 2026-09-04
Bring test suite line coverage to 100%

dblock:
### Description

Brings line coverage of the test suite to 100% (up from ~98.3%/93.0%).

Adds specs for previously-untested classes/methods and branches across the middleware, router, serve_stream, util, validations and exceptions namespaces.

Marks two branches that appear structurally unreachable through the public API with a `# simplecov:disable` / `# simplecov:enable` block and an explanatory comment (see inline PR comment tagging @ericproulx to confirm before anything is actually removed):
- `Grape::Middleware::Error#format_message`'s 406 "format not supported" throw, since `Grape::ErrorFormatter.formatter_for` always falls back to a registered `Txt` formatter (`Grape::ErrorFormatter::Base.inherited` registers every formatter subclass, including `Txt`), so `formatter` can never be falsy.
- `Grape::Router::Route#tag_utf8!`'s `else` branch, since Mustermann's `Pattern#params` only ever returns `String`/`Array` captures, and the only caller filters out `nil` first.

Branch coverage improved from 92.97% to 93.94% along the way, but was not brought to 100% as part of this PR given the scope of the remaining work.

### Verification

- [x] `bundle exec rspec` - full suite, 0 failures
- [x] `bundle exec rubocop` - clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@
 * [#2894](https://github.com/ruby-grape/grape/pull/2894): Read the request method once in `default_status` instead of asking through `post?` and `delete?` - [@ericproulx](https://github.com/ericproulx).
 * [#2895](https://github.com/ruby-grape/grape/pull/2895): Cut per-request work out of the endpoint, validation and router paths: scan the router's compiled union by capture number rather than by name, skip the Array boxing in `AttributesIterator` for a flat scope, read a coerced attribute once instead of three times, and resolve the formatter's config and the coercers' type checks once at build time - [@ericproulx](https://github.com/ericproulx).
 * [#2900](https://github.com/ruby-grape/grape/pull/2900): Remove the deprecations announced in 3.2 and 3.3: `Grape::Router.normalize_path`, Hash access on middleware `Options` and their `DEFAULT_OPTIONS` constants, the positional options Hash for `auth`/`http_basic`/`desc`, a Hash returned from a `rescue_from` handler, and `@option` on validators (see UPGRADING) - [@ericproulx](https://github.com/ericproulx).
+* [#2896](https://github.com/ruby-grape/grape/pull/2896): Bring test suite line coverage to 100% - [@dblock](https://github.com/dblock).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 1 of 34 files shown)


lib/grape/middleware/error.rb
```diff
@@ -83,14 +83,7 @@ def html_content_type?(content_type)
...
-        return formatter.call(error:, env:, include_backtrace:, include_original_exception:) if formatter
-
-        throw :error, Grape::Exceptions::ErrorResponse.new(
-          status: 406,
-          message: "The requested format '#{current_format}' is not supported.",
-          backtrace: error.backtrace,
-          original_exception: error.original_exception
-        )
+        formatter.call(error:, env:, include_backtrace:, include_original_exception:)
       end
```
dblock, line 93: @ericproulx I removed this branch as dead code: `Grape::ErrorFormatter.formatter_for` always falls back to `Txt` (registered via `ErrorFormatter::Base.inherited`), so `formatter` here can never be falsy and this 406 throw could never execute. Confirming this was unintentional dead code and not a bug elsewhere (e.g. in `formatter_for`'s fallback) that should instead make this branch reachable again. Happy to revert/adjust if the intent was different.

dblock (commented):
Flagging one change for @ericproulx to confirm intent (rest of the PR is coverage-only spec additions).

lib/grape/middleware/error.rb
ericproulx: oh thanks, I'll look closer for unreachable code.

ericproulx: #2901

lib/grape/router/route.rb
```diff
@@ -97,7 +97,14 @@ def tag_utf8!(value)
         when Array
           value.each { |v| tag_utf8!(v) }
         else
+          # simplecov:disable
+          # Unreachable: Mustermann's Pattern#params returns only String or
```
ericproulx: https://github.com/ruby-grape/grape/pull/2904

ericproulx: It's reachable apparently

ericproulx (approved):

ericproulx:
Committed through #2897

PR #2909, opened 2026-09-05
Standardize bare Ruby exception messages to lowercase, unpunctuated

dblock:
Lowercases and removes trailing periods from the 7 `ArgumentError` messages in `lib/` that deviated from the rest of the codebase's lowercase, unpunctuated style (`dsl/entity.rb`, `dsl/inside_route.rb`, `dsl/validations.rb`, `validations/types/dry_type_coercer.rb`), and documents the convention in CONTRIBUTING.md.

This matches Ruby's own core/stdlib exception style (e.g. `TypeError: no implicit conversion from nil to integer`), which reads naturally when printed after the exception class name and colon in a backtrace, and is consistent with how `Grape::Exceptions::*` messages are already written.

Updated the corresponding specs that asserted the old exact message text.

To help enforce this convention going forward, I created a new RuboCop plugin gem, [rubocop-exception_messages](https://github.com/dblock/rubocop-exception_messages), with `ExceptionMessages/Casing` and `ExceptionMessages/Punctuation` cops that flag/autocorrect exception messages violating this style. Added as a dev dependency here so CI catches any regressions.

Bumped to [0.2.0](https://github.com/dblock/rubocop-exception_messages/blob/master/CHANGELOG.md#020-2026090), which adds `ExceptionMessages/QuoteStyle`, enabled by default, wrapping interpolated values in backticks (e.g. `` "unknown type: `#{type}`" ``) for consistency across error messages. Autocorrected the 16 affected messages across `lib/` and `spec/`, and updated specs asserting exact message text accordingly.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,6 +8,7 @@ AllCops:
     - benchmark/**/*
 
 plugins:
+  - rubocop-exception_messages
   - rubocop-performance
   - rubocop-rspec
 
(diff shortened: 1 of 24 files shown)


ericproulx:
@dblock I'll make a release after this PR.

dblock:
@ericproulx ready

ericproulx (approved):

PR #2917, opened 2026-09-08
Read path captures out of the router's union match

ericproulx:
The router runs its compiled union to pick a route, then throws the `MatchData` away — and `Route#params_for` runs the route's *own* decode-aware regexp a second time to pull out the same captures. That second match is **1.01 µs of an 8.76 µs bare request (11.5%)**.

The union embeds each route's pattern verbatim, so those captures are already groups of the match that just ran. `compile!` now records, per route, the number each of its captures carries in the union, and `params_for` reads them straight off the match.

### Why the group numbers line up

The `(?<_N>…)` wrapper `to_regexp` adds is the group immediately before the pattern's own, and `Regexp.union` shifts numbering only by the groups preceding a branch — which is exactly what `regexp_capture_group` already counts. So the capture at position `n` of a route's own regexp is group `regexp_capture_group + n` of the union. On a three-route path-versioned API that is wrappers `1, 3, 5` and `version => [2, 4, 6]`.

### When it applies

Eligibility is settled once, at compile time, by two checks:

- every name occupies a **single position** in the route's own regexp — a multi-position name comes back from Mustermann as an Array, which one group number cannot stand for;
- `identity_params?` probed with empty strings. It is public on `Mustermann::AST::Pattern`, and its AST override folds in both `param_converters.empty?` and the `splat`/`captures` names Mustermann always answers with an Array. Probing with `''` leaves exactly the checks that hold for every request.

The one genuinely per-request condition is unescaping: `map_param` is `unescape(value, true)`, which returns the value untouched unless it contains `'%'`. No capture can contain `%` unless the path does, so the fast path is gated on `input.include?('%')` and anything else falls through to the full match — as do routes reached through `#rotation` or the ANY map (no `MatchData` to hand over) and `GreedyRoute`, whose `params_for` is `nil` anyway.

So `/dup/:id/x/:id` still yields `{"id":["1","2"]}`, `/star/*` still yields `{"splat":["x/y"]}`, and `/items/a%2Eb` still yields `a.b`. A *named* splat `*path` does take the fast path — `always_array?` lists only `splat` and `captures`.

### Two smaller allocations on the same path

- **`Instance#call`** destructured the router's tuple only to build an identical Array back up. Cascading is the default, so it now hands that tuple back. On the `cascade false` path the header copy is skipped when there is no `X-Cascade` to strip — but only for a `Grape::Util::Header`, which answers `key?` the way the copy's `delete` would. A mounted Rack app's plain Hash can spell the name in any case (it is matched case-sensitively there, while the copy folds case), so it still goes through the copy. `spec/grape/api/instance_spec.rb:132` pins this.
- **`Request#routing_args_as_params`** called `except` on routing args holding nothing but Grape's own keys, building an empty Hash for `make_params` to find blank and discard.

### Numbers

Ruby 4.0.6, no YJIT, median of 9 interleaved rounds against a worktree snapshot of `master`. Noise floor ±1.6%, established by A/B-ing identical code; the runner alternates which lib goes first, because running second gave a reproducible ~1% edge.

| scenario | delta |
| --- | ---: |
| bare GET (path-versioned) | **+16.7%** |
| POST | +6.7% |
| GET with params + validations | +4.7% |
| 400 validation error | +2.0% |
| unversioned static GET | +0.5% |

Allocations for a bare request: **23 → 18 objects**.

By route count, hitting the last route: +9.9% at 10 routes, +2.2% at 50, ~+1% at 200–500. The saving is constant per request, so it shrinks as a fraction once the union match itself dominates — that one is still the `uri_decode` expansion, untouched here.

These are laptop numbers on a single machine; worth a second opinion on other hardware.

### Notes for review

`transaction` runs the union match itself rather than going through `#match?`. The first cut passed a block through `#match?` and yielded, which cost the no-capture path ~1.6% (`unversioned` went to −1.6%); running it inline brings that back to +0.5%. `#match?` is otherwise unchanged, for the ANY and greedy callers. Inlining pushed `transaction` past `Metrics/CyclomaticComplexity` (17/15), so the ANY/greedy tail moved to `#neighbours`.

The three changes are independent and can be split if you'd rather review them separately.

### Verification

- Full suite 2847 examples, 0 failures; RuboCop clean over 362 files.
- Byte-identical responses (status, headers, body) across a 51-case matrix against a `git archive HEAD lib` snapshot: percent-encoded paths, dots in captures and `.format` suffixes, named and anonymous splats, duplicate capture names, regexp requirements, `anchor: false`, namespaces, path/header versioning, version cascading between mounted APIs, mounted plain Rack apps with and without `cascade false`, auto-OPTIONS, 405, HEAD and 404.

### DIFF
--- a/lib/grape/api/instance.rb
+++ b/lib/grape/api/instance.rb
@@ -116,18 +116,22 @@ module Grape
 
       # Handle a request. See Rack documentation for what `env` is.
       def call(env)
-        status, headers, response = @router.call(env)
-        unless @cascade
-          # +merge!+, not +merge+: the latter is a `dup` plus a `merge!`, so the
-          # Header built on this line would be allocated only to be discarded.
-          # The copy stays because +headers+ can come from a mounted Rack app,
-          # which is free to hand back a frozen or shared Hash that the delete
-          # below must not reach into.
-          headers = Grape::Util::Header.new.merge!(headers)
-          headers.delete('X-Cascade')
-        end
+        response = @router.call(env)
+        return response if @cascade
 
-        [status, headers, response]
+        headers = response[1]
+        # A Grape::Util::Header folds case the way the copy below does, so it can
+        # answer for it; any other Hash cannot, and still goes through the copy.
+        return response if headers.is_a?(Grape::Util::Header) && !headers.key?('X-Cascade')
+
+        # +merge!+, not +merge+: the latter is a `dup` plus a `merge!`, so the
+        # Header built on this line would be allocated only to be discarded.
+        # The copy stays because +headers+ can come from a mounted Rack app,
+        # which is free to hand back a frozen or shared Hash that the delete
+        # below must not reach into.
+        headers = Grape::Util::Header.new.merge!(headers)
+        headers.delete('X-Cascade')
+        [response[0], headers, response[2]]
       end
 
       # Some requests may return a HTTP 404 error if grape cannot find a matching
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [#2916](https://github.com/ruby-grape/grape/pull/2916): Pin `json` below 3 in the gemfiles whose dependencies cannot use it - [@ericproulx](https://github.com/ericproulx).
 * [#2915](https://github.com/ruby-grape/grape/pull/2915): Update rubocop to 1.90.0 and rubocop-performance to 1.27.0 - [@ericproulx](https://github.com/ericproulx).
 * [#2918](https://github.com/ruby-grape/grape/pull/2918): Skip the dry-types round trip when a value already is the declared type - [@ericproulx](https://github.com/ericproulx).
+* [#2917](https://github.com/ruby-grape/grape/pull/2917): Read path captures out of the router's union match instead of re-running the route's pattern - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 4.0.0 (2026-09-07)
--- a/lib/grape/request.rb
+++ b/lib/grape/request.rb
@@ -136,6 +136,10 @@ module Grape
       ["HTTP_#{header.upcase.tr('-', '_')}", header]
     end.freeze
 
+    # Routing-args keys that are Grape's own, not request params.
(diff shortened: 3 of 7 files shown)


lib/grape/api/instance.rb
```diff
@@ -116,18 +116,22 @@ def initialize
...
+        # A Grape::Util::Header folds case the way the copy below does, so it can
+        # answer for it; any other Hash cannot, and still goes through the copy.
+        return response if headers.is_a?(Grape::Util::Header) && !headers.key?('X-Cascade')
+
+        # +merge!+, not +merge+: the latter is a `dup` plus a `merge!`, so the
+        # Header built on this line would be allocated only to be discarded.
+        # The copy stays because +headers+ can come from a mounted Rack app,
+        # which is free to hand back a frozen or shared Hash that the delete
+        # below must not reach into.
+        headers = Grape::Util::Header.new.merge!(headers)
```
dblock, line 132: How is this faster than `Grape::Util::Header.new(headers)`?

dblock (approved):

lib/grape/api/instance.rb
ericproulx: It isn't faster — `Grape::Util::Header.new(headers)` is silently broken on Rack 3.

`Rack::Headers < Hash`, so the positional argument is consumed as the Hash's *default value*, not its contents:

```
rack 3.2.7, Grape::Util::Header == Rack::Headers

h = { 'content-type' => 'application/json', 'x-cascade' => 'pass' }

Grape::Util::Header.new(h)        # => size 0, {}
Grape::Util::Header.new.merge!(h) # => size 2, {"content-type"=>"application/json", "x-cascade"=>"pass"}
```

It happens to work on Rack 2.2, whose `HeaderHash#initialize` does `hash.each { |k, v| self[k] = v }`, so the failure would only show on the Rack 3.x half of the CI matrix — as a response that loses every header.

On allocations the two forms are equivalent (one Hash each). The form the comment is actually steering away from is `Grape::Util::Header.new.merge(h)`, which costs two, since `merge` is a `dup` plus a `merge!`.

For what it's worth, that line and its comment are pre-existing — this PR only moved them under the new early return.

PR #2919, opened 2026-09-09
Resolve a '%'-free path against a union without the encodings

ericproulx:
Every route pattern is compiled with Mustermann's `uri_decode: true`, which expands each literal character of a path into an alternation with its percent-encodings — `/resource7` compiles to `\/(?:r|%72)(?:e|%65)(?:s|%73)(?:o|%6F|%6f)…` — so that `/a%2Eb` still reaches a route declared as `/a.b`.

The router's per-method union is the disjunction of those patterns, so every request pays for the expansion. It roughly doubles the compiled source, and it hides the literal prefix each branch starts with, which is what would otherwise let Onigmo reject a non-matching path without walking the alternatives. On a 500-route API, failing to match cost 180 µs.

An input holding no `%` cannot take any of those alternatives, so for it the union without them accepts exactly the same strings, by the same path through the regexp, capturing the same substrings. `compile!` now derives that stand-in per union, and `#transaction`, `#match?` and `#greedy_match?` read it whenever the path has no `%`.

### Why source surgery rather than `uri_decode: false`

The stand-in is cut out of the union's compiled source rather than recompiled from each pattern with `uri_decode: false`. That is 20x cheaper — one `gsub` and one `Regexp.new` for the whole union, against a fresh Mustermann parse per route (13 ms vs 169 ms per 500 routes).

It is also the more faithful transformation. `uri_decode: false` drops the `\+` branch that lets a plus match a literal space, so `/with+space` — which carries no `%` anywhere and is therefore resolved by the stand-in — would miss a route the decode-aware union matches. Removing only the `%XX` branches keeps it.

### What the rewrite is allowed to touch

Only alternations are rewritten, and only their `%XX` alternatives are dropped, so a `requirements` Regexp inserted into the pattern verbatim survives whatever it holds. Two shapes are stepped over rather than rewritten:

* a **character class**, where `(?:` and `|` are literal characters and dropping them would shrink the set a route matches;
* an **escape**, so an escaped bracket cannot open a class.

The group itself is only unwrapped around a lone character — a quantifier binds to the group, so `(?:ab|%41)+` must not become `ab+`. Every path literal is a single character, and the bare form is what the prefix optimisation needs.

### Boot

GET routes are mirrored for HEAD and the greedy neighbours mirror both, so the same source is usually compiled several times over; the stand-in is built once per distinct source and shared. Without that, and with the rewrite memoised per distinct expansion rather than recomputed per match, boot for 500 routes grew 84% — with both, 10%.

| Full API instance build | before | after |
| --- | ---: | ---: |
| 50 routes | 7.3 ms | 8.1 ms |
| 200 routes | 29.5 ms | 32.3 ms |
| 500 routes | 77.4 ms | 85.1 ms |
| 1000 routes | 160.1 ms | 173.5 ms |

### Throughput

Ruby 4.0.6 without YJIT, `format :json` and a two-version path prefix.

| | before | after | |
| --- | ---: | ---: | ---: |
| 50 routes, last route | 43.3k i/s | 67.9k i/s | +57% |
| 200 routes, last route | 15.5k i/s | 31.7k i/s | +104% |
| 500 routes, last route | 6.9k i/s | 15.5k i/s | +126% |
| 500 routes, path matching nothing | 5.5k i/s | 433.0k i/s | 78x |
| 500 routes, percent-encoded path | 6.73k i/s | 6.71k i/s | −0.2% |

A path carrying a percent-encoding still goes through the decode-aware union and is unchanged, as is the first route of a union and a single-route API — those pay only the `String#include?`.

Beyond the specs, the two unions were checked for agreement over 2,223 `%`-free inputs against routes covering dots, unicode, splats, optionals, escaped bars, literal spaces, plus signs and `requirements` regexps: same match, same captures, same group contents throughout.

One cost worth naming: an extra compiled union per distinct source is held for the life of the router.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [#2915](https://github.com/ruby-grape/grape/pull/2915): Update rubocop to 1.90.0 and rubocop-performance to 1.27.0 - [@ericproulx](https://github.com/ericproulx).
 * [#2918](https://github.com/ruby-grape/grape/pull/2918): Skip the dry-types round trip when a value already is the declared type - [@ericproulx](https://github.com/ericproulx).
 * [#2917](https://github.com/ruby-grape/grape/pull/2917): Read path captures out of the router's union match instead of re-running the route's pattern - [@ericproulx](https://github.com/ericproulx).
+* [#2919](https://github.com/ruby-grape/grape/pull/2919): Resolve a path carrying no percent-encoding against a union compiled without them - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 4.0.0 (2026-09-07)
--- a/lib/grape/router.rb
+++ b/lib/grape/router.rb
@@ -10,12 +10,23 @@ module Grape
       # so request-time reads never mutate shared state (see #match? / #rotation).
       @map = {}
       @optimized_map = {}
+      # The same unions as @optimized_map / @union, with the percent-encoded
+      # alternatives of their path literals removed (see PlainUnion). Read for
+      # any input that carries no '%', which is nearly all of them.
+      @plain_map = {}
     end
 
     def compile!
       return if @compiled
 
+      # A route registered for GET is mirrored for HEAD, and the greedy
+      # neighbours mirror both, so the same source is usually compiled several
+      # times over. Its stand-in is built once and shared: a Regexp carries no
+      # per-match state, and unions with the same source number their groups
+      # the same way.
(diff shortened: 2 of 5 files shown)


ericproulx:
Closing: going with #2920 instead, which halves the cost of a 404 without rewriting the compiled union. The no-match figure in the description here was also inflated by a benchmark whose routes all shared one literal; with varied route names the gain per union walk was ~4x rather than 78x.

PR #2923, opened 2026-09-10
Read the versioners' version options off instance variables

ericproulx:
## Summary

`Versioner::Base` delegated `cascade`, `parameter`, `strict` and `vendor` through `version_options` into `config`, so each read went two Forwardable frames and two Data readers deep (about 150 ns) for a value fixed when the middleware was built. The header versioner reads `vendor` and `strict` on every request, the accept-version one `strict`, and the param one `parameter` twice.

They are now plain readers over instance variables set in `#initialize`, the pattern the formatter's options already use. The public readers keep their names.

## Benchmarks

| request | delta |
|---|---|
| header-versioned | +3.4% |
| accept-version | +1.7% |
| param-versioned | +2.8% |
| path-versioned (control, reads none of them) | −0.7% (noise) |

The path versioner carries the four extra variables on its per-request `dup`, but stays in the same 160-byte object slot (15 of its 18 embedded slots).

## Test plan

- [x] RuboCop clean.
- [x] Mutation-checked through existing specs: forcing `strict` to false, `vendor` to nil, a wrong `parameter` or an inverted `cascade` fails 10, 36, 18 and 34 specs respectively.
- [ ] CI green.


ericproulx:

PR #2924, opened 2026-09-10
Answer the formatter's common Accept headers from a table built once

ericproulx:
## Summary

- An API that pins no `format` (content negotiation, e.g. `default_format :json`) negotiated one from the Accept header on every request that sent one, through `Rack::Utils.best_q_match`: about 1.7 µs for `Accept: application/json`. The answer depends only on the header and the middleware's `mime_types`.
- `Formatter::FormatForAcceptCache` computes it once for every registered media type, `*/*` and no Accept header, and `format_from_header` looks the header up. Any other header misses and is negotiated as before. Only registered media types are keys, so client input cannot grow a table.
- Shared per `mime_types`, which `Grape::ContentTypes` already shares per content-type registry, since every endpoint builds its own formatter. An API that pins a `format` never negotiates from the header and builds no table.
- The negotiation moved into the class method `Formatter.format_for_accept` (`@api private`) so the cache and the miss path run the same function. The keys are valid strings that scrubbing leaves alone, so only a miss scrubs the header.

## Benchmarks

Median of 7 interleaved subprocess rounds against master, Ruby 4.0.6:

| request, API with no pinned format | no JIT | YJIT |
|---|---|---|
| `Accept: application/json` | **+40.3%** | **+54.1%** |
| `Accept: */*` | +5.3% | |
| pinned-format GET (control) | +0.2% (noise) | |

Boot, 500 negotiating endpoints: retained memory and compile time unchanged.

## Behaviour

Byte-identical to master over a 525-case matrix: 7 APIs (defaults, `default_format`, a custom `text/csv` type, a mixed-case registered type, a registered `application/*`, a parameterized `application/json; charset=utf-8`, a pinned format) × 25 Accept values (a browser Accept, axios's default, q-value lists, other casings, invalid bytes, binary-encoded headers) × 3 paths (plain, `.json` extension, `?format=xml`).

## Test plan

- [x] RuboCop clean.
- [x] Mutation-checked through existing specs: a table answering `:txt` for every key fails 7; skipping the scrub on a miss fails 1.
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#2918](https://github.com/ruby-grape/grape/pull/2918): Skip the dry-types round trip when a value already is the declared type - [@ericproulx](https://github.com/ericproulx).
 * [#2917](https://github.com/ruby-grape/grape/pull/2917): Read path captures out of the router's union match instead of re-running the route's pattern - [@ericproulx](https://github.com/ericproulx).
 * [#2921](https://github.com/ruby-grape/grape/pull/2921): Pin the router's request-time isolation regressions through requests instead of its instance variables - [@ericproulx](https://github.com/ericproulx).
+* [#2924](https://github.com/ruby-grape/grape/pull/2924): Answer the formatter's common Accept headers from a table built once instead of negotiating them per request - [@ericproulx](https://github.com/ericproulx).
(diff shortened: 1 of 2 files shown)


ericproulx:

PR #2925, opened 2026-09-10
Skip the qualifying-params lookup for scopes other than `given`

ericproulx:
## Summary

A nested scope resolves its params through `@parent.qualifying_params.presence`, which reads the request's `ParamScopeTracker` out of fiber storage and looks the parent up in its identity Hash. That runs twice per validator on a nested scope (once in `should_validate?`, once in the attributes iterator) on every request.

Only `#meets_dependency?` ever stores qualifying params, and it returns before doing so unless the scope has a dependency, i.e. is a `given`. Every other scope therefore always answered empty; it now answers `nil` without the lookups. The sole caller applies `.presence`, so `nil` and the previous empty Array behave the same.

## Benchmarks

| request | delta |
|---|---|
| JSON POST with a nested `requires :user, type: Hash do ... end` (3 attributes) | +5.4% |
| flat params GET (control) | +0.2% (noise) |

Per nested validator: 1.77 → 1.39 µs; per nested scope resolution, about 190 ns saved. The gain grows with nesting depth.

## Test plan

- [x] RuboCop clean.
- [x] Mutation-checked through existing specs: inverting the guard, or returning unconditionally, fails 2 `given` specs each.
- [ ] CI green.


ericproulx:

PR #2926, opened 2026-09-10
Skip the renamed-params lookup in `declared` when nothing is renamed

ericproulx:
## Summary

`DeclaredParamsHandler#build_memo_key` looks every declared param up in the route's renamed params, keyed by the param's whole path. That key is built fresh each time (two Arrays and a String per declared param) on every `declared` call. Only an API using `as:` has anything in that table; everywhere else it is empty and every lookup misses. The lookup is now skipped when the table is empty.

## Benchmarks

`declared` on five params (one nested Hash), isolated:

| call | master | branch |
|---|---|---|
| `declared(params, include_missing: false)` | 4.81 µs | 4.29 µs (−11%) |
| `declared(params)` | 6.16 µs | 5.59 µs (−9%) |

End to end, a JSON POST returning `declared(params, include_missing: false)` measured +1.4% (median of 7 interleaved subprocess rounds, Ruby 4.0.6, no JIT), which is inside that harness's noise floor: master against itself read −1.2% the same day. The saving is about 100 ns per declared param, so it is a small, fixed win that scales with how many params an endpoint declares rather than a headline number.

## Test plan

- [x] RuboCop clean.
- [x] Mutation-checked through existing specs: inverting the guard fails 14 `as:` specs.
- [ ] CI green.


ericproulx:

PR #2928, opened 2026-09-10
Settle the attributes iterator's per-scope state once per pass

ericproulx:
## Summary

For every element of an array scope, and for every validator on it, `AttributesIterator#do_each` re-derived things that depend only on the scope and the request: whether the scope iterates elements, whether its params are an Array, which scope an element's index is recorded against, and the request's `ParamScopeTracker` (a fiber-storage read). They are now settled once per `#each` and handed down, and the top-level call no longer allocates its empty `parent_indices`. An element-iterating scope whose params are not an Array returns before iterating, since it yielded and recorded nothing anyway.

One validator's iteration overhead on a ten-element array: 5.46 → 4.64 µs.

## Missing spec, added

The branch where a lateral scope (a `with` group inside an array) records each element's index against its nearest array ancestor had no coverage: removing it passed the whole suite, while errors then named the wrong element (`array[2][b]` for the first element, and two errors collapsed into one). The new spec fails a first and a last element; it passes before and after this change, and fails with that branch removed.

## Benchmarks

Median of 9 interleaved subprocess rounds against master, Ruby 4.0.6, no JIT:

| request | delta |
|---|---|
| JSON POST, 10 array elements × 3 attributes | +4.1% |
| GET with flat params (control) | +0.2% (noise) |

## Test plan

- [x] RuboCop clean.
- [x] Mutation-checked: never recording indices fails 23 specs; dropping parent indices fails 2; validating an array scope that did not get an Array fails 1; ignoring the nearest array ancestor fails 1 (the new spec).
- [ ] CI green.


ericproulx:

PR #2929, opened 2026-09-10
Skip three redundant steps on every error response

ericproulx:
## Summary

Three steps every error response took for nothing:

- **`Middleware::Error#rack_response` copied the headers into a `Grape::Util::Header`** only to hand them to `Rack::Response.new`, which copies its argument into a Headers of its own on every supported Rack: Rack 2.2's `HeaderHash[]` adopts only a Hash that already is a `HeaderHash`, and 3.x always builds a fresh `Rack::Headers`. #2911 kept this copy on the grounds that callers write into what they are given, but here the writer is `Rack::Response`, which writes into its own copy, and all three callers build the Hash fresh for this response. The copy #2911 kept in `API::Instance#call`, which can be handed a mounted app's frozen Hash, stays.
- **`run_rescue_handler` ran every handler through `instance_exec`**, turning a `Method` into a Proc first. The middleware's own handlers (`method(:error_response)`, `method(:default_rescue_handler)`) and the endpoint's for a `with:` Symbol are Methods, already bound to a receiver that `instance_exec` cannot change, so they are now called as they are. Blocks still run as the endpoint.
- **`ErrorFormatter::Json#ensure_utf8` re-encoded every String message.** For one that already is valid UTF-8, `encode` only returns a copy, for about 260 ns. Such a message is now returned as it is; anything else is converted as before.

## Benchmarks

| response | delta |
|---|---|
| 405 (raised `MethodNotAllowed`) | +4.9% |
| 401 via `error!` | +3.3% |
| 400 validation error | +2.6% |
| 404 via a `rescue_from` block calling `error!` | +2.6% |
| 500 via `rescue_from :all` | +2.3% |
| successful GET (control) | +0.5% (noise) |

## Behaviour

Byte-identical to master (status, headers and body) over a 228-case matrix: JSON, txt, XML and content-negotiated APIs × `error!` with a String, a Hash, custom headers, an HTML content type, and frozen, binary, malformed and US-ASCII messages; `rescue_from` with a block, a `with:` Symbol, an arity-0 block, a handler that returns junk, one that raises again and one that returns a `Rack::Response`; unhandled exceptions, validation errors, 405, OPTIONS and 404; each with three Accept headers.

## Missing spec, added

What `ensure_utf8` does with a message that is not valid UTF-8 had no spec: dropping the encoding half of the new check passed the whole suite while a binary message went out as raw bytes (json 3 warns this "will raise an encoding error"). `spec/grape/error_formatter/json_spec.rb` pins the three cases: UTF-8 as it is, binary converted with the unmappable bytes replaced, and malformed UTF-8 with the invalid bytes replaced. It passes before and after this change.

## Test plan

- [x] RuboCop clean.
- [x] Mutation-checked: calling every Method handler with the error fails 2 specs (arity-0 `with:` handlers); skipping the validity half of the UTF-8 check fails 2; skipping the encoding half fails 1 (the new spec).
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#2918](https://github.com/ruby-grape/grape/pull/2918): Skip the dry-types round trip when a value already is the declared type - [@ericproulx](https://github.com/ericproulx).
 * [#2917](https://github.com/ruby-grape/grape/pull/2917): Read path captures out of the router's union match instead of re-running the route's pattern - [@ericproulx](https://github.com/ericproulx).
 * [#2921](https://github.com/ruby-grape/grape/pull/2921): Pin the router's request-time isolation regressions through requests instead of its instance variables - [@ericproulx](https://github.com/ericproulx).
+* [#2929](https://github.com/ruby-grape/grape/pull/2929): Skip a header copy, a message re-encode and a Proc conversion on every error response - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 4.0.0 (2026-09-07)
--- a/lib/grape/error_formatter/json.rb
+++ b/lib/grape/error_formatter/json.rb
@@ -23,8 +23,11 @@ module Grape
           { error: ensure_utf8(message) }
         end
 
+        # Re-encoding a String that already is valid UTF-8 only copies it, for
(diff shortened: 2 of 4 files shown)


ericproulx:

PR #2932, opened 2026-09-11
Let a parser's Grape errors through the formatter without re-raising

ericproulx:
## Summary

`Formatter#read_rack_input` answers a parser's StandardErrors with a 400, and let Grape's own errors go on to the error middleware by rescuing them first: `rescue Grape::Exceptions::Base => e; raise e`. Re-raising an exception makes Ruby read its backtrace (`setup_exception` calls `rb_get_backtrace`, which builds the lazily captured backtrace as Strings), and at request depth that is the dearest part of the error: about 25 µs. It was paid by every malformed body, since the built-in parsers report one as `InvalidMessageBody`.

A matcher module in the `rescue` clause now selects the errors to answer here, StandardErrors that are not Grape errors, so a Grape error is never rescued at this frame and travels on as it is. Unwinding past a frame that does not rescue it costs nothing extra.

## Benchmarks

Median of 7 interleaved subprocess rounds, Ruby 4.0.6, no JIT, POST with a malformed JSON body:

| API | compared against | delta |
|---|---|---|
| no `rescue_from` | master | **+28.7%** |
| `rescue_from :all` | master + #2931 | **+29.8%** |
| `rescue_from :all` | master | −0.6% |

The last row is not a contradiction. A backtrace is built once and then cached, so the cost is paid by whichever step reads it first. On an API with `rescue_from :all`, master's default handler also reads the backtrace eagerly (fixed by #2931), so either change alone just moves the cost to the other step, and the gain lands once both are in. Without `rescue_from`, nothing else reads it and this change alone gets it all.

## Missing spec, added

A parser raising something that is not a StandardError, such as an Interrupt or a SystemExit, keeps propagating rather than being answered as a 400. Nothing pinned that: dropping the `StandardError` bound from the matcher passed the whole suite. The new spec raises a `NotImplementedError` from a custom parser; it passes before and after this change.

## Behaviour

Byte-identical to master over a 30-case matrix of `rescue_from :all` APIs, with and without `backtrace: true` and `original_exception: true`, in JSON, txt and XML, including malformed bodies. The rendered backtraces of `InvalidMessageBody` are identical once paths are normalized, so re-raising never altered them.

## Test plan

- [x] RuboCop clean.
- [x] Mutation-checked: the matcher also taking Grape errors fails 4 specs; the matcher dropping its `StandardError` bound fails 1 (the new spec).
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#2918](https://github.com/ruby-grape/grape/pull/2918): Skip the dry-types round trip when a value already is the declared type - [@ericproulx](https://github.com/ericproulx).
 * [#2917](https://github.com/ruby-grape/grape/pull/2917): Read path captures out of the router's union match instead of re-running the route's pattern - [@ericproulx](https://github.com/ericproulx).
 * [#2921](https://github.com/ruby-grape/grape/pull/2921): Pin the router's request-time isolation regressions through requests instead of its instance variables - [@ericproulx](https://github.com/ericproulx).
+* [#2932](https://github.com/ruby-grape/grape/pull/2932): Let a parser's Grape errors through the formatter instead of rescuing them to raise again - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 4.0.0 (2026-09-07)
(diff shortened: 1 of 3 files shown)


ericproulx:

PR #2934, opened 2026-09-11
Translate ValidationErrors#full_messages once instead of on every call

ericproulx:
## Summary

`ValidationErrors` builds its `#message` from `#full_messages` when it is created. `#full_messages` then translated every error again on each call: an I18n lookup for the `grape.errors.format` string (with interpolation) and one per attribute name. Each is a few microseconds on I18n 1.15.

The README's recipe for answering with the list, below, therefore paid for every translation twice:

```ruby
  error!({ messages: e.full_messages }, 400)
end
```

The list is now kept from the first call and handed out as a copy, so a caller changing it cannot change what the next caller gets. `#message` is still built through `#full_messages`, so a subclass overriding it keeps working as before.

## Benchmarks

Median of 7 interleaved subprocess rounds, Ruby 4.0.6. The API uses the README handler above, with `requires :id, type: Integer` and `requires :name, type: String`.

| request | no JIT | YJIT |
|---|---|---|
| one param missing | **+21.9%** | **+24.9%** |
| both params missing | **+27.3%** | **+34.4%** |
| default handler, missing param | −0.1% | |
| default handler, invalid Integer | −0.1% | |

## Behaviour

The list now stays what `#message` was built from. Before, a handler that switched the locale and then asked for the list got a mix of both locales. With a French format and a French `presence` message, `I18n.with_locale(:fr) { e.full_messages }` answered `["id : is missing", "nick : is missing"]` for an error raised in English: the French format and untranslated attribute names around the English messages, which were translated when each `Validation` was raised. It now answers `["id is missing", "Nickname is missing"]`, the same text as `#message`. I did not add an UPGRADING entry, since the old output was half-translated anyway; happy to add one.

Otherwise responses are byte-identical to master over:

- a 120-request matrix of `rescue_from ValidationErrors` handlers reading `full_messages` (once, twice, and after changing the list), `message`, both together, `as_json`, `to_json`, `errors`, `error!(e)`, and re-raising. It includes translated attribute names, a request in another locale, `mutually_exclusive` (several params in one error) and `values`. The locale-switching handler above is the only difference.
- the 228-case error-path matrix and the 104-case and 108-case validation matrices from earlier PRs.

## Missing spec, added

A caller changing the array `#full_messages` returned must not change the next answer. The spec passes on master, where each call builds a new array, and fails if the copy is dropped.

## Test plan

- [x] RuboCop clean.
- [x] Mutation-checked: dropping the copy fails the new spec. Dropping the memo is invisible by design, and the benchmark covers it.
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#2918](https://github.com/ruby-grape/grape/pull/2918): Skip the dry-types round trip when a value already is the declared type - [@ericproulx](https://github.com/ericproulx).
 * [#2917](https://github.com/ruby-grape/grape/pull/2917): Read path captures out of the router's union match instead of re-running the route's pattern - [@ericproulx](https://github.com/ericproulx).
 * [#2921](https://github.com/ruby-grape/grape/pull/2921): Pin the router's request-time isolation regressions through requests instead of its instance variables - [@ericproulx](https://github.com/ericproulx).
+* [#2934](https://github.com/ruby-grape/grape/pull/2934): Translate `ValidationErrors#full_messages` once instead of on every call - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 4.0.0 (2026-09-07)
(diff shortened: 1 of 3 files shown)


ericproulx:

PR #2955, opened 2026-09-18
Reject what a collection with multiple member types cannot coerce

ericproulx:
## Summary

A param declared as `type: Array[Integer, String]`, or with such a collection in its `types:` (including the README's `types: [Integer, String, Array[Integer, String]]`), is coerced by `VariantCollectionCoercer`. That coercer checks neither its input nor its output:

```ruby
params { requires :ids, type: Array[Integer, String] }
post('/x') { params[:ids].inspect }
```

| body | `Array[Integer]` | `Array[Integer, String]` before | after |
| --- | --- | --- | --- |
| `{"ids":"x"}` | 400 | **201, `ids` = `nil`** (despite `requires`) | 400 |
| `{"ids":{"a":1}}` | 400 | **201, `nil`** | 400 |
| `{"ids":[1,{}]}` | 400 | **201, `[1, #<InvalidValue>]`** | 400 |
| `{"ids":[1,"2"]}` | 400 | 201, `[1, 2]` | 201, `[1, 2]` |

- A non-Array returned `nil` instead of an `InvalidValue`, so `CoerceValidator` wrote `nil` over the param.
- Members went through the member coercer and the results were returned unchecked, so a member none of the types accepts reached the endpoint as an internal `InvalidValue` object.

Both now answer 400, like a single-type collection. `nil` and `''` still coerce to `nil`, as the existing specs pin.

A `coerce_with` method was also only ever called for an Array. A method that splits a String into the collection never ran, and the param came out `nil`. It now gets the value whatever it is, the same way `MultipleTypeCoercer` hands it over for `types:`.

The same behaviour shows up on 3.3.5 and 4.0.1. Found by fuzzing random `params` schemas.

## Backward compatibility

Both changes turn requests that used to pass into 400s, so UPGRADING has an entry. The unit spec that pinned `call('not an array')` → `nil` now expects an `InvalidValue`.

Members that are `nil` are still accepted (`[1, null]` → `[1, nil]`), because the member coercer itself answers `nil` for them. `Array[Integer]` rejects them. Changing that is a separate decision, so I left it alone.

## Test plan
- [x] Unit specs for the coercer and request specs in `coerce_validator_spec`. All five new examples fail on master and pass here.
- [x] Full RSpec suite passes locally (2916 examples).
- [ ] `spec/integration/hashie` not run locally (hashie isn't installed). It only covers valid input, which takes the same path as before, so it's left to CI.
- [x] RuboCop clean.
- [ ] CI green.


dblock (approved):
If I understood the behavior change correctly, this is a bug? Maybe state so.

PR #2970, opened 2026-09-19
Correct the README on the format of a route admitting dots

ericproulx:
## Summary

"Dots in a path segment" says a requirement that admits dots "claims the extension as well, so `GET /a.b.json` hands the endpoint `"a.b.json"` and format-by-extension no longer applies to that route". The endpoint does get `"a.b.json"`, but format-by-extension still applies. The Formatter reads the extension off the end of `PATH_INFO`, whatever the route captured, so the section's own example answers:

```ruby
route_param :name, requirements: /.+/ do
  get { params[:name] }
end
```

| Request | Status | Content-Type | Body |
|---|---|---|---|
| `GET /a.b.json` | 200 | `application/json` | `"a.b.json"` |
| `GET /a.b.xml` | 500 | `application/xml` | `cannot convert String to xml` |
| `GET /report.pdf` | 200 | `text/plain` | `report.pdf` |

The sentence came in with 8de63ca6 and shipped in 4.0.0; 4.0.1 answers the same way.

The paragraph now says the response format is still read from the path, and that declaring `format` settles it whatever the extension (with `format :json`, `GET /a.b.xml` answers `200` with `"a.b.xml"`). The sentence about `requirements: { n: Float }` was already right and stays.

Docs only; no code changes.


ericproulx:
Folded into #2969.

PR #2972, opened 2026-09-20
Stop the request path resolving params it was not asked for

ericproulx:
Three places where the request path resolved params nobody had asked for, or resolved the same ones again. They are independent; they are in one PR because they are all the same kind of find.

### A nested params scope is resolved from the scope above it

`ParamsScope#params` walks back to the root on every call, and `#should_validate?` asked every scope on the chain for its params in turn — so a scope nested four deep resolved ten scopes to answer one question, and did it again for each of its validators. The cost grows with the square of the nesting depth.

A scope that declares no dependency and has none above it can only be asked whether its own params were given: `given` is the one thing that makes a scope's answer depend on anything else, through the qualifying params it stores per request, and such a chain stores none. So the chain is resolved once from the root down, each scope from the one above it, and checked from there. Every check is pure, so the order it runs in does not change the answer. A chain that does hold a `given` takes the path it always took.

`#params` skips the qualifying-params lookup and the `presence` test that follows it on the same grounds, at every level of every call.

`#all_element_blank?` is only ever asked about params that are not blank, and a Hash that is not blank is never all blank: what `all?` yields for it is a `[key, value]` pair, an Array of two, which `blank?` is false for. Asking anyway walked the Hash and built one of those Arrays per entry.

### The query string is parsed for a `format` param only when it could name one

`?format=json` overrides the format an API pins, so the formatter read it out of the query params — which meant parsing the whole query string on every request that had one, before the endpoint had a say. Parsing is the most expensive thing such a request does (Rack splits the string, unescapes each key and value and builds the nested structure, about 1.2 µs per pair), and an endpoint that never reads its params paid all of it for one lookup that could not have hit.

Rack splits the query string on `&` and `=` and then unescapes each key, and unescaping only ever rewrites `%XX` (or a `+`, which becomes a space). So a query string that spells neither `format` nor a percent-escape cannot hold that key, and the parse is left to whoever reads the params. A request whose endpoint does read them parses exactly once, as before, from `Grape::Request#params`.

**This is the one visible change in the PR**, and UPGRADING spells it out: a query string Rack refuses to parse no longer answers `400` on an endpoint that reads no params.

```ruby
get('/ping') { 'pong' }

# Before
GET /ping?a[a][a]...=1 # => 400
# After
GET /ping?a[a][a]...=1 # => 200
```

An endpoint that reads the params still answers `400`, from the parse it asked for.

### Rack is asked for the form params only when the request can have some

`Rack::Request#params` is the query params merged with the form params, and Grape built its own params out of it. A request that carries no body has no form params to merge, but finding that out still cost a CONTENT_TYPE parse, an empty Hash written back into the env and the merged copy — on every GET that reads its params, which is most of them.

The env answers it first: form params Rack or the formatter's parsers have already read are memoized under Rack's own key, and without them only a body could hold any. The request method is deliberately **not** part of the test — Rack reads a form body whatever the method says, so a GET that sends one still has its params merged in (there is a spec for it). The query params are copied before the params builder sees them, which is what the merge handed over.

## Measured

Against master (`ffb05b21`), interleaved rounds, best of three per round, no JIT / YJIT:

| scenario | no JIT | YJIT |
|---|---|---|
| GET carrying tracking params the endpoint ignores | **+90%** | **+104%** |
| GET carrying one cache-buster param | **+45%** | **+43%** |
| POST nesting three Hash scopes | **+19%** | **+28%** |
| GET reading a nested filter Hash | +10% | +10% |
| GET with one query param | +10% | +9% |
| typed route param | +5.7% | +7.8% |
| `declared` with missing params | +6.6% | +5.0% |
| POST with a flat JSON body | +1.3% | +2.1% |

Controls — bare GET, a GET with no query string, 404, 405, a 422 error, a deeply namespaced route — stay within the ±1.8% the harness swings base-against-base. Boot time and retained memory on a 200-endpoint API with nested params are flat.

## Equivalence

Beyond the suite, each change was diffed against master over a matrix of responses, byte for byte:

- **612 validation cases** — 12 param declarations (optional/required Hash and Array scopes, `given` at the root and inside an Array scope, nested `given`, `at_least_one_of`, defaults with `values:`, `type: JSON`) × 51 bodies (blank, missing, wrong-typed, all-blank elements, `null`, top-level Array).
- **2340 negotiation cases** — a format-pinned and an unpinned API × 6 paths (including `.json`/`.txt`/`.xml` extensions) × 39 query strings (percent-escaped `format` keys, `format` as a value, `formats[]`, nested `a[format]`, repeats) × 5 Accept headers.
- **69 params cases** × 3 param builders — form-encoded, JSON, multipart and empty bodies, GET-with-a-body, route captures colliding with body keys.

All identical. The mutation testing that came with them is in the specs: every mutation of the new code that changes an answer is now killed by the suite, and the two that survive are the pure-speed branches, which the benchmark pins.


lib/grape/middleware/formatter.rb
```diff
@@ -269,9 +272,17 @@ def path_for_extension
...
+      # And a query string that spells neither +format+ nor a percent-escape
+      # cannot hold that key either: Rack splits the string on +&+ and +=+ and
+      # then unescapes each key, and unescaping only ever rewrites +%XX+ (or a
+      # +++, which becomes a space). So the parse -- the most expensive thing
+      # a request with a query string does, and one the endpoint may never ask
+      # for -- is left to whoever reads the params.
       def format_from_query
         query_string = env[Rack::QUERY_STRING]
         return if query_string.nil? || query_string.empty?
+        return unless query_string.include?(FORMAT_PARAM) || query_string.include?('%')
```
dblock, line 285: I think `?f%6frmat=json` is madness ;)

dblock (approved):

lib/grape/middleware/formatter.rb
ericproulx: Indeed, madness and a case I never thought of. Far fetch but still valid.

